PR #180, opened 2017-10-18
Create root directory on startup (if it does not exist)

zwass:


zwass:
The switch from private to public repo caused this PR to stop picking up new changes in my branch. Closing this PR and reopened it as https://github.com/kolide/launcher/pull/184.

PR #185, opened 2017-10-19
Improve runtime test reliability.

zwass:
Instead of waiting a predetermined time and then testing for instance health,
wait up to 30 seconds while constantly checking the health. This allows tests
to potentially run faster, and also not fail if on some runs it takes a bit
longer for the instance to become healthy.


marpaia:
Nice, I think this is a great idea.

PR #187, opened 2017-10-19
Add authenticated debug server

zwass:
- Add the runtime/pprof tools as an HTTP server
- Toggle server state by sending `SIGUSR1` (default off)
- Port randomly chosen from available ports on startup (for reliability)
- Auth token generated on startup (for security)
- Address with auth written to $launcher_root/debug_addr, and logs

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -45,7 +45,3 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
-
-[[constraint]]
-  branch = "master"
-  name = "github.com/e-dard/netbug"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -45,3 +45,7 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/e-dard/netbug"
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kolide/kit/fs"
 	"github.com/kolide/kit/version"
 	"github.com/kolide/launcher/autoupdate"
+	"github.com/kolide/launcher/debug"
 	"github.com/kolide/launcher/osquery"
 	"github.com/kolide/launcher/service"
 	"github.com/kolide/osquery-go/plugin/config"
@@ -90,6 +91,11 @@ func main() {
 
 	if _, err := osquery.DetectPlatform(); err != nil {
 		logFatal(logger, "err", errors.Wrap(err, "detecting platform"))
+	}
+
+	debugTokenPath := filepath.Join(rootDirectory, "debug_token")
+	if err := debug.StartDebugServer("localhost:5097", debugTokenPath, logger); err != nil {
+		logFatal(logger, "err", errors.Wrap(err, "starting debug server"))
 	}
 
 	httpClient := http.DefaultClient
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -94,9 +94,7 @@ func main() {
 	}
 
 	debugTokenPath := filepath.Join(rootDirectory, "debug_token")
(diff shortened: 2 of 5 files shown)


cmd/launcher/launcher.go
```diff
@@ -92,6 +93,11 @@ func main() {
 		logFatal(logger, "err", errors.Wrap(err, "detecting platform"))
 	}
 
+	debugTokenPath := filepath.Join(rootDirectory, "debug_token")
+	if err := debug.StartDebugServer("localhost:5097", debugTokenPath, logger); err != nil {
```
groob: the port should not be hardcoded. We're going to end up failing on a bunch of machines just because. It should be a random available port IMO.

cmd/launcher/launcher.go
```diff
@@ -92,6 +93,11 @@ func main() {
 		logFatal(logger, "err", errors.Wrap(err, "detecting platform"))
 	}
 
+	debugTokenPath := filepath.Join(rootDirectory, "debug_token")
```
marpaia: I think the debug server should be hidden by a flag. Perhaps it's own flag, or perhaps just `--debug`.

zwass: I'd like to reason about the possibility of leaving it on by default... We are not binding to any public address, and not exposing any particularly sensitive information. It's not as useful if someone encounters an issue and then cannot turn on the server for a running process.

zwass:
The compromise we came up with was to leave the debug server off by default, but toggle its state by sending `SIGUSR1`. The server will start with a random available port (for reliability), and a new randomly generated auth token (for security).

Gopkg.toml
```diff
@@ -45,3 +45,7 @@
 
 [[constraint]]
   name = "google.golang.org/grpc"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/e-dard/netbug"
```
marpaia: thoughts on moving the html template into this repo before merging?

debug/debug.go
```diff
@@ -0,0 +1,178 @@
...
+	level.Info(logger).Log(
+		"msg", "debug server started",
+		"addr", addr,
+	)
+
+	return &serv, nil
+}
+
+// The below handler code is adapted from MIT licensed github.com/e-dard/netbug
+func handler(token string, logger log.Logger) http.Handler {
```
groob: You can have http.HandlerFunc as the return arg. That returns a handler but you don't have to cast the function in the return.

debug/debug.go
```diff
@@ -0,0 +1,178 @@
...
+	return http.HandlerFunc(h)
+}
+
+func authHandler(token string, logger log.Logger) http.Handler {
+	h := handler(token, logger)
+	ah := func(w http.ResponseWriter, r *http.Request) {
+		if r.FormValue("token") == token {
+			h.ServeHTTP(w, r)
+		} else {
+			w.WriteHeader(http.StatusUnauthorized)
```
groob: Use http.Error

debug/debug.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, errors.Wrap(err, "generating debug token")
+	}
+
+	// Start the debug server
+	r := http.NewServeMux()
+	registerAuthHandler(token.String(), r, logger)
+	serv := http.Server{
+		Handler: r,
+	}
+	listener, err := net.Listen("tcp", "localhost:0")
```
groob: Can you add a comment explaining :0 ?

debug/debug.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, errors.Wrap(err, "opening socket")
+	}
+
+	go func() {
+		if err := serv.Serve(listener); err != nil && err != http.ErrServerClosed {
+			level.Info(logger).Log("msg", "debug server failed", "err", err)
+		}
+	}()
+
+	addr := fmt.Sprintf("http://%s/debug/?token=%s", listener.Addr().String(), token.String())
```
groob: What do you think about adding url.Parse on the addr for additional validation?

PR #190, opened 2017-10-20
Use public imports for Dep

### LINKED ISSUE #189: Update Gopkg.toml to use public github repo imports
Similar to [Issue #1580 on Fleet](https://github.com/kolide/fleet/issues/1580), the Gopkg.toml is using the git@ ssh style, which breaks builds if the user doesn't have SSH keys setup in git.

marpaia:
close #189

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -7,6 +7,12 @@
   revision = "0dd823580c78a79ae9696eb9b3650e400fff140f"
 
 [[projects]]
+  name = "github.com/Microsoft/go-winio"
+  packages = ["."]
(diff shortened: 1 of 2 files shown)


Gopkg.lock
```diff
@@ -7,6 +7,12 @@
   revision = "0dd823580c78a79ae9696eb9b3650e400fff140f"
 
 [[projects]]
+  name = "github.com/Microsoft/go-winio"
```
zwass: Was this pulled in as a recursive dep from `osquery-go`?

marpaia: Yeah, I believe the `Gopkg.lock` also declares transitive deps as well, but notice it's not in `Gopkg.toml`

PR #191, opened 2017-10-20
Limit log writes by number of bytes

zwass:
Previously, buffered log writes were limited by the total number of logs. Now
they are limited by total number of bytes, and logs that are over the max size
limit are dropped.

### DIFF
--- a/osquery/extension.go
+++ b/osquery/extension.go
@@ -3,6 +3,7 @@ package osquery
 import (
 	"context"
 	"encoding/binary"
+	"fmt"
 	"sync"
 	"time"
 
@@ -27,6 +28,7 @@ type Extension struct {
 	enrollMutex   sync.Mutex
 	done          chan struct{}
 	wg            sync.WaitGroup
+	logger        log.Logger
 }
 
 const (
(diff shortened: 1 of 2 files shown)


osquery/extension.go
```diff
@@ -44,9 +46,10 @@ const (
 	// DB key for last retrieved config
 	configKey = "config"
 
-	// Default maximum number of logs per batch (used if not specified in
-	// options)
-	defaultMaxLogsPerBatch = 500
+	// Default maximum number of bytes per batch (used if not specified in
+	// options). This 4MB limit is chosen based on the default grpc-go
+	// limit specified in https://github.com/grpc/grpc-go/blob/master/server.go#L51
+	defaultMaxBytesPerBatch = 4 * 1024 * 1024
```
groob: thoughts on using `4 << 20` here? It's a more common notation for constants.

osquery/extension.go
```diff
@@ -403,8 +404,25 @@ func (e *Extension) writeBufferedLogsForType(typ logger.LogType) error {
...
 		c := b.Cursor()
 		k, v := c.First()
-		for total := 0; k != nil && total < e.Opts.MaxLogsPerBatch; total++ {
-			logs = append(logs, string(v))
+		for totalBytes := 0; k != nil; {
+			if len(v) > e.Opts.MaxBytesPerBatch {
+				// Discard logs that are too big
+				level.Info(e.Opts.Logger).Log(
+					"msg",
+					fmt.Sprintf(
```
groob: use key/values instead of sprintf

zwass:
I'm going to merge this if there are no objections.

groob:
🚢

PR #194, opened 2017-10-25
Increase timeout to open osquery extension socket

zwass:
Occasionally during tests the 5s timeout would expire. Bumping to 10s still
seems like a reasonable limit, and may prevent startup from failing under
unusual circumstances.


groob:
No objections here.

PR #195, opened 2017-10-25
Refactor logging and expose ability to toggle level in running process

zwass:
- Create a new logger that defaults to info level logging.
- Add the ability to toggle info/debug logging by sending `SIGUSR2`.
- Refactor some logging.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -8,23 +8,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 )
 
-// Logger is the interface exposed to users of this launcher logger package.
-type Logger interface {
-	kitlog.Logger
-	// Debug logs keyvals at the "debug" level
-	Debug(keyvals ...interface{}) error
-	// Info logs keyvals at the "info" level
-	Info(keyvals ...interface{}) error
-	// Fatal logs keyvals at the "info" level, and then exits with error code 1.
-	Fatal(keyvals ...interface{}) error
-
-	// AllowDebug sets the allowed log level to include debug logs and above (info).
-	AllowDebug()
-	// AllowInfo sets the allowed log level to include info logs.
-	AllowInfo()
-}
-
-type logger struct {
+type Logger struct {
 	// baseLogger stores the un-leveled logger. Writes should not be done
 	// directly to this logger.
 	baseLogger kitlog.Logger
@@ -34,18 +18,18 @@ type logger struct {
 	swapLogger kitlog.SwapLogger
 }
 
-func (l *logger) AllowDebug() {
+func (l *Logger) AllowDebug() {
 	l.allowedLevel(level.AllowDebug(), "debug")
 }
 
-func (l *logger) AllowInfo() {
+func (l *Logger) AllowInfo() {
 	l.allowedLevel(level.AllowInfo(), "info")
 }
 
-func (l *logger) allowedLevel(lev level.Option, name string) {
+func (l *Logger) allowedLevel(lev level.Option, name string) {
 	newLogger := level.NewFilter(l.baseLogger, lev)
 	l.swapLogger.Swap(newLogger)
-	l.Info(
+	level.Info(&l.swapLogger).Log(
 		"msg", "allowed log level set",
 		"allowed_level", name,
 	)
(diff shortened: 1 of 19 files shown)


service/server_grpc.go
```diff
@@ -26,7 +26,6 @@ func parseUUID() grpctransport.ServerOption {
 
 func NewGRPCServer(endpoints KolideClient, logger log.Logger) pb.ApiServer {
 	options := []grpctransport.ServerOption{
-		grpctransport.ServerErrorLogger(logger),
```
zwass: Note I killed this because the docs say
> ServerErrorLogger is used to log non-terminal errors. By default, no errors are logged.
and I think we should leave it at the default.

log/log.go
```diff
@@ -0,0 +1,87 @@
...
+	// Call this directly instead of using l.Info so that we get the
+	// correct caller.
+	level.Info(l.swapLogger).Log(keyvals...)
+	os.Exit(1)
+	// never hit
+	return nil
+}
+
+func NewLogger(w io.Writer) Logger {
+	base := log.NewJSONLogger(log.NewSyncWriter(w))
```
groob: I often prefer logfmt for reading logs locally. This hardcodes the output format too.

service/server_grpc.go
groob: I'm not sure I agree. This method is for server side implementations so it would not affect the launcher client. Is there a reason we wouldn't want to log if a grpc error happened (unable to decode, request failed etc)

IMO if you're removing it, the `NewGRPCServer` should accept a list of `grpctransport.ServerOption` because there's no way to add a logger if it's needed.

service/server_grpc.go
```diff
@@ -3,8 +3,8 @@ package service
 import (
 	"context"
 
-	"github.com/go-kit/kit/log"
 	grpctransport "github.com/go-kit/kit/transport/grpc"
+	"github.com/kolide/launcher/log"
```
groob: This is a breaking change for servers. since the `log.Logger` in this file has changed to a much larger interface, any `NewGRPCServer` will require to instantiate a new launcher/log.Logger rather than the go-kit one. I'm not sure that's a good change.

service/logging.go
```diff
@@ -25,7 +25,7 @@ type logmw struct {
 func (mw logmw) RequestEnrollment(ctx context.Context, enrollSecret, hostIdentifier string) (errcode string, reauth bool, err error) {
 	defer func(begin time.Time) {
 		uuid, _ := uuid.FromContext(ctx)
-		mw.logger.Log(
+		mw.logger.Debug(
```
groob: These logging middleware are also used by the server and should not be enforcing a level. The level was enforced when the logger is created. 

See the pull request that removed `Debug` for more info https://github.com/kolide/launcher/pull/181

log/log.go
```diff
@@ -0,0 +1,96 @@
...
+	// Fatal logs keyvals at the "info" level, and then exits with error code 1.
+	Fatal(keyvals ...interface{}) error
+
+	// AllowDebug sets the allowed log level to include debug logs and above (info).
+	AllowDebug()
+	// AllowInfo sets the allowed log level to include info logs.
+	AllowInfo()
+}
+
+type logger struct {
```
groob: Do you think it would be possible to scale this package back in this way?

- Export the `logger` struct. 
- Return the struct. 
- Don't declare an interface in this package. 
- Only make any necessary changes in main, leaving other packages to only consume a go-kit logger? 

The reason I'm asking is because the logger is now required by all the packages in this repo, including ones meant to be imported by server-side implementations. 

 I wouldn't be opposed to discussing/iterating on the proposed interface here, but I think it should be more carefully considered before it propagates.

groob: For example, for GCP servers I have created this helper https://github.com/kolide/kit/blob/566c8f56a6ff7daba204818fbab0f2cb854b3310/logutil/logutil.go#L22 to rewrite the `level` key. I'm not sure how that will work with this logger.

zwass:
@groob I scoped this back in line with your comments and I feel better about it. Please let me know what you think.

groob:
🚢

PR #202, opened 2017-11-01
add pathstate to launchd template

### LINKED ISSUE #199: update launchd to wait until secret is present
Some users would like to add the enrollment secret out of band from the install. To do that, a suggested solution is to add an `OnDemand` conditional to launchd so the process only starts after the secret is added at the right path.

groob:
Keeps the launchd process up only if the enroll secret is present.

Closes #199

### DIFF
--- a/tools/packaging/packaging.go
+++ b/tools/packaging/packaging.go
@@ -453,7 +453,13 @@ func renderLaunchDaemon(w io.Writer, options *launchDaemonTemplateOptions) error
         <key>RunAtLoad</key>
         <true/>
         <key>KeepAlive</key>
-        <true/>
+        <dict>
+            <key>PathState</key>
+            <dict>
+                <key>{{.SecretPath}}</key>
+                <true/>
+            </dict>
+        </dict>
         <key>ThrottleInterval</key>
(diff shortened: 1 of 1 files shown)


tools/packaging/packaging.go
```diff
@@ -450,19 +450,27 @@ func renderLaunchDaemon(w io.Writer, options *launchDaemonTemplateOptions) error
...
         <key>ThrottleInterval</key>
         <integer>60</integer>
         <key>ProgramArguments</key>
         <array>
             <string>{{.LauncherPath}}</string>
-            <string>--debug</string>{{if .InsecureGrpc}}
-            <string>--insecure_grpc</string>{{end}}{{if .Insecure}}
-            <string>--insecure</string>{{end}}{{if .Autoupdate}}
-            <string>--autoupdate</string>{{end}}
+            {{if .InsecureGrpc}}
```
marpaia: The above makes the resultant launch daemons look nicer even though it's kind of ugly in the template. This is not in scope of this PR, but I'm ok with this change as long as you've verified all of the conditions result in the appropriate flags.

PR #204, opened 2017-11-01
JSON format osquery logs at debug level

zwass:


zwass:
For some reason the caller is being overwritten strangely:
```
{"caller":"I1101 15:43","component":"osquery","level":"debug","msg":":25.484685 163426304 distributed.cpp:133] Executing distributed query: kolide:populate:foo: SELECT COUNT(*) AS result FROM (SELECT * FROM time);","ts":"2017-11-01T22:43:25.484755Z"}
```

zwass:
That problem is now fixed, but we could regex to properly extract the caller information from the osquery logs...

cmd/launcher/launcher.go
```diff
@@ -149,6 +149,8 @@ func main() {
 	ext.Start()
 	defer ext.Shutdown()
 
+	osqueryLogger := &kolidelog.OsqueryLogAdapter{level.Debug(log.With(logger, "component", "osquery"))}
```
groob: Sweet. But would benefit from creating a multiline struct.

zwass:
The caller is now extracted into the JSON:
```
{"caller":"distributed.cpp:133","component":"osquery","level":"debug","msg":"I1101 19:39:01.540799 41111552 distributed.cpp:133] Executing distributed query: kolide:populate:time: SELECT * from time;","ts":"2017-11-02T02:39:01.540916Z"}
```

log/log.go
```diff
@@ -77,8 +79,20 @@ type OsqueryLogAdapter struct {
 	kitlog.Logger
 }
 
+var callerRegexp = regexp.MustCompile(`[\w.]+:\d+]`)
+
+func extractOsqueryCaller(msg string) string {
+	if match := callerRegexp.FindString(msg); len(match) > 0 {
+		return strings.TrimSuffix(string(match), "]")
+	} else {
```
groob: No need for else if there's a return in the previous branch. 

I think there's a linter that catches this case.

zwass: We can actually remove the `if` entirely ;) We only ended up here because I was tinkering around with the implementation.

PR #215, opened 2017-11-29
Osquery Simulator for Load Testing

marpaia:
This PR contains an implementation of an osquery launcher load test simulator.


simulator/simulator.go
```diff
@@ -0,0 +1,124 @@
+package simulator
+
+import (
+	"errors"
+	"sync"
+)
+
+// FakeHost is the interface which defines the pluggable behavior of a simulated
+// host. Each "type" of host may have their own implementation of this interface.
+type FakeHost interface {
```
zwass: I think we should probably rename this interface to `Osquery` or `FakeOsquery`.

groob: Why not QueryRunner?

groob: I'm curious as to the strategy here. I would think that the Host interface would have the same methods as the osquery launcher. It needs to exercise all endpoints that the launcher does(enroll, log, config, read, write).

simulator/simulator.go
```diff
@@ -32,6 +34,9 @@ type HostSimulation struct {
 	//   h.state.lock.Lock()
 	//   defer h.state.lock.Unlock()
 	state *hostSimulationState
+
+	shutdown chan bool
```
groob: You can accomplish the shutdown with a single field, which I think is a bit cleaner: 

```
type Runner struct {
        shutdown   chan chan struct{}
}

func (r *Runner) loop() {
...
        case quit := <-r.shutdown:
            close(quit)
            return
        }   
}

func (r *Runner) Shutdown() {
    quit := make(chan struct{})
    r.shutdown <- quit
    <-quit
}
```

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+	"sync"
+
+	"github.com/ghodss/yaml"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+)
+
+type Host struct {
```
marpaia: Add comment

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type matcher struct {
+	// Pattern is a regexp for the query patterns this should match.
+	Pattern regexp.Regexp `json:"pattern"`
+	// Results is the results to return for matched queries
+	Results []map[string]string `json:"results"`
+}
+
+type hostYAML struct {
```
marpaia: Perhaps we can rename this to something like `hostSimulationSpecification` or something more related to the content of the file?

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+	Name       string `json:"name"`
+	ParentName string `json:"parent"`
+	Queries    []struct {
+		Pattern string              `json:"pattern"`
+		Results []map[string]string `json"results"`
+	} `json:"queries"`
+}
+
+func LoadHosts(dir string, logger log.Logger) (map[string]*Host, error) {
```
marpaia: add comment

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		host.parent = parent
+	}
+
+	// TODO check for cycles
+
+	return hostMap, nil
+}
+
+func (h *Host) RunQuery(sql string) (rows []map[string]string, err error) {
```
marpaia: add comment

cmd/launcher-pummel/README.md
```diff
@@ -0,0 +1,21 @@
...
+./build/launcher-pummel --help
+```
+
+## Usage Instructions
+
+```
+./build/launcher-pummel \
+  --host_path=./path/to/my/host/templates \
+	--server_url=fleet.acme.co \
+	--enroll_secret=mB3XE5wwLt3YryD9FAanjwhm02HoOqll \
```
groob: would love to see enroll secret path instead.

zwass: Does the path allow this to be more easily used for cloud/fleet testing?

cmd/launcher-pummel/launcher-pummel.go
```diff
@@ -0,0 +1,128 @@
...
+		"msg", "all hosts started",
+	)
+
+	sleep := make(chan struct{})
+	<-sleep
+}
+
+func newLogger(w io.Writer) log.Logger {
+	logger := log.NewJSONLogger(log.NewSyncWriter(w))
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC)
```
groob: `log.With` is variadic, can combine both `ts` and `component` fields. Maybe we can do that and keep each k/v on a newline?

cmd/launcher-pummel/README.md
groob: Yes, because we generally save the secret to a file, and our configurations all use the `enroll_secret_path` field. 

At least that's been my personal expectation so far.

cmd/launcher-pummel/README.md
```diff
@@ -0,0 +1,21 @@
...
+	mac:100 windows:20 linux:5000
```
groob: Thoughts on having the host types as a flag with csv values instead of an argument?

```--hosts=mac:100,windows:20,linux:5000```

Not a big preference but having only flags would appear to be more consistent.

File: cmd/launcher-pummel/launcher-pummel.go [hunk 10]
```diff
@@ -0,0 +1,128 @@
...
+		// Start hosts
+		for i := 0; i < count; i++ {
+			simulator.LaunchSimulation(
+				host,
+				*flServerURL,
+				fmt.Sprintf("%s_%d", hostType, i),
+				*flEnrollSecret,
+				simulator.WithInsecure(),
+			)
+			time.Sleep(10 * time.Millisecond)
```
groob: Maybe a POST MVP action but I'd love to limit enrollment in batches. So you could enroll 50 hosts at once or 500 etc. 

I could see the for loop here start go-routines semaphored by a buffered channel

cmd/launcher-pummel/launcher-pummel.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+	level.Info(logger).Log(
+		"msg", "all hosts started",
+	)
+
+	sleep := make(chan struct{})
```
groob: Let's use the pattern here:

https://github.com/kolide/launcher/blob/bf16180aa067695b7cd6befa5a2a198a8dbf1696/cmd/osquery-extension/osquery-extension.go#L19-L21

simulator/hosts.go
groob: hostSpec sounds good to me.

simulator/hosts.go
groob: Any reason for `map[string]*Host` vs `map[string]Host`?

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+				re, err := regexp.Compile(strings.ToLower(q.Pattern))
+				if err != nil {
+					return nil, errors.Wrapf(err, "compile regexp for %s", path)
+				}
+				host.Queries = append(host.Queries, matcher{*re, q.Results})
+			}
+
+			// Check for duplicate host type name
+			if _, exists := hostMap[host.Name]; exists {
+				return nil, errors.Errorf("duplicate host type: %s", host.Name)
```
groob: Would love to see in the comment on line 89 why you want to return an error instead of logging and ignoring the duplicate.

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			hostMap[host.Name] = &host
+		}
+	}
+
+	// Link parents
+	for _, host := range hostMap {
+		if host.ParentName == "" {
```
groob: Is a empty string even possible outside of operator error? maybe log or fail here?

simulator/hosts.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+
+		parent, ok := hostMap[host.ParentName]
+		if !ok {
+			return nil, errors.Errorf("missing parent named: %s", host.ParentName)
+		}
+		host.parent = parent
+	}
+
+	// TODO check for cycles
```
groob: Is this still relevant?

simulator/hosts/base_darwin.yaml
```diff
@@ -0,0 +1,31 @@
+name:
+  base_darwin
+queries:
+  - pattern: select 1$
+    results: [
+      {"1":"1"}
+    ]
+
+  - pattern: select \* from interface_details id join interface_addresses ia
```
groob: is there a way to avoid escaping the `*`?

simulator/simulator.go
```diff
@@ -0,0 +1,329 @@
...
+	config, invalid, err := h.state.serviceClient.RequestConfig(context.Background(), h.state.nodeKey)
+	if err != nil {
+		return errors.Wrap(err, "transport error retrieving config")
+	}
+	if invalid {
+		return errors.New("enrollment invalid in request config")
+	}
+
+	// Ignore actual config. Someday we may do something with the config.
+	_ = config
```
groob: ignore it on line 78 instead.

simulator/simulator.go
```diff
@@ -0,0 +1,329 @@
...
+	//   h.state.lock.RLock()
+	//   defer h.state.lock.RUnlock()
+	//
+	// To write state based on the on-going simulation:
+	//
+	//   h.state.lock.Lock()
+	//   defer h.state.lock.Unlock()
+	state *hostSimulationState
+
+	shutdown chan bool
```
groob: see my previous comment about using a single `chan chan struct{}` for the shutdown method.

simulator/simulator.go
```diff
@@ -0,0 +1,329 @@
...
+func (h *HostSimulation) Enroll() error {
+	h.state.lock.Lock()
+	defer h.state.lock.Unlock()
+
+	nodeKey, invalid, err := h.state.serviceClient.RequestEnrollment(context.Background(), h.enrollSecret, h.uuid)
+	if err != nil {
+		return errors.Wrap(err, "transport error in enrollment")
+	}
+	if invalid {
+		return errors.New("enrollment invalid")
```
groob: this message is too generic to see where the error came from. maybe `simulator: enrollment invalid` and `fmt.Errorf` to include some information about which host failed.

cmd/launcher-pummel/README.md
marpaia: if we wanted to make it more like launcher (which has both with an order of precedence), we could accept both: https://github.com/kolide/launcher/blob/bf16180aa067695b7cd6befa5a2a198a8dbf1696/cmd/launcher/launcher.go#L123-L132

simulator/hosts.go
marpaia: the parent would be empty if the query result file was not inheriting anything (the most senior parent).

simulator/hosts.go
marpaia: Yeah, ideally the data structure that we just assembled is a proper DAG, but we have not actually verified that the graph is acyclic. Ideally the tool would identify the invalid relationship (two files each saying the other is it's parent), but right now that will cause undetermined behavior.

marpaia: We can just try our best to not give it invalid values for `parent` for now though and get back to this some other time IMO.

simulator/simulator.go
marpaia: I don't know, I am not super convinced that a channel of channels is cleaner than two channels really. Do you have any further reading on this topic?

simulator/hosts.go
zwass: To be precise, if there is a cycle the `RunQuery` call may end up infinite looping if none of the hosts in the cycle have a match for the given query.

simulator/hosts/base_darwin.yaml
zwass: Not that I know of... These are regexes, and `*` is interpreted as usual.

simulator/simulator.go
zwass: We currently don't log this error in any case... Maybe we should in `LaunchSimulation`?

simulator/simulator.go
```diff
@@ -0,0 +1,329 @@
...
+// WithPublishLogsInterval is a functional option which allows the user to
+// specify how often the simulation host should log status and results logs
+func WithPublishLogsInterval(interval time.Duration) SimulationOption {
+	return func(i *HostSimulation) {
+		i.publishLogsInterval = interval
+	}
+}
+
+// WithInsecure is a functional option which allows the user to declare that the
+// remote API should be connected to over an insecure channel
```
zwass: This comment (and the one on line 199) don't really accurately explain what these options do. IIRC `WithInsecure()` will skip cert verification but still use TLS, while `WithInsecureGrpc()` will turn off TLS entirely.

simulator/simulator.go
groob: No reading, just examples 
https://github.com/oklog/oklog/blob/d78d4a9b54097c70a3abb4a793dca5f30cf9f4b4/pkg/ingest/api.go#L107-L111

There's a larger pattern within that code that has to do with creating actors which accept functions on a channel too that would be nice to discuss separately. I have some reading on the actor pattern specifically, not on the `stop chan chan struct{}` primitive. 

https://github.com/oklog/oklog/blob/d78d4a9b54097c70a3abb4a793dca5f30cf9f4b4/pkg/ingest/api.go#L32

simulator/hosts/base_darwin.yaml
marpaia: maybe we should me these patterns `from interface_details id join interface_addresses ia` and omit the columns?

PR #218, opened 2017-12-01
add mdm profile info table

groob:
excuse the exec jank, https://github.com/facebook/osquery/issues/1666#issuecomment-160812564

```
osquery> select * from kolide_mdm_info;
                 enrolled = true
               server_url = https://dev.micromdm.io/mdm/connect
              checkin_url = https://dev.micromdm.io/mdm/checkin
            access_rights = 8191
             install_date = 2017-12-01 22:53:00 +0000
       payload_identifier = com.github.micromdm.micromdm.enroll
                    topic = com.apple.mgmt.External.50C91AEB-9506-484C-A83A-86CBDE06A95F
             sign_message = true
identity_certificate_uuid = f8f9b353-393a-45eb-b70d-51c64910e943
         has_scep_payload = true
```

### DIFF
--- a/osquery/mdm.go
+++ b/osquery/mdm.go
@@ -1,0 +1,113 @@
+package osquery
+
+import (
+	"context"
+	"os/exec"
+	"strconv"
+
+	"github.com/go-kit/kit/log"
+	"github.com/groob/plist"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+func MDMInfo(logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("enrolled"),
+		table.TextColumn("server_url"),
+		table.TextColumn("checkin_url"),
+		table.IntegerColumn("access_rights"),
+		table.TextColumn("install_date"),
+		table.TextColumn("payload_identifier"),
+		table.TextColumn("topic"),
+		table.TextColumn("sign_message"),
+		table.TextColumn("identity_certificate_uuid"),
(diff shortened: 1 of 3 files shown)


osquery/mdm.go
```diff
@@ -0,0 +1,113 @@
...
+				continue
+			}
+			if item.PayloadType == "com.apple.mdm" {
+				enrollProfile := item.PayloadContent
+				enrollProfileItems = payload.ProfileItems
+				mdmResults = map[string]string{
+					"enrolled":                  "true",
+					"server_url":                enrollProfile.ServerURL,
+					"checkin_url":               enrollProfile.CheckInURL,
+					"access_rights":             fmt.Sprintf("%d", enrollProfile.AccessRights),
```
marpaia: `strconv.Atoi` if there is no other formatting going on here

groob: strconv.Atoi has an `err` return making it more annoying to since it can't be used on the same line.

zwass: Maybe you mean `strconv.Itoa`?

groob: Ah good call. I'll use `strconv.Itoa`.

osquery/mdm.go
```diff
@@ -0,0 +1,113 @@
...
+					"install_date":              payload.ProfileInstallDate,
+					"payload_identifier":        payload.ProfileIdentifier,
+					"sign_message":              fmt.Sprintf("%v", enrollProfile.SignMessage),
```
zwass: If we are preferring `strconv` to `fmt`, I think we should use `strconv.FormatBool` here.

PR #222, opened 2017-12-05
Kill osqueryd spawned children during restart

### LINKED ISSUE #221: osqueryd restart method not restarting osquery with autoupdate
Looks like a regression from https://github.com/kolide/launcher/commit/f7ea6538353bfd3c9aa1aebc6f95fb894faee26d but still investigating. 

The restart method when osqueryd is autoupdate no longer restart the instance. Launcher is still running, but there's no osqueryd process. 
```
{"binary":"/usr/local/kolide-corp/bin/osqueryd","caller":"client.go:291","component":"launcher","file":"/usr/local/kolide-corp/bin/osqueryd-staging/darwin/osqueryd-beta.tar.gz","level":"info","msg":"new staged tuf file","target":"darwin/osqueryd-beta.tar.gz","ts":"2017-12-04T23:23:55.244934Z"}
{"binary":"/usr/local/kolide-corp/bin/osqueryd","caller":"client.go:291","component":"launcher","level":"info","msg":"completed update for binary","ts":"2017-12-04T23:23:55.439691Z"}
...
```

groob:
fix osquery restart process

    cmd.Process.Kill() only kills the child process, but not those spawned by osqueryd.
    When Restart() was called, osqueryd would die, but the extension process would remain, causing
    a new osqueryd process to not start again. Killing the whole process group fixes the error.

    Fixed the wait_healty test helper, which was shadowing the error and terminating early.
    Added errgroup.Wait() in the restart method to wait for the old osquery extension to die.

Closes #221 

Used https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773 for reference.

### DIFF
--- a/osquery/runtime.go
+++ b/osquery/runtime.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -444,6 +445,9 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "couldn't create osqueryd command")
 	}
 
+	// Assign a PGID that matches the PID. This lets us kill the entire process group later.
+	o.cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
 	// Launch osquery process (async)
 	err = o.cmd.Start()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


osquery/runtime.go
```diff
@@ -565,10 +570,12 @@ func (r *Runner) launchOsqueryInstance() error {
 // instance with the same configurations.
 func (r *Runner) Restart() error {
 	r.instanceLock.Lock()
-	defer r.instanceLock.Unlock()
 	// Cancelling will cause all of the cleanup routines to execute, and a
 	// new instance will start.
 	r.instance.cancel()
+	r.instance.errgroup.Wait()
+	r.instanceLock.Unlock()
```
zwass: Why switch this from defer?

groob: tests weren't passing initially, but maybe this is not needed anymore. Let me rerun the test with a defer.

PR #224, opened 2017-12-06
Update Notary repo URL

### LINKED ISSUE #196: Notary repo moving
Hi!

We were extremely pleased to find out you're using Notary. I wanted to give you a heads up that with acceptance in to CNCF, notary will be moving to `GitHub.com/theupdateframework/notary` and it would be good to update your imports after that transition.

I will be moving the repo in the next 48 hours which will include updating all the internal imports. There will be a redirect in place and it looks like you're currently pinned to a specific git commit so nothing should break for you immediately with the move.

zwass:
Closes #196

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -2,10 +2,10 @@
   name = "github.com/boltdb/bolt"
 
 [[constraint]]
-  name = "github.com/docker/notary"
+  name = "github.com/theupdateframework/notary"
   # pin to the revision because there has not been a new release
   # in 6 months and we depend on changes in master.
-  revision = "fab4c6708bc200c4cfe56c74f902d6e8bb26c803"
+  revision = "332094e0c1c45a46b4c3be9fe239672d900a2fea"
 
 [[constraint]]
   name = "github.com/go-kit/kit"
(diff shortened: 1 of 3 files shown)


Gopkg.toml
```diff
@@ -2,7 +2,7 @@
   name = "github.com/boltdb/bolt"
 
 [[constraint]]
-  name = "github.com/docker/notary"
+  name = "github.com/theupdateframework/notary"
```
groob: don't you have to also rename all the imports?

zwass: Yes, good point (though I'm not sure why CI passed anyway). Also had to bump the commit we pin so that it doesn't transitively import `docker/notary`.

groob: I think CI passed because we use a cache https://github.com/kolide/launcher/blob/6e3b4226054aaa432d58fba4f66871dd6b4f3786/.circleci/config.yml#L9 and the old dep is still there.

zwass: Could also be because we still ended up with `docker/notary` in the `Gopkg.lock` (it passed for me locally as well) due to the transitive dependency. Anyway, thanks for pointing this out.

PR #230, opened 2017-12-07
kill process group without calling getpgid

### LINKED ISSUE #229: If you SIGKILL the osqueryd process, the launcher doesn't come back up
... because the extension process is not cleaned up properly. 

Repro:

```
ps  xao pid,ppid,pgid,comm |grep kolide

13023     1 13023 /usr/local/kolide-corp/bin/launcher
33947 13023 33947 /usr/local/kolide-corp/bin/osqueryd
33956 33947 33947 /usr/local/kolide-corp/bin/osquery-extension.ext
```

kill the osqueryd process (33947). The launcher doesn't detect that osqueryd is gone and because the extension is still running, osqueryd fails to come back up.

groob:
If the original exec process is dead, then syscall.Getpgid fails and syscall.Kill is never called.
Updated the method to just call syscall.Kill directly.

Closes #229

### DIFF
--- a/osquery/runtime.go
+++ b/osquery/runtime.go
@@ -630,9 +630,6 @@ func (o *OsqueryInstance) Query(query string) ([]map[string]string, error) {
 
 // kill process group kills a process and all its children.
 func killProcessGroup(cmd *exec.Cmd) error {
-	pgid, err := syscall.Getpgid(cmd.Process.Pid)
-	if err == nil {
-		return syscall.Kill(-pgid, 15)
-	}
-	return errors.Wrap(err, "get PGID")
+	err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+	return errors.Wrapf(err, "kill process group %s", cmd.Process.Pid)
 }
--- a/osquery/runtime_test.go
+++ b/osquery/runtime_test.go
@@ -1,12 +1,18 @@
 package osquery
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"io"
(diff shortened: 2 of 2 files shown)


osquery/runtime.go
```diff
@@ -630,9 +630,6 @@ func (o *OsqueryInstance) Query(query string) ([]map[string]string, error) {
 
 // kill process group kills a process and all its children.
 func killProcessGroup(cmd *exec.Cmd) error {
-	pgid, err := syscall.Getpgid(cmd.Process.Pid)
-	if err == nil {
-		return syscall.Kill(-pgid, 15)
-	}
-	return errors.Wrap(err, "get PGID")
+	err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```
zwass: Do we know that the pgid is the same as the process pid?

groob: Yes, at least for posix this behavior is true when you apply 
https://github.com/kolide/launcher/blob/8a58d284a1f0152da0523374f53c8d7260323a96/osquery/runtime.go#L448-L449

groob: >  The setpgid() function shall either join an existing process group or create a new process group within the session of the calling process. The process group ID of a session leader shall not change. Upon successful completion, the process group ID of the process with a process ID that matches pid shall be set to pgid. As a special case, if pid is 0, the process ID of the calling process shall be used. Also, if pgid is 0, the process ID of the indicated process shall be used.

zwass: Okay cool thank you. There's already a lot of POSIX dependency in here.

Issue #241, opened 2018-01-24
Allow the client to publish the results of a server-dictated set of queries


marpaia:
## Overview

Let's say your organization instruments hosts with osquery and you would like to write a tool/app which operates on or displays information about the local host. It would be useful if the osquery gRPC server could provide a set of queries that were OK for the Launcher to execute and make the results available via some local interface.

#### Who can decide to execute a query?

Local apps may only consume query results. Which queries are executed, if any, can only be dictated by the osquery server.

#### Are the query results authenticated?

For the MVP implementation of this feature, no. Any local user on the system can see the results of the queries. Future iterations of this feature may allow the server to specify a per-query authentication mechanism.

PR #242, opened 2018-01-25
Client-side query result broadcast

marpaia:
This is a really naive implementation of #241. It has a lot of parts that are kind of rough, but it works end to end, so I wanted to put it up and gather feedback.


marpaia:
I don't really like this PR for the following reasons:

- I don't like adding a new method to `osquery.Extension` that isn't actually apart of the extension interface.
- I don't like having to share that method with the osquery runtime so that the runtime restarting can maintain an active instance
- I don't like using the functional option in the osquery runtime to enable query retrieval

For these reasons, I'm going to close this PR and think about how to do this more robustly.

PR #244, opened 2018-01-26
create macho table

groob:


marpaia:
This is cool. I wonder is there are more utilities in `"debug/macho"` that we can use to add more things to this table while we're doing this.

groob:
yeah it would definitely be nice to expose more data. I'll see what other columns are useful. 

It would also be interesting to call this `debug_info` and have a linux version with elf.

PR #249, opened 2018-02-02
Adding restart to launcher systemd unit file

synapsis2112:
Built new package and verified systemd unit file text was accurate. This will allow the launcher to overcome an intermittent issue noticed with systemd-resolved when first starting and throwing a "connection refused" error upon trying to resolve a domain by simply retrying after a few seconds.


synapsis2112:
Changed from `systemctl start` to `systemctl restart` in the installer script so we can pick up any process configuration changes in addition to the unit file changes which get loaded with `systemctl daemon-reload`

PR #250, opened 2018-02-07
add gcp pkgbuild

groob:
Not intending to merge but putting it up for discussion. 

The idea behind this is that the cloudbuild steps below can be used to build the binary then tar it to a storage bucket. 

Once that is done, a second build job can be triggered to create mac or linux packages based on the binary. See https://github.com/kolide/gcp-pkgbuild for the mac package part. 

```
steps:
    - name: 'gcr.io/cloud-builders/go'
      entrypoint: 'sh'
      args: ['-c', '/workspace/build.sh']
      env: ['PROJECT_ROOT=github.com/kolide/launcher']

    - name: 'gcr.io/cloud-builders/gsutil'
      entrypoint: 'bash'
      args:
      - '-c'
      - '/workspace/tarbin.sh'

     # a third step here could kick off the package build
     # or the pkgbuild steps could be copy pasted into here
     # or something could watch the bucket.  
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+. /builder/prepare_workspace.inc
+prepare_workspace || exit
+make deps
+make xp
--- a/simulator/testdata/bad_symlink/symlink.yaml
+++ b/simulator/testdata/bad_symlink/symlink.yaml
@@ -1,1 +1,0 @@
-foo/bar/bing/bang/boom
--- a/tarbin.sh
+++ b/tarbin.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+version="$(git rev-parse --short HEAD)"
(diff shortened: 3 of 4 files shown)


simulator/testdata/bad_symlink/symlink.yaml
```diff
@@ -1 +0,0 @@
-foo/bar/bing/bang/boom
```
groob: had to remove this, it _really_ didn't like symlinks like this for the upload. Would have to work around.

tarbin.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
marpaia: can you move this to a `make` command

build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
marpaia: can you move this to a `make` command

PR #255, opened 2018-02-09
Add cert pinning support to package tooling

zwass:


zwass:
I tested this successfully on macOS, Ubuntu16 and Centos7.

Issue #231, opened 2017-12-08
Allow providing a custom TLS CA to the package builder


stouset:
For our infrastructure, we use a custom CA that's not necessarily installed into the default system trust store location for the underlying OS. Most applications that deal with TLS allow providing a path to trusted CA.

`osqueryd` allows this through the `--tls_server_certs` flag. Would it be possible to add support for this flag in the package builder so it can be provided to the underlying osqueryd binary?

monkykap:
@stouset I don't think there is support in grpc implementation to handle a specified cert path at the moment. There is a section in the packaging.go file with postinstall script to handle systemctl/service. I think this is a good place to grab your bundle and move it into the appropriate default directory as trusted CA. `postInstallLauncherContents := `#/bin/bash
sudo service launcher restart`.

zwass:
This is possible through our custom TLS dialer (https://github.com/kolide/launcher/blob/master/cmd/launcher/launcher.go#L466)... We can pass root CAs through the `tls.Config` (https://golang.org/src/crypto/tls/common.go?#L414). I'm going to implement this.

PR #257, opened 2018-02-21
Add support for specifying root CAs

zwass:
First step on the path to #231.


zwass:
I manually tested that the launcher will connect when the appropriate root cert is provided, and will not connect when it does not have the appropriate root.

PR #261, opened 2018-02-22
expose tables through a standard osquery extension

groob:
Replace development extension with `launcher.ext` which does the same thing but implies others can use it outside of local development purposes.


groob:
Docks and maybe a blog post would be great. I'll merge this and then find the time to work on usage :)

PR #263, opened 2018-02-22
Add packaging support for specifying root CAs

### LINKED ISSUE #231: Allow providing a custom TLS CA to the package builder

zwass:
Closes #231.

### DIFF
--- a/tools/packaging/packaging.go
+++ b/tools/packaging/packaging.go
@@ -31,13 +31,13 @@ type PackagePaths struct {
 
 // CreatePackages will create a launcher macOS package. The output paths of the
 // packages are returned and an error if the operation was not successful.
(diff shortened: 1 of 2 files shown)


tools/packaging/packaging.go
```diff
@@ -141,6 +141,15 @@ func CreateLinuxPackages(osqueryVersion, hostname, secret string, insecure, inse
 		}
 	}
 
+	rootPEMPath := ""
```
groob: nit: I would prefer to use the zero value then initializing an empty string like this in Go code.

zwass:
CI passed but I cancelled it while it waited for SSH so it would not use up resources.

I manually tested this by creating packages with and without the appropriate roots in the PEM.

PR #283, opened 2018-04-16
Updating circle ci to use go 1.10.1 during tests

wstewartii:


tools/builders/launcher-builder/1.10.1/cloudbuild.yml
```diff
@@ -0,0 +1,29 @@
+steps:
+- name: 'gcr.io/cloud-builders/git'
```
groob: I don't think we need the clone steps here. It's just a Dockerfile that is being submitted. Let's have the makefile use `gcloud container builds submit --config cloudbuild.yaml .`.

wstewartii: container can't see the Dockerfile without cloning the repo.

groob: Why not? it should get uploaded when you submit.

wstewartii: actually looks like I may be able to pass "." to get it to work.

PR #287, opened 2018-04-25
Expand osquery types

marpaia:
This PR expands the types in `osquery/osquery.go`.

### DIFF
--- a/osquery/osquery.go
+++ b/osquery/osquery.go
@@ -19,6 +19,9 @@ type QueryContent struct {
 // Queries is a helper which represents the format of a set of queries in a pack
 type Queries map[string]QueryContent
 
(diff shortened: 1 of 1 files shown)


osquery/osquery.go
```diff
@@ -51,14 +54,24 @@ type OsqueryConfig struct {
 type OsqueryResultLog struct {
 	Name           string `json:"name"`
 	HostIdentifier string `json:"hostIdentifier"`
-	UnixTime       string `json:"unixTime"`
+	UnixTime       int    `json:"unixTime"`
```
marpaia: This is an osquery 3.0 change I believe. Epoch and Counter would also have been a 3.0 change, but this struct didn't have them before. IMO we should just make these ints.

PR #292, opened 2018-06-05
add macOS appstore autoupdate table

groob:
* not complete
* looking for feedback on schema
* blows up on 10.14, we'll have to switch on the platform version to return results from two different sources. 

```
osquery> select * from kolide_macos_software_update;
+--------------------+--------------------+---------------------+---------------------------+------------------+------------------+
| autoupdate_managed | autoupdate_enabled | background_download | app_store_apps_autoupdate | macos_autoupdate | critical_updates |
+--------------------+--------------------+---------------------+---------------------------+------------------+------------------+
| 0                  | 1                  | 1                   | 1                         | 1                | 1                |
+--------------------+--------------------+---------------------+---------------------------+------------------+------------------+
```

I'd like to add a few more columns for last_successful_update and for catalog_url, since there's different sources updates can be pulled from.


rmcvey:
This is fantastic, we are currently pulling this from various plist values and it is kind of a pain.

couple thoughts:
- `macos_autoupdate` seems a little verbose when the table is specifying that it is for `macos`, `os_autoupdate` might be sufficient.
- separate but related: a lot of these properties are relevant on Windows systems as well (and map directly to Windows settings). I'm not suggesting that you implement the Windows checks; however, it would be ideal to name the table and columns in a way that allows for cross-platform usage - even if a column or two is null on one platform or the other.

Thanks!

groob:
Updated to schema a bit based on feedback:

```
osquery> select * from kolide_macos_software_update;
+--------------------+--------------------+----------+-------------+------------+------------------+
| autoupdate_managed | autoupdate_enabled | download | app_updates | os_updates | critical_updates |
+--------------------+--------------------+----------+-------------+------------+------------------+
| 0                  | 1                  | 1        | 1           | 1          | 1                |
+--------------------+--------------------+----------+-------------+------------+------------------+
```

working on adding `last_sync_date` and `catalog_url`.

groob:
Table is now complete, and works with 10.14:

```
osquery> select * from kolide_macos_software_update;
+--------------------+--------------------+----------+-------------+------------+------------------+---------------------------------+
| autoupdate_managed | autoupdate_enabled | download | app_updates | os_updates | critical_updates | last_successful_check_timestamp |
+--------------------+--------------------+----------+-------------+------------+------------------+---------------------------------+
| 0                  | 1                  | 1        | 1           | 1          | 1                | 1528531761                      |
+--------------------+--------------------+----------+-------------+------------+------------------+---------------------------------+
```

osquery/sus_darwin.go
```diff
@@ -0,0 +1,151 @@
...
+		*doesBackgroundDownload = 1;
+	}
+
+	val = [manager doesAppStoreAutoUpdates];
+	if (val) {
+		*doesAppStoreAutoUpdates = 1;
+	}
+
+	// before 10.13 the method was called doesOSXAutoUpdates, since 10.14 it's called doesMacOSAutoUpdate.
+	if (os_version >=14) {
```
marpaia: mega nit: there is a space before `>=` but not after

osquery/sus_darwin.go
```diff
@@ -0,0 +1,151 @@
...
+	return;
+}
+*/
+import "C"
+import (
+	"context"
+	"fmt"
+	"strconv"
+
+	osquery "github.com/kolide/osquery-go"
```
marpaia: you can remove the osquery rename here

PR #295, opened 2018-06-10
reorganize extension structure

groob:
Move table plugin implementations into a new subdirectory structure. 
Move `runtime` implementation into its own package. 

Feedback welcome. IMO as the code for tables grows, it's easier to reason about each table individually. I also felt that `runtime` should be separate from the launcher extension. The runtime package was already completely decoupled from the rest of the extension.


marpaia:
I'm down with breaking apart the runtime from the tables but I think one package per table is too much. Maybe we can start with a package for just tables and see how that grows?

groob:
Back to a single table package:

```
osquery/
├── extension.go
├── extension_test.go
├── osquery.go
├── platform.go
├── platform_test.go
├── runtime
│   ├── runtime.go
│   └── runtime_test.go
└── table
    ├── SUSharedPrefs.h
    ├── best_practices.go
    ├── best_practices_test.go
    ├── email_addresses.go
    ├── kolide_vulnerabilities.go
    ├── launcher_identifier.go
    ├── launcher_info.go
    ├── macho.go
    ├── mdfind_darwin.go
    ├── mdm.go
    ├── mdm_test.go
    ├── munki_darwin.go
    ├── platform_tables.go
    ├── platform_tables_darwin.go
    ├── spotlight.go
    ├── sus_darwin.go
    └── table.go
```

osquery/table/sus_darwin.go
```diff
@@ -75,7 +75,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func MacUpdate(client *osquery.ExtensionManagerClient) *table.Plugin {
+func MacOS(client *osquery.ExtensionManagerClient) *table.Plugin {
```
marpaia: Did you mean to change this back?

PR #296, opened 2018-06-10
refactor service package to use per-method files

groob:
Refactored the go-kit code to be more accessible, by grouping all the related method code together. 
Also cleaned up the uuid context helpers which now exist in kolide/kit.


groob:
I agree, things are somewhat verbose, but this PR isn't meant to address that, just to make future refactoring more ergonomic. 

I'm interested in keeping the existing structure, but iterating on `transport/twirp` and also on some streaming implementation of the extension in the near future.

PR #300, opened 2018-06-18
add airdrop table

groob:
Query Airdrop discovery for the primary user.

```
osquery> select * from kolide_airdrop_preferences;
+----------+-------------+
| username | discover_by |
+----------+-------------+
| victor   | Everyone    |
+----------+-------------+

```

### DIFF
--- a/osquery/table/corefoundation_helpers_darwin.go
+++ b/osquery/table/corefoundation_helpers_darwin.go
@@ -11,12 +11,16 @@ import (
 	"unsafe"
 )
 
+// Functions with "Create" or "Copy" in the name return references that need to
+// be CFReleased. See
+// https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029
+
 func copyPreferenceValue(key, domain, username string) interface{} {
-	keyCFString := CFStringRef(key)
+	keyCFString := cFStringRef(key)
 	defer C.CFRelease((C.CFTypeRef)(keyCFString))
-	domainCFString := CFStringRef(domain)
+	domainCFString := cFStringRef(domain)
 	defer C.CFRelease((C.CFTypeRef)(domainCFString))
-	usernameCFString := CFStringRef(username)
+	usernameCFString := cFStringRef(username)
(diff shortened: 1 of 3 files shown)


terracatta:
Once merged would love to ship to customers ASAP so we can fix the bad practice.

osquery/table/corefoundation_helpers_darwin.go
```diff
@@ -0,0 +1,76 @@
...
+*/
+import "C"
+import (
+	"fmt"
+	"reflect"
+	"unsafe"
+)
+
+func copyValue(key, domain, username string) C.CFPropertyListRef {
+	return C.CFPreferencesCopyValue(
```
zwass: Do we need an associated `CFRelease` with this? https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029

osquery/table/corefoundation_helpers_darwin.go
```diff
@@ -0,0 +1,76 @@
...
+		CFStringRef(key),
+		CFStringRef(domain),
+		CFStringRef(username),
+		C.kCFPreferencesAnyHost,
+	)
+}
+
+func CFStringRef(s string) C.CFStringRef {
+	return C.CFStringCreateWithCString(C.kCFAllocatorDefault, C.CString(s), C.kCFStringEncodingUTF8)
```
zwass: I think by the same "Create Rule" we need to `CFRelease` this string.

osquery/table/corefoundation_helpers_darwin.go
```diff
@@ -0,0 +1,83 @@
...
+
+	val := C.CFPreferencesCopyValue(
+		keyCFString, domainCFString, usernameCFString, C.kCFPreferencesAnyHost,
+	)
+	defer C.CFRelease((C.CFTypeRef)(val))
+	return goValueFromCFPlistRef(val)
+}
+
+// CFStringRef returns a C.CFStringRef which must be released with C.CFRelease
+func CFStringRef(s string) C.CFStringRef {
```
zwass: should this and the below functions be private?

groob: Yeah, no reason for them to be public, I copied them from my own gist i found somewhere.

PR #305, opened 2018-06-21
add entrypoint for printing preference in a user context.

groob:
Allows querying user defaults which can only be read as the user.

Documentation: https://github.com/groob/launcher/blob/371527d702fbe8c33cf7c481e52d3d6d9864e3cf/docs/architecture/2018-06-21-exec-to-drop-privilege.md

### DIFF
--- a/osquery/table/airdrop_darwin.go
+++ b/osquery/table/airdrop_darwin.go
@@ -41,14 +41,15 @@ func (t *airdropTable) generateAirdrop(ctx context.Context, queryContext table.Q
 		username = t.primaryUser
 	}
 
-	discover := "Unknown"
-	if val, ok := copyPreferenceValue("DiscoverableMode", "com.apple.sharingd", username).(string); ok {
-		discover = val
+	out, err := execPreferenceAsUser(ctx, username, "DiscoverableMode", "com.apple.sharingd")
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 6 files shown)


osquery/table/airdrop_darwin.go
```diff
@@ -41,14 +41,15 @@ func (t *airdropTable) generateAirdrop(ctx context.Context, queryContext table.Q
 		username = t.primaryUser
 	}
 
-	discover := "Unknown"
-	if val, ok := copyPreferenceValue("DiscoverableMode", "com.apple.sharingd", username).(string); ok {
-		discover = val
+	out, err := execPreferenceAsUser(ctx, username, "DiscoverableMode", "com.apple.sharingd")
```
marpaia: `username = q.Constraints[0].Expression` this isn't going to work if you, for example, join this table against the users table with more than 1 user.

groob: That is true. I don't know if joining against multiple users is ever desirable here, and would end up in a large number of sequential exec calls.

I don't know if it's the right decision, but I thought this query should only work for a single user. The default is to query the primary user anyway.

groob:
I'm getting a number of failures through this, going to refactor to pass the username as an arg and call setuid in the exec, see if that makes a difference.

PR #311, opened 2018-06-29
create a standalone grpc extension

groob:
this is my windows experiment.


cmd/grpc.ext/grpc.go
```diff
@@ -0,0 +1,137 @@
...
+		certPins,
+		rootPool,
+		logger,
+	)
+	remote := service.New(conn, level.Debug(logger))
+
+	extOpts := grpcext.ExtensionOpts{
+		EnrollSecret:    enrollSecret,
+		Logger:          logger,
+		LoggingInterval: 30 * time.Second,
```
marpaia: ideally this could be configurable.

cmd/grpc.ext/grpc.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := server.Run(); err != nil {
+		logger.Fatal("err", err)
+	}
+}
+
+type queryier struct {
+	client *osquery.ExtensionManagerClient
+}
+
+func (q *queryier) Query(query string) ([]map[string]string, error) {
```
marpaia: Copy paste? This doesn't seem to be used.

groob: It's used on line 37. It's required to make the RequestEnrollment method add initial enrollment data from the client. 

Launcher has this method on the osquery runner.

cmd/grpc.ext/grpc.go
groob: 👍

PR #317, opened 2018-07-06
Add Remote Shell / Control

loganmac:
Continuation of @marpaia's earlier PR


control/wsrelay/client.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// construct the URL to connect to
+	u := url.URL{
+		Scheme: scheme,
+		Host:   brokerAddr,
+		Path:   path,
+	}
+
+	// set the secret in the query params
```
directionless: Where does this drop that secret?

putting it into the URL parameters makes it visible to logs and various intermediates.

control/control.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return c.doWithHeaders(verb, path, params, headers)
+}
+
+func (c *Client) doWithHeaders(verb, path string, params interface{}, headers map[string]string) (*http.Response, error) {
+	var bodyBytes []byte
+	var err error
+	if params != nil {
+		bodyBytes, err = json.Marshal(params)
```
groob: why marshal into a byte slice and then create a buffer when you can use the encoder?

control/wsrelay/client.go
```diff
@@ -0,0 +1,74 @@
...
+// Client is a websocket client
+type Client struct {
+	conn *websocket.Conn
+}
+
+// NewClient creates a new websocket client that can be interrupted
+// via SIGINT
+func NewClient(brokerAddr, path string, useTLS bool, insecure bool) (*Client, error) {
+	// determine the scheme
+	var scheme string
```
groob: set a default value and omit an else.

control/control.go
loganmac: question for @marpaia lol

marpaia: This was copy paste from the Fleet API client.

PR #321, opened 2018-07-12
Added os switch for package-builder

SeppBender:
That makes building package-builder from linux possible.


SeppBender:
The CLA Agreement is empty.

groob:
We're refactoring the package-builder code and it should address this issue.

PR #329, opened 2018-07-20
Add actors for individual components of launcher

loganmac:
This allows individual pieces of Launcher to be started and stopped with all of the other components and helps ensure the proper shutdown of started goroutines.


cmd/launcher/control.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/boltdb/bolt"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/actor"
+	"github.com/kolide/launcher/pkg/control"
+	kolideLog "github.com/kolide/launcher/pkg/log"
+	"github.com/pkg/errors"
+)
+
+func createControl(ctx context.Context, db *bolt.DB, logger *kolideLog.Logger, opts *options) (*actor.Actor, error) {
```
marpaia: IMO this should accept a `github.com/go-kit/kit/log.Logger`. The swap logger functionality was abstracted into `github.com/kolide/kit`: https://github.com/kolide/kit/blob/45f7c9242522e5dd787e33bdec670fd8aec4ed60/logutil/logutil.go#L45

cmd/launcher/control.go
```diff
@@ -0,0 +1,43 @@
...
+		controlOpts = append(controlOpts, control.WithDisableTLS())
+	}
+	controlClient, err := control.NewControlClient(db, opts.controlServerURL, controlOpts...)
+	if err != nil {
+		logger.Fatal(errors.Wrap(err, "creating control client"))
+	}
+
+	return &actor.Actor{
+		Execute: func() error {
+			println("\ncontrol started\n")
```
marpaia:
```
level.Info(logger).Log("msg", "control started")
```

cmd/launcher/control.go
```diff
@@ -0,0 +1,43 @@
...
+			controlClient.Start(ctx)
+			return nil
+		},
+		Interrupt: func(err error) {
+			println("\ncontrol interrupted\n")
```
marpaia:
```
level.Info(logger).Log("msg", "control interrupted")
```

cmd/launcher/enabler.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, err
+	}
+
+	// create a variable to hold the stop function that the updater returns,
+	// so that we can pass it into the interrupt function
+	var stop func()
+
+	return &actor.Actor{
+		Execute: func() error {
+			println("\nupdater started\n")
```
marpaia:
```
level.Info(logger).Log("msg", "updater started")
```

cmd/launcher/enabler.go
```diff
@@ -0,0 +1,82 @@
...
+			stop, err = updater.Run(tuf.WithFrequency(config.AutoupdateInterval))
+			if err != nil {
+				return errors.Wrap(err, "running updater")
+			}
+			return nil
+		},
+		Interrupt: func(err error) {
+			println("\nupdater interrupted\n")
```
marpaia:
```
level.Info(logger).Log("msg", "updater interrupted")
```

cmd/launcher/extension.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+			ext.Start()
+
+			// TODO: remove when underlying libs are refactors
+			// everything exits right now, so block this actor on the context finishing
+			<-ctx.Done()
+			return nil
+		},
+		Interrupt: func(err error) {
+			println("\nextension interrupted\n")
```
marpaia:
```
level.Info(logger).Log("msg", "extension interrupted")
```

cmd/launcher/extension.go
```diff
@@ -0,0 +1,156 @@
...
+	osqueryLogger := &kolidelog.OsqueryLogAdapter{Logger: level.Debug(log.With(logger, "component", "osquery"))}
+
+	var runner *runtime.Runner
+
+	return &extension{
+		// add the runtime to the extension
+		Runner: runner,
+		// and the methods for starting and stopping the extension
+		Execute: func() error {
+			println("\nextension started\n")
```
marpaia: I would suggest putting this in between `ext.Start()` and `<-ctx.Done()`. And also changing to:

```
level.Info(logger).Log("msg", "extension started")
```

cmd/launcher/extension.go
```diff
@@ -0,0 +1,156 @@
...
+			_, invalid, err := ext.Enroll(ctx)
+			if err != nil {
+				logger.Fatal("err", errors.Wrap(err, "enrolling host"))
+			}
+			if invalid {
+				logger.Fatal(errors.Wrap(err, "invalid enroll secret"))
+			}
+			ext.Start()
+
+			// TODO: remove when underlying libs are refactors
```
marpaia: :+1: We could also make `ext.Start` take a context and have it block on this channel if you want `Start()` to block.

cmd/launcher/extension.go
```diff
@@ -0,0 +1,160 @@
...
+	// create the extension
+	ext, err := osquery.NewExtension(launcherClient, db, extOpts)
+	if err != nil {
+		return nil, errors.Wrap(err, "starting grpc extension")
+	}
+
+	// create the logging adapter for osquery
+	osqueryLogger := &kolidelog.OsqueryLogAdapter{Logger: level.Debug(log.With(logger, "component", "osquery"))}
+
+	var runner *runtime.Runner
```
groob: can you add a comment explaining why a nil runner is added to the struct, rather than creating the runner here? 

I understand after reading the code, but it would help with readability.

cmd/launcher/main.go
```diff
@@ -270,186 +266,102 @@ func main() {
...
+			config,
 		)
 		if err != nil {
 			logger.Fatal(err)
 		}
-		defer stopLauncher()
+		runGroup.Add(launcherUpdater.Execute, launcherUpdater.Interrupt)
 	}
 
+	// start the rungroup
```
groob: starting the group should block. 
Add the Signal handler as an actor as well instead of launching this goroutine. 

Something like this:

```
			var g run.Group
			{
				// this actor handles an os interrupt signal and terminates the server.
				sig := make(chan os.Signal, 1)
				g.Add(func() error {
					signal.Notify(sig, os.Interrupt)
					select {
					case <-sig:
						level.Info(logger).Log("msg", "beginnning shutdown")
						return nil
					}
				}, func(err error) {
					level.Info(logger).Log("msg", "interrupted", "err", err)
					close(sig)
				})
			}
```

cmd/launcher/extension.go
```diff
@@ -0,0 +1,160 @@
...
+
+// extension is the osquery extension and osquery runtime
+type extension struct {
+	*runtime.Runner
+	Execute   func() error
+	Interrupt func(err error)
+}
+
+// TODO: the extension, runtime, and client are all kind of entangled here. Untangle the underlying libraries and separate into units
+func createExtension(ctx context.Context, rootDirectory string, db *bolt.DB, logger log.Logger, opts *options) (*extension, error) {
```
groob: maybe `createExtensionRuntime`?

marpaia:
This looks good to me. I'm cool with merging this assuming that it's been pretty aggressively tested.

PR #331, opened 2018-07-26
Add missing control flags to packaging

loganmac:
I missed adding the new flag to disable TLS in the control component in the packaging tooling.

### DIFF
--- a/cmd/package-builder/package-builder.go
+++ b/cmd/package-builder/package-builder.go
@@ -79,6 +79,11 @@ func runMake(args []string) error {
 			env.String("CONTROL_HOSTNAME", ""),
 			"the value that should be used when invoking the launcher's --control_hostname flag",
 		)
+		flDisableControlTLS = flagset.Bool(
+			"disable_control_tls",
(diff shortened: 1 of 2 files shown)


cmd/package-builder/package-builder.go
```diff
@@ -79,6 +79,11 @@ func runMake(args []string) error {
 			env.String("CONTROL_HOSTNAME", ""),
 			"the value that should be used when invoking the launcher's --control_hostname flag",
 		)
+		flDisableControlTLS = flagset.Bool(
+			"disable_control_tls",
+			env.Bool("DISABLE_CONTROL_TLS", false),
```
groob: should this be true for now at least? Lots of people building launcher packages, but the control server is kolide specific atm.

loganmac: Control is disabled by default I believe, so no need to have the TLS disabled by default for it

groob: ignore this comment.

PR #333, opened 2018-07-26
Add macOS uninstall instructions

0xmachos:
I've run `sudo find / -name 'Kolide'` on my system, with the Kolide launcher installed, the only two directories which are present appear to be:
- `$HOME/Applications/Kolide.app`
- `$HOME/Library/Application Support/Kolide`

`rm -r`ing both of these directories constitutes a successful uninstall as far as I can tell.

### DIFF
--- a/docs/uninstalling.md
+++ b/docs/uninstalling.md
@@ -44,16 +44,26 @@ Based on the configurations used when the Launcher package was created, the spec
 Directories:
 - `$HOME/Applications/Kolide.app`
 - `$HOME/Library/Application Support/Kolide`
+- `/usr/local/kolide`
+- `/var/kolide`
+- `/etc/kolide`
+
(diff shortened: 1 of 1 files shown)


cwhits:
Is this still true for the latest version of Launcher? Here's what my package is creating:
image

0xmachos:
@cwhits Nope. Thanks for the correction!

0xmachos:
Added the following:

Directories:
- `/usr/local/kolide`
- `/var/kolide`
- `/etc/kolide`

Files:
- `/Library/LaunchDaemons/com.kolide.launcher.plist`

0xmachos:
@directionless I've added the `launchctl unload` commands. I don't have time to work out the other path variants so someone else will have to do that.

PR #336, opened 2018-07-31
Reduce shell check interval, increase error reporting

loganmac:
The default shell interval was too long and would often close due to timeout before even connecting. Also there were a few spots we forgot to log errors.

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -348,6 +348,9 @@ func main() {
 		level.Info(logger).Log("msg", "beginnning shutdown")
 		return nil
 	}, func(err error) {
+		if err != nil {
+			level.Info(logger).Log("err", err)
+		}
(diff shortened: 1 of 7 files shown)


cmd/launcher/main.go
```diff
@@ -348,6 +348,9 @@ func main() {
 		level.Info(logger).Log("msg", "beginnning shutdown")
 		return nil
 	}, func(err error) {
+		if err != nil {
+			level.Info(logger).Log("err", err)
```
marpaia: maybe include a `msg` here with where the error occurred? or just include `err` in the next log (and it can just be nil when it's nil)?

loganmac: I think if it's nil it panics, but good idea about the msg

loganmac: Sorry, I was wrong! I'll do that.

PR #338, opened 2018-08-06
Fix - MacOS distribution package

m6a-UdS:
Some MDMs (like SimpleMDM) require the OSX package (`.pkg`) to contain a "distribution" file, and therefore require the command `productbuild` to be run after `pkgbuild` in order for the package to be accepted by SimpleMDM.

I was unsure whether it would be better to output both versions (with and without the distribution file), the choice is done at the line 553, but could be modified to include both.

I can confirm it works and gets distributed and installed fine with SimpleMDM, but it would be a good idea to also check with other MDMs.

### DIFF
--- a/pkg/packaging/packaging.go
+++ b/pkg/packaging/packaging.go
@@ -512,7 +512,11 @@ func CreateMacPackage(
 		return "", errors.Wrap(err, "could not create final output directory for package")
 	}
 
-  distributionFile, err := os.Create(filepath.Join(packageRoot, "distribution.xml"))
+	distributionFileName := "distribution.xml"
+	distributionFilePath := filepath.Join(outputPathDir, distributionFileName)
+  distributionFile, err := os.Create(distributionFilePath)
+	distrOutputFile := fmt.Sprintf("launcher-darwin-%s-distribution.pkg", packageVersion)
+	distrOutputPath := filepath.Join(outputPathDir, distrOutputFile)
 	distributionOpts := &distributionTemplateOptions{
 		PackageVersion:    packageVersion,
 		PackageFileName:   outputFile,
@@ -534,7 +538,19 @@ func CreateMacPackage(
 		return "", errors.Wrap(err, "could not create macOS package")
 	}
 
-	return outputPath, nil
+	// Add the distribution file
+	err = productbuild(
+		outputPathDir,
+		distributionFileName,
+		outputFile,
+		macPackageSigningKey,
+		distrOutputPath,
+	)
+	if err != nil {
+		return "", errors.Wrap(err, "could not add the distribution file")
+	}
+
+	return distrOutputPath, nil
 }
 
 // systemdTemplateOptions is a struct which contains dynamic systemd
@@ -762,6 +778,29 @@ sleep 5
 		return errors.Wrap(err, "not able to parse postinstall template")
 	}
 	return t.ExecuteTemplate(w, "postinstall", options)
+}
+
+func productbuild(packageDir, distributionFile, packageFile, macPackageSigningKey, outputPath string) error {
+	args := []string{
+		"--distribution", distributionFile,
+		"--package-path", packageFile,
+	}
+
+	if macPackageSigningKey != "" {
+		args = append(args, "--sign", macPackageSigningKey)
+	}
+
+	args = append(args, outputPath)
+  fmt.Printf("productbuild %s", args)
+	cmd := exec.Command("productbuild", args...)
+	cmd.Dir = packageDir
+	stderr := new(bytes.Buffer)
+	cmd.Stderr = stderr
+	if err := cmd.Run(); err != nil {
+		io.Copy(os.Stderr, stderr)
+		return err
+	}
+	return nil
 }
 
 // pkgbuild runs the following pkgbuild command:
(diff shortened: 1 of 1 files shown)


erikng:
If it works with SimpleMDM it will work with others.

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
...
+<installer-gui-script minSpecVersion="1">
+    <pkg-ref id="com.launcher.launcher"/>
+    <options customize="never" require-scripts="false"/>
+    <choices-outline>
+        <line choice="default">
+            <line choice="com.launcher.launcher"/>
+        </line>
+    </choices-outline>
+    <choice id="default"/>
+    <choice id="com.launcher.launcher" visible="false">
```
marpaia: This should be parameterized with the `identifier` option so that it's `com.{{.Identifier}}.launcher`

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
...
+	PackageVersion    string
+	PackageFileName   string
+}
+
+// renderDistributionFile renders a distribution file to add to launcher's template.
+func renderDistributionFile(w io.Writer, options *distributionTemplateOptions) error {
+	distributionTemplate :=
+		`<?xml version="1.0" encoding="utf-8"?>
+<installer-gui-script minSpecVersion="1">
+    <pkg-ref id="com.launcher.launcher"/>
```
marpaia: This should be parameterized with the `identifier` option so that it's `com.{{.Identifier}}.launcher`

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
...
+    <options customize="never" require-scripts="false"/>
+    <choices-outline>
+        <line choice="default">
+            <line choice="com.launcher.launcher"/>
```
marpaia: This should be parameterized with the `identifier` option so that it's `com.{{.Identifier}}.launcher`

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
...
+        </line>
+    </choices-outline>
+    <choice id="default"/>
+    <choice id="com.launcher.launcher" visible="false">
+        <pkg-ref id="com.launcher.launcher"/>
```
marpaia: This should be parameterized with the `identifier` option so that it's `com.{{.Identifier}}.launcher`

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
...
+    </choice>
+    <pkg-ref id="com.launcher.launcher" version="{{.PackageVersion}}" onConclusion="none">{{.PackageFileName}}</pkg-ref>
```
marpaia: This should be parameterized with the `identifier` option so that it's `com.{{.Identifier}}.launcher`

pkg/packaging/packaging.go
```diff
@@ -506,11 +506,25 @@ func CreateMacPackage(
...
 	if err != nil {
 		return "", errors.Wrap(err, "could not create final output directory for package")
 	}
 
+	distributionFileName := "distribution.xml"
+	distributionFilePath := filepath.Join(outputPathDir, distributionFileName)
+	distributionFile, err := os.Create(distributionFilePath)
+	distrOutputFile := fmt.Sprintf("launcher-darwin-%s-distribution.pkg", packageVersion)
+	distrOutputPath := filepath.Join(outputPathDir, distrOutputFile)
+	distributionOpts := &distributionTemplateOptions{
```
marpaia: Include `Identifier: identifier` in this struct.

pkg/packaging/packaging.go
```diff
@@ -784,6 +838,38 @@ func pkgbuild(packageRoot, scriptsRoot, identifier, version, macPackageSigningKe
 	return nil
 }
 
+// distributionTemplateOptions is a struct which contains dynamic distribution
+// file parameters that will be rendered into a template.
+type distributionTemplateOptions struct {
```
marpaia: Include `Identifier string` in this struct

zwass:
Is it sufficient to include the `distribution.xml` in the generated package? Or must we also run `productbuild`?

m6a-UdS:
No, I tried just adding `distribution.xml` at the root of the original package, but it does not work with SimpleMDM. I reproduced these instructions in order to make it work: https://medium.com/the-c4-dev/building-an-installer-part-3-d4d145ac2bbc

erikng:
Yes you must run productbuild to get a distribution package. Otherwise it is just a flat package.

Thanks,
Erik Gomez
________________________________

zwass:
sgtm. Thanks @m6a-UdS and @erikng.

groob:
To solve this all you have to do is take the initial package and run `productbuild` to build it. I don't think the distribution file is necessary. We can just implement pkgbuild output to a temp location and take that path to run productbuild in a followup step. 

Long term, I think that would be more reliable than maintaining the distribution.xml ourselves. 

Thoughts?

groob:
Closing due to the refactoring of all the package tools.

PR #339, opened 2018-08-24
store logs to file on windows.

groob:


groob:
I'm closing this. My plan is to update the logger on windows to use the windows event log. Everything else is not idiomatic and leads to full disks.

PR #344, opened 2018-08-29
add flare binary

groob:
Updated:

Adds a launcher subcommand `launcher flare` which builds a debug report running various tests on the system, and saving the output to a `.tar.gz` file which the user can then pass to support. 

Currently the following tests are run:
- connect to grpc server and try serveral endpoints (with a fake node key). 
- launch an osquery instance and connect to it with osquery-go, running a process info query.
- connect to the notary server and perform a healthcheck. 

Additional tests will be added, including os specific ones. This is just the start, but provides a consistent UI to start debugging launcher issues.


directionless:
Yes. So happy to see this getting started. I'm thinking it should be build into launcher -- `launcher flare` so we don't need to ship people another binary. But that's kinda a nit

groob:
Agreed. A subcommand will also be more representative in case of binary whitelisting and related issues.

groob:
I'm also thinking the output should all go to a tar.gz which the user is prompted to upload to us.

directionless:
Yes. Tarball gives us a simple way to have different files of output

cmd/launcher/flare.go
```diff
@@ -0,0 +1,331 @@
...
+	fileOnly outputDestination = iota
+	stdout
+)
+
+func fatal(w io.Writer, err error) {
+	output(w, stdout, "error: %s\n", err)
+	os.Exit(1)
+}
+
+func output(w io.Writer, printTo outputDestination, f string, a ...interface{}) error {
```
directionless: What is printTo if not "stdout" ?

cmd/launcher/flare.go
```diff
@@ -0,0 +1,331 @@
...
+		return errors.Wrapf(err, "creating osquerygo client with socket path %s", socketPath)
+	}
+	defer client.Close()
+
+	logger.Log(
+		"flare", "reportOsqueryProcessInfo",
+		"msg", "running query with osquery-go",
+	)
+
+	const query = `select * from processes where name like '%osqueryd%' OR name like '%launcher%';`
```
directionless: Doing the data analysis via osquery is a clever way to get the cross platform ability. +100

cmd/launcher/flare.go
```diff
@@ -0,0 +1,331 @@
...
+func output(w io.Writer, printTo outputDestination, f string, a ...interface{}) error {
+	if printTo == stdout {
+		fmt.Printf(f, a...)
+	}
+
+	_, err := fmt.Fprintf(w, f, a...)
+	return err
+}
+
+// starts an osqueryd runtime, and then connects an osquery client and runs queries to check and log process info.
```
directionless: Building this, and running `./launcher flare`, my logs show:
```
reportOsqueryProcessInfo error: creating osquery instance for process info query: starting instance: osqueryd not supplied and not found: exec: "osqueryd": executable file not found in $PATH
```

groob: do you not have an osqueryd in your path?

groob: I will try to first find a binary is `/usr/local/kolide/bin/osqueryd`, since that's the default in our packages. 
Its not full proof with custom launcher builds though, so we can parametrize it.

directionless: Correct. I don't have an osquery in my path.

cmd/launcher/flare.go
groob: fileOnly. 

The way this output helper works is it lets me decide if I want to send some output to the users terminal, or if I want to silently write it to the log file in the report. I want to capture everything in the log file, but the stdout output should be a bare minimum.

cmd/launcher/flare.go
groob: We probably want a simple `ps` output as well, but here I wanted to see if we can stand up and communicate with an osquery process over the extension.

cmd/launcher/flare.go
directionless: Ah. I'd reasonable about that as a boolean. But that makes sense

groob: boolean could work, I was thinking there might be a third case down the road, but I also like functions that look like this:

`foo(true, true, false, true)`. It's hard to read and know the meaning of these booleans. An enum can be more descriptive.

PR #346, opened 2018-08-31
Creating documentation for launcher using systemd to persistence

hackunagi:
This pull adds a new file with instructions to create a unit file to use launcher with systemd.
There's also a change to launcher.md to point to this new document.

I tried to follow as much as possible the same documentation that exists on Fleet.

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -13,7 +13,6 @@ Below is a sample unit file.
  ExecStart= $LauncherPath \
  --hostname=$FleetServer:FleetPort \
  --enroll_secret=$FleetSecret \
- --insecure \
  --autoupdate \
 --osqueryd_path=$OsquerydPath
 Restart=on-failure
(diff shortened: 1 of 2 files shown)


docs/systemd.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+## Making changes
+Sometimes you'll need to update the systemd unit file defining the service. To do that, first open `/etc/systemd/system/launcher.service` in a text editor, and make your modifications.
+
+Then, run
+
+```
+sudo systemctl daemon-reload
+sudo systemctl restart fleet.service
```
groob: launcher.service

docs/systemd.md
```diff
@@ -0,0 +1,43 @@
...
+```
+[Unit]
+ Description=The Kolide Launcher
+ After=network.service syslog.service
+
+ [Service]
+ ExecStart= $LauncherPath \
+ --hostname=$FleetServer:FleetPort \
+ --enroll_secret=$FleetSecret \
+ --insecure \
```
groob: I would advise not including this flag as a default suggestion. You can document it, just dont make it copy/pasted.

docs/systemd.md
hackunagi: Adjusted.

docs/systemd.md
hackunagi: Completely agree. Added a line informing this flag as optional and only for test purposes.

groob:
Great. Thank you for the PR!

PR #350, opened 2018-09-14
implement log rotation

groob:
LaunchD does not rotate the file defined under StandardErrorPath.
Here, we render a newsyslogd config under /etc/newsyslog.d/<identifier>.conf to rotate the log once it grows past a certain size.
Upon rotation, newsyslogd will also send a SIGTERM to the launcher pid file, which will be under the launcher root directory.

### DIFF
--- a/pkg/packaging/packaging.go
+++ b/pkg/packaging/packaging.go
@@ -778,7 +778,7 @@ type newSyslogTemplateOptions struct {
 
 func renderNewSyslogConfig(w io.Writer, options *newSyslogTemplateOptions) error {
 	syslogTemplate := `# logfilename          [owner:group]    mode count size when  flags [/pid_file] [sig_num]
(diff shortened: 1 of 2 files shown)


pkg/packaging/packaging.go
```diff
@@ -754,6 +771,21 @@ sleep 5
...
 }
 
+type newSyslogTemplateOptions struct {
+	LogPath string
+	PidPath string
+}
+
+func renderNewSyslogConfig(w io.Writer, options *newSyslogTemplateOptions) error {
+	syslogTemplate := `# logfilename          [owner:group]    mode count size when  flags [/pid_file] [sig_num]
+{{.LogPath}}               640  4000     3   *   G  {{.PidPath}} 15`
```
directionless: This says keep 4,000 3k files? That seems like a weird choice.

groob: 🤦‍♂️ 

I meant 3 4000 kb files

PR #351, opened 2018-09-17
Add new table kolide_gdrive_sync_config

FritzX6:
Major shoutouts to @groob who was instrumental in holding my hand through the process of building this table.
## What does this table do?

This PR adds a new table gdrive_sync_config which outputs the `user_email` and `local_sync_root_path` from a local device's Google Drive configuration database: `/Users/username/Library/Application Support/Google/Drive/user_default/sync_config.db`. 

Example output below:
```
osquery> SELECT * FROM kolide_gdrive_sync_config;
+-----------------+--------------------------------+
| user_email      | local_sync_root_path           |
+-----------------+--------------------------------+
| fritz@kolide.co | /Users/fritz-imac/Google Drive |
+-----------------+--------------------------------+
```

## Why do we need this table?

Customers have asked us to collect information on the configuration of their user's cloud storage usage. This is done both to determine:

- data-exfiltration risk (eg. when a user is storing company IP on a personal cloud storage account)
- compliance with IT practice / onboarding (having a properly configured cloud storage account)

Furthermore, I would like to see the output of this table bolster the results set that we are currently collecting in the `kolide_email_addresses` to perform user:device assignments.

## Why are we adding this as an osquery-go table and not using ATC functionality?

Currently, we do not have an existing workflow setup for distributing new atc tables via custom osquery configuration files.

Furthermore, the existing implementation of ATC suffers from an issue which prohibits parsing SQLite DB's which utilize the WAL type Journal Mode.
https://github.com/facebook/osquery/issues/5225

### DIFF
--- a/pkg/osquery/table/gdrive_sync.go
+++ b/pkg/osquery/table/gdrive_sync.go
@@ -2,32 +2,19 @@ package table
 
 import (
 	"context"
-	"encoding/json"
-	"io/ioutil"
-	"strings"
+	"database/sql"
+	"path/filepath"
 
 	"github.com/kolide/osquery-go"
 	"github.com/kolide/osquery-go/plugin/table"
 	"github.com/pkg/errors"
+
 	_ "github.com/mattn/go-sqlite3"
 )
 
-func GDrivePlugin(client *osquery.ExtensionManagerClient) *table.Plugin {
-	t := &gDriveTable{client: client}
-	paths, err := queryDbPath(t.client)
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
-	database, err := sql.Open("sqlite3", path)
-	if err != nil {
-		fmt.Println(err)
-		os.Exit(1)
-	}
-	defer db.Close()
-
+func GDriveSyncConfig(client *osquery.ExtensionManagerClient) *table.Plugin {
 	g := &gdrive{
-		db: database,
+		client: client,
 	}
 
 	columns := []table.ColumnDefinition{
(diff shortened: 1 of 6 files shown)


pkg/osquery/table/gdrive_sync.go
```diff
@@ -0,0 +1,61 @@
+func GDrivePlugin(client *osquery.ExtensionManagerClient) *table.Plugin {
```
groob: add the imports and package name.

pkg/osquery/table/gdrive_sync.go
```diff
@@ -0,0 +1,61 @@
...
+	database, _ := sql.Open("sqlite3", path)
+	g := &gdrive{
+		db: database,
+	}
+
+	columns := []table.ColumnDefinition{
+		table.TextColumn("user_email"),
+		table.TextColumn("local_sync_root_path"),
+	}
+	return table.NewPlugin("kolide_gdrive_sync_config", columns, generate)
```
groob: should be `g.generate`, since you're calling the generate function on the `g` struct`

pkg/osquery/table/gdrive_sync.go
```diff
@@ -0,0 +1,61 @@
+func GDrivePlugin(client *osquery.ExtensionManagerClient) *table.Plugin {
+	paths := queryDbPath...
+	database, _ := sql.Open("sqlite3", path)
```
groob: should not ignore error. However not sure what to do about it, lets look at other example.

pkg/osquery/table/gdrive_sync.go
```diff
@@ -0,0 +1,61 @@
+func GDrivePlugin(client *osquery.ExtensionManagerClient) *table.Plugin {
+	paths := queryDbPath...
```
groob: call the queryDbPath function with the correct argument, check for error.

pkg/osquery/table/gdrive_sync.go
FritzX6: Thanks I just saw this was missing the sqlite import!

I thought the package name was supposed to be table? (per your example)

FritzX6:
In trying to extend this table structure to another SQLite DB I am running into database locked errors:
```
2018/09/18 08:58:57 panic in processor: database is locked: goroutine 22 [running]:
```
This is something that Mitchell had to address when building his ATC function:
https://github.com/facebook/osquery/pull/4414

I am not sure if it is something that we will need to address with this table as well.

PR #353, opened 2018-09-21
Add 1Password Teams Configuration osquerygo virtual table

FritzX6:
## What does this table do?
This PR adds a new virtual table kolide_onepassword_account_config which outputs the **OnePassword SaaS** account config from the 1Password configuration database located at: `/Users/username/Library/Application Support/1Password 4/Data/B5.sqlite`

To avoid database locking issues (preventing the associated app from functioning properly without access to its respective db) we make a temporary local copy of the db file which we open to read and then subsequently delete.

## Example output below:

```
osquery> select * from kolide_onepassword_account_config;
+-----------+-----------------+-----------------+----------------+
| team_name | user_email      | user_first_name | user_last_name |
+-----------+-----------------+-----------------+----------------+
| Kolide    | fritz@kolide.co | Fritz           | Ifert-Miller   |
+-----------+-----------------+-----------------+----------------+
```
## Why do we need this table?
To output the configuration of a 1Password Teams account present on a local device.

This is also part of the greater initiative to scrape email addresses to bolster the results set that we are currently collecting in the kolide_email_addresses to perform user:device assignments.

## Why are we adding this as an osquery-go table and not using ATC functionality?
Currently, we do not have an existing workflow setup for distributing new atc tables via custom osquery configuration files.

Furthermore, the existing implementation of ATC suffers from an issue which prohibits parsing SQLite DB's which utilize the WAL type Journal Mode.
facebook/osquery#5225

## What are the limitations of this table and possible improvements?
It is currently written in a way that will only return results for the Primary User and as such does not require a JOIN, but also means that it is naturally limited in a multi-user environment.

The table could be improved by gathering the paths for every user directory so that you could perform a CROSS JOIN and output the sync configurations for every user or specific users.

### DIFF
--- a/pkg/osquery/table/email_addresses.go
+++ b/pkg/osquery/table/email_addresses.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/kolide/osquery-go"
@@ -27,11 +28,32 @@ type emailAddressesTable struct {
 }
 
 func (t *emailAddressesTable) generateEmailAddresses(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
-	results := []map[string]string{}
+	var results []map[string]string
 
 	// add results from chrome profiles
+	chromeResults, err := generateEmailAddressesFromChromeState()
+	if err != nil {
+		return nil, errors.Wrap(err, "get email addresses from chrome state")
+	}
+	results = append(results, chromeResults...)
+
+	// add results from 1password
+	onePassResults, err := t.onePasswordAccountConfig.generate(ctx, queryContext)
+	if err != nil {
+		return nil, errors.Wrap(err, "adding email results from 1password config")
+	}
+	results = append(results, onePassResults...)
+
+	return results, nil
+}
+
+func generateEmailAddressesFromChromeState() ([]map[string]string, error) {
+	var results []map[string]string
 	for _, stateFilePath := range findChromeStateFiles() {
 		fileContent, err := ioutil.ReadFile(stateFilePath)
+		if os.IsNotExist(err) {
+			break
+		}
 		if err != nil {
 			return nil, errors.Wrapf(err, "could not read file %s", stateFilePath)
 		}
@@ -45,13 +67,6 @@ func (t *emailAddressesTable) generateEmailAddresses(ctx context.Context, queryC
 			results = addEmailToResults(profile.Username, results)
 		}
 	}
-
-	// add results from 1password
-	results, err := t.onePasswordAccountConfig.generate(ctx, queryContext, results)
-	if err != nil {
-		return nil, errors.Wrap(err, "adding email results from 1password config")
-	}
-
 	return results, nil
 }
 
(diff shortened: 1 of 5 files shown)


zwass:
Needs a `dep ensure` to fix missing dependency in CI?

pkg/osquery/table/onepassword_config.go
```diff
@@ -0,0 +1,83 @@
...
+	return results, nil
+}
+
+func queryDbPath(client *osquery.ExtensionManagerClient) (string, error) {
+	query := `select username from last where username not in ('', 'root') group by username order by count(username) desc limit 1`
+	row, err := client.QueryRow(query)
+	if err != nil {
+		return "", errors.Wrap(err, "querying for primaryUser version")
+	}
+	var username string
```
zwass: Lines 77-80 currently have the same effect as `username := row["username"]`. Is this intentional? If so, I'd prefer the simpler implementation.

pkg/osquery/table/email_addresses.go
```diff
@@ -16,32 +16,43 @@ func EmailAddresses(client *osquery.ExtensionManagerClient) *table.Plugin {
...
-			}
+func (t *emailAddressesTable) generateEmailAddresses(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	results := []map[string]string{}
 
-			var parsedStateFileContent chromeLocalStateFile
-			if err := json.Unmarshal(fileContent, &parsedStateFileContent); err != nil {
-				return nil, errors.Wrap(err, "could not unmarshal json file")
-			}
+	// add results from chrome profiles
+	for _, stateFilePath := range findChromeStateFiles() {
```
zwass: I would feel better if this were abstracted into a `generate` function like the 1password portion.

pkg/osquery/table/onepassword_config.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/kolide/osquery-go/plugin/table"
+)
+
+type onePasswordAccountConfig struct {
+	client *osquery.ExtensionManagerClient
+}
+
+// OnePasswordAccountConfigGenerate will be called whenever the table is queried. It should return
+// a full table scan.
+func (o *onePasswordAccountConfig) generate(ctx context.Context, queryContext table.QueryContext, results []map[string]string) ([]map[string]string, error) {
```
zwass: I'm a little uneasy with the API here. I'd prefer to see it be
```
func (o *onePasswordAccountConfig) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error)
```
and then do the append at the top level.

FritzX6:
I am no longer able to test this table successfully by building osqueryi locally within the repo having run: `make osqueryi`

Not sure if I am doing something wrong.

zwass:
Regarding @FritzX6' last comment, it looks like @groob refactored this to expose the onepassword account email address as part of the `kolide_email_addresses` table. Do we need to reach some consensus on how we'd like to expose this data?

groob:
@zwass we discussed this in slack with fritz. we will only expose the email info via the  `kolide_email_addresses` table. 

We can come to a different conclusion later if needed, but IMO this is good to merge as is.

zwass:
Thanks for resolving that, folks.

PR #355, opened 2018-10-02
Initial differential query

groob:


vivekuppal:
https://app.clubhouse.io/kolide/story/4537/built-in-queries-take-a-while-to-populate-from-the-launcher

pkg/osquery/extension.go
```diff
@@ -144,12 +155,19 @@ func NewExtension(client service.KolideService, db *bolt.DB, opts ExtensionOpts)
 		return nil, errors.Wrap(err, "creating DB buckets")
 	}
 
+	identifier, err := IdentifierFromDB(db)
+	if err != nil {
+		return nil, errors.Wrap(err, "get host identifier from db when creating new extension")
+	}
+	initialRunner := &initialRunner{identifier: identifier, db: db, enabled: opts.RunDifferentialQueriesImmediately}
```
NickTitle: you could skip this whole instantiation if you check the bool first - not sure if that matters

groob: it matters because we still want to cache queries on first seen.

NickTitle: smart, makes sense!

PR #358, opened 2018-10-17
Retrieve SQLite databases from user home directory paths

zwass:
Replaces the old method of determining "primary user" by querying the `last`
table. This will return more results and avoid the expensive query to the
`last` table.

### DIFF
--- a/pkg/osquery/table/chrome_login_keychain.go
+++ b/pkg/osquery/table/chrome_login_keychain.go
@@ -3,11 +3,12 @@ package table
 import (
 	"context"
 	"database/sql"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/go-kit/kit/log"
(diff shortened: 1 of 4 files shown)


pkg/osquery/table/chrome_login_keychain.go
```diff
@@ -85,3 +78,22 @@ func (c *ChromeLoginKeychain) generate(ctx context.Context, queryContext table.Q
...
+	paths, err := findFileInUserDirs("Library/Application Support/Google/Chrome/Default/Login Data")
+	if err != nil {
+		return nil, errors.Wrap(err, "find chrome login data sqlite DBs")
+	}
+
+	var results []map[string]string
+	for _, path := range paths {
+		res, err := c.generateForPath(ctx, path)
+		if err != nil {
+			fmt.Println("Error generating result for path ", path)
```
groob: use a logger?

PR #359, opened 2018-10-18
updating broken server spec link

securityclippy:
Just updated a broken link to documentation that exists, instead of a github 404


groob:
Thanks!

PR #362, opened 2018-10-18
Log proper keyvals in new virtual table error logs

zwass:


pkg/osquery/table/gdrive_sync_history.go
```diff
@@ -97,7 +97,11 @@ func (g *GDriveSyncHistory) generate(ctx context.Context, queryContext table.Que
 	for _, path := range paths {
 		res, err := g.generateForPath(ctx, path)
 		if err != nil {
-			level.Error(g.logger).Log("Error generating gdrive history result for path %s: %s", path, err.Error())
+			level.Error(g.logger).Log(
```
groob: can we log as info?

PR #368, opened 2018-11-01
handle grpc errors

groob:
Redo the server code to correctly encode status codes around errors. 
Update the client code to respect Unauthenticated as Node Invalid. 
Update the logging middleware to differentiate levels when err != nil.


pkg/osquery/extension.go
```diff
@@ -229,6 +231,15 @@ func NodeKeyFromDB(db *bolt.DB) (string, error) {
 	}
 }
 
+func isGRPCInvalid(err error) bool {
```
NickTitle: should you call this isNodeInvalidError(err)? using gRPCInvalid sounds like there's something wrong with the transport itself

pkg/service/error_handling.go
```diff
@@ -0,0 +1,28 @@
...
+	error
+	NodeInvalid() bool
+}
+
+func isNodeInvalidErr(err error) bool {
+	oe, ok := errors.Cause(err).(osqueryError)
+	return ok && oe.NodeInvalid()
+}
+
+func encodeResponse(err error, resp interface{}) (interface{}, error) {
```
NickTitle: could you swap resp and err? that feels like a footgun to me

pkg/osquery/extension.go
```diff
@@ -355,15 +373,14 @@ func (e *Extension) generateConfigsWithReenroll(ctx context.Context, reenroll bo
 			return "", errors.Wrap(err, "enrollment invalid, reenrollment errored")
 		}
 		if invalid {
-			return "", errors.New("enrollment invalid, reenrollment invalid")
+			return "", reenrollmentInvalidErr
```
NickTitle: since you only use it in here, i would stick with the string and pull the comment inside, instead of that var

pkg/service/publish_results.go
```diff
@@ -145,7 +146,11 @@ func (mw logmw) PublishResults(ctx context.Context, nodeKey string, results []di
 	defer func(begin time.Time) {
 		resJSON, _ := json.Marshal(results)
 		uuid, _ := uuid.FromContext(ctx)
-		mw.logger.Log(
+		logger := level.Debug(mw.logger)
```
NickTitle: nit: you could drop an assignment per method with declaring a var, and conditionally assigning

groob: How do you mean?

pkg/osquery/extension.go
groob: I pulled it out so it could be checked. The todo refers to the checking once we decide how to handle that.

pkg/service/publish_results.go
NickTitle:
```golang
var logger log.logger
if err != nil {
  logger := level.Info(mw.logger)
} else {
  logger := level.Debug(mw.logger)
}
```

groob: resolved in slack. I'm going to leave as is and assume the compiler / runtime does the right thing.

PR #372, opened 2018-11-05
bypass log buffer for initial run queries.

groob:
Most of these are there to populate initial enrollment information. It makes for bad UX to wait for the logging interval.

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -871,7 +871,12 @@ func (i *initialRunner) Execute(configBlob string, writeFn func(ctx context.Cont
 			return errors.Wrap(err, "encoding initial run result")
 		}
 		if err := writeFn(cctx, logger.LogTypeString, []string{buf.String()}, true); err != nil {
-			return errors.Wrap(err, "writing encoded initial result log")
+			level.Debug(i.logger).Log(
+				"msg", "writing initial result log to server",
+				"query_name", result.Name,
+				"err", err,
+			)
+			continue
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/osquery/extension.go
```diff
@@ -805,7 +805,7 @@ type initialRunner struct {
 	db         *bolt.DB
 }
 
-func (i *initialRunner) Execute(configBlob string, writeFn func(ctx context.Context, l logger.LogType, s string) error) error {
+func (i *initialRunner) Execute(configBlob string, writeFn func(ctx context.Context, l logger.LogType, s []string, reeenroll bool) error) error {
```
zwass: Is `s` the query result? Can you provide a more descriptive name?

pkg/osquery/extension.go
```diff
@@ -862,13 +862,21 @@ func (i *initialRunner) Execute(configBlob string, writeFn func(ctx context.Cont
...
+	defer cancel()
+
 	for _, result := range initialRunResults {
 		var buf bytes.Buffer
 		if err := json.NewEncoder(&buf).Encode(result); err != nil {
 			return errors.Wrap(err, "encoding initial run result")
 		}
-		if err := writeFn(context.Background(), logger.LogTypeString, buf.String()); err != nil {
-			return errors.Wrap(err, "writing encoded initial result log")
+		if err := writeFn(cctx, logger.LogTypeString, []string{buf.String()}, true); err != nil {
```
zwass: Are there other places where `writeFn` is called with multiple strings in the slice? If not, let's make the type `string`.

groob: That would be more difficult to change because the function its actually calling is `writeResultsWithReenroll`. that is called in multiple places and is `[]string`

PR #375, opened 2018-11-22
experimental: convert build system to bazel

groob:
just exploring this space. (opinion: it's not super intuitive but I like it)


groob:
This was fun, and mostly works fine, but for a purely Go project doesn't provide much benefit over the simple Makefile we have.

PR #378, opened 2018-11-27
compile launcher on windows

groob:
moves all the incompatible syscalls behind build tags, but leaves them as TODOs.

my plan is to merge this then pick it up again on windows and begin an idiomatic runtime & logger implementation.

```
env GOOS=windows go build -o build/launcher.exe ./cmd/launcher
ls build |grep exe
launcher.exe
```

### DIFF
--- a/pkg/debug/signal_debug_windows.go
+++ b/pkg/debug/signal_debug_windows.go
@@ -1,0 +1,9 @@
+// +build windows
+
+package debug
+
+import "github.com/go-kit/kit/log"
+
+func AttachDebugHandler(addrPath string, logger log.Logger) {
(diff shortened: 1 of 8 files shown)


pkg/debug/signal_debug_windows.go
```diff
@@ -0,0 +1,9 @@
+// +build windows
+
+package debug
+
+import "github.com/go-kit/kit/log"
+
+func AttachDebugHandler(addrPath string, logger log.Logger) {
+	// noop for now
```
marpaia: maybe add a `// TODO` so that we have something to grep for?

PR #381, opened 2018-11-28
update runtime with windows specific helpers. 

groob:


pkg/osquery/runtime/runtime.go
```diff
@@ -481,10 +482,14 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "starting osqueryd process")
 	}
 	o.errgroup.Go(func() error {
-		if err := o.cmd.Wait(); err != nil {
+		err := o.cmd.Wait()
+		switch {
+		case err == nil, isExitOk(err):
+			// TODO: should this return nil?
```
marpaia: If `err == nil` and/or `isExitOk(err)` this reads like it should probably return `nil`.

groob: The reason I chose to return an error and add that TODO is because that is the current behavior in master. I'm hoping to return nil and test it more extensively to see whats up.

marpaia: SGTM

PR #382, opened 2018-12-02
simplify package builder options

groob:
remove a bunch of duplication. wip


groob:
Closing in favor of #383 which is a richer refactor of the package-builder.

PR #383, opened 2018-12-04
Packing Tools

directionless:
As part of a general packaging cleanup and , this create a set of `kit` style packaging functions.


pkg/packagekit/package_deb.go
```diff
@@ -0,0 +1,67 @@
...
+
+	outputFilename := fmt.Sprintf("%s-%s.deb", po.Name, po.Version)
+
+	outputPathDir, err := ioutil.TempDir("/tmp", "packaging-deb-output")
+	if err != nil {
+		return errors.Wrap(err, "making TempDir")
+	}
+	defer os.RemoveAll(outputPathDir)
+
+	fpmCommand := []string{
```
groob: the only difference between deb and rpm is `-t`

pkg/packagekit/package_pkg.go
```diff
@@ -0,0 +1,97 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+type PkgOptions struct {
+	SigningKey string
```
groob: the option fields should be private, they're only exposed via the functional args.

pkg/packagekit/package_pkg.go
```diff
@@ -0,0 +1,97 @@
...
+}
```
groob: Another option to add is `WithEnrollSecret`

pkg/packagekit/render_launchd.go
```diff
@@ -0,0 +1,62 @@
...
+	Environment       map[string]string      `plist:"EnvironmentVariables"`
+	Args              []string               `plist:"ProgramArguments"`
+	Label             string                 `plist:"Label"`
+	ThrottleInterval  int                    `plist:"ThrottleInterval"`
+	StandardErrorPath string                 `plist:"StandardErrorPath"`
+	StandardOutPath   string                 `plist:"StandardOutPath"`
+	KeepAlive         map[string]interface{} `plist:"KeepAlive"`
+}
+
+type lOption func(*launchdOptions)
```
groob: this is private, but RenderLaunchd is public. decide on one. IMO `renderLaunchd` should be private, there's arguably no need to call it independently.

pkg/packagekit/render_launchd.go
```diff
@@ -0,0 +1,62 @@
...
+	keepAlive := map[string]interface{}{
+		"PathState": pathState,
+	}
+
+	lOpts := &launchdOptions{
+		Environment:       initOptions.Environment,
+		Args:              append([]string{initOptions.Path}, initOptions.Flags...),
+		Label:             fmt.Sprintf("com.%s.launcher", initOptions.Identifier),
+		ThrottleInterval:  60,
+		StandardErrorPath: fmt.Sprintf("/var/log/%s/launcher-stderr.log", initOptions.Identifier),
```
groob: use filepath.Join

pkg/packagekit/render_launchd.go
directionless: I think this is different now. I took a stab using https://derekchiang.com/posts/reusable-and-type-safe-options-for-go-apis/

pkg/packagekit/render_systemd.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/pkg/errors"
+	"go.opencensus.io/trace"
+)
+
+type systemdOptions struct {
+	Restart    string
+	RestartSec int
+}
+
+//type sOption func(*systemdOptions)
```
groob: rm

pkg/packaging/target.go
```diff
@@ -0,0 +1,57 @@
...
+	Rpm               = "rpm"
+	Msi               = "msi"
+)
+
+func (t *Target) String() string {
+	return fmt.Sprintf("%s-%s-%s", t.Platform, t.Init, t.Package)
+}
+
+func (t *Target) Extension() string {
+	switch t.Package {
```
groob: you can return `t.Package` here since they map 1:1. Consider a `type PackageFlavor int` with iota and a String() method instead.

pkg/packagekit/package.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"io"
+)
+
+// PackageOptions is the superset of all packaging options. Not all
+// packages will support all options.
+type PackageOptions struct {
+	Name       string    // What's the name for this package
+	Postinst   io.Reader // script to run after package install
```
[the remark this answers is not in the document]
directionless: I'm thinking about passing scriptdir instead.

pkg/packaging/target.go
directionless: `Target` is already a type, with methods. Feels off to have `PackageFlavor` be one. I stuck this here as switch anticipating some future need about changing the extension based on something more complicated than just a string(). But simplify for now,

pkg/make/builder.go
```diff
@@ -124,8 +124,8 @@ func (b *Builder) ExtBinary(input string) string {
 	}
 }
 
-// PlatformBinary is a helper to return the platform specific binary suffix.
-func (b *Builder) PlatformBinary(input string) string {
+// BinExtension is a helper to return the platform specific binary suffix.
+func (b *Builder) BinExtension(input string) string {
```
NickTitle: BinExtension immediately made me think this was for making the osquery extension binary. Would you consider calling this BuildFileName, BinaryPath, etc?

cmd/package-builder/package-builder.go
```diff
@@ -36,18 +40,28 @@ func runMake(args []string) error {
...
 			env.String("ENROLL_SECRET", ""),
 			"the string to be used as the server enrollment secret",
 		)
-		flMacPackageSigningKey = flagset.String(
+		flSigningKey = flagset.String(
 			"mac_package_signing_key",
-			env.String("MAC_PACKAGE_SIGNING_KEY", ""),
-			"the name of the key that should be used to sign mac packages",
+			env.String("SIGNING_KEY", ""),
+			"the name of the key that should be used to packages. Will be platform specific",
```
NickTitle: i think you mean `the key that should be used to sign packages`, since you're not dynamically fetching from env based on a name

pkg/make/builder.go
directionless: Sure. These helpers are now `PlatformExtensionName` and `PlatformBinaryName`

pkg/packagekit/package_fpm.go
```diff
@@ -0,0 +1,147 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"go.opencensus.io/trace"
+)
+
+type outputType string
+
+const (
+	Deb outputType = "deb"
```
NickTitle: would you consider taking advantage of the `PackageFlavor` enum instead of redefining these guys?

cmd/package-builder/package-builder.go
directionless: Maybe. For OSX you do. I don't know the other platforms yet.

pkg/packaging/fetch.go
```diff
@@ -0,0 +1,119 @@
...
+	"strings"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/fs"
+	"github.com/kolide/launcher/pkg/contexts/ctxlog"
+	"github.com/pkg/errors"
+)
+
+var (
+	localCacheDir string
```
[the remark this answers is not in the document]
directionless: I pulled this code unmodified from before. I'm not sure why it was implemented this way. I'll leave this comment, and try to figure it out

pkg/packagekit/package_fpm.go
directionless: TBH I'm not really happy with any of that. Talking aloud:

PackageFlavor is part of `packaging`, not `packagekit` 

PackageFlavor is (probably) broader than what fpm can handle. So re-using it may lead to fpm errors down the road. 

This is a mandatory thing, which might fit better as a positional argument, and not as functional parameter.

pkg/packaging/fetch.go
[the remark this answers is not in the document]
directionless: Removed from the global, and pulled up a layer to the callers.

pkg/packagekit/helpers.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+func isDirectory(d string) error {
+	if dStat, err := os.Stat(d); os.IsNotExist(err) {
+		return errors.Wrapf(err, "missing packageRoot %s", d)
+	} else {
```
groob: no need for `else` after a `return`.

pkg/packagekit/package.go
```diff
@@ -0,0 +1,14 @@
+package packagekit
+
+// PackageOptions is the superset of all packaging options. Not all
+// packages will support all options.
+//
+
```
groob: rm empty line

pkg/packagekit/package_fpm.go
```diff
@@ -0,0 +1,120 @@
...
+	RPM            = "rpm"
+	Tar            = "tar"
+)
+
+type fpmOptions struct {
+	outputType outputType
+}
+
+type fpmOpt func(*fpmOptions)
```
groob: export this type declaration, otherwise it's impossible to wrap the PackageFPM function in any way.

pkg/packagekit/package_pkg.go
```diff
@@ -0,0 +1,79 @@
...
+
+	cmd := exec.CommandContext(ctx, "pkgbuild", args...)
+
+	stderr := new(bytes.Buffer)
+	cmd.Stderr = stderr
+	if err := cmd.Run(); err != nil {
+		return errors.Wrapf(err, "creating pkg package: %s", stderr)
+	}
+
+	outputFH, err := os.Open(filepath.Join(outputPathDir, outputFilename))
```
groob: don't forget to close

pkg/packagekit/render_launchd.go
```diff
@@ -0,0 +1,64 @@
...
+	Environment       map[string]string      `plist:"EnvironmentVariables"`
+	Args              []string               `plist:"ProgramArguments"`
+	Label             string                 `plist:"Label"`
+	ThrottleInterval  int                    `plist:"ThrottleInterval"`
+	StandardErrorPath string                 `plist:"StandardErrorPath"`
+	StandardOutPath   string                 `plist:"StandardOutPath"`
+	KeepAlive         map[string]interface{} `plist:"KeepAlive"`
+}
+
+// type lOption func(*launchdOptions)
```
groob: 👀

pkg/packagekit/render_launchd.go
```diff
@@ -0,0 +1,64 @@
...
+		Environment:       initOptions.Environment,
+		Args:              append([]string{initOptions.Path}, initOptions.Flags...),
+		Label:             fmt.Sprintf("com.%s.launcher", initOptions.Identifier),
+		ThrottleInterval:  60,
+		StandardErrorPath: filepath.Join("/var/log", initOptions.Identifier, "launcher-stderr.log"),
+		StandardOutPath:   filepath.Join("/var/log", initOptions.Identifier, "launcher-stdout.log"),
+		KeepAlive:         keepAlive,
+	}
+
+	enc := plist.NewEncoder(w)
```
groob: set an indent on the encoder, otherwise the xml is even more awful.

PR #388, opened 2018-12-10
Move make into a package

directionless:
* Move make into a package
* Add some tests
* Adding some logging

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -1,5 +1,5 @@
 /* Package make provides some simple functions to handle build and go
-dependancies.
+dependencies.
 
 We used to do this with gnumake rules, but as we added windows
 compatibility, we found make too limiting. Moving this into go allows
@@ -49,7 +49,7 @@ type Builder struct {
 
 type Option func(*Builder)
 
-func WithOs(o string) Option {
+func WithOS(o string) Option {
 	return func(b *Builder) {
 		b.os = o
 	}
@@ -210,7 +210,7 @@ func (b *Builder) InstallTools(ctx context.Context) error {
 			)
 			continue
 		}
-		// TODO for go 1.12, this can be parallized.
+
 		g.Go(func() error {
 			return b.installTool(ctx, toolPath)
 		})
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -1,0 +1,429 @@
+/* Package make provides some simple functions to handle build and go
+dependancies.
+
+We used to do this with gnumake rules, but as we added windows
+compatibility, we found make too limiting. Moving this into go allows
+us to write cleaner cross-platform code.
+*/
+
+package make
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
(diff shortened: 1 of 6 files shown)


pkg/make/builder.go
```diff
@@ -0,0 +1,429 @@
...
+	race         bool
+	stampVersion bool
+
+	goVer  *semver.Version
+	cmdEnv []string
+}
+
+type Option func(*Builder)
+
+func WithOs(o string) Option {
```
groob: capitalize OS

pkg/make/builder.go
```diff
@@ -0,0 +1,429 @@
+/* Package make provides some simple functions to handle build and go
+dependancies.
```
groob: `dependencies`

pkg/make/builder.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+func WithStampVersion() Option {
+	return func(b *Builder) {
+		b.stampVersion = true
+	}
+}
+
+// execCommand is a var to allow mocking in tesst
+var execCommand = exec.Command
```
groob: instead of these package globals, add `exec` as a field to Builder and swap it with a private option.

pkg/make/builder.go
```diff
@@ -0,0 +1,429 @@
...
+		if err == nil {
+			level.Debug(ctxlog.FromContext(ctx)).Log(
+				"target", "install tools",
+				"tool", tool,
+				"exists", true,
+				"path", path,
+			)
+			continue
+		}
+		// TODO for go 1.12, this can be parallized.
```
groob: this is concurrent. hence the errgroup.

pkg/make/builder_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"testing"
+
+	"github.com/Masterminds/semver"
+	"github.com/stretchr/testify/require"
+)
+
+func helperCommandContext(ctx context.Context, command string, args ...string) (cmd *exec.Cmd) {
+	cs := []string{"-test.run=TestHelperProcess", "--", command}
+	cs = append(cs, args...)
+	if ctx != nil {
```
groob: ctx should never be nil. eliminate the non context branch from here and always pass the ctx param.

pkg/make/builder_test.go
```diff
@@ -0,0 +1,180 @@
...
+		cmd = exec.CommandContext(ctx, os.Args[0], cs...)
+	} else {
+		cmd = exec.Command(os.Args[0], cs...)
+	}
+	cmd.Env = []string{"GO_WANT_HELPER_PROCESS=1"}
+	return cmd
+}
+
+func helperCommand(command string, args ...string) *exec.Cmd {
+	return helperCommandContext(nil, command, args...)
```
groob: > Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use.

from https://golang.org/pkg/context/#pkg-overview

pkg/make/builder.go
groob: Also, just eliminate the non context version. it's never needed.

PR #390, opened 2018-12-18
add querytarget  proto, kolide_target_membership table

NickTitle:
- what?
  - adds a new proto for accessing targeting information from the device server
  - adds a virtual table called kolide_target_membership, which pulls down an updated label membership state on an interval


pkg/pb/kt/ktarget.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package kolide.launcher;
+
+option go_package = "kt";
```
marpaia: what is `kt`? can we get a more descriptive name perhaps like `pkg/pb/targets` or something?

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,123 @@
...
+		table.TextColumn("id"),
+	}
+
+	return table.NewPlugin("kolide_target_membership", columns, tm.generate())
+}
+
+func (tm *TargetMembership) generate() table.GenerateFunc {
+	return func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+
+		targets, err := tm.GetTargetMembership(ctx)
```
marpaia: Consider making this `tm.getTargetMembership` unless there is a reason to have this symbol exported.

pkg/osquery/extension.go
```diff
@@ -62,6 +62,9 @@ const (
 	// Bucket name to use for buffered result logs.
 	resultLogsBucket = "result_logs"
 
+	// the bucket which we push values into from server-backed tables, like kolide_target_membership
+	ServerProvidedDataBucket = "server_provided_data"
```
marpaia: I don't think this symbol needs to be exported.

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,123 @@
...
+		results := make([]map[string]string, len(targets))
+		for i, t := range targets {
+			results[i] = map[string]string{"id": t.GetId()}
+		}
+
+		return results, nil
+	}
+}
+
+func (tm *TargetMembership) Run(ctx context.Context) error {
```
marpaia: It's a little odd to me that this is in `pkg/osquery/table`, even on the same struct as the table... IMO `Run`, `GetTargetMembership`, and `UpdateTargetMembership` should be on their own struct. If necessary, an instance of that type can be a member of this struct.

cmd/launcher/extension.go
```diff
@@ -89,6 +93,7 @@ func createExtensionRuntime(ctx context.Context, rootDirectory string, db *bolt.
 		runtime.WithOsqueryExtensionPlugin(osquerylogger.NewPlugin("kolide_grpc", ext.LogString)),
 		runtime.WithOsqueryExtensionPlugin(distributed.NewPlugin("kolide_grpc", ext.GetQueries, ext.WriteResults)),
 		runtime.WithOsqueryExtensionPlugin(table.LauncherIdentifierTable(db)),
+		runtime.WithOsqueryExtensionPlugin(tm.Plugin()),
```
marpaia: This table should be registered via the normal `table.PlatformTables` mechanism in the runtime, not instantiated in main. See this for more info on where this happens: https://github.com/kolide/launcher/blob/ac1a399d3e0dbdb894b1e3ea41c0fc5aeb1e8d8d/pkg/osquery/runtime/runtime.go#L547-L549

pkg/pb/kt/ktarget.proto
```diff
@@ -0,0 +1,21 @@
...
+message GetTargetsRequest {
+  string node_key = 1;
+}
+
+message GetTargetsResponse {
+  repeated Target targets = 1;
+}
+
+message Target {
+  string id = 1;
```
directionless: This drops the type enum?

pkg/osquery/table/kolide_target_membership.go
directionless: That seems unlike the other examples here. Certainly `GetTargetMembership` feels like it be on the `TargetMembership` struct. 

I dunno if we should make another struct for the fetches. Maybe depends where we want to go.

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,123 @@
...
+	return TargetMembership{
+		logger:       logger,
+		db:           db,
+		targetClient: targetClient,
+	}
+}
+
+func (tm *TargetMembership) Plugin() *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("id"),
```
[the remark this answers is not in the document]
marpaia: We always call this attribute "id" as far as I know.

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (tm *TargetMembership) Run(ctx context.Context) error {
+	timeChan := time.NewTicker(30 * time.Second).C
+
+	for {
+		select {
+		case <-timeChan:
+			if err := tm.UpdateTargetMemberships(ctx); err != nil {
+				level.Debug(tm.logger).Log(
```
blaedj: Do we try to minimize log output in launcher? I would think this would be an info or error log, but just curios

pkg/osquery/table/kolide_target_membership.go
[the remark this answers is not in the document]
NickTitle: the field is always called id, which may be numeric, uuid, or ulid

pkg/pb/kt/ktarget.proto
NickTitle: renaming to `querytarget` which will remain semantically correct when we extend it beyond the simple id-serving functionality

pkg/osquery/table/kolide_target_membership.go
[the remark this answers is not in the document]
directionless: Works for me.

pkg/pb/kt/ktarget.proto
directionless: I feel like we're talking in circles. What happens when I want to add SQL support?

NickTitle: proto is always forward compatible, so at that point you'd just add extra fields you need to the `Target` message (including the type enum)

 ```proto
message Target {
  string id = 1;
  TargetType target_type = 2;
  string sephs_sql_field = 3;
}

enum TargetType {
   <...>
}
```

File: pkg/osquery/table/kolide_target_membership.go (hunk 8 above)
Comment (NickTitle) [reply pkg/osquery/table/kolide_target_membership.go]: you're right - it looks like we log errors in the other cases. i'll update the level here as well

File: pkg/osquery/table/kolide_target_membership.go [hunk 9]
```diff
@@ -0,0 +1,123 @@
...
+			}
+		case <-ctx.Done():
+			return nil
+		}
+	}
+
+	return nil
+}
+
+func (tm *TargetMembership) GetTargetMembership(ctx context.Context) ([]*kt.Target, error) {
```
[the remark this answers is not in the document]
NickTitle: i've pulled that functionality into the caller instead, and killed the GetTargetMembership method 👍

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,123 @@
...
+		for i, t := range targets {
+			results[i] = map[string]string{"id": t.GetId()}
+		}
+
+		return results, nil
+	}
+}
+
+func (tm *TargetMembership) Run(ctx context.Context) error {
+	timeChan := time.NewTicker(30 * time.Second).C
```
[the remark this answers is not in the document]
NickTitle: could you elaborate on what you mean with `the other gRPC status`?

pkg/osquery/extension.go
NickTitle: this does, so that we can refer to the bucket from the table package - i'll actually have to do the same with the launcher identifier bucket when i move that table out into the 'platform tables' section

pkg/osquery/table/kolide_target_membership.go
NickTitle: i've refactored the Plugin and generate methods to be funcs instead, which just take a db handle and are no longer part of this type

pkg/pb/kt/ktarget.proto
directionless: I think the value of TargetType changes how string is handled.

NickTitle: no issue there - it's future functionality, so you can handle it however you like

cmd/launcher/extension.go
NickTitle: i talked through this with victor, and we came to a slightly different approach - the tables are launcher specific, and not related to platform functionality. I've abstracted them into `LauncherTables` which are loaded alongside launcher plugins

pkg/osquery/table/kolide_target_membership.go
```diff
@@ -0,0 +1,52 @@
...
+		var cachedResp qt.GetTargetsResponse
+		if err := proto.Unmarshal(targetRespBytes, &cachedResp); err != nil {
+			return nil, errors.Wrap(err, "unmarshaling target resp")
+		}
+
+		targets := cachedResp.GetTargets()
+
+		results := make([]map[string]string, len(targets))
+		for i, t := range targets {
+			results[i] = map[string]string{"id": t.GetId()}
```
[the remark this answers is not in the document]
NickTitle: you're protected actually!

in the event we send down stuff the launcher doesn't know about, the deserialization will ignore those fields (json works the same way in go) and the launcher will operate as usual. 

if instead launcher has an updated proto message here with more fields than the server sends, their values in the resultant struct will just be initialized.

in proto all fields are optional, so nothing here will prevent your extending the functionality later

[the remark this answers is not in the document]
directionless: I may not understand.  Today, we'll send down something like
```
targets{
  Target{ id: "x" },
  Target{ id: "y" },
}
```

and those are copied into a `results` at the above code.

If we add a type column that changes behavior what happens?

If we do something like:
```
targets{
  Target{ id: "x", sql: "select ..." },
  Target{ id: "y", "sql: "select ..." },
  Target{ id: "z"},
}
```

The sql is ignored by the older clients, and they have an results array with all three.

If we do something like:

```
targets{
  Target{ sqlid: "x", sql: "select ..." },
  Target{ sqlid: "y", "sql: "select ..." },
  Target{ id: "z"},
}
```

Then older clients would ignore the sqlid, but the results array would be `{nil, nil, z}` (right?) also, this feels weird.

Are you thinking of something else?

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,84 @@
...
+		select {
+		case <-timeChan:
+			if err := qtu.updateTargetMemberships(ctx); err != nil {
+				level.Error(qtu.logger).Log(
+					"msg", "updating kolide_target_membership data",
+					"err", err,
+				)
+			}
+		case <-ctx.Done():
+			return nil
```
groob: return ctx.Err()

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,84 @@
...
+
+		return errors.Wrap(err, "updating target memberships")
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (qtu *QueryTargetUpdater) Start(ctx context.Context) error {
```
groob: Call this `Run`, Start implies it's async.

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,84 @@
...
+	ctx, qtu.cancel = context.WithCancel(ctx)
+	timeChan := time.NewTicker(time.Minute).C
```
groob: call this `ticker`, per convention. 
also add `defer ticker.Stop()`

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+		case <-ctx.Done():
+			return nil
+		}
+	}
+
+	return nil
+}
+
+func (qtu *QueryTargetUpdater) Stop() {
```
groob: Stop should not be necessary. You're calling `Run(ctx)` in main, you also get to call Cancel in main. pass a cancelable context to Run and remove the stop method.

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/gogo/protobuf/proto"
+	"github.com/kolide/launcher/pkg/osquery"
+	"github.com/kolide/launcher/pkg/osquery/table"
+	qt "github.com/kolide/launcher/pkg/pb/querytarget"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+)
+
+type QueryTargetUpdater struct {
+	cancel       context.CancelFunc
```
groob: rm this, use the ctx passed to Run to cancel.

pkg/osquery/table/platform_tables_darwin.go
```diff
@@ -4,27 +4,27 @@ package table
 
 import (
 	"github.com/go-kit/kit/log"
-	"github.com/kolide/osquery-go"
+	osquery "github.com/kolide/osquery-go"
```
groob: no need to alias, the import was already there and worked.

pkg/osquery/table/platform_tables.go
```diff
@@ -4,13 +4,12 @@ package table
 
 import (
 	"github.com/go-kit/kit/log"
-	"github.com/kolide/osquery-go"
+	osquery "github.com/kolide/osquery-go"
```
groob: no need to alias

pkg/osquery/table/platform_tables_darwin.go
NickTitle: ^ weird, i did not do this manually. i'll remove it

pkg/osquery/table/kolide_target_membership.go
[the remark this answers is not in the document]
directionless: I think I'm thinking about future work that doesn't help us here.

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,88 @@
...
+
+		return errors.Wrap(err, "updating target memberships")
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (qtu *QueryTargetUpdater) Run() error {
```
groob: why does run not take a ctx? why is interrupt needed?

pkg/querytarget/query_target.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/kolide/launcher/pkg/osquery"
+	"github.com/kolide/launcher/pkg/osquery/table"
+	qt "github.com/kolide/launcher/pkg/pb/querytarget"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc"
+)
+
+type QueryTargetUpdater struct {
+	runnerCtx    context.Context
+	cancelFunc   func()
```
groob: i would remove the context field and cancel func from here

cmd/launcher/query_target_updater.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/actor"
+	"github.com/kolide/launcher/pkg/querytarget"
+	"google.golang.org/grpc"
+)
+
+func createQueryTargetUpdater(logger log.Logger, db *bolt.DB, grpcConn *grpc.ClientConn) *actor.Actor {
+	updater := querytarget.NewQueryTargeter(logger, db, grpcConn)
+
```
groob: create a context here, call run with the context.

cmd/launcher/query_target_updater.go
```diff
@@ -0,0 +1,26 @@
...
+	return &actor.Actor{
+		Execute: func() error {
+			level.Info(logger).Log("msg", "query target updater started")
+			updater.Run()
+			return nil
+		},
+		Interrupt: func(err error) {
+			level.Info(logger).Log("msg", "query target updater interrupted")
+			updater.Interrupt()
```
groob: remove this, and call cancel() on the context you created above.

pkg/querytarget/query_target.go
NickTitle: misunderstanding over our convo before (just about where the context lives)

i'll resolve all the comments with my next commit

PR #395, opened 2019-01-04
add config table

NickTitle:
- what?
  - adds a table to query the config of the osquery instance

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -257,6 +257,28 @@ func NodeKeyFromDB(db *bolt.DB) (string, error) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/osquery/extension.go
```diff
@@ -257,6 +257,28 @@ func NodeKeyFromDB(db *bolt.DB) (string, error) {
...
+	}
+
+	var key []byte
+	err := db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(configBucket))
+		key = b.Get([]byte(configKey))
+		return nil
+	})
+	if err != nil {
+		return "", errors.Wrap(err, "error reading node key from db")
```
groob: fix wrapped message.

PR #398, opened 2019-01-07
Support for upstart

directionless:
Add support for upstart to packagekit. 

Rework packagekit tests. Expand them to cover more of the options.

Split the render functions to have different signatures. 

Start passing in conflicts/replaces options to FPM

Add Upstart into the package-builder builds.

### DIFF
--- a/pkg/packagekit/package_fpm.go
+++ b/pkg/packagekit/package_fpm.go
@@ -24,6 +24,7 @@ const (
 
 type fpmOptions struct {
 	outputType outputType
+	replaces   []string
 }
 
 type FpmOpt func(*fpmOptions)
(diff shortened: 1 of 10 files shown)


pkg/packagekit/package_fpm.go
```diff
@@ -46,6 +47,12 @@ func AsTar() FpmOpt {
 	}
 }
 
+func WithReplaces(r []string) FpmOpt {
```
groob: ad doc comment? what is this?

pkg/packaging/target.go
```diff
@@ -19,6 +19,7 @@ const (
 	LaunchD InitFlavor = "launchd"
 	SystemD            = "systemd"
 	Init               = "init"
+    Upstart            = "upstart"
```
groob: gofmt

PR #408, opened 2019-01-23
Add --offset flag to launcher-pummel

zwass:
Allows creating a large number of hosts with only some running at a given time.
This makes it easier to fill up a test DB on a local development machine, where
large numbers of hosts running simultaneously would overwhelm the resources
available.


zwass:
This has been superseded by https://github.com/fleetdm/osquery-perf.

PR #414, opened 2019-02-04
Multi Platform support for Chrome Login Keychain

directionless:
Update findFileInUserDirs to be cross-platform. Initially, a fairly minimal set of search locations.

Update the Chrome profile stuff to be multiplatform.

Not sure if we should return an error on unknown platforms, or just an empty set.


directionless:
A quick test on my windows test box (a recent datacenter edition on cloud cadence) the forward slashes are fine. I'm not sure if it's golang or windows handling them, but that part seems fine.

FritzX6:
We should also include the paths for Chromium on linux:

`~/snap/chromium/%/.config/chromium/Local State` 
(*%* represents an unknown snap integer eg. 470)

`~/.config/chromium/Local State`

pkg/osquery/table/chrome_login_keychain.go
```diff
@@ -14,18 +15,23 @@ import (
...
 	c := &ChromeLoginKeychain{
 		client: client,
 		logger: logger,
 	}
 	columns := []table.ColumnDefinition{
-		table.TextColumn("origin_url"),
-		table.TextColumn("action_url"),
+		table.TextColumn("username"),
 		table.TextColumn("username_value"),
+		table.BigIntColumn("frequency"),
```
directionless: Is this a frequency or a count?

pkg/osquery/table/chrome_login_keychain.go
```diff
@@ -65,41 +71,65 @@ func (c *ChromeLoginKeychain) generateForPath(ctx context.Context, path string)
...
+		osProfileDirs, ok = profileDirs["default"]
+		if !ok {
+			return results, errors.New("No profileDir for this platform")
 		}
-		results = append(results, res...)
 	}
 
+	for _, username := range usernames {
+		for _, profileDir := range osProfileDirs {
+			userPaths, err := findFileInUserDirs(filepath.Join(profileDir, "Default/Login Data"), WithUsername(username))
```
directionless: What is this `filepath.Join`?

pkg/osquery/table/chrome_login_keychain.go
```diff
@@ -55,7 +61,7 @@ func (c *ChromeLoginKeychain) generateForPath(ctx context.Context, path string)
 
 	db.Exec("PRAGMA journal_mode=WAL;")
 
-	rows, err := db.Query("SELECT origin_url, action_url, username_value FROM logins")
+	rows, err := db.Query("SELECT username_value, count(*) AS frequency FROM logins GROUP BY lower(username_value)")
```
directionless: How can we add test functionality around `generateForPath` ?

pkg/osquery/table/chrome_login_keychain.go
JonathanNogueira: `Default/Login Data` is the sqlite db.

JonathanNogueira: Are you asking why filepath.Join vs path.Join?

directionless: Oh, that's the sqlite db? The space in the filename made me suspect it was darwin specific. If that's correct on all platforms, it fine. Else it should get dropped into the constants at the top

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, err
+	}
+
+	db, err := sql.Open("sqlite3", dst)
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close()
+
+	db.Exec("PRAGMA journal_mode=WAL;")
```
blaedj: perhaps a comment on why we need this would be warranted here?

pkg/osquery/table/table_util.go
```diff
@@ -4,25 +4,80 @@ import (
...
-	var res []string
-	for _, dir := range userDirs {
-		fullPath := filepath.Join(userDir, dir.Name(), path)
+
+	var foundDirs []string
+
+	homedirRoots, ok := homedirLocations[runtime.GOOS]
+	if !ok {
+		homedirRoots, ok = homedirLocations["default"]
+		if !ok {
```
blaedj: this seems odd, it should never return !ok, because the homeDirLocations explicitly defines a value for the `"default"` key. Probably not a big deal, although the error message is a bit misleading, it should perhaps reference the fact that the `"default"` key was removed from the homeDirLocations map.

pkg/osquery/table/table_util.go
```diff
@@ -4,25 +4,80 @@ import (
...
+		opt(ff)
 	}
-	var res []string
-	for _, dir := range userDirs {
-		fullPath := filepath.Join(userDir, dir.Name(), path)
+
+	var foundDirs []string
+
+	homedirRoots, ok := homedirLocations[runtime.GOOS]
+	if !ok {
```
blaedj: should we add a debug log stating that we are going to search the default location since we didn't recognize the platform?

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+	c := &ChromeLoginKeychainEmailsTable{
+		client: client,
+		logger: logger,
+	}
+	columns := []table.ColumnDefinition{
+		table.TextColumn("username"),
+		table.TextColumn("email"),
+		table.BigIntColumn("count"),
+	}
+	return table.NewPlugin("kolide_chrome_login_keychain_emails", columns, c.generate)
```
directionless: `login` and `keychain` feel redundant.  @FritzX6 Any thoughts?

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+func (c *ChromeLoginKeychainEmailsTable) generateForPath(ctx context.Context, path string, username string) ([]map[string]string, error) {
+	dir, err := ioutil.TempDir("", "kolide_chrome_login_keychain_emails")
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(dir) // clean up
+
+	dst := filepath.Join(dir, "tmpfile")
+	if err := fs.CopyFile(path, dst); err != nil {
+		return nil, err
```
directionless: Please wrap err.

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	db, err := sql.Open("sqlite3", dst)
+	if err != nil {
+		return nil, err
```
directionless: Please wrap err.

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// loop through all the sqlite rows and add them as osquery rows in the results map
+	for rows.Next() { // we initialize these variables for every row, that way we don't have data from the previous iteration
+		var username_value string
+		var username_count string
+		if err := rows.Scan(&username_value, &username_count); err != nil {
+			return nil, errors.Wrap(err, "scanning chrome login keychain db row")
+		}
+		// append anything that could be an email
+		if strings.Contains(username_value, "@") {
```
directionless: Nit: I'd invert this, I think early returns for the exception case makes for more readable go. (do an early return if the string did _not_ contain `@`. But this is fine. )

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+		for _, constraint := range q.Constraints {
+			usernames = append(usernames, constraint.Expression)
+		}
+	} else {
+		return nil, errors.New("The kolide_chrome_login_keychain table requires that you specify a constraint WHERE username =")
+	}
+
+	var results []map[string]string
+	osProfileDirs, ok := profileDirs[runtime.GOOS]
+	if !ok {
```
directionless: Chatting on go slack, I'm still not sure there's something more idiomatic. We can condense a to single line ifs, but I'm not sure it's a win:

```
	var name string
	var ok bool

	if name, ok = nameMap[runtime.GOOS]; !ok {
		if name, ok = nameMap["default"]; !ok {
			panic("No name for this platform")
		}
	}
```

pkg/osquery/table/table_util.go
directionless: This one is me being super defensive. I hate ignore `ok`. Agree on a more clear error

pkg/osquery/table/chrome_login_keychain_emails.go
```diff
@@ -0,0 +1,140 @@
...
+		osProfileDirs, ok = profileDirs["default"]
+		if !ok {
+			return results, errors.New("No profileDir for this platform")
+		}
+	}
+
+	for _, username := range usernames {
+		for _, profileDir := range osProfileDirs {
+			userPaths, err := findFileInUserDirs(filepath.Join(profileDir, "Default/Login Data"), WithUsername(username))
```
directionless: @blaedj is right. We should be looking in all the profiles, not just `Default`. For example, I don't have a `Default` but I do have 7 profiles.

Probably enumerate the directories here, and check anything that starts with Default or Profile

pkg/osquery/table/chrome_login_keychain_emails.go
JonathanNogueira: This was carried over from the chrome_login_keychain.go I will do some digging.

pkg/osquery/table/chrome_login_keychain.go
```diff
@@ -82,18 +82,18 @@ func (c *ChromeLoginKeychain) generateForPath(ctx context.Context, path string)
 }
 
 func (c *ChromeLoginKeychain) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
-	paths, err := findFileInUserDirs("Library/Application Support/Google/Chrome/Default/Login Data")
+	files, err := findFileInUserDirs("Library/Application Support/Google/Chrome/*/Login Data", c.logger)
```
blaedj: Nice, I like how this allows us to sidestep caring about # of profiles etc.

PR #415, opened 2019-02-06
Windows services and move to flag files

directionless:
Now that we support flagfiles for configuration (https://github.com/kolide/launcher/pull/426) let's start using them. 

* moves the launcher options into a flag file
* adds things to wix to correctly create services
* Fixes some bugs around wix template variables


pkg/packagekit/wix/wix.go
```diff
@@ -160,14 +164,63 @@ func (wo *wixOptions) Package(ctx context.Context, pkgOutput io.Writer) error {
...
+	defer heatWrite.Close()
+
+	lines := strings.Split(string(heatContent), "\n")
+	for _, line := range lines {
+		heatWrite.WriteString(line)
+		heatWrite.WriteString("\n")
+		for _, service := range wo.services {
+			isMatch, err := service.Match(line)
+			if err != nil {
+				return err
```
blaedj: wrap err?

pkg/packaging/packaging.go
```diff
@@ -92,59 +92,54 @@ func (p *PackageOptions) Build(ctx context.Context, packageWriter io.Writer, tar
...
 	}
 
 	if p.DisableControlTLS {
-		launcherFlags = append(launcherFlags, "--disable_control_tls")
+		launcherBoolFlags = append(launcherBoolFlags, "disable_control_tls")
 	}
 
 	if p.InsecureGrpc {
-		launcherFlags = append(launcherFlags, "--insecure_grpc")
+		launcherBoolFlags = append(launcherBoolFlags, "insecure_grpc")
```
blaedj: Are we going to want to support the `-insecure_jsonrpc` flag?

pkg/packaging/packaging.go
```diff
@@ -92,59 +92,54 @@ func (p *PackageOptions) Build(ctx context.Context, packageWriter io.Writer, tar
...
+		launcherBoolFlags = append(launcherBoolFlags, "with_initial_runner")
 	}
 
 	if p.Control && p.ControlHostname != "" {
-		launcherEnv["KOLIDE_LAUNCHER_CONTROL_HOSTNAME"] = p.ControlHostname
+		launcherMapFlags["control_hostname"] = p.ControlHostname
 	}
 
 	if p.Autoupdate && p.UpdateChannel != "" {
-		launcherFlags = append(launcherFlags, "--autoupdate")
```
blaedj: interesting, so if we're using a flagfile we just put a flag per line and skip the `-`/`--`?

pkg/packaging/packaging.go
directionless: Erm. Yes, but I feel like we might want to merge `-insecure_grpc` and `-insecure_jsonrpc` into `insecure_transport`. Thoughts?

blaedj: Yea that makes sense I think

pkg/packaging/packaging.go
directionless: Empirically, yes. That seems to be how `ff` works. (it also supports json and yaml, but this was good enough)

pkg/packagekit/wix/service.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for _, opt := range opts {
+		opt(s)
+	}
+
+	return s
+}
+
+// Match returns a bool if there's a match
+func (s *Service) Match(line string) (bool, error) {
```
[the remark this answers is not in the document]
directionless: Yes, exactly. I wanted to add an error if you mad a regex too broad. Like, if you set a regexp to make all `.exe$` into services, that should throw an error here.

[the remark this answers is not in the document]
directionless: I've updated the comment

pkg/packaging/packaging.go
blaedj: interesting choice by `ff` there. Thanks for the explanation

pkg/packaging/packaging.go
directionless: Though, I'd probably rather re-plumb that in a separate PR

blaedj: yea it can go in another PR

PR #416, opened 2019-02-07
Log to debug, not error if GetTargets not implemented by server

blaedj:
This commit:
 - Checks errors returned during calls to `GetTargets`, and logs
 w/'debug' severity instead of 'error' if the error code is
 'Unimplemented..'

Why?
 - There are many clients (most importantly fleet and cloud) that
 don't support this endpoint, and we can safely ignore this error in
 that case. This should reduce the number of unactionable errors that
 launcher produces when used with fleet or other servers that don't
 implement the GetTargets endpoint.

### DIFF
--- a/pkg/querytarget/query_target.go
+++ b/pkg/querytarget/query_target.go
@@ -3,7 +3,6 @@ package querytarget
 import (
 	"context"
 	"time"
-	"strings"
 
 	"github.com/boltdb/bolt"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


pkg/querytarget/query_target.go
```diff
@@ -65,10 +66,18 @@ func (qtu *QueryTargetUpdater) Run(ctx context.Context) error {
 		select {
 		case <-ticker.C:
 			if err := qtu.updateTargetMemberships(ctx); err != nil {
-				level.Error(qtu.logger).Log(
-					"msg", "updating kolide_target_membership data",
-					"err", err,
-				)
+				if strings.Contains(err.Error(), "code = Unimplemented") {
```
directionless: strings o_O

blaedj: 😅

PR #419, opened 2019-02-08
Cross compilation version detection

### LINKED ISSUE #418: Generating packages with package-builder for targets other than darwin fails
**1. What version of launcher are you using (launcher --version)?**
`launcher - version 0.8.0
  branch: 	master
  revision: 	fa45344613462c7955543564efc258ba48e9856a
  build date: 	2019-02-07
  build user: 	seph (seph)
  go version: 	go1.11.5`
**2. What operating system are you using?** 
macOS 10.14.2
**3. What did you do?**
`./build/package-builder make -enroll_secret=“enrollsecret” -hostname=fleet.example.com:443 -targets="linux-systemd-deb" -debug`
**4. What did you expect to happen?**
Ability to build packages for alternate platforms (i.e. debian)
**5. What happened instead?**
`could not generate packages: version detection: Failed to exec. Perhaps -- Can't autodetect while cross compiling. (): run command /var/folders/f6/mcfrxynd42j2982thj6qhrm00000gn/T/package.packageRoot230576456/usr/local/kolide/bin/launcher [-version], stderr=: fork/exec /var/folders/f6/mcfrxynd42j2982thj6qhrm00000gn/T/package.packageRoot230576456/usr/local/kolide/bin/launcher: exec format error`

Unable to build packages for alternate targets other than darwin. Tested with deb, rpm, and Windows with similar result.

davepgreene:
This PR resolves the issue noted in #418 where package version detection will fail in a scenario where a user is cross-compiling and the `package_version` isn't specified. 

This change adds a check around whether the user is cross-compiling and, if so, pulls down the host architecture's launcher to detect the launcher version to use.

### DIFF
--- a/cmd/package-builder/package-builder.go
+++ b/cmd/package-builder/package-builder.go
@@ -228,6 +228,8 @@ func runMake(args []string) error {
 		return err
 	}
 
+	var builtTargets []packaging.Target
+
 	for _, target := range targets {
 		outputFileName := fmt.Sprintf("launcher.%s.%s", target.String(), target.PkgExtension())
 		outputFile, err := os.Create(filepath.Join(outputDir, outputFileName))
@@ -239,9 +241,13 @@ func runMake(args []string) error {
 		if err := packageOptions.Build(ctx, outputFile, target); err != nil {
 			return errors.Wrap(err, "could not generate packages")
 		}
+		builtTargets = append(builtTargets, target)
 	}
 
-	fmt.Printf("Built you packages in %s\n", outputDir)
+	fmt.Printf("\nBuilt the following packages in %s:\n", outputDir)
+	for _, target := range builtTargets {
+		fmt.Println(target.String())
+	}
 	return nil
 }
 
--- a/pkg/packaging/packaging.go
+++ b/pkg/packaging/packaging.go
@@ -599,7 +599,14 @@ func (p *PackageOptions) setupDirectories() error {
 
 // BUG This doesn't work on windows
 func (p *PackageOptions) detectLauncherVersion(ctx context.Context) error {
-	launcherPath := filepath.Join(p.packageRoot, p.binDir, p.target.PlatformBinaryName("launcher"))
+	var launcherPath string
+
+	if p.crossCompiling {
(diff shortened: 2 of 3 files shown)


directionless:
Thanks for sending this in! I'm going to think on what approach is good. Sometimes the different platforms have different releases.

directionless:
I think there are a couple of bugs here. One of them, is that the current stable windows build does not support `-version`. This is fixed in https://github.com/kolide/launcher/pull/426

The other one, is that if you're building a linux package on darwin, the detection fails. I think most packaging is going to need to happen on the native platform, thus I'm unsure if I want to fix this. I'd appreciate hearing more about how people expect to use this.

davepgreene:
I have a thought: what if version detection posted a target platform to an endpoint that just returned the current `latest`, `stable`, etc. version for that platform? 

This way platform target version determination can be separated from pulling the launcher binary. It'll also avoid the issue you noted around the stable windows build not supporting the `-version` flag.

davepgreene:
Also, separately, it would be helpful for someone to define constraints around compilation origin and target platforms. 

For instance, we know that Darwin can build against Linux but not vice versa. Similarly, nothing can build Windows except Windows afaik. I'm unclear if Windows can build against any other targets. 

My initial approach was to promote cross-platform compilation but if it's easier to define the boundaries and enforce them in much happier with that approach provided it's documented.

directionless:
> I have a thought: what if version detection posted a target platform to an endpoint that just returned the current `latest`, `stable`, etc. version for that platform?

Yeah, I've been thinking about that too. It ought be extractable from the notary server. But, as things move to less cross-platform, it has less value.

> Also, separately, it would be helpful for someone to define constraints around compilation origin and target platforms.

There are 2 pieces. First one, is whether launcher itself can be build cross platform. Today, it can be (and we do). But the general expectation is that as we add OS specific tables, we'll need to move to building on said OSes.

Second, packaging-building. This is mostly a function of the tools. We build mac packages with pkgbuilder, and this will only ever run on osx. We build windows package with wix, which will only ever run on wix. We build linux packages using fpm, which runs in docker probably anywhere.

directionless:
Sorry this has languished so long. I don't want to add this, I think it will prove fragile. I think if you want this kind of functionality, a better place is wrapper scripts above package-builder. (or a client that uses the same interfaces)

PR #420, opened 2019-02-11
Subcommand test should bail early if no subcommand or flag is provided.

davepgreene:
This PR fixes a bug where the launcher will get a `panic`:
```
Feb 11 11:29:10 hostname launcher[103732]: panic: runtime error: index out of range
Feb 11 11:29:10 hostname launcher[103732]: goroutine 1 [running]:
Feb 11 11:29:10 hostname launcher[103732]: main.isSubCommand(0xa52c20)
Feb 11 11:29:10 hostname launcher[103732]:         /Users/seph/go/src/github.com/kolide/launcher/cmd/launcher/main.go:69 +0xf3
Feb 11 11:29:10 hostname launcher[103732]: main.main()
Feb 11 11:29:10 hostname launcher[103732]:         /Users/seph/go/src/github.com/kolide/launcher/cmd/launcher/main.go:23 +0x83
```
if the launcher is executed without command line args and all relevant settings being provided via ENV var.

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -56,7 +56,7 @@ func main() {
 }
 
 func isSubCommand() bool {
-	if len(os.Args) != 2 {
+	if len(os.Args) == 1 {
 		return false
 	}
 


directionless:
what version are you seeing a panic on? I think the panic was fixed in https://github.com/kolide/launcher/pull/402

dgreene-r7:
This is on `master`.

The solution from #402 doesn't cover the case where no args are provided to the launcher (`os.Args` will equal 1). In that case the code will range over the list of possible subcommands and then attempt to determine if `os.Args[1]` (which doesn't exist) is one of those subcommands. 

Does that make sense?

davepgreene:
This is on `master`.

The solution from #402 doesn't cover the case where no args are provided to the launcher (`os.Args` will equal 1). In that case the code will range over the list of possible subcommands and then attempt to determine if `os.Args[1]` (which doesn't exist) is one of those subcommands. 

Does that make sense?

davepgreene:
@directionless I'm seeing this because `stable` on Notary is still `0.7.0`.

directionless:
I like to let things hang out on nightly and beta a bit before pushing to stable. Hopefully I can update this week.

directionless:
0.8.0 is out as stable

PR #421, opened 2019-02-12
Fix binary-bundle

directionless:
We moved a bunch of the launcher make tooling _from_ make, and into a go helper. Unfortunately, this broke the `binary-bundle` target. While not strictly required, it does help with the launcher releases. 

Update the binary-bundle target to work in the new build system

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ builder:
 	cd tools/builders/launcher-builder/1.11/ && gcloud builds submit --project=kolide-public-containers --config=cloudbuild.yml
 
 binary-bundle: VERSION = $(shell git describe --tags --always --dirty)
-binary-bundle: #xp-codesign
+binary-bundle: xp-codesign
 	rm -rf build/binary-bundle
 	$(MAKE) -j $(foreach p, darwin linux windows, build/binary-bundle/$(p))
 	cd build/binary-bundle && zip -r "launcher_${VERSION}.zip" *
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ PATH := $(GOPATH)/bin:$(PATH)
 
 export GO111MODULE=on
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -102,15 +102,15 @@ push-containers: $(CONTAINERS)
 builder:
 	cd tools/builders/launcher-builder/1.11/ && gcloud builds submit --project=kolide-public-containers --config=cloudbuild.yml
 
-binary-bundle: xp-codesign
+binary-bundle: VERSION = $(shell git describe --tags --always --dirty)
+binary-bundle: #xp-codesign
```
blaedj: rm comment

Makefile
```diff
@@ -102,15 +102,15 @@ push-containers: $(CONTAINERS)
...
-	mkdir -p build/binary-bundle/linux
-	mkdir -p build/binary-bundle/darwin
-	cp build/linux/launcher build/binary-bundle/linux/launcher
-	cp build/linux/osquery-extension.ext build/binary-bundle/linux/osquery-extension.ext
-	go run ./tools/download-osquery.go --platform=linux --output=build/binary-bundle/linux/osqueryd
-	cp build/darwin/launcher build/binary-bundle/darwin/launcher
-	cp build/darwin/osquery-extension.ext build/binary-bundle/darwin/osquery-extension.ext
-	go run ./tools/download-osquery.go --platform=darwin --output=build/binary-bundle/darwin/osqueryd
-	cd build/binary-bundle && zip -r "launcher_${VERSION}.zip" linux/ darwin/
+	$(MAKE) -j $(foreach p, darwin linux windows, build/binary-bundle/$(p))
```
blaedj: TIL make has a foreach construct

Makefile
directionless: Otherway here -- it should be uncommented. Nice catch

PR #423, opened 2019-02-13
Travis-CI ands test updates

directionless:
Some initial work to get travis-ci running. This includes updates to the make the tests more reliable, and have fewer dependancies on on the initial container setup. Travis-ci is desirable for it's multiple-platforms.

* Added travis-ci
* Updated runtime_test to download osqueryd as part of a test setup step.
* moved downloadOsqueryInBinDir to test setup and removed race conditions

This does mean that we have duplicate CI systems

### DIFF
--- a/pkg/osquery/runtime/runtime_test.go
+++ b/pkg/osquery/runtime/runtime_test.go
@@ -4,23 +4,48 @@ import (
 	"bufio"
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"syscall"
 	"testing"
 	"time"
 
+	"github.com/kolide/kit/fs"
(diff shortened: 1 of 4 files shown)


pkg/osquery/runtime/runtime_test.go
```diff
@@ -4,38 +4,75 @@ import (
...
+	"github.com/kolide/launcher/pkg/packaging"
 	osquery "github.com/kolide/osquery-go"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+var testOsqueryBinaryDirectory string
+
+// TestMain overrides the default test main function. This allows us to share setup/teardown.
```
blaedj: is there anything that the default test main function does for us? or is it just the `os.Exit(m.Run())` by default?

directionless: Docs imply that's all it is, and suggest it as the place to hang setup/teardown tasks.

PR #425, opened 2019-02-19
Debug Enrollment errors

directionless:
Replace error.New with error.Wrap

Add stacktraces to exit conditions

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -338,7 +338,10 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 		return "", true, errors.Wrap(err, "transport error in enrollment")
 	}
 	if invalid {
-		return "", true, errors.Errorf("enrollment invalid (orig_err: %v)", err)
+		if err == nil {
+			err = errors.New("no further error")
+		}
+		return "", true, errors.Wrap(err, "enrollment invalid")
 	}
 
 	// Save newly acquired node key if successful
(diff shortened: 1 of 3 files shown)


pkg/osquery/extension.go
```diff
@@ -330,14 +330,15 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+	// note that we set invalid two ways. Via the return, _or_ via isNodeInvaliderr
 	keyString, invalid, err := e.serviceClient.RequestEnrollment(ctx, e.Opts.EnrollSecret, identifier, enrollDetails)
 	if isNodeInvalidErr(err) {
 		invalid = true
 	} else if err != nil {
 		return "", true, errors.Wrap(err, "transport error in enrollment")
 	}
 	if invalid {
-		return "", true, errors.Wrap(err, "enrollment invalid")
+		return "", true, errors.Errorf("enrollment invalid (orig_err: %v)", err)
```
blaedj: don't we lose the backtrace etc if return a new error here?

directionless: Yeah. I'm thinking I'm going to have to do something really ugly

directionless: Okay. How's this look

blaedj: much better 😃

directionless:
@blaedj is `if Rails.env.production? || ENV.fetch('USE_STACKDRIVER', false)` better?

PR #426, opened 2019-02-19
Rework Windows Services

directionless:
We have a couple of hard to support windows corner cases. I think these stem from the difference between windows services, and traditional unix daemons. This PR aims to make them more approachable.

**Moves the windows startup process**  to being explicit, instead of autotected on whether it's in windows  and non-interactive. Now there are `svc` and `svc-debug` subcommands.  This should remove a source of strange errors when interactivity is mis-detected, and also makes it easier to run launcher in various modes for debugging.

**Remove the service management** This removes the service management tools. That's better handled by ther packaging.

**Change the flag parsing** from `kit/env` to `ff`. This should allow us to more easily pass in a flag file. 

Refactor several things in support of these.

I think this is mergeable now.


cmd/launcher/options.go
```diff
@@ -52,141 +50,60 @@ const (
...
+		flEnrollSecretPath  = flagset.String("enroll_secret_path", "", "Optionally, the path to your enrollment secret")
+		flGetShellsInterval = flagset.Duration("control_get_shells_interval", 3*time.Second, "The interval at which the get shells request will be made")
+		flInitialRunner     = flagset.Bool("with_initial_runner", false, "Run differential queries from config ahead of scheduled interval.")
+		flKolideServerURL   = flagset.String("hostname", "", "The hostname of the gRPC server")
+		flLoggingInterval   = flagset.Duration("logging_interval", 60*time.Second, "The interval at which logs should be flushed to the server")
+		flOsquerydPath      = flagset.String("osqueryd_path", "", "Path to the osqueryd binary to use (Default: find osqueryd in $PATH)")
+		flRootDirectory     = flagset.String("root_directory", "", "The location of the local database, pidfiles, etc.")
+		flRootPEM           = flagset.String("root_pem", "", "Path to PEM file including root certificates to verify against")
+		flVersion           = flagset.Bool("version", false, "Print Launcher version and exit")
+		_                   = flagset.String("config", "", "config file ro parse options from (optional)")
```
blaedj:
```suggestion
		_                   = flagset.String("config", "", "config file to parse options from (optional)")
```

cmd/launcher/options.go
```diff
@@ -52,141 +50,60 @@ const (
...
-			"The interval at which logs should be flushed to the server",
-		)
+	var (
+		// Primary options
+		flCertPins          = flagset.String("cert_pins", "", "Comma separated, hex encoded SHA256 hashes of pinned subject public key info")
+		flControl           = flagset.Bool("control", false, "Whether or not the control server is enabled (default: false)")
+		flControlServerURL  = flagset.String("control_hostname", "", "The hostname of the control server")
+		flEnrollSecret      = flagset.String("enroll_secret", "", "The enroll secret that is used in your environment")
+		flEnrollSecretPath  = flagset.String("enroll_secret_path", "", "Optionally, the path to your enrollment secret")
+		flGetShellsInterval = flagset.Duration("control_get_shells_interval", 3*time.Second, "The interval at which the get shells request will be made")
```
blaedj:
```suggestion
		flGetShellsInterval = flagset.Duration("control_get_shells_interval", 3*time.Second, "The interval at which the 'get shells' request will be made")
```

cmd/launcher/options.go
```diff
@@ -52,141 +50,60 @@ const (
...
-		)
-		flUpdateChannel = flag.String(
-			"update_channel",
-			env.String("KOLIDE_LAUNCHER_UPDATE_CHANNEL", "stable"),
-			"The channel to pull updates from (options: stable, beta, nightly)",
-		)
+		flAutoupdate         = flagset.Bool("autoupdate", false, "Whether or not the osquery autoupdater is enabled (default: false)")
+		flNotaryServerURL    = flagset.String("notary_url", autoupdate.DefaultNotary, "The Notary update server (default: https://notary.kolide.co)")
+		flMirrorURL          = flagset.String("mirror_url", autoupdate.DefaultMirror, "The mirror server for autoupdates (default: https://dl.kolide.co)")
+		flAutoupdateInterval = flagset.Duration("autoupdate_interval", 1*time.Hour, "The interval to check for updates (default: once every hour)")
```
blaedj: perhaps we can punt on it for this PR, but AFAICT the only way to figure out the expected format for this value is to read the code.

cmd/launcher/svc_windows.go
```diff
@@ -0,0 +1,103 @@
...
+		level.Info(logger).Log("err", err)
+		os.Exit(1)
+	}
+
+	run := svc.Run
+
+	return run(serviceName, &winSvc{logger: logger, opts: opts})
+}
+
+func runWindowsSvcDebug(args []string) error {
```
blaedj: should this just be a debug flag? Or should we call it `foreground-svc` or something? I feel like there's going to be some confusion around this vs the `-debug` flag.

cmd/launcher/svc_windows.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			level.Info(w.logger).Log("err", err)
+			changes <- svc.Status{State: svc.Stopped, Accepts: cmdsAccepted}
+			os.Exit(1)
+		}
+	}()
+
+	for {
+		select {
+		case c := <-r:
```
blaedj: more of a gripe than anything, but the go's idiomatic single letter var names have yet to resonate with me.

cmd/launcher/options.go
directionless: I say punt. This is mostly an advanced debug option, it's not usually displayed. And, if you get it wrong:
```
dover:launcher seph$ ./build/launcher -autoupdate_interval 5
invalid value "5" for flag -autoupdate_interval: time: missing unit in duration 5
```

cmd/launcher/svc_windows.go
directionless: I thought about a flag, but given how the flag options are constructed, that felt awkward. Like, what would that flag do for not-windows? Meanwhile, I renamed to `svc-fg` to make the distinction clear. I think it is an exceedingly uncommon debugging mode

cmd/launcher/svc_windows.go
directionless: Yeah... sometimes I like them, other times I have the same reaction.

But this code was just moved from main_windows.go to svc_windows.go.

directionless:
Correct -- it removes the service management commands. I think those are better handled by the package tooling. I'm not opposed to bringing them back if we find we need them.

Yes, the tests are trying to very both `-` and `--` work. I've added a comment. The tests are a bit silly, since that should be tested by the `ff` library. But it seemed good to ensure we don't break something along the way

cmd/launcher/svc_windows.go
blaedj: Yea I saw this after I commented on that. 👍

cmd/launcher/options.go
blaedj: ah ok, good enough for now then.

PR #427, opened 2019-02-20
Add deferred cleanups

directionless:
I noticed we were making temp files and not cleaning them up.

Add the deferred cleanup routines.


directionless:
Breaks tests. Skipping this for now

PR #434, opened 2019-02-26
Change a bunch of sleep to timers

directionless:
Sleep is blocking, whereas timers are not. Change a bunch of sleep instances to timers.


groob:
Your timers are still blocking on the channel. What’s the difference?

Also the extension startup sleep is no longer necessary. We fixed osquery-go to wait until the socket is available.

directionless:
Hrm. I had thought timer was preferred to sleep, and that `time.Sleep` blocked, but testing I cannot reproduce that. So I'm either out of date, or just wrong. Either way, I'll close this.

I might revisit some of the sleeps, for possible removal. But I don't think I want to touch them right now. There's been a lot of fragility around it

PR #437, opened 2019-02-28
Work around json being unable to unmarshal errors

directionless:
We cannot directly unmarshal errors. But, we seem to get them sent over. So, bounce through an intermediary struct and create the error.

This feels very roundabout.

### DIFF
--- a/pkg/service/request_enrollment.go
+++ b/pkg/service/request_enrollment.go
@@ -37,7 +37,8 @@ type EnrollmentDetails struct {
 type enrollmentResponse struct {
 	NodeKey     string `json:"node_key"`
 	NodeInvalid bool   `json:"node_invalid"`
-	Err         error  `json:"error_code"`
+	ErrorCode   string `json:"error_code"`
+	Err         error
 }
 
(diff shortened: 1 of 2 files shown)


pkg/service/request_enrollment.go
```diff
@@ -70,12 +70,25 @@ func decodeJSONRPCEnrollmentResponse(_ context.Context, res jsonrpc.Response) (i
...
 	}
-	var result enrollmentResponse
-	err := json.Unmarshal(res.Result, &result)
+
+	// Can't unmarshal errors, as they're interfaces. So we unmarshal to
+	// an intermediary first.
+	var tmpResult = struct {
+		NodeKey     string `json:"node_key"`
+		NodeInvalid bool   `json:"node_invalid"`
+		Err         string `json:"error_code"`
```
blaedj: we could also just update the `enrollmentResponse` type to have a field `ErrorCode`, and a method 
```go
func (e enrollmentResponse) Error() error {
  if e.ErrorCode != nil {
    return errors.New(errorCode)
  }
return nil
}
```
OTOH, that's likely to require more work to ensure we don't break existing things..

directionless: Yeah. I wasn't sure what all that might break.

PR #438, opened 2019-03-02
Add go-based platform info to the initial enrollment (and fix bugs)

directionless:
A motley assortment of small fixes as I get this enrolling:
* Plumb build tags through make.go, to allow for fake data. Fix the version string appending.
* Add some go based platform indicators to kolide_launcher_info, as well as to the initial enrollment.
* Move kolide_launcher_info to being in both launcher, and tables.ext
* Fix error handling on the call to `getEnrollDetails`
* Attempt to fix suspected flakey test after https://github.com/kolide/launcher/pull/342

### DIFF
--- a/cmd/osquery-extension/osquery-extension.go
+++ b/cmd/osquery-extension/osquery-extension.go
@@ -6,14 +6,24 @@ import (
 	"os"
 	"os/signal"
 	"time"
+
+	"github.com/kolide/kit/version"
 )
 
 func main() {
-	flag.String("socket", "", "")
-	flag.Int("timeout", 0, "")
-	flag.Int("interval", 0, "")
-	flag.Bool("verbose", false, "")
+	var (
+		_         = flag.Bool("verbose", false, "")
+		_         = flag.Int("interval", 0, "")
+		_         = flag.Int("timeout", 0, "")
+		_         = flag.String("socket", "", "")
+		flVersion = flag.Bool("version", false, "Print  version and exit")
+	)
 	flag.Parse()
+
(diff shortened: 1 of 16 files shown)


cmd/osquery-extension/osquery-extension.go
```diff
@@ -6,15 +6,25 @@ import (
...
+	"github.com/kolide/kit/version"
 )
 
 func main() {
-	flag.String("socket", "", "")
-	flag.Int("timeout", 0, "")
-	flag.Int("interval", 0, "")
-	flag.Bool("verbose", false, "")
+	var (
+		_         = flag.Bool("verbose", false, "")
```
blaedj: are these flags just defined to preserve compatibility with osquery's flag set? 

Mostly I'm trying to figure out why we define them if we don't use them.

pkg/osquery/extension.go
```diff
@@ -788,6 +791,7 @@ func (e *Extension) writeResultsWithReenroll(ctx context.Context, results []dist
 }
 
 func getEnrollDetails(client Querier) (service.EnrollmentDetails, error) {
+	fmt.Println("seph 1")
```
blaedj: remove the `seph` `println`s in this file? I see 3 of them

directionless: oops, I missed pushing up a commit. Yeah

cmd/osquery-extension/osquery-extension.go
directionless: Dunno. I assume so.

PR #444, opened 2019-03-12
Windows Update Support

directionless:
Windows update support. Though not a big change, it is a bit subtle. 

In short, the old autoupdate functionality is based on replacing the launcher binary, and then exec'ing the new one. Unfortunately, this does not work on windows. Both the replace and the exec fail for different reasons. So instead, we do two moves, and then exit letting the service manager restart. 

There's a bunch more detail in the [ADR](https://github.com/directionless/launcher/blob/seph/windows-update/docs/architecture/2019-03-11_autoupdate.md)


cmd/launcher/updater-finalizer.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+)
+
+// launcherFinalizer finalizes a launcher update. It assume the new
+// binary has been copied into place, and calls exec, so we start a
+// new running launcher in our place.
+func launcherFinalizer(logger log.Logger, shutdownOsquery func() error) func() error {
```
blaedj: `updateFinalizer` or `launcherUpdateFinalizer` would be a better name for this perhaps.

docs/architecture/2019-03-11_autoupdate.md
```diff
@@ -0,0 +1,76 @@
...
+The update process on Windows is based on the service manager
+restarting the service. We don't believe there's a downside here, but
+it does increase restart counts. Due to implementation limitations of
+WiX, the shortest recovery period is 1 day.
+
+Due to the nature of this update process, updates that depend on
+command line flag changes, require a re-installation of Launcher. They
+are handled outside this update process.
+
+This process can result in running binaries that are no longer on disk.
```
blaedj: can you expand on this? afaict this would only happen if the error that we raise to kill the service weren't properly handled, or is it because we _move_ the existing binary to a new place in order to move the new binary into the expected launcher location?

pkg/autoupdate/autoupdate.go
```diff
@@ -103,14 +103,16 @@ func NewUpdater(binaryPath, rootDirectory string, logger log.Logger, opts ...Upd
...
 func (u *Updater) createLocalTufRepo() error {
 	if err := os.MkdirAll(u.settings.LocalRepoPath, 0755); err != nil {
 		return err
 	}
 	localRepo := filepath.Base(u.settings.LocalRepoPath)
 	assetPath := path.Join("pkg", "autoupdate", "assets", localRepo)
-	if err := createTUFRepoDirectory(u.settings.LocalRepoPath, assetPath, AssetDir); err != nil {
+
+	if err := u.createTUFRepoDirectory(u.settings.LocalRepoPath, assetPath, AssetDir); err != nil {
 		return err
```
blaedj: wrap this error?

pkg/autoupdate/autoupdate.go
```diff
@@ -138,7 +140,7 @@ func createTUFRepoDirectory(localPath string, currentAssetPath string, assetDir
 			// files not yet yet there -- Generating an invalid state. Note:
 			// this does not check the validity of the files, they might be
 			// corrupt.
-			if _, err := os.Stat(fullAssetPath); !os.IsNotExist(err) {
```
blaedj: was this the bug that you mentioned related to tuf not replacing directories?

pkg/autoupdate/autoupdate_test.go
```diff
@@ -18,7 +18,8 @@ func TestCreateTUFRepoDirectory(t *testing.T) {
 	localTUFRepoPath, err := ioutil.TempDir("", "")
 	require.NoError(t, err)
 
-	require.Nil(t, createTUFRepoDirectory(localTUFRepoPath, "pkg/autoupdate/assets", AssetDir))
+	u := &Updater{}
+	require.Nil(t, u.createTUFRepoDirectory(localTUFRepoPath, "pkg/autoupdate/assets", AssetDir))
```
blaedj: nit: `require.NoError(...)` is slightly more expressive here, but this is fine.

directionless: Not my code. But that's reasonable, I fixed it

pkg/autoupdate/autoupdate.go
directionless: Yes. It should have been `fullLocalPath` Might have only been broken if you were in the root of a launcher checkout. Not sure

cmd/launcher/updater-finalizer.go
directionless: reasonable

docs/architecture/2019-03-11_autoupdate.md
directionless: Heh. Neither. this happens on the unix side, when we move the new binary over the old one. I'll update this

PR #452, opened 2019-03-22
Add slack config table

JonathanNogueira:
This Commit: 
- added a new table kolide_slack_config

The new table requires a team_id constraint to be part of the query
```
select * from kolide_slack_config where team_id='T08V7PM9C';
+-----------+-----------+----------------------------+-----------+-------------+-----------+
| team_id   | team_name | team_url                   | logged_in | user_handle | user_id   |
+-----------+-----------+----------------------------+-----------+-------------+-----------+
| T08V7PM9C | osquery   | https://osquery.slack.com/ | 1         | jonathan    | 000000000 |
+-----------+-----------+----------------------------+-----------+-------------+-----------+
```

### DIFF
--- a/pkg/osquery/table/slack_config.go
+++ b/pkg/osquery/table/slack_config.go
@@ -17,7 +17,7 @@ import (
 
 var slackConfigDirs = map[string][]string{
 	"windows": []string{"AppData/Roaming/Slack/storage"},
-	"darwin":  []string{"/Library/Application Support/Slack/storage"},
+	"darwin":  []string{"Library/Application Support/Slack/storage"},
 }
 var slackConfigDirDefault = []string{".config/Slack/storage/"}
 
--- a/pkg/osquery/table/slack_config.go
+++ b/pkg/osquery/table/slack_config.go
@@ -26,7 +26,7 @@ func SlackConfig(client *osquery.ExtensionManagerClient, logger log.Logger) *tab
 		table.TextColumn("team_id"),
 		table.TextColumn("team_name"),
 		table.TextColumn("team_url"),
-		table.TextColumn("logged_in"),
+		table.IntegerColumn("logged_in"),
 		table.TextColumn("user_handle"),
 		table.TextColumn("user_id"),
 	}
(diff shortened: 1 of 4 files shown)


pkg/osquery/table/slack_config.go
```diff
@@ -0,0 +1,114 @@
...
+			continue
+		}
+		for _, file := range files {
+			res, err := t.generateForPath(ctx, file)
+			if err != nil {
+				level.Info(t.logger).Log(
+					"msg", "Generating slack team result",
+					"path", file.path,
+					"err", err,
+				)
```
groob: should continue, otherwise you'll append a nil result

pkg/osquery/table/slack_config.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	osquery "github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+var slackConfigDirs = map[string][]string{
+	"windows": []string{"AppData/Roaming/Slack/storage"},
+	"darwin":  []string{"/Library/Application Support/Slack/storage"},
```
directionless:
```suggestion
	"darwin":  []string{"Library/Application Support/Slack/storage"},
```
Why the leading slash

pkg/osquery/table/slack_config.go
```diff
@@ -0,0 +1,123 @@
...
+func (t *SlackConfigTable) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+	// Prevent this table from being used to easily enumerate a user's slack teams
+	q, ok := queryContext.Constraints["team_id"]
+	if ok && len(q.Constraints) == 0 {
+	return results, errors.New("The kolide_slack_config table requires that you specify a constraint WHERE team_id =")
+	}
+	if ok { // If we have a constraint on team_id limit it to the = operator
+			for _, constraint := range q.Constraints {
+				if constraint.Operator != table.OperatorEquals {
```
directionless: I feel weird about this. I think we should be enforcing this constraint in how we generate the table. Eg, just take what they're asking for, and pull the data. If they want to get info for `%`, then they get data for that as a literal.

pkg/osquery/table/slack_config.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	osquery "github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+var slackConfigDirs = map[string][]string{
+	"windows": []string{"AppData/Roaming/Slack/storage"},
+	"darwin":  []string{"Library/Application Support/Slack/storage"},
```
directionless: app store distributions are `Library/Containers/com.tinyspeck.slackmacgap/Data/Library/Application Support/Slack/storage/slack-teams`

pkg/osquery/table/slack_config.go
```diff
@@ -0,0 +1,123 @@
...
+					return results, errors.New("The kolide_slack_config table only accepts = constraints on the team_id column")
+				}
+			}
+	}
+	osProfileDirs, ok := slackConfigDirs[runtime.GOOS]
+	if !ok {
+		osProfileDirs = slackConfigDirDefault
+	}
+	for _, profileDir := range osProfileDirs {
+		files, err := findFileInUserDirs(filepath.Join(profileDir, "slack-teams"), t.logger)
```
directionless: Seems weird to have this `filepath.Join` here, and not just sticking `slack-teams` in the strings above. But 🤷‍♂️

PR #457, opened 2019-03-27
Update to windows service names

directionless:
In https://github.com/kolide/launcher/pull/442 I updated the logic to strip out some characters. But, that’s in the wrong place. It needs to remove them from both the inferred `defaultName`, but also from a specified `ServiceName`. At least, probably… We might argue that if one specifies a bad service name, we shouldn’t correct it and instead through an error. However, correcting it here, allows the identifier to be set consistently across all platforms and modified in the narrow place it’s needed.

### DIFF
--- a/pkg/packagekit/wix/service.go
+++ b/pkg/packagekit/wix/service.go
@@ -110,10 +110,10 @@ type ServiceOpt func(*Service)
 
 func ServiceName(name string) ServiceOpt {
 	return func(s *Service) {
-		s.serviceControl.Id = name
-		s.serviceControl.Name = name
-		s.serviceInstall.Id = name
-		s.serviceInstall.Name = name
+		s.serviceControl.Id = cleanServiceName(name)
+		s.serviceControl.Name = cleanServiceName(name)
+		s.serviceInstall.Id = cleanServiceName(name)
+		s.serviceInstall.Name = cleanServiceName(name)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/packagekit/wix/service.go
```diff
@@ -241,3 +230,18 @@ func (s *Service) Xml(w io.Writer) error {
 	return nil
 
 }
+
+// cleanServiceName removes characters windows doesn't like in
+// services names, and converts everything to snake case.
```
blaedj: > `converts everything to snake case`
Doesn't this conflict with the returned `defaultName`, which is the return value of `snaker.SnakeToCamel()`? seems like this returns camelCase strings

pkg/packagekit/wix/service.go
```diff
@@ -241,3 +230,18 @@ func (s *Service) Xml(w io.Writer) error {
...
+func cleanServiceName(in string) string {
+	r := strings.NewReplacer(
+		"-", "_",
+		" ", "_",
+		".", "_",
+	)
+
+	snakeName := r.Replace(strings.TrimSuffix(in, ".exe") + "_svc")
+	defaultName := snaker.SnakeToCamel(snakeName)
```
groob: you can `return snaker.SnakeToCamel` and avoid the extra variable assignment.

pkg/packagekit/wix/service.go
directionless: Fixed comment.

pkg/packagekit/wix/service.go
directionless: Yeah... I probably didn't for some now removed debugging. Simplified, thanks!

PR #462, opened 2019-04-04
Fix windows service name appending

directionless:
I noticed that we were getting a doubled `SvcSvc` on builds. This was clearly wrong.

Digging, it was because we _always_ striped a trailing `.exe` and then appended a `_svc`. Thinking about it, I think it’s better to do that if, and only if. We’re setting a default name. An explicit ServiceName call should get what it expects. (After the character cleaning, and Camel Casing)

### DIFF
--- a/pkg/packagekit/wix/service.go
+++ b/pkg/packagekit/wix/service.go
@@ -237,12 +237,16 @@ func (s *Service) Xml(w io.Writer) error {
 }
 
 // cleanServiceName removes characters windows doesn't like in
-// services names, and converts everything to camel case.
+// services names, and converts everything to camel case. Right now,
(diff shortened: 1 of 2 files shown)


pkg/packagekit/wix/service.go
```diff
@@ -240,6 +245,5 @@ func cleanServiceName(in string) string {
 		".", "_",
 	)
```
JonathanNogueira: forward and back slashes are also invalid

directionless: Yeah... I should really look up the full rules here, and then make a whitelist. But removing the likely characters was quicker. I added those two.

PR #465, opened 2019-04-18
Require update_channel flag with autoupdate flag in package-builder

zwass:
Now users set `update_channel` and autoupdate will be enabled.


zwass:
That seems fine with me. I will change this to do as you describe.

directionless:
Awesome! Changes around the package builder stuff is pretty free. I don't know that's it's worth getting into refactoring launcher  options directly. Ping me for review of questions

zwass:
This is updated as discussed.

PR #466, opened 2019-04-18
Add table to expose autoupdate configuration

zwass:
kolide_launcher_autoupdate_config table exposes the autoupdate config
via live query or scheduled query when Launcher is running.

Required some refactoring of the options (making struct members public).

### DIFF
--- a/pkg/osquery/table/kolide_launcher_autoupdate_config.go
+++ b/pkg/osquery/table/kolide_launcher_autoupdate_config.go
@@ -1,0 +1,44 @@
+package table
+
+import (
+	"context"
+
+	"github.com/kolide/launcher/pkg/launcher"
+	"github.com/kolide/osquery-go/plugin/table"
+)
+
+const launcherAutoupdateConfigTableName = "kolide_launcher_autoupdate_config"
+
+func LauncherAutoupdateConfigTable(opts *launcher.Options) *table.Plugin {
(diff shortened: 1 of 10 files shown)


zwass:
Screen Shot 2019-04-17 at 5 04 01 PM

pkg/osquery/table/kolide_launcher_autoupdate_config.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+
+	"github.com/kolide/launcher/pkg/launcher"
+	"github.com/kolide/osquery-go/plugin/table"
+)
+
+const launcherAutoupdateConfigTableName = "kolide_launcher_autoupdate_config"
+
+func LauncherAutoupdateConfigTable(opts *launcher.Options) *table.Plugin {
```
directionless: Why here  and not the existing `kolide_launcher_info` table? (I feel agnostic)

zwass: The `kolide_launcher_info` table exposes information that was compiled into Launcher at build time. This table exposes configuration that was supplied at runtime. I think they are somewhat separate concerns.

In general, I'd prefer to keep tables smaller and more single purpose. Users can always join them if they want info from both.

PR #470, opened 2019-04-26
ssh key parsing (initial spike)

directionless:
This creates an `pkg/keyidentifier` that returns key information about arbitrary keys on the filesystem. (This may be better extracted to it's own repo)

Add two osquery tables leveraging it:
* `kolide_keyinfo` which, given a path, returns key info.
* `kolide_ssh_keys` finds keys in user homedirs, and returns them with key info


directionless:
I'm going to break this up a bit, into some other PRs

directionless:
Replaced by:
* https://github.com/kolide/launcher/pull/473
* https://github.com/kolide/launcher/pull/475

PR #476, opened 2019-04-28
Update systemd.md

MarkVLK:
Minor spacing updates and added `=true` after `--autoupdate` option

What version of launcher are you using (launcher --version)?
0.9.4

What operating system are you using?
Ubuntu 18.04

What did you do?
Minor updates to launcher systemd service doc

What did you expect to happen?
N/A

What happened instead?
N/A

### DIFF
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -13,7 +13,7 @@ After=network.service syslog.service
 ExecStart= $LauncherPath \
   --hostname=$FleetServer:FleetPort \
   --enroll_secret=$FleetSecret \
-  --autoupdate=true \
+  --autoupdate \
   --osqueryd_path=$OsquerydPath
 Restart=on-failure
 RestartSec=3
--- a/docs/systemd.md
+++ b/docs/systemd.md
@@ -6,18 +6,19 @@ Below is a sample unit file.
 
 ```
 [Unit]
- Description=The Kolide Launcher
- After=network.service syslog.service
+Description=The Kolide Launcher
+After=network.service syslog.service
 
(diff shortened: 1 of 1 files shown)


docs/systemd.md
```diff
@@ -6,18 +6,19 @@ Below is a sample unit file.
...
- --autoupdate \
---osqueryd_path=$OsquerydPath
+Description=The Kolide Launcher
+After=network.service syslog.service
+
+[Service]
+ExecStart= $LauncherPath \
+  --hostname=$FleetServer:FleetPort \
+  --enroll_secret=$FleetSecret \
+  --autoupdate=true \
```
directionless: This doesn't need `=true` why add it?

MarkVLK: I initially thought it was the cause of an error I was getting when trying to enable the service so added this. The main objective in this update was the messy spacing though, I just left this in because I personally prefer explicit to implicit.

directionless: I admire the desire for explicitness. I'm leery, since I think the history of option parsing has often had boolean options throw errors when supplied with an argument. Golang supports this, but it feels very modern to me. 

I'm happy to just accept the spacing cleanup. I can try to decide whether to accept this as is, or push for it to remain.

MarkVLK: Totally understood. I have no problems with removing the `=true` and just keeping the space updates, if desired.

directionless: Yes, do that please.

MarkVLK: Done

PR #479, opened 2019-05-01
Test proxy server

directionless:
Enable the proxy server. Let's see what the builds do!


directionless:
meh. Will revisit as needed

Issue #445, opened 2019-03-12
error while marshaling: proto: field \"kolide.agent.LogCollection.Log.Data\" contains invalid UTF-8


cyberonstian:
We are in the process of deploying Kolide to our customers and we get this error on multiple of the windows clients. When the error is logged launcher stops sending data to the server.

We have tried updating osqueryd (3.3.2) and launcher (9.0.2) to latest versions but the error still persists.

If we delete the osquery.db files and restart launcher it works for a while and then the error reappears and the log traffic stops.

We have done some investigation and nailed it down to this query:
SELECT * FROM programs 

This works perfectly in Kolide fleet dashboard, but when running it scheduled launcher fails to send the logs due to the error message. We have seen the error message on multiple computers but are still investigation what queries triggers those endpoints.

To nail it down a bit further we found the software in the list that had the characters that triggered the error:
SELECT * FROM programs WHERE name = "HxD Hex Editor version 1.7.7.0"

### What version of `fleet` are you using (`fleet version --full`)?
fleet - version 2.0.2
  branch:       master
  revision:     8ca0358bf28173685815b79d8683a4239d629a14
  build date:   2019-01-18T00:39:40Z
  build user:   zwass
  go version:   go1.11.3¨
### What operating system are you using?
Windows 10 Pro 1893
### What did you do?
1. Install Kolide Fleet agent on the client
2. Start in debug mode
3. Wait a couple of minutes, we see that the server is recieving data from the client
4. The error appears in the logs and no data is recieved on the server.
### What did you expect to see?
Data sent from launcher.exe to kolide fleet
### What did you see instead?

The two first log entries are the error message, the last one is the software that we think is causing them.

{"caller":"publish_logs.go:157","err":"rpc error: code = Internal desc = grpc: error while marshaling: proto: field \"kolide.agent.LogCollection.Log.Data\" contains invalid UTF-8","errcode":"","logType":"string","log_count":3,"message":"","method":"PublishLogs","reauth":false,"severity":"info","took":"0s","ts":"2019-03-12T12:34:15.8004054Z","uuid":"bde64b33-697d-4782-b8a2-866e0a44e71a"}

{"caller":"extension.go:494","err":"sending string logs: writing logs: transport error sending logs: rpc error: code = Internal desc = grpc: error while marshaling: proto: field \"kolide.agent.LogCollection.Log.Data\" contains invalid UTF-8","severity":"info","ts":"2019-03-12T12:34:15.8014026Z"}

[...]

PR #481, opened 2019-05-03
Patch utf8 handling

directionless:
Osquery sometimes mis-encodes utf8 data https://github.com/facebook/osquery/issues/5288

This is a broad attempt to repair log files that exhibit that issue. This runs against the entire log file. Hopefully, there isn’t going to be a case where it misfires.

I'm not sure if this is great. On one hand, it fixes a bug in osquery. On the other hand, it introduces a discrepancy between osquery logs and launcher logs. Feel free to vote with 👍 or 👎. I'm going to let this sit a bit

Fixes: https://github.com/kolide/launcher/issues/445


directionless:
I'm not confident this will behave correctly, and there's one downvote. So I'm going to close this.

directionless:
Sounds like we probably want to revisit this. Probably only triggering repairs on utf8 errors

directionless:
In light of the additional conversation on #445, let's revisit this. 

This is a pretty heavy handed attempt to repair the data, and failing that, redact it, 

I don't have much for a test case though.

directionless:
This was fixed upstream, and I haven't seen this occur in ages. So I'm going to close this PR.

PR #489, opened 2019-05-14
Windows Signing

directionless:
Add package-kit support for authenticode signing through `signtool.exe`. This should be seen as an initial pass at getting general APIs and structure right. Tools may change here.

Reworks the options. Signing keys now have platform specific configuration 

Makefile updated to include launcher code push snippets. These are Kolide specific, and part of how we update notary.

Makefile updated to use `osslsigncode` to sign windows binaries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,6 +175,7 @@ dockerpush-%: docker-%
 
 # Porter is a kolide tool to update notary, part of the update framework
 porter-%: codesign
+	@if [ -z "${NOTARY_DELEGATION_PASSPHRASE}" ]; then echo "Missing NOTARY_DELEGATION_PASSPHRASE"; exit 1; fi
 	for p in darwin linux windows; do \
 	  echo porter mirror -debug -channel $* -platform $$p -launcher-all; \
 	  echo porter mirror -debug -channel $* -platform $$p -extension-tarball -extension-upload; \
--- a/Makefile
+++ b/Makefile
@@ -60,14 +60,15 @@ codesign-darwin: xp
 	codesign --force -s "${CODESIGN_IDENTITY}" -v ./build/darwin/launcher
 	codesign --force -s "${CODESIGN_IDENTITY}" -v ./build/darwin/osquery-extension.ext
 
-
 # Using the `osslsigncode` we can sign windows binaries from
-# non-windows platforms. This is part of our notary upload pipeline.
-codesign-windows-%:
-	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha1 -t http://timestamp.verisign.com/scripts/timstamp.dll -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -askpass
-	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha256 -nest -ts http://sha256timestamp.ws.symantec.com/sha256/timestamp -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -askpass
+# non-windows platforms.
+codesign-windows: codesign-windows-launcher.exe  codesign-windows-osquery-extension.exe
+codesign-windows-%: xp
+	@if [ -z "${AUTHENTICODE_PASSPHRASE}" ]; then echo "Missing AUTHENTICODE_PASSPHRASE"; exit 1; fi
+	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha1 -t http://timestamp.verisign.com/scripts/timstamp.dll -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -pass "${AUTHENTICODE_PASSPHRASE}"
+	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha256 -nest -ts http://sha256timestamp.ws.symantec.com/sha256/timestamp -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -pass "${AUTHENTICODE_PASSPHRASE}"
 
-codesign: codesign-darwin
+codesign: codesign-darwin codesign-windows
 
 package-builder: .pre-build deps
 	go run cmd/make/make.go -targets=package-builder -linkstamp
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,13 @@ codesign-darwin: xp
 	codesign --force -s "${CODESIGN_IDENTITY}" -v ./build/darwin/launcher
 	codesign --force -s "${CODESIGN_IDENTITY}" -v ./build/darwin/osquery-extension.ext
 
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -60,7 +60,15 @@ codesign-darwin: xp
...
 	codesign --force -s "${CODESIGN_IDENTITY}" -v ./build/darwin/osquery-extension.ext
 
-codesign: codesign-darwin
+# Using the `osslsigncode` we can sign windows binaries from
+# non-windows platforms.
+codesign-windows: codesign-windows-launcher.exe  codesign-windows-osquery-extension.exe
+codesign-windows-%: xp
+	@if [ -z "${AUTHENTICODE_PASSPHRASE}" ]; then echo "Missing AUTHENTICODE_PASSPHRASE"; exit 1; fi
+	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha1 -t http://timestamp.verisign.com/scripts/timstamp.dll -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -pass "${AUTHENTICODE_PASSPHRASE}"
+	osslsigncode -in build/windows/$*  -out build/windows/$*  -i https://kolide.com -h sha256 -nest -ts http://sha256timestamp.ws.symantec.com/sha256/timestamp -pkcs12 ~/Documents/kolide-codesigning-2019.p12  -pass "${AUTHENTICODE_PASSPHRASE}"
```
blaedj: should perhaps document the requirement for the `~/Documents/kolide-codesigning-2019.p12` file location in the comment above.

pkg/packagekit/authenticode/authenticode_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+//const notepadeExe = `C:\Windows\System32\Notepad.exe`
```
blaedj: rm this line?

pkg/packagekit/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,97 @@
...
+		if err := so.signtoolSign(ctx, file, "/fd", "sha1", "/t", so.timestampServer); err != nil {
+			return errors.Wrap(err, "signing msi with sha1")
+		}
+	} else {
+		if err := so.signtoolSign(ctx, file, "/fd", "sha1", "/t", so.timestampServer); err != nil {
+			return errors.Wrap(err, "signing msi with sha1")
+		}
+
+		if err := so.signtoolSign(ctx, file, "/as", "/fd", "sha256", "/td", "sha256", "/tr", so.rfc3161Server); err != nil {
+			return errors.Wrap(err, "signing msi with sha1")
```
blaedj:
```suggestion
			return errors.Wrap(err, "signing msi with sha256")
```

pkg/packagekit/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,97 @@
...
+	// signatures. _But_ it uses different arguments for the subsequent
+	// signatures. So, multiple calls.
+	//
+	// _However_ it's not clear this is supported for MSIs, which maybe
+	// only have a single slot for signing.
+	//
+	// References:
+	// https://knowledge.digicert.com/generalinformation/INFO2274.html
+	if strings.HasSuffix(file, ".msi") {
+		if err := so.signtoolSign(ctx, file, "/fd", "sha1", "/t", so.timestampServer); err != nil {
```
blaedj: The linked document seems to recommend only using sha1 to sign if it's necessary due to 'legacy systems'. I'm not 💯 on what 'legacy system' entails, but do we need to use sha1 for a particular reason?

pkg/packagekit/authenticode/authenticode_windows.go
```diff
@@ -0,0 +1,97 @@
...
+			return errors.Wrap(err, "signing msi with sha1")
+		}
+	} else {
+		if err := so.signtoolSign(ctx, file, "/fd", "sha1", "/t", so.timestampServer); err != nil {
+			return errors.Wrap(err, "signing msi with sha1")
```
blaedj: this and line 66 probably shouldn't say `signing msi..` seeing as the file doesn't end in `.msi`, right?

pkg/packagekit/package_wix.go
```diff
@@ -82,8 +88,31 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
...
+	}
+
+	// Sign?
+	if po.WindowsUseSigntool {
+		if err := authenticode.Sign(
+			ctx, msiFile,
+			authenticode.WithExtraArgs(po.WindowsSigntoolArgs),
+			authenticode.WithSigntoolPath(signtoolPath),
+		); err != nil {
+			return errors.Wrap(err, "authencode signing")
```
blaedj:
```suggestion
			return errors.Wrap(err, "authenticode signing")
```

pkg/packagekit/package_wix.go
```diff
@@ -82,8 +88,31 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
...
-	if err := wixTool.Package(ctx, w); err != nil {
-		return errors.Wrap(err, "running light")
+	msiFile, err := wixTool.Package(ctx)
+	if err != nil {
+		return errors.Wrap(err, "wix packaging")
+	}
+
+	// Sign?
+	if po.WindowsUseSigntool {
+		if err := authenticode.Sign(
```
blaedj: nit, but if we're doing a multi-line method call, it seems a bit convoluted to use the conditional assignment instead of 

```go
err := authenticode.Sign(...)
if err != nil {
// ...
}

Makefile
directionless: The `codesign-` targets have a slew of assumptions. I'm going to leave them, the errors are reasonable straight forward, and it's all slowly being replaced by the packagekit side.

pkg/packagekit/authenticode/authenticode_test.go
directionless: oops. yeah

pkg/packagekit/package_wix.go
directionless: TBH I like the conditional assignment, solely because I don't have to think about whether `err` was already defined.

Makefile
blaedj: ok, yea as long as the errors are clear I'm fine with it.

pkg/packagekit/authenticode/authenticode_windows.go
directionless: Testing, I think you're right. we can use sha256 for MSIs. I've updated

PR #492, opened 2019-05-28
Make GUN prefix configurable

jalseth:
Adds the ability to change the GUN prefix from a hardcoded `kolide/` using a `--gun_prefix` configuration flag.


pkg/autoupdate/autoupdate.go
```diff
@@ -64,7 +65,7 @@ func NewUpdater(binaryPath, rootDirectory string, logger log.Logger, opts ...Upd
 	strippedBinaryName := strings.TrimSuffix(binaryName, ".exe")
 	tufRepoPath := filepath.Join(rootDirectory, fmt.Sprintf("%s-tuf", strippedBinaryName))
 	stagingPath := filepath.Join(filepath.Dir(binaryPath), fmt.Sprintf("%s-staging", binaryName))
-	gun := fmt.Sprintf("kolide/%s", strippedBinaryName)
+	gun := fmt.Sprintf("%s%s", gunPrefix, strippedBinaryName)
```
zwass: Prefer `path.Join`

pkg/launcher/options.go
```diff
@@ -55,6 +55,8 @@ type Options struct {
 	AutoupdateInterval time.Duration
 	// UpdateChannel is the channel to pull options from (stable, beta, nightly).
 	UpdateChannel autoupdate.UpdateChannel
+	// The GUN prefix for where the launcher and osqueryd collections are
```
zwass: GUNPrefix is the path prefix used to store launcher and osqueryd binaries on the Notary server

jalseth:
Yes, that's the motivation for adding this flag.

cmd/launcher/options.go
```diff
@@ -53,6 +53,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		flMirrorURL          = flagset.String("mirror_url", autoupdate.DefaultMirror, "The mirror server for autoupdates (default: https://dl.kolide.co)")
 		flAutoupdateInterval = flagset.Duration("autoupdate_interval", 1*time.Hour, "The interval to check for updates (default: once every hour)")
 		flUpdateChannel      = flagset.String("update_channel", "stable", "The channel to pull updates from (options: stable, beta, nightly)")
+		flGUNPrefix          = flagset.String("gun_prefix", autoupdate.DefaultGUNPrefix, "The prefix for the GUN that contains the collections (default: kolide/)")
```
directionless: I would be inclined to call this `notary_prefix`. It's notary specific, and I find the `GUN` terminology somewhat distasteful.

pkg/autoupdate/autoupdate.go
```diff
@@ -54,7 +55,7 @@ type Updater struct {
 
 // NewUpdater creates a unstarted updater for a specific binary
 // updated from a TUF mirror.
-func NewUpdater(binaryPath, rootDirectory string, logger log.Logger, opts ...UpdaterOption) (*Updater, error) {
+func NewUpdater(binaryPath, rootDirectory string, gunPrefix string, logger log.Logger, opts ...UpdaterOption) (*Updater, error) {
```
directionless: This should be an `UpdaterOption`, not a change to the function signature. For example, https://github.com/kolide/launcher/blob/4c0320d61d2b05fdd79bc32e905e2bf670937cfb/pkg/autoupdate/autoupdate.go#L217

seph:
Not sure why I’m being included in this thread. This isn’t a project I’m familiar with.

directionless:
Sorry @seph -- they meant to tag me with that.

pkg/autoupdate/autoupdate.go
```diff
@@ -33,8 +33,9 @@ const (
 )
 
 const (
-	DefaultMirror = "https://dl.kolide.co"
-	DefaultNotary = "https://notary.kolide.co"
+	DefaultMirror       = "https://dl.kolide.co"
+	DefaultNotary       = "https://notary.kolide.co"
+	DefaultNotaryPrefix = "kolide/"
```
directionless: If you're using `path.Join`, skip the `/`
```suggestion
	DefaultNotaryPrefix = "kolide"
```

pkg/autoupdate/autoupdate_test.go
```diff
@@ -68,6 +68,7 @@ func TestNewUpdater(t *testing.T) {
 				WithUpdateChannel(Beta),
 				WithNotaryURL("https://notary"),
 				WithMirrorURL("https://mirror"),
+				WithNotaryPrefix("kolide/"),
```
directionless: Shouldn't need this. Better to test the default without
```suggestion
```

pkg/autoupdate/autoupdate.go
```diff
@@ -64,7 +65,7 @@ func NewUpdater(binaryPath, rootDirectory string, logger log.Logger, opts ...Upd
 	strippedBinaryName := strings.TrimSuffix(binaryName, ".exe")
 	tufRepoPath := filepath.Join(rootDirectory, fmt.Sprintf("%s-tuf", strippedBinaryName))
 	stagingPath := filepath.Join(filepath.Dir(binaryPath), fmt.Sprintf("%s-staging", binaryName))
-	gun := fmt.Sprintf("kolide/%s", strippedBinaryName)
+	gun := path.Join(DefaultNotaryPrefix, strippedBinaryName)
```
directionless: I know this is the existing code, but it seems cleaner to skip the `gun` assignment, and just stick the `path.Join(...)` into the `tuf.Settings` struct a couple lines down

Issue #488, opened 2019-05-13
Table Request: TouchID configuration


FritzX6:
Describe the Feature being requested:
---
Since late 2016 macOS has included the capability for unlocking devices using their TouchID biometric authentication. While the bulk of TouchID is handled in the Secure Enclave of the T1 and T2 chips, we can still introspect the state of TouchID configuration using the `bioutil` command line utility.

How does this Feature help?
---
Many institutions have internal security policies regarding the utilization of biometric authentication and the ability to display this information can be utilized to ensure the feature is enabled or disabled.

Sample Output of bioutil command:
---
```
➜  ~ bioutil -r
User Touch ID configuration:
    Touch ID for unlock: 1
    Touch ID for ApplePay: 1
    Effective Touch ID for unlock: 1
    Effective Touch ID for ApplePay: 1
Operation performed successfully.
➜  ~ bioutil -c
User 501:    2 fingerprint(s)
Operation performed successfully.
➜  ~ bioutil -r -s
System Touch ID configuration:
    Touch ID functionality: 1
    Touch ID for unlock: 1
Operation performed successfully.
➜  ~ sudo bioutil -c -s
Password:
User 501:    2 fingerprint(s)
Operation performed successfully.
```

How would we want to expose this data?
---
Two tables: 
- [ ] `touchid_system_config` reads whether a device is capable of TouchID and its respective system-wide settings.
- [ ] `touchid_user_config` reads the individual configuration settings of TouchID for each user

Example output of proposed tables:
---
```
Kolide-iMac-Pro> osqueryi

osquery> SELECT * FROM touchid_system_config;
+--------------------+--------------------+-----------------+----------------+
| touchid_compatible | secure_enclave_cpu | touchid_enabled | touchid_unlock |
+--------------------+--------------------+-----------------+----------------+
| false              | T2                 | false           | false          |
+--------------------+--------------------+-----------------+----------------+
```

```
Kolide-Touchbar-MacbookPro> osqueryi

osquery> SELECT * FROM touchid_user_config;
+-----+----------------+------------------+------------------+--------------------+-------------------------+
| uid | touchid_unlock | touchid_applepay | effective_unlock | effective_applepay | fingerprints_registered |
+-----+----------------+------------------+------------------+--------------------+-------------------------+
| 501 | true           | true             | true             | true               | 2                       |
+-----+----------------+------------------+------------------+--------------------+-------------------------+
```

bioutil man Output
---
```
[...]

PR #502, opened 2019-07-26
Add TouchID tables

jalseth:
This adds the ability to read a Mac's TouchID configuration as well as each user's TouchID configuration, addressing #488.

I'm not thrilled about how `enumerateUIDs()` works, so I would especially appreciate your input on that function.


pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,168 @@
...
+
+func enumerateUIDs() ([]int, error) {
+	// Enumerate all of the files in /Users
+	files, err := ioutil.ReadDir("/Users")
+	if err != nil {
+		return nil, errors.Wrap(err, "enumerating files and folders in /Users")
+	}
+
+	// If it's a directory, add it to the array
+	user_folders := make([]string, 0)
```
groob: Use camelCase for Go

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,168 @@
...
+		}
+	}
+
+	// Run dscl on each directory to get its owner's UID and add it to the slice
+	uids := make([]int, 0)
+	for _, folder := range user_folders {
+		var stdout bytes.Buffer
+		cmd := exec.Command("/usr/bin/dscl", ".", "-read", folder, "UniqueID")
+		cmd.Stdout = &stdout
+		cmd.Run()
```
groob: Error handling with logs here would potentially help with debugging failures

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,168 @@
...
+	files, err := ioutil.ReadDir("/Users")
+	if err != nil {
+		return nil, errors.Wrap(err, "enumerating files and folders in /Users")
+	}
+
+	// If it's a directory, add it to the array
+	user_folders := make([]string, 0)
+	for _, file := range files {
+		if file.IsDir() {
+			user_folders = append(user_folders, filepath.Join("/Users", file.Name()))
```
groob: /Users is not necessarily the correct place for the user directories

terracatta:
@jalseth Thanks so much for the PR!

I am getting a panic when I run `SELECT * FROM touchid_user_config;`

```
osquery> select * from touchid_user_config;
osquery> 2019/07/26 15:34:41 panic in processor: runtime error: index out of range: goroutine 11 [running]:
runtime/debug.Stack(0xc0000eb858, 0x479d880, 0x4dec9a0)
	/usr/local/Cellar/go/1.12/libexec/src/runtime/debug/stack.go:24 +0x9d
git.apache.org/thrift.git/lib/go/thrift.(*TSimpleServer).processRequests.func1()
	/Users/jmeller/go/pkg/mod/git.apache.org/thrift.git@v0.12.0/lib/go/thrift/simple_server.go:198 +0x57
panic(0x479d880, 0x4dec9a0)
	/usr/local/Cellar/go/1.12/libexec/src/runtime/panic.go:522 +0x1b5
github.com/kolide/launcher/pkg/osquery/table.enumerateUIDs(0x4764100, 0x4e1f080, 0x8, 0x473c080, 0xc0000ba6a0)
	/Users/jmeller/go/src/github.com/kolide/launcher/pkg/osquery/table/touchid_user_darwin.go:76 +0x70b
github.com/kolide/launcher/pkg/osquery/table.(*touchIDUserConfigTable).generate(0xc00000e380, 0x4942f20, 0xc0000d8078, 0xc0000b1d40, 0xc0000c0198, 0x7, 0xc0000c0190, 0x2, 0x8)
	/Users/jmeller/go/src/github.com/kolide/launcher/pkg/osquery/table/touchid_user_darwin.go:95 +0x37
github.com/kolide/osquery-go/plugin/table.(*Plugin).Call(0xc0001a2570, 0x4942f20, 0xc0000d8078, 0xc0000b1d10, 0xc000022558, 0x1, 0x0, 0xc00022ad30)
	/Users/jmeller/go/pkg/mod/github.com/kolide/osquery-go@v0.0.0-20190113061206-be0a8de4cf1d/plugin/table/table.go:68 +0x313
github.com/kolide/osquery-go.(*ExtensionManagerServer).Call(0xc0001a6000, 0x4942f20, 0xc0000d8078, 0xc0000c00d1, 0x5, 0xc0001d6020, 0x13, 0xc0000b1d10, 0xc000000000, 0x4014c77, ...)
	/Users/jmeller/go/pkg/mod/github.com/kolide/osquery-go@v0.0.0-20190113061206-be0a8de4cf1d/server.go:241 +0x110
github.com/kolide/osquery-go/gen/osquery.(*extensionProcessorCall).Process(0xc00006a2c0, 0x4942f20, 0xc0000d8078, 0x0, 0x4950d80, 0xc000196510, 0x4950d80, 0xc0001965a0, 0x4335ed9, 0x47aa0a0, ...)
	/Users/jmeller/go/pkg/mod/github.com/kolide/osquery-go@v0.0.0-20190113061206-be0a8de4cf1d/gen/osquery/osquery.go:1165 +0x2ed
github.com/kolide/osquery-go/gen/osquery.(*ExtensionProcessor).Process(0xc00000e400, 0x4942f20, 0xc0000d8078, 0x4950d80, 0xc000196510, 0x4950d80, 0xc0001965a0, 0x0, 0x0, 0xc00007e480)
	/Users/jmeller/go/pkg/mod/github.com/kolide/osquery-go@v0.0.0-20190113061206-be0a8de4cf1d/gen/osquery/osquery.go:1084 +0x347
git.apache.org/thrift.git/lib/go/thrift.(*TSimpleServer).processRequests(0xc0001a6900, 0x49475e0, 0xc0001a2810, 0x0, 0x0)
	/Users/jmeller/go/pkg/mod/git.apache.org/thrift.git@v0.12.0/lib/go/thrift/simple_server.go:213 +0x27c
git.apache.org/thrift.git/lib/go/thrift.(*TSimpleServer).innerAccept.func1(0xc0001a6900, 0x49475e0, 0xc0001a2810)
	/Users/jmeller/go/pkg/mod/git.apache.org/thrift.git@v0.12.0/lib/go/thrift/simple_server.go:143 +0x73
created by git.apache.org/thrift.git/lib/go/thrift.(*TSimpleServer).innerAccept
	/Users/jmeller/go/pkg/mod/git.apache.org/thrift.git@v0.12.0/lib/go/thrift/simple_server.go:141 +0x140
```

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -0,0 +1,88 @@
...
+		logger: logger,
+	}
+	columns := []table.ColumnDefinition{
+		table.IntegerColumn("touchid_compatible"),
+		table.TextColumn("secure_enclave_cpu"),
+		table.IntegerColumn("touchid_enabled"),
+		table.IntegerColumn("touchid_unlock"),
+	}
+
+	return table.NewPlugin("touchid_system_config", columns, t.generate)
```
terracatta: Please prepend all table names with `kolide_` so we don't accidentally conflict with other plugins

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,168 @@
...
+	columns := []table.ColumnDefinition{
+		table.IntegerColumn("uid"),
+		table.IntegerColumn("fingerprints_registered"),
+		table.IntegerColumn("touchid_unlock"),
+		table.IntegerColumn("touchid_applepay"),
+		table.IntegerColumn("effective_unlock"),
+		table.IntegerColumn("effective_applepay"),
+	}
+
+	return table.NewPlugin("touchid_user_config", columns, t.generate)
```
terracatta: Please prepend all table names with `kolide_` so we don't accidentally conflict with other plugins

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// TouchIDSystemConfigGenerate will be called whenever the table is queried.
+func (t *touchIDSystemConfigTable) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+	var touchIDCompatible, secureEnclaveCPU, touchIDEnabled, touchIDUnlock string
+
+	// Read the security chip from system_profiler
+	var stdout bytes.Buffer
+	cmd := exec.Command("/usr/sbin/system_profiler", "SPiBridgeDataType")
```
directionless: Consider using `-xml` and parsing that. I think there's more info

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -0,0 +1,88 @@
...
+
+	// Read the system's bioutil configuration
+	stdout.Reset()
+	cmd = exec.Command("/usr/bin/bioutil", "-r", "-s")
+	cmd.Stdout = &stdout
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrap(err, "calling bioutil for system configuration")
+	}
+	configOutStr := string(stdout.Bytes())
+	configSplit := strings.Split(configOutStr, ":")
```
directionless: This feels fragile to line order changes. Is there another way to parse this?

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,143 @@
...
+		if err := cmd.Run(); err != nil {
+			level.Debug(t.logger).Log(
+				"msg", "Failed to run bioutil for configuration",
+				"uid", uid,
+				"err", err,
+			)
+			continue
+		}
+		configOutStr := string(stdout.Bytes())
+		configSplit := strings.Split(configOutStr, ":")
```
directionless: The exec and parse `bioutil` is repeated. What do you think about pulling into a shared function with an arg for system or user? (not a full package)

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,143 @@
...
+				"msg", "bioutil -r returned unexpected output",
+				"uid", uid,
+				"err", "bad output",
+			)
+			continue
+		}
+
+		// Grab the fingerprint count
+		stdout.Reset()
+		cmd = exec.Command("/usr/bin/bioutil", "-c")
```
directionless: I wonder if this would ever want to be run as root (returns the count for the whole system)

pkg/osquery/table/touchid_system_darwin.go
directionless: `exec.CommandContext` please

jalseth: I saw this flag and have considered this, but I'm just more familiar with regex than parsing XML in Go so I went this route. The only other info is the firmware version for the chip, which I believe (could be wrong) is directly tied to the macOS version so that information can be inferred that way.

pkg/osquery/table/touchid_system_darwin.go
jalseth: It is fragile, but `bioutil` does not have an option for structured output. I considered using multiple regexs to parse the values out which would be resilient to small changes in the output format, but it would not work with output in languages other than English. I am working under the assumption that bioutil outputs in other languages, I haven't tested that.

pkg/osquery/table/touchid_user_darwin.go
jalseth: I don't think that is a good idea, at least for now. To my knowledge, with the current way of parsing the output, it would just move the code into a different function and maybe save a few lines of code. I think the way it is laid out currently is easier to read even if slightly less efficient.

pkg/osquery/table/touchid_user_darwin.go
jalseth: I think we would be better served by updating the code to not require a UID specified so you can do something like `SELECT * FROM kolide_touchid_user_config` to retrieve TouchID config for multiple users without knowing their UIDs ahead of time and then use the SQL `SUM()` function to add them together if you want the total fingerprint count for a given system, or systems.

pkg/osquery/table/touchid_system_darwin.go
directionless: Yeah... Anything going line by line and matching the key, would be subject to issues around language. Maybe leave a comment explaining why?

pkg/osquery/table/touchid_user_darwin.go
directionless: I'm thinking of the regext parsing code, and positional parameters. I'm pretty sure if I find a bug in 6 months, I'll forget to change one of them

pkg/osquery/table/touchid_user_darwin.go
directionless: I was thinking adding the as-root part to the `TouchIDSystemConfig` table.

But that raises a good question -- my bias for this sort of data, is to require uid, and let people join against the users table if desired. It's also how you've written it. 👍

directionless:
The CLA tool is acting funny. I checked the existing signatures, and @jalseth signed it previously

ibakshay:
> The CLA tool is acting funny. I checked the existing signatures, and @jalseth signed it previously

Hi @directionless, Could you please tell what exactly happened so that I can give you insights about the CLA Assistant tool?

directionless:
@ibakshay -- https://github.com/cla-assistant/cla-assistant/issues/465 it's no longer clear if this is my error or a delay in processing. Discussion should happen on that ticket, not here, however.

terracatta:
@jalseth Thank for for the changes...

I am on a T1 Macbook Pro (15-inch, 2017) and `select * from kolide_touchid_user_config;` returns no results. I also tried `select * from kolide_touchid_user_config where uid = 501` and still no results.

Here is the output from running `bioutil -c` as my local user in the terminal

```
☁  launcher [add-touchid-tables] ⚡ /usr/bin/bioutil -c
User 501:	1 fingerprint(s)
Operation performed successfully.
```

Here is the output from running `bioutil -r` as my local user in the terminal

```
☁  launcher [add-touchid-tables] ⚡ /usr/bin/bioutil -r
User Touch ID configuration:
	Touch ID for unlock: 1
	Touch ID for ApplePay: 1
	Effective Touch ID for unlock: 1
	Effective Touch ID for ApplePay: 1
Operation performed successfully.
```

jalseth:
@terracatta In the current implementation, `touchid_user_config` will only work if the uid is specified as it does not have a way to enumerate the uids for all the users on the system itself. 

It's working for me on a 2017 MBP with a T1 chip. I'd be interested in the launcher debug output from your machine as it should log exactly where it's running into issues. 

```
SELECT * 
FROM kolide_touchid_user_config
WHERE uid = 501
```

Returns

```
"host_hostname","effective_applepay","effective_unlock","fingerprints_registered","touchid_applepay","touchid_unlock","uid"
"ml-jalseth-01.local","1","1","2","1","1","501"
```

And 

```
SELECT * 
FROM kolide_touchid_system_config
```

Returns

```
"host_hostname","secure_enclave_cpu","touchid_compatible","touchid_enabled","touchid_unlock"
"ml-jalseth-01.local","T1","1","1","1"
```

terracatta:
@jalseth here is the verbose output.

```
         select * from kolide_touchid_user_config where uid = 501;
{"caller":"touchid_user_darwin.go:85","err":"fork/exec /usr/bin/bioutil: operation not permitted","msg":"Failed to run bioutil for configuration","severity":"debug","ts":"2019-07-27T17:54:22.836378Z","uid":501}
```

if I run the extension with sudo it works, but interestingly, I do not need sudo to run the underlying `bioutil -r` command. Is this expected?

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,143 @@
...
+	uid                    int
+	fingerprintsRegistered int
+	touchIDUnlock          int
+	touchIDApplePay        int
+	effectiveUnlock        int
+	effectiveApplePay      int
+}
+
+func (t *touchIDUserConfigTable) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	q, _ := queryContext.Constraints["uid"]
```
groob: Check ok and return an err if uid is missing.

directionless:
I suspect I know what's going on with @terracatta's reported failure. 

The code as written has a really nice use of `cmd.SysProcAttr.Credential` to invoke `bioutil` as the target uid. That call is going to fail when you're not running as root. But, this presents a bit of a thorny question about what should happen.

If we're running as root, we can invoke as a target user and get desired data. 
But if we're not running as root, what should happen? We can only inspect for ourselves.

One possibility is to compare the current user with the uid requested, and short cut if they match. (and error if they don't) If folks think that makes sense, I'm happy to try writing up a quick exec library for that. I've got some ideas I wouldn't mind getting in

groob:
macOS also behaves differently in regards to user context switches when running as a launchdaemon vs when invoked with `sudo`. It needs to be better tested. `launchctl asuser`  could be useful here. 

https://derflounder.wordpress.com/2016/03/25/running-processes-in-os-x-as-the-logged-in-user-from-outside-the-users-account/

groob:
According to a friend who poked at the binary in Hopper, bioutil is looking for the euid.

jalseth:
Honest question: How often do we **_not_** run launcher/osquery as root? More osquery features than this one require root privileges to work correctly.

terracatta:
> Honest question: How often do we not run launcher/osquery as root? More osquery features than this one require root privileges to work correctly.

Honest answer: The reason I caught this is we hardly ever run launcher as root when testing the launcher on our commercial product in local development. I will typically run launcher with `sudo` when I am explicitly testing the following virtual tables (all of which are osquery)

* Gatekeeper Status
* Gatekeeper Approved Apps
* Sudoers

I also very rarely run `osqueryi` or our `make osqueryi-tables` commands as root unless I know I am testing a table where that matters. It makes it much easier for me to just quickly get a sense of data without the cognitive overhead of reasoning about joins for certain tables like `chrome_extensions` 

As a general rule of thumb, if I can access the data without root on the command line, I expect that the osquery equivalent virtual table has the same authorization requirements. This is why I was confused by the lack of output.

It's ultimately up to you. Is this useful to us in its current with it only working on root? Yes. 
Is it a better contribution for the community without the artificial root restriction? Yes.

jalseth:
Added support for this, as well as moved running the command into its own function to reduce duplicated code.

pkg/osquery/table/touchid_user_darwin.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// runCommand runs a given command and arguments as the supplied user
+func runCommandContext(ctx context.Context, uid int, cmd string, args ...string) (string, error) {
+	// Set up the command
+	var stdout bytes.Buffer
+	c := exec.CommandContext(ctx, cmd, args...)
+	c.Stdout = &stdout
+
+	// Check if the supplied UID is that of the current user
```
directionless: Nice <3

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -0,0 +1,88 @@
...
+	touchIDEnabled    int
+	touchIDUnlock     int
+}
+
+// TouchIDSystemConfigGenerate will be called whenever the table is queried.
+func (t *touchIDSystemConfigTable) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+	var touchIDCompatible, secureEnclaveCPU, touchIDEnabled, touchIDUnlock string
+
+	// Read the security chip from system_profiler
```
directionless: The system_profiler stuff should never change. So it could be moved up into TouchIDSystemConfig. (I'll accept this as-is, and am willing to do that in future work)

directionless:
I filed something with apple about the bioutil issue.

https://feedbackassistant.apple.com/feedback/7214386 (though I suspect you need my account to read that)

directionless:
Digging at the bioutil thing some more, it looks like bioutil spits back 1, for users that have never configured it. I think this is clearly an apple bug. It does make those values suspect if fingerprint count is 0. 

I wonder if it makes sense force all the values to 0, if there are no fingerprints enabled

jalseth:
That's very odd and definitely a bug. Even with a brand new macOS user with TouchID never set up `bioutil -r` reports it is enabled. 

What is more interesting, is running it as root with `sudo bioutil -r` it dumps an empty config as expected for a user without Touch ID enabled, which is why I check for a split length of 2 at https://github.com/kolide/launcher/pull/502/files#diff-9dc6cceb8739b1d49bcf678326e50937R92.

For now, I'll add a check for if the fingerprint count is 0 then the effective Touch ID configuration should always return 0. This way when Apple fixes the issue, the code will return the correct values without any changes.

directionless:
The build failure is due to a thrift go change. I'm trying to resolve it in https://github.com/kolide/launcher/pull/516

pkg/osquery/table/touchid_system_darwin.go
jalseth: I'd like to look into this before we merge. Please hold off on merging for now.

jalseth:
Cherry-picked the commit. Should be all good now, but I'd like to look into your other comment before you merge.

PR #494, opened 2019-05-31
Add env variable support to builder.go for NotaryURL and NotaryPrefix

jalseth:
This adds environment variable support for `NotaryURL` and `NotaryPrefix` so they can be specified as part of a build pipeline without needing to change files on disk.

PR #503, opened 2019-07-29
Rework generate TUF

directionless:
This reworks the generate-TUF routine to solve several annoyances. The underlying change, is a shift away from some hardcode notary configs, to generating them at build time. The underlying `bootstrapFromNotary` remains much the same.

* No longer hard codes a path, thus removing the dependency on GOPATH
* Allows replacement of the notary server (an implicit request in https://github.com/kolide/launcher/pull/494)


directionless:
@jalseth I'm curious if this works for you. I, uh, haven't tested it yet. :)

directionless:
This is super stale, and I don't think we're going to do this

PR #515, opened 2019-09-04
Upgrade osquery-go dependancies (change the thrift location)

directionless:
Upgrade osquery-go, to pull in https://github.com/kolide/osquery-go/pull/72/ and fix the thrift git location.


directionless:
This should happen off a personal fork

Issue #467, opened 2019-04-25
Windows install permissions issue


directionless:
I've had a couple reports that the windows MSI installs, but then fails to start. Producing an error like:

![image (1)](https://user-images.githubusercontent.com/179542/56749099-11c96780-674f-11e9-85c0-0d44f1e8f834.png)

Launcher needs admin permission. I _think_ I've seen this when the MSI installs, but then it tries to start the service as the user, not as admin.

directionless:
We're installing the service with  `NT AUTHORITY\SYSTEM`, we suspect it should get installed using `LocalSystem` instead. I'm currently testing a build that uses `[SERVICEACCOUNT]` which I think is a wix shortcut.

directionless:
Digging a lot more into this, this sounds like a generic windows error. If the service install fails, for any reason, it ends up presenting this way.

The internet suggests common reasons to be:
* Service already exists
* Missing dependency

directionless:
I was able to get a msi exec install log file, but it doesn't seem very helpful. From the service install:

```
MSI (s) (70:7C) [10:13:50:092]: Executing op: ServiceInstall(Name=LauncherKolideK2Svc,DisplayName=LauncherKolideK2Svc,ImagePath="C:\Program Files\Kolide\Launcher-kolide-k2\bin\launcher.exe" svc -config "C:\Program Files\Kolide\Launcher-kolide-k2\conf\launcher.flags",ServiceType=16,StartType=2,ErrorControl=32769,,Dependencies=[~],,,Password=**********,Description=The Kolide Launcher (kolide-k2),,)
MSI (s) (70:7C) [10:13:50:093]: Note: 1: 2205 2:  3: Error 
MSI (s) (70:7C) [10:13:50:093]: Note: 1: 2228 2:  3: Error 4: SELECT `Message` FROM `Error` WHERE `Error` = 1923 
MSI (c) (60:2C) [10:13:50:096]: Font created.  Charset: Req=0, Ret=0, Font: Req=MS Shell Dlg, Ret=MS Shell Dlg

Error 1923. Service 'LauncherKolideK2Svc' (LauncherKolideK2Svc) could not be installed.  Verify that you have sufficient privileges to install system services.
```

Picking this apart:

* docs -- https://docs.microsoft.com/en-us/windows/desktop/Msi/serviceinstall-table
* `ServiceType=16` -- looks right -- own process
* `StartType=2` -- start on boot. Looks right
[...]

PR #517, opened 2019-09-06
Adjust MSI Service Settings

directionless:
If there are service install problems, say because a prior install left the service in a weird state, `Vital=yes` means the MSI install will fail.

Adjust to `Vital=no`, and `Wait=yes` to see how that behaves

Relates To: https://github.com/kolide/launcher/issues/467


directionless:
I'm not sure this is a great idea. If the service fails to install for some other reason, it will still be a successful MSI install

directionless:
I'm going to close this, but we can always re-open or grab pieces.

PR #518, opened 2019-09-06
Helper osquery commands in the Makefile

directionless:
I often need to start osqueryi with the tables.ext in place.

This adds various forms of doing so to the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,12 @@ table.ext: .pre-build
 	go run cmd/make/make.go -targets=table-extension -linkstamp
 
 osqueryi-tables: table.ext
-	osqueryd -S --extension ./build/darwin/tables.ext
+	osqueryd -S --allow-unsafe --extension ./build/darwin/tables.ext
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,7 +35,12 @@ table.ext: .pre-build
 	go run cmd/make/make.go -targets=table-extension -linkstamp
 
 osqueryi-tables: table.ext
-	osqueryd -S --extension ./build/darwin/tables.ext
+	osqueryd -S --allow-unsafe --extension ./build/darwin/tables.ext
+sudo-osqueryi-tables: table.ext
```
zwass: Is there a reason to prefer this over `sudo make osqueryi-tables`?

directionless: That would trigger some weird build and mod caching as root, likely leaving some random files on disk not owned by the user.

I am, generally speaking, against running commands like `make` under sudo, they have a lot unexpected effect. 

I'm not totally sure we need these targets. They're here primarily for convinience

PR #524, opened 2019-10-10
Autoupdate v2

directionless:
Substantial rework of autoupdate.  From the [ADR]:(https://github.com/directionless/launcher/blob/seph/autoupdate-v2/docs/architecture/2019-09-05_autoupdate_v2.md)

> Our current update process has several flaws. These lead towards an unreliable autoupdate process.
>
> Instead of replacing the running binary, we will create a directory to store updates in. These updates will be launched from there.

I'm amenable to splitting this up, though the pieces may not make sense on their own.


cmd/launcher/launcher.go
```diff
@@ -4,183 +4,37 @@ import (
...
-	signal.Notify(sig, os.Interrupt, os.Kill, syscall.Signal(15))
-	<-sig
-
-	// allow for graceful termination.
-	runner.Shutdown()
-
-	return nil
-}
+// runLauncher is the entry point into running launcher. It creates a
+// rungroups with the various options, and goes! Autoupdate will kill the entire rungroup, and trigger FIXME
```
blaedj: fill in `FIXME`?

pkg/autoupdate/findnew.go
```diff
@@ -0,0 +1,177 @@
...
+		opt(newestSettings)
+	}
+
+	updateDir := getUpdateDir(fullBinaryPath)
+	binaryName := filepath.Base(fullBinaryPath)
+
+	fileGlob := filepath.Join(updateDir, "*", binaryName)
+
+	possibleUpdates, err := filepath.Glob(fileGlob)
+	if err != nil {
```
blaedj: `Glob(..)` returns nil if no matches are found, should we return early or if there are no matches found _and_ the err is nil? Not sure what that case means (probably that launcher is running but it's on-disk binary is moved or missing..?)
Reading the docs, I think we should probably handle this case, since errors are only thrown for invalid input:
`Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed.`

Also, further on in the code it looks like this will actually be handled fine, but I think I'd want some logs for this if I was debugging it.

pkg/autoupdate/findnew.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// iterate backwards over files, looking for a suitable binary
+	foundCount := 0
+	foundFile := ""
+	for i := len(possibleUpdates) - 1; i >= 0; i-- {
+		file := possibleUpdates[i]
+
+		// If we've already found at least 2 files, (newest, and presumed
+		// current), trigger delete routine
+		if newestSettings.deleteOld && foundCount >= 2 {
```
blaedj: I assume that in empirical testing this works as expected, but do we have any guarantees about how the matches from `Glob` are sorted across platforms?

blaedj: it could be that I just need to read the test files still..

pkg/autoupdate/findnew_test.go
```diff
@@ -0,0 +1,258 @@
...
+	emptySetup setupState = iota
+	emptyUpdateDirs
+	nonExecutableUpdates
+	executableUpdates
+)
+
+// setupTestDir function to setup the test dirs. This work is broken
+// up in stages, allowing test functions to tap into various
+// points. This is setup this way to allow simpler isolation on test
+// faiulures.
```
blaedj:
```suggestion
// failures.
```

pkg/autoupdate/handler.go
```diff
@@ -1,52 +1,115 @@
...
+		updateDir := filepath.Join(u.updatesDirectory, strconv.FormatInt(time.Now().Unix(), 10))
+
+		// Note that this is expecting the binary in the
+		// tarball to be named binaryName. There some some
+		// extension weirdness issues on windows vs posix.
+		outputBinary := filepath.Join(updateDir, u.binaryName)
+
+		if err := os.MkdirAll(updateDir, 0755); err != nil {
+			level.Error(u.logger).Log(
+				"msg", "making updated directory",
```
blaedj: should this read: 
```suggestion
				"msg", "making updates directory",
```

pkg/autoupdate/handler.go
```diff
@@ -1,52 +1,115 @@
...
-// the remote metadata.
+// handler is called by the tuf package in two cases. First, and
+// confusingly, it's used as an error reporting channel for any kind
+// of issue. In this case, it's called with an err set.
+//
+// Second, it's called when tuf detects a change with the remote metadata.
 // The handler method will do the following:
 // 1) untar the staged staged file,
-// 2) replace the existing binary,
+// 2) into the updated directory
```
blaedj: is this untar-ing the staged file into `<binaryPath>-updates/`? I think it's being put into a `<binaryPath>-updates/<somedatehere>/` directory right?

pkg/autoupdate/handler.go
```diff
@@ -1,52 +1,115 @@
...
+		)
+
 		if err := u.finalizer(); err != nil {
-			level.Info(u.logger).Log("msg", "calling restart function for updated binary", "binary", u.destination, "err", err)
+			// Some kinds of updates, require a full
+			// launcher restart. For example, windows
+			// doesn't have an exec, so launcher exits so
+			// the service manager will restart it. There
+			// may be others.
+			if IsLauncherRestartNeededErr(err) {
```
blaedj: could this be done in the finalizer method? Perhaps there's value in having it outside the finalizer too, not sure.

pkg/autoupdate/handler.go
```diff
@@ -1,52 +1,115 @@
...
+				u.sigChannel <- os.Interrupt
+				return
+			}
+
+			level.Error(u.logger).Log(
+				"msg", "calling restart function for updated binary",
+				"binary", outputBinary,
+				"err", err)
 			return
+			// FIXME: I'm pretty sure if we get here, launcher hangs
```
blaedj: Thinking about this.. my instinct here is that if we think it's going to hang then we should fail loudly, but I'm not sure what affect that will have in relation to tuf.. What would happen if we hypothetically did an `os.Exit(1)` here?

pkg/contexts/ctxlog/ctxlog.go
```diff
@@ -21,6 +21,11 @@ func FromContext(ctx context.Context) log.Logger {
 		return log.NewNopLogger()
 	}
 	span := trace.FromContext(ctx).SpanContext()
+
+	if span.TraceID.String() == "00000000000000000000000000000000" {
```
blaedj: perhaps a comment explaining this? Is the exact number of zero's significant?

pkg/osquery/runtime/runtime.go
```diff
@@ -477,17 +502,41 @@ func (r *Runner) launchOsqueryInstance() error {
...
 	}
 
+	// If we're on windows, ensure that we're looking for the .exe
+	if runtime.GOOS == "windows" && !strings.HasSuffix(o.opts.binaryPath, ".exe") {
+		o.opts.binaryPath = o.opts.binaryPath + ".exe"
+	}
+
+	// before we start osqueryd, check with the update system to
+	// see if we have the newest version. Do this everytime. If
+	// this prove undesirable, we can expost a function to set
```
blaedj:
```suggestion
	// this prove undesirable, we can expose a function to set
```

Hey, at least you know I'm reading all this 😛

pkg/autoupdate/findnew.go
directionless: sort.strings is a couple lines up. https://github.com/kolide/launcher/pull/524/files/df0e753ec23419adb1d0cab980735e038146d562#diff-130b1554d3f180a250c35f46b1473c8fR96

blaedj: ha such a 🤦‍♂  - I was _looking_ for that, but missed it.

pkg/autoupdate/findnew.go
directionless: I don't think an early return makes sense. If there are no matches, then we return either `fullBinaryPath` or `""` depending. I don't think duplicating that logic makes sense.

I did, however, add a comment.

pkg/autoupdate/handler.go
directionless: Better would be to signal a shutdown. Probably no harm

pkg/contexts/ctxlog/ctxlog.go
directionless: Refactored. Probably more self describing now

pkg/autoupdate/handler.go
directionless: I think moving it into the finalizer would be more thinking. I want to kick that forward.

PR #528, opened 2019-10-17
 Add ssh key fingerprints to kolide_ssh_keys table

blaedj:
This commit:
 - adds a 'fingerprint' string column to the kolide_ssh_keys table
 - adds a Fingerprint attr to the KeyInfo type
 - populates the fingerprint attr (in the KeyInfo struct and the
 table) if we can use the method FingerprintSHA256 from x/crypto/ssh
 to get the fingerprint

Why?
 - This gets the 'easy' fingerprints populated in the table. There are
 other steps we can take to get the fingerprints for other
 keys (though perhaps not all of them) that will require a little more
 work. It remains to be seen if we'll do that work in this PR, or in a 
follow-up.


pkg/keyidentifier/puttyppk.go
```diff
@@ -44,6 +46,32 @@ func ParsePuttyPrivateKey(keyBytes []byte) (*KeyInfo, error) {
 		}
 	}
 
-	return ki, nil
+	// this can probably be done much more elegantly
+	var publicKeyLines strings.Builder
+	inPublicLines := false
+	for _, line := range strings.Split(keyString, "\n") {
```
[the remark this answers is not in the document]
blaedj: Yea, I accidentally committed this, I meant to keep it local until I proved out the strategy, then clean it up. If I end up doing this I'll absolutely consolidate the 2 loops.

pkg/keyidentifier/keyidentifier.go
```diff
@@ -15,13 +15,14 @@ import (
...
-	Comment    string // comments attached to the key
-	Parser     string // what parser we used to determine information
+	Type        string // Key type. rsa/dsa/etc
+	Format      string // file format
+	Bits        int    // number of bits in the key
+	Encryption  string // key encryption algorythem
+	Encrypted   *bool  // is the key encrypted
+	Comment     string // comments attached to the key
+	Parser      string // what parser we used to determine information
+	Fingerprint string // the fingerprint of the key, in SHA256 format
```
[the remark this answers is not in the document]
blaedj: hmm, I'm not sure on that either. Largely because I think that the fingerprint we report should match whatever `$TOOL` reports, where `$TOOL` is the standard program that would generally be used to get the fingerprint. It doesn't add a lot of value to report the md5 sha fingerprint of a modern openssh key, when the standard is the sha256 and the md5 is just for backwards compatibility. 

One option that could strike a nice balance between simplicity and extensibility is to _not_ export the Fingerprint attr, define a `Fingerprint()` method on `KeyInfo`, and then in the future we can add additional methods as we see fit. Right now we don't need additional fingerprint types (as we only report fingerprints modern openssh-gen'd keys), but we would still be able to add additional fingerprint types in the future. We can decide how to store the different fingerprints (array of objects, map etc) when we get to that point.

Does that seem like a reasonable strategy?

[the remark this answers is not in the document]
directionless: The issue, is that different tools use different fingerprints. 

In the webui, github displays the md5 fingerpringts.

At the command line, ssh-keygen defaults to the sha256.

So I'm pretty sure we need both.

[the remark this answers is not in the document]
directionless: Using a `.Fingerprint()` method is interesting. Doing that implies that we'd also keep private data to facilitate it.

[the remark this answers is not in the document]
blaedj: > In the webui, github displays the md5 fingerpringts.
>
> At the command line, ssh-keygen defaults to the sha256.

Ahh ok, that clears things up a bit then. Yea if that's the case we _should_ record both. I'd even say that we should expose both in the osquery table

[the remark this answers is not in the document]
blaedj: I'm thinking that we just have SHA256Fingerprint _and_ MD5Fingerprint on the KeyInfo type as well as columns for both in the table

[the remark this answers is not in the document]
directionless: So you think we should have them as straight fields on the `KeyInfo` struct?  (the first suggestion above) I'm good with that. Though I prefer to name them `Fingerprint<Type>` so they sort together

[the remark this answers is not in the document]
blaedj: Yea that's a good idea!

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -13,15 +14,24 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type spec struct {
```
directionless: Why not just use `KeyInfo`?

pkg/keyidentifier/testdata/make_temp.sh
```diff
@@ -0,0 +1,77 @@
...
+
+
+# FIXME: this entire file should be deleted and merged with make.sh
+set -e
+
+function rand {
+    dd if=/dev/random bs=1 count=16 2> /dev/null  | base64
+}
+
+function makeUUID {
```
directionless: Erm. When I said "uuid" I meant "random string". Please don't shell to ruby. Just use rand as defined above. Or something even simpler, so filter out the / characters. Or change rand to do both. We don't need strong randomness for any of this

pkg/keyidentifier/testdata/make_temp.sh
```diff
@@ -0,0 +1,77 @@
...
+function makeOpensshKeyAndSpec {
+    type=$1
+    bits=$2
+    format="openssh7"
+
+    keyfile=$(makeUUID)
+    keypath="fingerprints/$keyfile"
+
+    ssh-keygen -t $type -b $bits  -C "" -f $keypath -P /dev/null
+    ssh-keygen -t $type -b $bits  -C "" -f $keypath-encrypted -P "$(rand)"
```
directionless: I'd recommend changing this to being two calls. `makeOpensshKeyAndSpec` should take a password boolean.

pkg/keyidentifier/testdata/make_temp.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    for key in $keyfile $keyfile-encrypted; do
+        fingerprint=$(ssh-keygen -l -f fingerprints/$key | awk '{print $2}')
+        md5fingerprint=$(ssh-keygen -l -E md5 -f fingerprints/$key | awk '{print $2}' | sed 's/^MD5://')
+
+        encrypted=false
+        if [ "$key" == "$keyfile-encrypted" ]; then
+            encrypted=true
+        fi
+        echo "{ \"ExpectedFingerprintSHA256\": \"$fingerprint\",\
```
directionless: I'd love to also get the command line that generated this in here.

(doesn't need to be in the go struct, just the json)

directionless: This will be cleaner with a heredoc. Something like:

```
cat << EOF
{
"ExpectedFingerprintSHA256": "$fingerprint",
...
}
EOF | jq .
```

pkg/keyidentifier/testdata/make_temp.sh
blaedj: ha I figured the shell out to ruby would make you wince, meant to take it out in favor of some commented out shell script similar to the `rand` function. 

I'll simplify this vastly though, good call.

pkg/keyidentifier/testdata/make_temp.sh
blaedj: ahh yea for sure, I'll do that.

pkg/keyidentifier/testdata/make_temp.sh
blaedj: Yea that makes sense, I think that might help with debugging why I'm getting unexpected failing tests as well

pkg/keyidentifier/keyidentifier_test.go
blaedj: Mostly because I didn't want to try shoe-horn the expected data into the KeyInfo struct, especially if I wanted to add debugging info to the spec type.

directionless: The json is a superset of the KeyInfo struct, right? KeyInfo just specifies what's parsed from it. Or do I have that backwards?

blaedj: Hmm, yea I suppose it is a superset of `KeyInfo`.. There are some things that are convenient to parse from the json to a go struct, and this allows for that nicely.

Do you think that there is significant benefit to using the same type for the test specification and the type that holds the key information? I personally don't see much value, but I could just be missing something.

directionless: IME having less stuff is always less confusing. Fewer things to keep in my head. Fewer LoC. When they're different I immediately start trying to understand _why_ and whether the difference is important. 

I also think the resulting code is cleaner.

Compare:
```go
fromJson := readFileFromJson()

if includeField1 {
  require.Equal(t, fromJson.Field1, actual.Field1)
}
require.Equal(t, fromJson.Field2, actial.Field2)
require.Equal(t, fromJson.Field3, actial.Field3)
```

with
```go
fromJson := readFileFromJson()

if !includeField1 {
  fromJson.Field1 = ""
}

require.EqualValues(t,fromJson, actial.Field3)
```

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,170 @@
...
+# attributes of a key, plus the name of the key described. This spec file is
+# consumed by golang tests, and the output of the keyidentifier package is
+# compared against these expected values.
 
 set -e
 #set -x
 
 function rand {
-    dd if=/dev/random bs=1 count=16 2> /dev/null  | base64
+    dd if=/dev/random bs=1 count=16 2> /dev/null  | base64 | sed 's/[^0-9a-zA-Z]//g'
```
directionless:
```suggestion
   LC_ALL=C tr -dc 'A-Za-z0-9' </dev/urandom | head -c 16
```

For nitpicky reasons I'm not fond of this. the `count=16`, means you always have two trailing `=`, which sed just strips out anyhow. And you get inconsistent file lengths. The internet helped me with this code golf

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,170 @@
...
-makeOpensshKey rsa 2048
-makeOpensshKey rsa 4096
+    keyfile=$(rand)
+    keypath="$DATA_DIR/$keyfile"
 
-makeOpensshKey dsa 1024
 
-makeOpensshKey ecdsa 256
-makeOpensshKey ecdsa 521
+    if [ $encrypted == true ]; then
```
directionless: I think this would be cleaner, if the if block just set a password and the keygen  and command were consistent. Something like:
```
PASS=""
if [ $encrypted == true ]; then
  PASS=$(rand)
fi
```

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,170 @@
...
+        ssh-keygen -t $type -b $bits  -C "" -f $keypath -P "$(rand)"
+        cmd='ssh-keygen -t $type -b $bits  -C \"\" -f $keypath -P $(rand)'
```
directionless: This duplication feels weird. Maybe bash arrays make this a little cleaner? (This pattern should be fine in all the exec places. Also, this is where I start getting sad we're using bash and not go, even though it's probably correct)

```shell
type=rsa
bits=2048
pass='""'
cmd=(ssh-keygen -t "$type" -b "$bits" -f /tmp/testkey -P "$pass")

rm -f /tmp/testkey*

eval ${cmd[*]}
echo ${cmd[*]}
echo returned $?
```

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,170 @@
...
+        putty_format="-$format"
     fi
 
-    dir_fragment="$type/$bits/$format"
-    mkdir -p {encrypted,plaintext}"/$dir_fragment"
+    keyfile=$(rand)
+    keypath="$DATA_DIR/$keyfile"
+
+    passphrase=""
+    if [ $encrypted == true ]; then
```
directionless: Same comments as above WRT if blocks and command executions

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -13,15 +13,26 @@ import (
...
+// file.
 func TestIdentifyFiles(t *testing.T) {
-	kIdentifer, err := New(WithLogger(logutil.NewCLILogger(true)))
+	kIdentifier, err := New(WithLogger(logutil.NewCLILogger(true)))
 	require.NoError(t, err)
 
 	testFiles := []string{}
 
-	err = filepath.Walk("testdata", func(path string, info os.FileInfo, err error) error {
+	err = filepath.Walk("testdata/specs", func(path string, info os.FileInfo, err error) error {
```
directionless: This is now a single level, you can get simpler than Walk. glob for json is probably simple?

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -30,105 +41,88 @@ func TestIdentifyFiles(t *testing.T) {
...
-		Format: pathComponents[4],
-	}
+func testIdentifyFile(t *testing.T, kIdentifer *KeyIdentifier, specFilePath string) {
+	// load the json file
+	data, err := ioutil.ReadFile(specFilePath)
+	require.NoError(t, err, "reading spec file")
+	var example spec
+	err = json.Unmarshal(data, &example)
+	require.NoError(t, err, "parsing json spec file: %s", specFilePath)
+	keyPath := "testdata/specs/" + example.KeyPath
```
directionless: Shouldn't need this. The name is always `strings.TrimSuffix(".json", specFilePath)`

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,170 @@
...
+    md5fingerprint=$(ssh-keygen -l -E md5 -f $keypath | awk '{print $2}' | sed 's/^MD5://')
+
+    cat <<EOF > $keypath.json
+    {
+      "ExpectedFingerprintSHA256": "$fingerprint",
+      "ExpectedFingerprintMD5": "$md5fingerprint",
+      "Type": "$type",
+      "Bits": $bits,
+      "Encrypted": $encrypted,
+      "KeyPath": "$keyfile",
```
directionless: Shouldn't need this. The name is always strings.TrimSuffix(".json", specFilePath)

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -30,105 +41,88 @@ func TestIdentifyFiles(t *testing.T) {
...
-	pathComponents := strings.Split(path, "/")
-	expected := &KeyInfo{
-		Format: pathComponents[4],
-	}
+func testIdentifyFile(t *testing.T, kIdentifer *KeyIdentifier, specFilePath string) {
+	// load the json file
+	data, err := ioutil.ReadFile(specFilePath)
+	require.NoError(t, err, "reading spec file")
+	var example spec
+	err = json.Unmarshal(data, &example)
```
directionless: Why change `expected` to `example`?

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -30,105 +41,88 @@ func TestIdentifyFiles(t *testing.T) {
...
 	}
 
+	// Test correct 'bits' reporting
 	if keyInfo.Format == "openssh" && *keyInfo.Encrypted {
-		expected.Bits = 0
-	}
-	if expected.Type == "ecdsa" && *keyInfo.Encrypted {
-		expected.Bits = 0
+		// Can't get keys from encrypted openssh keys. At least not yet.
+	} else if example.Type == "ecdsa" && *keyInfo.Encrypted {
```
directionless: consider `switch` over `else if`

pkg/keyidentifier/keyidentifier_test.go
blaedj: no good reason, mostly because I was piecemeal replacing old code with new, and it was simpler to have a different variable name. You want me to change it back?

pkg/keyidentifier/testdata/make.sh
blaedj: I thought we were moving _away_ from 'test data encoded in filepath' 😛 

I don't _really_ have a strong opinion here, but it seems more consistent to encode everything in the json spec file.

pkg/keyidentifier/keyidentifier_test.go
directionless: If you don't feel strong, yes. I think `actual` and `expected` are more clear than `example`

blaedj: makes sense!

pkg/keyidentifier/testdata/make.sh
directionless: I guess I can go either way here. It's not really test _data_ per say. The existing of `key` and `key.json` define the test. It's never right to have one without the other...

pkg/keyidentifier/keyidentifier_test.go
directionless: (updated example code)

pkg/keyidentifier/keyidentifier_test.go
```diff
@@ -1,134 +1,107 @@
...
-		}
-
-		return nil
-	})
-	require.NoError(t, err, "filepath.Walk")
-
-	for _, path := range testFiles {
-		testIdentifyFile(t, kIdentifer, path)
+	testFiles, err = filepath.Glob("testdata/specs/*.json")
+	require.NoError(t, err, "finding spec files")
```
directionless:
```suggestion
	require.NoError(t, err, "error in glob")
	require.Greater(t, len(testFiles), 0, "must have more than 0 tests")
```

pkg/keyidentifier/keyidentifier_test.go
blaedj: Hmm, ok continuing our slack conversation, the place that this becomes burdensome is the loss of the `Source` attribute. Knowing that a key is generated from `puttygen` vs `ssh-keygen` is really handy when testing the fingerprints, because `puttygen` only seems to do the old-style hex fingerprints. see https://github.com/kolide/launcher/pull/528/files#diff-2bb52927af21dd3478f1948bed54f864R96

blaedj: I suppose I could unmarshal the json a second time into a hash to get the source, but at that point we might as well just use the original type..

pkg/keyidentifier/testdata/make.sh
blaedj: that's a fair assessment, and it simplifies the make script as well

pkg/keyidentifier/keyidentifier_test.go
directionless: That's a good reason not to just reuse. Bummer.

I'd still really avoid the duplication. Either with a nested type, or with the type embedding

blaedj: The spec type is now essentially a wrapper around the KeyInfo type w/an additional 'Source' attr

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,165 @@
...
 
-    puttygen -t $type -b $bits -C "" -o "plaintext/$dir_fragment/seph-putty" -O private --new-passphrase /dev/null
-    puttygen -t $type -b $bits -C "" -o "encrypted/$dir_fragment/seph-putty" -O private --new-passphrase <(rand)
+    passphrase=""
+    if [ $encrypted == true ]; then
+        passphrase=$(rand)
+        puttygen -t $type -b $bits -C "" -o $keypath -O private$putty_format --new-passphrase <(echo $passphrase)
+        cmd="puttygen -t $type -b $bits -C '' -o $keypath -O private$putty_format --new-passphrase $passphrase"
+    else
+        puttygen -t $type -b $bits -C "" -o $keypath -O private$putty_format --new-passphrase <(cat /dev/null)
```
directionless:
```suggestion
        puttygen -t $type -b $bits -C "" -o $keypath -O private$putty_format --new-passphrase /dev/null
```

`<(cat /dev/null)` is redundant. just `/dev/null`. The `<(...)` is making a file handle out of stdout. `/dev/null` is already a file.

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,165 @@
...
+        cmd="puttygen -t $type -b $bits -C '' -o $keypath -O private$putty_format --new-passphrase '<(cat /dev/null)'"
```
directionless:
```suggestion
        cmd="puttygen -t $type -b $bits -C '' -o $keypath -O private$putty_format --new-passphrase /dev/null"
```

duplication still bad

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,162 @@
...
+
+    passphrase=""
+    if [ $encrypted == true ]; then
+        passphrase=$(rand)
+    fi
 
-    puttygen -t $type -b $bits -C "" -o "plaintext/$dir_fragment/seph-putty" -O private --new-passphrase /dev/null
-    puttygen -t $type -b $bits -C "" -o "encrypted/$dir_fragment/seph-putty" -O private --new-passphrase <(rand)
+    puttygen -t $type -b $bits -C "" -o $keypath -O private$putty_format --new-passphrase <(echo $passphrase)
+    cmd="puttygen -t $type -b $bits -C '' -o $keypath -O private$putty_format --new-passphrase $passphrase"
```
directionless: The comment mismatch makes it hard to check

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,162 @@
...
+    keypath="$DATA_DIR/$(rand)"
+
+    passphrase=""
+    if [ $encrypted == true ]; then
+        passphrase=$(rand)
+    fi
 
-    puttygen -t $type -b $bits -C "" -o "plaintext/$dir_fragment/seph-putty" -O private --new-passphrase /dev/null
-    puttygen -t $type -b $bits -C "" -o "encrypted/$dir_fragment/seph-putty" -O private --new-passphrase <(rand)
+    puttygen -t $type -b $bits -C "" -o $keypath -O private$putty_format --new-passphrase <(echo $passphrase)
```
directionless: It's worth checking that this does what you need. I wonder if that needs to be an `echo -n`

pkg/keyidentifier/testdata/make.sh
blaedj: ah right, I see. This is rendered moot w/the commit I have locally

pkg/keyidentifier/testdata/make.sh
blaedj: interesting, I'm not familiar with using bash arrays like that. This works nicely

pkg/keyidentifier/testdata/make.sh
blaedj: it works fine for executing the command, but outputs the file handle (`/dev/fd/63`) for the documentation. which is fine imo.

directionless: Agree that's fine for the docs. And, I guess the trailing new line doesn't matter here. If the password gets munged, it doesn't matter for this

pkg/keyidentifier/testdata/make.sh
```diff
@@ -3,97 +3,202 @@
...
 
-function makePuttyKey {
-    type=$1
-    bits=$2
-    formats="$3"
+    # make the public key pair
+    puttygen -L --old-passphrase <(echo -n "$passphrase") $keypath > $keypath.pub
 
-    makePuttyKeyPuttyFormat $type $bits
+    # puttygen does not directly support thge sha256 fingerprint. But, we can use ssh-keygen for it.
```
blaedj:
```suggestion
    # puttygen does not directly support the sha256 fingerprint. But, we can use ssh-keygen for it.
```

PR #529, opened 2019-10-21
Git Version Parsing

directionless:
A very heavy route to parsing the git version. The split and recombine is due to some expected work towards supporting windows versions. For now, it's overkill sanity checking.

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -20,6 +20,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/make/builder.go
```diff
@@ -467,6 +468,50 @@ func (b *Builder) BuildCmd(src, output string) func(context.Context) error {
...
+	versionRegex, err := regexp.Compile(`^v?(\d+)\.(\d+)(?:\.(\d+))(?:(-.+))?$`)
+	if err != nil {
+		return "", errors.Wrap(err, "bad regex")
+	}
+
+	// regex match and check the results
+	matches := versionRegex.FindAllStringSubmatch(gitVersion, -1)
+
+	if len(matches) == 0 {
+		return "", errors.Errorf(`Version "%s" did not match expected format`, gitVersion)
```
blaedj: appending something like `: <major>.<minor>.<patch>.<additional>` to the error message might be helpful

pkg/make/builder_test.go
```diff
@@ -122,6 +128,68 @@ func TestExecOut(t *testing.T) {
...
+		{
+			in:  "",
+			err: true,
+		},
+		{
+			in:  "1",
+			err: true,
+		},
+		{
+			in:  "0.1",
```
blaedj: this is an error because it should be `0.1.0`?

directionless: I think that's a bug, albeit an unlikely one. I'm going to update the logic

pkg/make/builder.go
directionless: Okay

PR #530, opened 2019-10-22
Postinstall script to generate an info file

directionless:
We often want to understand what happened during installation. So the postinstall script should drop a config file onto disk with some basic info.

This is expected to be json, though on osx we additionally keep it as a plist. 

To make this simpler, the postinstall templates have been extracted into the asset generation flow.


blaedj:
> ... so while the conf file is generated, the overall install fails.

Does this mean that this PR will break installation on centos/ubuntu? Or just that installing in the docker container breaks, but that's fine because we use an image rather than installing into docker containers ?

blaedj:
Ok cool, that makes sense. I agree, the issue w/docker is not in-scope for this PR

PR #539, opened 2019-11-04
Add a kolide_json table

directionless:
Today, both osquery and launcher have plist tables. But, we often find we want a generalized json table as well.

As the underlying `pkg/dataflatten` supports json, this is a fairly small PR to allow creating both json and plist tables using it.


pkg/osquery/tables/dataflattentable/tables.go
```diff
@@ -56,12 +77,12 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
 			for _, constraint := range q.Constraints {
-				plistQuery := constraint.Expression
+				dataQuery := constraint.Expression
 
-				data, err := dataflatten.PlistFile(filePath,
-					append(flattenOpts, dataflatten.WithQuery(strings.Split(plistQuery, "/")))...)
+				data, err := t.dataFunc(filePath,
+					append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
 				if err != nil {
-					fmt.Println("parse failjure")
```
blaedj: should this be sent to the logger (if available) instead of println-ed out? it feels like a 'puts-debugging' line..

pkg/osquery/tables/dataflattentable/tables.go
```diff
@@ -33,7 +42,19 @@ func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *tab
...
-	return table.NewPlugin("kolide_plist", columns, t.generate)
+	switch dataSourceType {
+	case PlistType:
+		t.dataFunc = dataflatten.PlistFile
+		t.tableName = "kolide_plist"
+	case JsonType:
+		t.dataFunc = dataflatten.JsonFile
+		t.tableName = "kolide_json"
+	default:
+		panic("Unknown data soure type")
```
blaedj:
```suggestion
		panic("Unknown data source type")
```

pkg/osquery/tables/dataflattentable/tables_test.go
```diff
@@ -0,0 +1,94 @@
...
+// TestDataFlattenTable_Animals tests the basic generation
+// functionality for both plist and json parsing using the mock
+// animals data.
+func TestDataFlattenTable_Animals(t *testing.T) {
+	t.Parallel()
+
+	// Test this with both plist and json
+	testTables := map[string]Table{
+		"plist": Table{dataFunc: dataflatten.PlistFile},
+		"xml":   Table{dataFunc: dataflatten.PlistFile},
```
blaedj: is this a placeholder for if/when we do a generic 'xml' table?

pkg/osquery/tables/dataflattentable/tables_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+			require.NoError(t, err)
+
+			// delete the path and query keys, so we don't need to enumerate them in the test case
+			for _, row := range rows {
+				delete(row, "path")
+				delete(row, "query")
+			}
+
+			// Despite being an array. data is returned
```
blaedj: Nit: put this comment on one line?

pkg/osquery/tables/dataflattentable/tables.go
directionless: good catch.

pkg/osquery/tables/dataflattentable/tables.go
directionless: why didn't `misspell` catch this. Anyhow, fixed.

pkg/osquery/tables/dataflattentable/tables_test.go
directionless: xml is a valid format for plist. It's testing thge xml plist format.

directionless:
> I was a little unsure about the 'dynamic' table creation (i.e. the dataflattentable.TablePlugin returning a different table based on it's params), but the more I think about it, any introspection we want to do on the table plugins defined (e.g. generating some sort of list of table schemas for documentation) can be better implemented at a different (probably higher) level.

My guess is that we're going to have to do it differently. All the tables are somewhat dynamic, this one just more so. Don't really know until we start trying to write that code.

PR #542, opened 2019-11-07
Fix windows osqueryd updates

directionless:
Since the 0.11 autoupdate merge, windows osqueryd updates had not worked. This turns out to be because the osqueryd binary was specified as `osqueryd`, however it was actually `osqueryd.exe`. This mismatch caused the autoupdater routine and the findnewest function to use different paths.

autoupdate was downloading to `osqueryd-updates` while findnewest was looking in `osqueryd.exe-updates`

I think the underlying bug was because windows allows one to omit the `.exe` on executions. Which has the effect of masking an incorrect binary path.

This PR fixes this be ensure that the osqueryd path always has the trailing `.exe`. It also adds some additional logging.

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -23,6 +23,12 @@ func main() {
 	// re-leveled during options parsing.
 	logger := logutil.NewServerLogger(env.Bool("LAUNCHER_DEBUG", false))
 
+	level.Info(logger).Log(
+		"msg", "Launcher starting up",
+		"version", version.Version().Version,
+		"revision", version.Version().Revision,
+	)
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
@@ -47,6 +53,8 @@ func main() {
 			logutil.Fatal(logger, err, "exec")
 		}
 		panic("how")
+	} else {
+		level.Info(logger).Log("msg", "Nothing new")
 	}
 
 	// if the launcher is being ran with a positional argument, handle that
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -107,6 +107,11 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


cmd/launcher/main.go
```diff
@@ -23,6 +23,12 @@ func main() {
 	// re-leveled during options parsing.
 	logger := logutil.NewServerLogger(env.Bool("LAUNCHER_DEBUG", false))
 
+	level.Info(logger).Log(
```
blaedj: This also runs in launcher.go correct? Not saying it's wrong to have it here, but it does result in some duplicate info in the logs. Perhaps we could remove it from `launcher.go`?

cmd/launcher/options.go
```diff
@@ -107,6 +107,11 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		}
 	}
 
+	// On windows, we should make sure osquerydPath ends in .exe
+	if runtime.GOOS == "windows" && !strings.HasSuffix(osquerydPath, ".exe") {
+		osquerydPath = osquerydPath + ".exe"
```
blaedj: Should we add some debug logging here? I _think_ this change could break some launcher invocations, if they were successfully running with suffix-less osqueryd binaries? Might be helpful to at least point them in the right direction.

directionless: Ideally yes. But there's no logger instantiated at this point. :(

cmd/launcher/main.go
directionless: I think they're slightly different. At least, there's a bunch of other stuff that might happen in between them. See also, tracking down that boltdb hang

cmd/launcher/options.go
blaedj: ah that's a good point. I suppose the full osqueryd path _is_ logged out later on anyhow

directionless: It's part of why I made the later logging lines more explicit. But yes, I'd love to log this. I guess future refactor can move that to main, and then log about it

PR #552, opened 2019-12-13
Seph/initial runner snapshots

directionless:
This fixes the support for snapshot queries in the initial runner. 

The first commit splits this off into it's own file, so the simpler way to see the diff is [via filtering](https://github.com/kolide/launcher/pull/552/files/86b99490cbae39624bc434f1ee09bf990bfdf9fe..6231e62982ed3ce4c5e6a1725d186325f8fff31c)

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -1,13 +1,9 @@
 package osquery
 
 import (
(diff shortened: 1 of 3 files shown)


directionless:
Don't really want this

PR #557, opened 2020-01-14
Use /opt and /var/lib for Linux packaging

wbolster:
The directories /opt/{identifier} and /var/lib/{identifier} are the
proper places to install 3rd party software and its state.

### DIFF
--- a/pkg/packaging/packaging.go
+++ b/pkg/packaging/packaging.go
@@ -602,9 +602,9 @@ fi`
 func (p *PackageOptions) setupDirectories() error {
 	switch p.target.Platform {
 	case Linux, Darwin:
-		p.binDir = filepath.Join("/usr/local", p.Identifier, "bin")
+		p.binDir = filepath.Join("/opt", p.Identifier, "bin")
 		p.confDir = filepath.Join("/etc", p.Identifier)
-		p.rootDir = filepath.Join("/var", p.Identifier, sanitizeHostname(p.Hostname))
+		p.rootDir = filepath.Join("/var/lib", p.Identifier, sanitizeHostname(p.Hostname))
 	case Windows:
 		// On Windows, these paths end up rooted not at `c:`, but instead
 		// where the WiX template says. In our case, that's `c:\Program


wbolster:
related:

- https://github.com/wbolster/kolide-k2-launcher-arch-linux-packaging
- https://twitter.com/twseph/status/1217192799633858565

pkg/packaging/packaging.go
```diff
@@ -602,9 +602,9 @@ fi`
 func (p *PackageOptions) setupDirectories() error {
 	switch p.target.Platform {
 	case Linux, Darwin:
-		p.binDir = filepath.Join("/usr/local", p.Identifier, "bin")
+		p.binDir = filepath.Join("/opt", p.Identifier, "bin")
 		p.confDir = filepath.Join("/etc", p.Identifier)
-		p.rootDir = filepath.Join("/var", p.Identifier, sanitizeHostname(p.Hostname))
+		p.rootDir = filepath.Join("/var/lib", p.Identifier, sanitizeHostname(p.Hostname))
```
directionless: I think this one makes sense

pkg/packaging/packaging.go
```diff
@@ -602,9 +602,9 @@ fi`
 func (p *PackageOptions) setupDirectories() error {
 	switch p.target.Platform {
 	case Linux, Darwin:
-		p.binDir = filepath.Join("/usr/local", p.Identifier, "bin")
+		p.binDir = filepath.Join("/opt", p.Identifier, "bin")
```
directionless: I feel like I want to sleep on this one. It's not so clearly correct to me. `/usr/local` is host specific data, why do you think that's wrong?

wbolster:
Some rationale:

Arch:

>  Packages should never be installed to /usr/local/
 – https://wiki.archlinux.org/index.php/Arch_package_guidelines

Debian:

> Most third-party software installs itself in the /usr/local directory hierarchy. On Debian this is reserved for private use by the system administrator, so packages must not use directories such as /usr/local/bin but should instead use system directories such as /usr/bin, obeying the Filesystem Hierarchy Standard (FHS). – https://www.debian.org/doc/manuals/maint-guide/modify.en.html

pkg/packaging/packaging.go
wbolster: it's not host-specific. `/etc/...` contains host-specific config. `/usr/local` is for stuff _outside_ the packaging world. see also other comment in this pr with quotes and references to official packaging guides.

wbolster:
in fact is should/could even be `/etc/opt/kolide-k2`, not `/etc/kolide-k2` (see packaging guides). for example, google chrome packages use the same approach.

pkg/packaging/packaging.go
directionless: These are vendor or site package, not things in the underlying package repositories. 

My read on these docs, is that an official distribution package must not use `/usr/local` because it's reserved for host, site, and third part software. Generally I think of launcher as fitting into that. 

If the goal were to get packages into the upstream distributions, then I'd completely agree. However, there is a small amount of additional work to make that feasible. This set of packaging is rooted in some ideas that wouldn't translate to being upstreamed.

All that said, I'm not sure I feel strongly about it. So I'm weighing re-training finger macros and updating docs, against making things feel better to a couple of people.

wbolster: from the [debian packaging docs](https://www.debian.org/doc/manuals/maint-guide/modify.en.html):

> On Debian this is reserved for private use by the system administrator

this means _local_ use, so stuff _outside_ packages.

this really means packages should _never_ put stuff in `/usr/local`, no matter who built the package.

wbolster:
closing b/c lack of activity

PR #558, opened 2020-01-14
Use conventional names for systemd service names

wbolster:
A systemd service with a name like launcher.foo-bar is confusing and
non-standard. virtually all software uses something along the lines of
‘package-foo’, with ‘package’ being the main name of the software, and
‘foo’ some detail about the specific service (there can be more than
one).

This PR changes the systemd service name to ‘%s-launcher’, e.g.
‘kolide-k2-launcher’.

### DIFF
--- a/pkg/packaging/packaging.go
+++ b/pkg/packaging/packaging.go
@@ -418,11 +418,11 @@ func (p *PackageOptions) setupInit(ctx context.Context) error {
 		if p.target.Package == Rpm {
 			dir = "/usr/lib/systemd/system"
 		}
-		file = fmt.Sprintf("launcher.%s.service", p.Identifier)
+		file = fmt.Sprintf("%s-launcher.service", p.Identifier)
 		renderFunc = packagekit.RenderSystemd
 	case p.target.Platform == Linux && p.target.Init == Upstart:
 		dir = "/etc/init"
-		file = fmt.Sprintf("launcher-%s.conf", p.Identifier)
+		file = fmt.Sprintf("%s-launcher.conf", p.Identifier)
 		renderFunc = func(ctx context.Context, w io.Writer, io *packagekit.InitOptions) error {
 			return packagekit.RenderUpstart(ctx, w, io)
 		}


pkg/packaging/packaging.go
```diff
@@ -418,11 +418,11 @@ func (p *PackageOptions) setupInit(ctx context.Context) error {
 		if p.target.Package == Rpm {
 			dir = "/usr/lib/systemd/system"
 		}
-		file = fmt.Sprintf("launcher.%s.service", p.Identifier)
+		file = fmt.Sprintf("%s-launcher.service", p.Identifier)
```
directionless: Replacing the `.` with `-` makes sense. Do you think this makes more sense as `launcher-identifier` or `identifier-launcher` ? I feel like either is justified.

wbolster: the actual software is `kolide` / `kolide-k2`, and this is what it is referred to by users. that is has a `launcher` component is fine, but that's not the primary identifier. so i argue that `kolide-k2-launcher` makes the most sense.

(from personal experience: i had trouble remembering the unit name in the past and the `launcher` (very generic word) at the beginning didn't help me find it with tab completion either.)

semi-random examples following the same pattern:

```
$ ls /usr/lib/systemd/system/{systemd,fwupd,btrfs}* | sort | head -n 15
/usr/lib/systemd/system/btrfs-scrub@.service
/usr/lib/systemd/system/btrfs-scrub@.timer
/usr/lib/systemd/system/fwupd-offline-update.service
/usr/lib/systemd/system/fwupd-refresh.service
/usr/lib/systemd/system/fwupd-refresh.timer
/usr/lib/systemd/system/fwupd.service
/usr/lib/systemd/system/systemd-ask-password-console.path
/usr/lib/systemd/system/systemd-ask-password-console.service
/usr/lib/systemd/system/systemd-ask-password-wall.path
/usr/lib/systemd/system/systemd-ask-password-wall.service
/usr/lib/systemd/system/systemd-backlight@.service
/usr/lib/systemd/system/systemd-binfmt.service
/usr/lib/systemd/system/systemd-bless-boot.service
/usr/lib/systemd/system/systemd-boot-check-no-failures.service
/usr/lib/systemd/system/systemd-boot-system-token.service
```

(in fact i couldn't find any counterexamples)

directionless: As I think of it, `launcher` is a piece of endpoint software. People use it to connect to the Kolide SaaS products (mostly "k2", but also our depreciated SaaS "cloud"), as well as a [Fleet](https://github.com/kolide/fleet) (an OSS osquery fleet manager).

The current model is to have a package per setup. So, if your machine was connected to both k2, and a fleet install, you might have this installed twice, with appropriate identifiers.

I think, from that perspective, "launcher-identifier" makes more sense. 

But even with that context, "identifier-launcher" isn't bad, so I think I'm up for changing it. 

I don't think there's much impact. People who install versionX get one unit name, people who install versionY get another.

directionless: Yeah, exactly. To be really formal, `kolide-launcher-kolide-k2` would designate this, but that's just unwieldy. 

I think I'm concluding that your original suggestion of `identifier-launcher` is the best solution today. Specifically, I think this paragraph is the right model.

> the actual software is kolide / kolide-k2, and this is what it is referred to by users. the fact that it has a launcher component is all fine, but that's not the first thing that comes to mind. so i argue that kolide-k2-launcher makes the most sense.

So, put differently, I'm in favor of:
1. As we're building fully baked packages, we can name them `identifier-launcher`.
  a.  So for example, `kolide-k2-launcher`. 
  b. Or as a local site's fleet install `example-fleet-launcher`
2. If and when we move to a split binary package / config model, the binary package would obviously be named `kolide-launcher`

This approach might not be the most ideal, but it's a manageable change and fits with the existing assumptions in this code base. (Moving to something like `kolide-launcher-k2` would be a huge amount of work.

This may have some larger changes, I don't know if I'll merge this, or pull it into a new branch. I'm going to need to think about testing the various permutations for it.

I am unlikely to build anything to handle changing from one to the other. Packaging should generally handle it, and the usual model for launcher is to use it's auto-update and not package updates.

directionless:
To pull this out of the thread, and into the main comments --

I'm in favor of this. But will need to work through testing on the various init scripts to make sure all the right bits get changed. I'm not sure if I'll do it in this PR or cut new one. Either way huge thanks for speaking up and pushing for it

pkg/packaging/packaging.go
wbolster: cool, i agree with your reasoning and pragmatism :+1:

wbolster:
closing b/c lack of activity, and this isn't good enough to merge

Issue #565, opened 2020-01-24
Table Request: Add DEP Enrolled column to existing Kolide MDM Info table


FritzX6:
@grahamgilbert wrote a change to the existing Kolide MDM Info table and posted it to the [MacAdmins/#osquery Slack](https://macadmins.slack.com/archives/C07DGNRGS/p1579892869000800)

I would love to see this get pushed upstream:

```go
package main
​
import (
	"bytes"
	"context"
	"fmt"
	"os/exec"
	"strconv"
​
	"github.com/groob/plist"
	"github.com/kolide/osquery-go/plugin/table"
	"github.com/pkg/errors"
)
​
type profilesOutput struct {
	ComputerLevel []profilePayload `plist:"_computerlevel"`
}
​
type profilePayload struct {
	ProfileIdentifier  string
	ProfileInstallDate string
	ProfileItems       []profileItem
}
​
type profileItem struct {
	PayloadContent *payloadContent
	PayloadType    string
}
​
type payloadContent struct {
	AccessRights            int
	CheckInURL              string
	ServerURL               string
	ServerCapabilities      []string
	Topic                   string
	IdentityCertificateUUID string
	SignMessage             bool
}
​
type profileStatus struct {
	DEPEnrolled  bool
	UserApproved bool
}
​
type depStatus struct {
	DEPCapable bool
}
​
func MDMInfoColumns() []table.ColumnDefinition {
	return []table.ColumnDefinition{
		table.TextColumn("enrolled"),
		table.TextColumn("server_url"),
		table.TextColumn("checkin_url"),
		table.IntegerColumn("access_rights"),
		table.TextColumn("install_date"),
		table.TextColumn("payload_identifier"),
		table.TextColumn("topic"),
		table.TextColumn("sign_message"),
		table.TextColumn("identity_certificate_uuid"),
		table.TextColumn("has_scep_payload"),
		table.TextColumn("installed_from_dep"),
		table.TextColumn("user_approved"),
		table.TextColumn("dep_capable"),
	}
}
​
func MDMInfoGenerate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
	profiles, err := getMDMProfile()
	if err != nil {
		return nil, err
	}
​
	depEnrolled, depCapable, userApproved := "unknown", "unknown", "unknown"
	status, err := getMDMProfileStatus()
	if err == nil { // only supported on 10.13.4+
		depEnrolled = strconv.FormatBool(status.DEPEnrolled)
[...]

PR #566, opened 2020-01-24
Add DEP Enroll column to MDM table

FritzX6:
## What is being changed?

This PR adds @grahamgilbert 's DEP enrollment column to the existing MDM table #565.

### How is this helpful?
Per [Apple's Documentation:](https://www.apple.com/business/docs/site/DEP_Guide.pdf)
>Device Enrollment Program or DEP is a program available through Apple which helps businesses easily deploy and configure Apple devices. DEP simplifies initial setup by automating mobile device management (MDM) enrollment and supervision of devices during setup, which enables you to configure your organization’s devices without touching them. To further simplify the process, you can skip certain Setup Assistant screens so users can start using their devices right out of the box. 

Understanding whether a device is enrolled in DEP can be critical to guaranteeing the desired configuration of a device.

Prior to this column being added we have not had a reliable method to determine whether devices were DEP enrolled.

### What will this look like?
```SQL
osquery> SELECT * FROM kolide_mdm_info;
                 enrolled = true
               server_url = https://acme.jamfcloud.com//computer/mdm
              checkin_url = https://acme.jamfcloud.com//computer/mdm?invitation=xxxx
            access_rights = ****
             install_date = 2019-02-18 00:11:15 +0000
       payload_identifier = 00000000-0000-0000-0000-***********
                    topic = com.apple.mgmt.External.****
             sign_message = true
identity_certificate_uuid = 00000000-0000-0000-0000-***********
         has_scep_payload = true
       installed_from_dep = true
            user_approved = true
             dep_enrolled = true
```

### DIFF
--- a/pkg/osquery/table/mdm.go
+++ b/pkg/osquery/table/mdm.go
@@ -3,6 +3,7 @@ package table
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"os/exec"
 	"strconv"
 
@@ -11,6 +12,40 @@ import (
 	"github.com/kolide/osquery-go/plugin/table"
 	"github.com/pkg/errors"
 )
+
+type profilesOutput struct {
+	ComputerLevel []profilePayload `plist:"_computerlevel"`
+}
+​
+type profilePayload struct {
+	ProfileIdentifier  string
+	ProfileInstallDate string
+	ProfileItems       []profileItem
+}
+​
+type profileItem struct {
+	PayloadContent *payloadContent
+	PayloadType    string
+}
+​
+type payloadContent struct {
+	AccessRights            int
+	CheckInURL              string
+	ServerURL               string
+	ServerCapabilities      []string
+	Topic                   string
+	IdentityCertificateUUID string
+	SignMessage             bool
+}
+​
+type profileStatus struct {
+	DEPEnrolled  bool
+	UserApproved bool
+}
+​
+type depStatus struct {
+	DEPCapable bool
+}
 
 func MDMInfo(logger log.Logger) *table.Plugin {
 	columns := []table.ColumnDefinition{
@@ -36,11 +71,16 @@ func generateMDMInfo(ctx context.Context, queryContext table.QueryContext) ([]ma
 		return nil, err
 	}
 
-	depEnrolled, userApproved := "unknown", "unknown"
+	depEnrolled, depCapable, userApproved := "unknown", "unknown", "unknown"
 	status, err := getMDMProfileStatus()
 	if err == nil { // only supported on 10.13.4+
 		depEnrolled = strconv.FormatBool(status.DEPEnrolled)
 		userApproved = strconv.FormatBool(status.UserApproved)
+	}
+
+	depstatus, err := getDEPStatus()
+	if err == nil {
+		depCapable = strconv.FormatBool(depstatus.DEPCapable)
 	}
 
 	var enrollProfileItems []profileItem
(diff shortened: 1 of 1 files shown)


weswhet:
@FritzX6 we would definitely be interested in this change :)

FritzX6:
Tested post changes and it appears the results are coming in as expected:
```shell
➜  launcher git:(fritz_add_dep_column) sudo make osqueryi-tables                             

go run cmd/make/make.go -targets=table-extension -linkstamp
osqueryd -S --allow-unsafe --extension ./build/darwin/tables.ext
Using a virtual database. Need help, type '.help'

osquery> select * from kolide_mdm_info;
                 enrolled = true
               server_url = https://mdm.*****/mdm/connect
              checkin_url = https://mdm.*****/mdm/checkin
            access_rights = 8191
             install_date = 2018-01-02 17:24:33 +0000
       payload_identifier = com.github.micromdm.micromdm.mdm
                    topic = com.apple.mgmt.External.*****
             sign_message = true
identity_certificate_uuid = 2b9a09ea-7063-*****
         has_scep_payload = true
       installed_from_dep = true
            user_approved = true
              dep_capable = true
```

pkg/osquery/table/mdm.go
```diff
@@ -66,6 +106,7 @@ func generateMDMInfo(ctx context.Context, queryContext table.QueryContext) ([]ma
 					"identity_certificate_uuid": enrollProfile.IdentityCertificateUUID,
 					"installed_from_dep":        depEnrolled,
 					"user_approved":             userApproved,
+					"has_dep_profile":           depCapable,
```
groob: That's a better name than "dep_capable" and IMO you should rename the other variables/struct fields in this code. All macs are "capable" of DEP. Some serial numbers are not possible to add due to geographic regions/legal but that information can't be derived from the device itself.

pkg/osquery/table/mdm.go
```diff
@@ -36,13 +71,18 @@ func generateMDMInfo(ctx context.Context, queryContext table.QueryContext) ([]ma
...
-	depEnrolled, userApproved := "unknown", "unknown"
+	depEnrolled, depCapable, userApproved := "unknown", "unknown", "unknown"
 	status, err := getMDMProfileStatus()
 	if err == nil { // only supported on 10.13.4+
 		depEnrolled = strconv.FormatBool(status.DEPEnrolled)
 		userApproved = strconv.FormatBool(status.UserApproved)
 	}
 
+	depstatus, err := getDEPStatus()
+	if err == nil {
```
groob: if err != nil, return it. The logic here doesn't make much sense because the error is ignored.

pkg/osquery/table/mdm.go
```diff
@@ -145,7 +161,20 @@ func getMDMProfileStatus() (profileStatus, error) {
 	}, nil
 }
 
-type profileStatus struct {
-	DEPEnrolled  bool
-	UserApproved bool
+func getDEPStatus() (depStatus, error) {
```
groob: what this function appears to be checking is if the device is on a OS version where the OS supports these flags. I don't think there's actual correlation between this command having 3 lines of output and DEP profile status. 

I would solicit more output from different systems and update this helper to more reliably print a status.

pkg/osquery/table/mdm.go
```diff
@@ -12,6 +12,40 @@ import (
...
+	IdentityCertificateUUID string
+	SignMessage             bool
+}
+
+type profileStatus struct {
+	DEPEnrolled  bool
+	UserApproved bool
+}
+
+type depStatus struct {
```
groob: does this struct serve any purpose? Why doesn't `depStatus` return a simple boolean?

FritzX6:
Pretty sure this change is made redundant with the addition of the new `kolide_profiles` table. Closing PR.

PR #574, opened 2020-02-19
macOS system profiler table

directionless:
Create a `systemprofile` table using `dataflatten` to process the output from `system_profile`


pkg/osquery/tables/systemprofiler/systemprofiler.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	var detailLevel string
+	if q, ok := queryContext.Constraints["detaillevel"]; ok && len(q.Constraints) != 0 {
+		if len(q.Constraints) > 1 {
+			level.Info(t.logger).Log("msg", "WARNING: Only using the first detaillevel request")
+		}
+
+		dl := q.Constraints[0].Expression
+		for _, known := range knownDetailLevels {
```
JonathanNogueira: why make `knownDetailLevels` an array instead of a set(map)?

pkg/osquery/tables/systemprofiler/systemprofiler.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+
+	}
+
+	systemProfilerOutput, err := t.execSystemProfiler(ctx, detailLevel, requestedDatatypes)
+	if err != nil {
+		return results, errors.Wrap(err, "exec")
+	}
+
+	if q, ok := queryContext.Constraints["query"]; ok && len(q.Constraints) != 0 {
```
JonathanNogueira: Wouldn't it be simpler to avoid trying to parse the rows out of the output in the table generate func by using `system_profiler -json -detailLevel <level> [dataType1 ... dataTypeN]` just returning the JSON back. The consumer can then use sqlite's JSON1 extension to query the output or not.

FritzX6: I believe @directionless mentioned to me that the `-json` flag is not supported/available on versions of macOS prior to 10.15 _(Catalina)_

pkg/osquery/tables/systemprofiler/systemprofiler.go
```diff
@@ -0,0 +1,236 @@
...
+		table.TextColumn("parentdatatype"),
+
+		table.TextColumn("query"),
+		table.TextColumn("datatype"),
+		table.TextColumn("detaillevel"),
+	}
+
+	t := &Table{
+		client:    client,
+		logger:    level.NewFilter(logger, level.AllowInfo()),
```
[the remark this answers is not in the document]
directionless: This used to be like that. But the dataflatten stuff is intentionally really noisy to debug the object recursion. So the table wrappers whomp it.

pkg/osquery/tables/systemprofiler/systemprofiler.go
directionless: AFAIK this is the preferred go idiom. 

If I needed to do something with the resultant strings (like map them to numeric values) I think a `map` would be cleaner, but for checking known values this is super common

pkg/osquery/tables/systemprofiler/systemprofiler.go
directionless: I'm not totally sure I understand your question.

As fritz mentions, `-json` is a catalina option. But I could work around that. I think the underlying question here is what does datatable's `query` do that sqlite's `json1` doesn't already do? If that's it, I think that's a great question. 

I wrote dataflatten to be able to take arrays of maps, and re-write them into a map. Thus allowing simpler manipulations in EAV style tables. When I wrote it, I couldn't find good prior art that felt right. xpath is the right idea. but big and hard to use in a simple select interface. Similar for some existing json things.

I did not evaluate sqlite's `json1` one at the time, entirely because I wasn't aware of it :(

Looking at it now, I'm not sure it works. While it handles some use cases, we have a pretty common case of taking the aforementioned array-of-maps, converting it into a nested map, and then pivoting it into a wide tuple. `json_extract` might do that, but I don't immediately see how. 

I'm happy to consider the json1 stuff in the future. Though I think there several reasons to kick that forward:
* unsure about the feasibility, we'd need to play with some of our queries
* We'd need to make a general plist -> json routine
* We already have this

PR #584, opened 2020-02-26
Arch's lib is /usr/lib

directionless:
Arch has `/lib/` as a symlink to `/usr/lib`. This means packages cannot install files to `/lib` and must use `/usr/lib` instead.

See https://bbs.archlinux.org/viewtopic.php?id=192018 for more information


directionless:
(test failure is unrelated, and due to an issue in GCP)

PR #587, opened 2020-02-28
Use mxj to implement an xml dataflattening

directionless:
Implement an xml dataflattening using mxj.

The general issue with xml encoding, is that because of the difference between elements and attributes, it's hard to write a fully generic parser. Luckily, other people already have. This uses `mxj` as a generalized xml parser.

An alternative would be to implement our own.


directionless:
(not ready for review. Misclicked)

PR #590, opened 2020-03-21
Improve verification on downloaded updates

directionless:
We occasionally see downloaded updates get corrupt. This attempts to mitigate that by testing the update before assuming it's good.

This test is implemented by execing with `--version` (accepted by both osquery and launcher), and examining the error code. An error of 0, 1, or 2 is considered fine. We're trying to detect exec failures, not process failures. 

As windows semantics are a bit different, this also updates and repairs the `pkg/autoupdate` tests to run in windows. That library passes in my VM. CI still doesn't have windows support.

### DIFF
--- a/pkg/autoupdate/findnew_test.go
+++ b/pkg/autoupdate/findnew_test.go
@@ -90,12 +90,12 @@ func TestFindBaseDir(t *testing.T) {
 		out string
 	}{
 		{in: "", out: ""},
-		{in: "/a/path/launcher", out: "/a/path"},
-		{in: "/a/path/launcher-updates/1569339163/launcher", out: "/a/path"},
+		{in: "/a/path/launcher", out: filepath.Clean("/a/path")},
+		{in: "/a/path/launcher-updates/1569339163/launcher", out: filepath.Clean("/a/path")},
 	}
 
 	for _, tt := range tests {
-		require.Equal(t, tt.out, FindBaseDir(tt.in), "input: %s", filepath.Clean(tt.in))
+		require.Equal(t, tt.out, FindBaseDir(tt.in), "input: %s", tt.in)
 	}
 
 }
@@ -127,6 +127,10 @@ func TestFindNewestEmptyUpdateDirs(t *testing.T) {
 
 func TestFindNewestNonExecutable(t *testing.T) {
 	t.Parallel()
+
+	if runtime.GOOS == "windows" {
+		t.Skip("Windows doesn't use executable bit")
+	}
 
 	tmpDir, binaryName, cleanupFunc := setupTestDir(t, nonExecutableUpdates)
 	defer cleanupFunc()
--- a/pkg/autoupdate/findnew_test.go
+++ b/pkg/autoupdate/findnew_test.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"testing"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/autoupdate/findnew_test.go
```diff
@@ -253,6 +295,138 @@ func setupTestDir(t *testing.T, stage setupState) (string, string, func()) {
...
+
+	require.Error(t,
+		checkExecutable(context.TODO(), truncatedBinary.Name(), "-test.run=TestHelperProcess", "--", "exit0"),
+		"truncated binary")
+}
+
+// TestHelperProcess isn't a real test. It's used as a helper process
+// to make a portableish binary. See
+// https://github.com/golang/go/blob/master/src/os/exec/exec_test.go#L724
+// and https://npf.io/2015/06/testing-exec-command/
```
blaedj: Wow, this took a bit of reading to grok (thanks for providing the explanatory blog post), but it's an interesting/clever way to test exec!

pkg/autoupdate/handler.go
```diff
@@ -81,6 +82,24 @@ func (u *Updater) handler() tuf.NotificationHandler {
...
+		// Check that it all came through okay
+		if err := checkExecutable(context.TODO(), outputBinary, "--version"); err != nil {
+			level.Error(u.logger).Log(
+				"msg", "Broken updated binary. Removing",
+				"target", u.target,
+				"outputBinary", outputBinary,
+				"err", err,
+			)
+			if err := os.RemoveAll(updateDir); err != nil {
+				level.Error(u.logger).Log(
```
blaedj: Just brainstorming here.. Suppose we _do_ fail to remove the corrupt update, the next time launcher starts up we'll just run through the same process. Continually failing to remove the corrupt update won't prevent the launcher binary from updating, so it probably doesn't make sense to try other mitigation methods (I was initially thinking of re-downloading the latest update and putting it a 'newer' directory or something..).

In any case, we'll (hopefully) be able to look at the logs on the device if we need to investigate further.

kinda makes me wish we had a canary style test, e.g. a launcher that pulls updates and fails very loudly somehow if the update is corrupt. Though, the corruption problem is probably _not_ related to us serving corrupt updates..

blaedj: (that got long and rambly...)

pkg/autoupdate/helpers_test.go
```diff
@@ -4,41 +4,54 @@ import (
...
 )
 
-func TestCheckExecutable(t *testing.T) {
+func TestCheckExecutablePermissions(t *testing.T) {
 	t.Parallel()
 
-	require.Error(t, checkExecutable(""), "passing empty string")
-	require.Error(t, checkExecutable("/random/path/should/not/exist"), "passing empty string")
+	require.Error(t, checkExecutablePermissions(""), "passing empty string")
+	require.Error(t, checkExecutablePermissions("/random/path/should/not/exist"), "passing empty string")
```
blaedj:
```suggestion
	require.Error(t, checkExecutablePermissions("/random/path/should/not/exist"), "passing non-existent file path")
```

pkg/autoupdate/findnew_test.go
directionless: This pattern appears in a couple other places in the `launcher` base. I go back and forth on whether it's too clever, or the best there is. I always have to stare at it to remind myself how it works

pkg/autoupdate/handler.go
directionless: Yeah... I don't really know the history of the current update model, but I kinda want to majorly re-write it. Right now, it has an asynchronous thread which downloads an update, and calls a `handler` to trigger a restart. 

This design makes it hard to have clear control over how things are happening. Or even to trigger a `check now` style function. It's all very `fire and forgot` style.

But I agree -- this error shouldn't end up being fatal. It means there's some weird detritus on disk.

PR #591, opened 2020-03-24
Circleci updates

directionless:


directionless:
Launcher no longer uses circle

PR #592, opened 2020-03-26
kolide WMI table

directionless:
This is a general `wmi` table. It requires you to list the properties, and it uses dataflatten.

There are some tests, and as we don't have windows CI, I've manually run them.


pkg/wmi/wmi.go
```diff
@@ -0,0 +1,130 @@
...
+
+func Query(ctx context.Context, className string, properties []string) ([]map[string]interface{}, error) {
+	logger := log.With(ctxlog.FromContext(ctx), "caller", "wmi.Query")
+	handler := NewOleHandler(ctx, properties)
+
+	// If we query for the exact fields, _and_ one of the property
+	// names is wrong, we get no results. (clearly an error. but I
+	// can't find it) So query for `*`, and then fetch the
+	// property. More testing might show this needs to change
+	queryString := fmt.Sprintf("SELECT * FROM %s", className)
```
directionless: This needs a untaint routine

directionless: added.

blaedj, line 42: if the query specifies one of the properties that doesn't show up in `select *`, should we log that or throw an error? Perhaps that's better handled one level higher though..

blaedj: ah never mind, I see the handling for it below in HandleVariant

pkg/wmi/wmi.go
```diff
@@ -0,0 +1,130 @@
...
+	item := v.ToIDispatch()
+	defer item.Release()
+
+	result := make(map[string]interface{})
+
+	for _, p := range oh.properties {
+		val, err := oleutil.GetProperty(item, p)
+		defer val.Clear()
+		if err != nil {
+			level.Debug(oh.logger).Log("msg", "Got error looking for property", "property", p, "err", err)
```
blaedj, line 113: there's a tradeoff here I think between erroring loudly to avoid confusion (e.g. if there's a typo, this fails quietly I think) and annoyance from over-strict requirements. I almost feel like the lower-level library here should raise an error to be handled by the caller, but I'm not sure..

That being said, I think we should stick with just logging, and we can iterate as needed.

pkg/wmi/wmi.go
directionless: It doesn't seem to matter. I'm not totally sure how it works under the WMI hood. But I think the higher level API is reasonable, and this may change

pkg/wmi/wmi.go
directionless: Unclear. If you wanted to write something that hits different versions, it would be nice if it just kinda returned the data it had. 

I don't think we have a good error channel for reporting non-fatal errors. The osquery-go stuff swallows errors. So logging is mostly it.

PR #595, opened 2020-04-02
Fix an infinite exec bug

directionless:
The FindNewestSelf function calls FindNewest. Which, in turns, calls `checkExecutable`. Which launchers the binary. Which goes back to the beginning. This bug only manifests for `launcher`, but not for the `osquery` side. This isn't a fatal error, as there's a 5s timeout on it. But it is slow.

Add an option to skip the final check, solving this. And add comments.

Convert the finalizers over to using `FindNewestSelf`. First, because that's what it's there for. Second, because `os.Executable()` is more correct than `os.Args[0]`

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -41,7 +41,7 @@ func main() {
 	//
 	// launcher is _also_ called when we're checking update
 	// validity (with autoupdate.checkExectuable). This is
-	// somewhat awkward as we emnd up with extra call layers.
+	// somewhat awkward as we end up with extra call layers.
 	newerBinary, err := autoupdate.FindNewestSelf(ctx)
 	if err != nil {
 		logutil.Fatal(logger, err, "checking for updated version")
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -40,7 +40,7 @@ func main() {
 	// a straight forward exec.
 	//
 	// launcher is _also_ called when we're checking update
-	// validity (with autoupdate.checkExectuable). This is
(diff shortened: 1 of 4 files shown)


cmd/launcher/main.go
```diff
@@ -38,6 +38,10 @@ func main() {
 	// This does not call DeleteOldUpdates, on the theory that
 	// it's better left to the service to handle cleanup. This is
 	// a straight forward exec.
+	//
+	// launcher is _also_ called when we're checking update
+	// validity (with autoupdate.checkExectuable). This is
+	// somewhat awkward as we emnd up with extra call layers.
```
blaedj, line 44:
```suggestion
	// somewhat awkward as we end up with extra call layers.
```

cmd/launcher/main.go
```diff
@@ -38,6 +38,10 @@ func main() {
 	// This does not call DeleteOldUpdates, on the theory that
 	// it's better left to the service to handle cleanup. This is
 	// a straight forward exec.
+	//
+	// launcher is _also_ called when we're checking update
+	// validity (with autoupdate.checkExectuable). This is
```
blaedj, line 43:
```suggestion
	// validity (with autoupdate.checkExecutable). This is
```

directionless:
> I'm trying to figure out why we don't need to use this option in other calls to findNewest, e.g. updater-finalizer.go:29

Good catch, I think that needs this option as well. Though I wonder why it's not using `FindNewestSelf` instead. I mean, I wrote the code...

PR #596, opened 2020-04-02
Autoupdate should not exec updates older than itself

directionless:
As implemented in https://github.com/kolide/launcher/pull/524, the autoupdater will exec the last downloaded updated. This is _incorrect_ when there are updates on disk, and a device has installed a new package.

This adds a build timestamp, which can be used to ignore (and delete) old updates in disk

Note that it is comparing the _build timestamp_ with the update's _download_ timestamp. This is an important distinction, as it allows an intentional version downgrade.

### DIFF
--- a/pkg/autoupdate/findnew_test.go
+++ b/pkg/autoupdate/findnew_test.go
@@ -428,9 +428,7 @@ func TestBuildTimestamp(t *testing.T) {
 	for _, tt := range tests {
 		t.Run("buildTimestamp="+tt.buildTimestamp, func(t *testing.T) {
 			tmpDir, binaryName, cleanupFunc := setupTestDir(t, executableUpdates)
-			fmt.Println(tmpDir)
-			//defer cleanupFunc()
-			_ = cleanupFunc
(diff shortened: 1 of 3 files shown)


pkg/autoupdate/findnew_test.go
```diff
@@ -396,6 +397,69 @@ func TestCheckExecutableTruncated(t *testing.T) {
...
+			buildTimestamp: "6",
+			expectedOnDisk: 0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("buildTimestamp="+tt.buildTimestamp, func(t *testing.T) {
+			tmpDir, binaryName, cleanupFunc := setupTestDir(t, executableUpdates)
+			fmt.Println(tmpDir)
+			//defer cleanupFunc()
```
blaedj, line 432: do we want to run the cleanup func or remove this line?

directionless: oops. That's development crud I left in. Nice catch.

PR #601, opened 2020-04-13
Fork Bomb Fixes

directionless:
In #590 we expanded `checkExecutable` to invoke the executable to verify correctness. When taken with `launcher's` inclination to exec newer versions, this created a fork bomb. There was a fix added in #595, but that's not adequate. 

This PR attempts to address this in three ways. While any one of them should be adequate, we may as one run all three. 

First, we add an environmental variable that _skips_ the initial re-exec functionality. This isn't documented, and is meant for internal use only. Frustratingly, we cannot just add it to flags, as we can't parse flags prior to exec.

Second, this risk primarily comes from the `FindNewestSelf` code path. This introduces `withRunningExectuable` as a mechanism to indicate that we should _not_ check a given path.

Third, if we can determine `os.Executable`, don't exec it. This is duplicative with the second option. But, it comes at it from a different direction. Seems better to have both

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -42,23 +42,30 @@ func main() {
 	// launcher is _also_ called when we're checking update
 	// validity (with autoupdate.checkExecutable). This is
 	// somewhat awkward as we end up with extra call layers.
-	newerBinary, err := autoupdate.FindNewestSelf(ctx)
-	if err != nil {
-		logutil.Fatal(logger, err, "checking for updated version")
-	}
-
-	if newerBinary != "" {
-		level.Debug(logger).Log(
-			"msg", "preparing to exec new binary",
-			"oldVersion", version.Version().Version,
-			"newBinary", newerBinary,
-		)
-		if err := execwrapper.Exec(ctx, newerBinary, os.Args, os.Environ()); err != nil {
-			logutil.Fatal(logger, err, "exec")
+	//
+	// Allow a caller to set `LAUNCHER_SKIP_UPDATES` as a way to
+	// skip exec'ing an update. This helps prevent launcher from
+	// fork-bombing itself. This is an ENV, because there's no
+	// good way to pass it through the flags.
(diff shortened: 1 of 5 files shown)


blaedj:
Is there a good reason not to parse options before running the `findNewest` check? 

It would be nice if the `--version` option/subcommand (and also `--usage`, really) skipped checking for updates, since if I'm running the binary with that command I really just want to know the version of the binary that I called

blaedj:
If we _don't_ want to parse the options first, could we put the `isSubCommand` before the update check? If we did that, we could call `launcher version` instead of `launcher --version` when we see if the binary is executable, the `version` subcommand does the same thing as the `--version` flag without requiring us to parse options.

Note that I may be missing something fundamental here too..

directionless:
I struggled a bunch with these questions...

The standard library's flag parsing has, AFAIK, no way to ignore unknown options. They will trigger an `os.Exit(2)` This means that we cannot nicely parse flags or subcommands before invoking the newest. Else new options trigger fatal errors. (I would _love_ to parse it and ignore the errors)

I thought about whether to move the `--version` prior to FindNewest. I deliberately have them in this order. I think the most common use case, even for us, is trying to understand what command will be operating. Thus, it pulls the newest first. I see this as a small step towards some deeper refactoring. Launcher's top level should be seen as a _launcher_ and an updater. Which components for osquery and tables. (Even if it's the same binary, thinking as these as split helps)

And while I understand the appeal of making `version` different from `--version` I think that confusion is a bad idea. Consistent is probably better.

blaedj:
Ok that all makes sense! We could reason about a dedicated 'test executable' subcommand (the subcommand check doesn't require flag parsing) that might be too trivial an execution path to be useful..

directionless:
> about a dedicated 'test executable' subcommand

I thought about that too! As is, the code path is the same for testing `osqueryd` and `launcher`. Making a dedicated launcher test command (even if `--version-no-updates`) requires splitting those code paths. The ENV was a workaround.

PR #612, opened 2020-06-08
Fix Munki tables when munki isn't installed

directionless:
When munki isn't installed, it's plist file obviously doesn't exist. Other osquery tables return no results for non-exist files. Now this does as well.

Along the way, refactor and add some tests.

### DIFF
--- a/pkg/osquery/tables/munki/munki_test.go
+++ b/pkg/osquery/tables/munki/munki_test.go
@@ -90,9 +90,3 @@ func TestGenerateMunkiReport(t *testing.T) {
 		})
 	}
 }
-
-func TestGenerateMunkiInstalls(t *testing.T) {
-	t.Parallel()
-
-	//m := New()
-}
--- a/pkg/osquery/tables/munki/munki_test.go
+++ b/pkg/osquery/tables/munki/munki_test.go
@@ -1,0 +1,98 @@
+// +build darwin
+
+package munki
+
+import (
+	"context"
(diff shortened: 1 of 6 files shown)


pkg/osquery/tables/munki/munki_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		})
+	}
+}
+
+func TestGenerateMunkiInstalls(t *testing.T) {
+	t.Parallel()
+
+	//m := New()
+}
```
blaedj, line 98:
```suggestion
```

directionless: oops. Good catch

blaedj:
@directionless are you able to repro the original issue locally? The error doesn't seem to pop up for me locally on master, but I may be doing something wrong. I assumed building on master, then running `make osqueryi-tables` and `select * from kolide_munki_installs;` would generate an error, but I don't see one.

directionless:
> @directionless are you able to repro the original issue locally? The error doesn't seem to pop up for me locally on master, but I may be doing something wrong. I assumed building on master, then running `make osqueryi-tables` and `select * from kolide_munki_installs;` would generate an error, but I don't see one.

Yes. 

The code is pretty clear about this error. But I did reproduce locally.

blaedj:
Yep I thought the error source was pretty clear too, but I was a bit surprised I couldn't verify the error.

PR #613, opened 2020-06-08
Allow set and override of osquery flags

### LINKED ISSUE #96: Add osquery flagfile option
We start osqueryd with an `exec` command and pass all configuration via CLI flags. 
Sometimes it is necessary to adjust the configuration osqueryd launches with, but that would mean modifying the launcher source code -- for example, starting osqueryd with `--tls_dump`, `--verbose` or `--debug` flags. 

I propose we write the osqueryd flags into `/etc/kolide/flags` and allow osquery to be restarted with adjusted options. 

Thoughts?

zwass:
Adds a new option `--osquery_flag` that allows the user to provide a
flag that will be passed directly to osquery. These flags are appended
to all other default and computed flags, allowing them to potentially
override the default and computed flags. This allows users to set flags
that would not otherwise be possible to use with Launcher (like
`windows_event_channels`), and also opens up additional uses of Launcher
with overridden options, plugins, etc.

Support is also added to the package tooling to build packages with this
flag set.

Closes #96


cmd/package-builder/package-builder.go
```diff
@@ -26,6 +26,19 @@ func runVersion(args []string) error {
 	return nil
 }
 
+// Adapted from
```
zwass, line 29: Was not sure if you would want to add this to the public API of one of the packages, or just have a slight bit of duplication.

cmd/launcher/options.go
```diff
@@ -200,6 +216,8 @@ func shortUsage(flagset *flag.FlagSet) {
 	fmt.Fprintf(os.Stderr, "\n")
 	printOpt("version")
 	fmt.Fprintf(os.Stderr, "\n")
+	printOpt("osquery_flag")
```
directionless, line 219: I think this belongs in `developerUsage` not `shortUsage`. I can be convinced otherwise, but my gut sense is pretty clear. I think this is advanced/uncommon usage.

pkg/osquery/runtime/runtime.go
```diff
@@ -181,6 +182,12 @@ func (opts *osqueryOptions) createOsquerydCommand(osquerydBinary string, paths *
 		cmd.Stderr = opts.stderr
 	}
 
+	// Apply user-provided flags last so that they can override any other flags
```
directionless, line 185: I am _very_ leery of this footgun, I spend some time debugging footguns.

What settings do you want to override? I probably feel most protective of:
```
		fmt.Sprintf("--pidfile=%s", paths.pidfilePath),
		fmt.Sprintf("--database_path=%s", paths.databasePath),
		fmt.Sprintf("--extensions_socket=%s", paths.extensionSocketPath),
		fmt.Sprintf("--extensions_autoload=%s", paths.extensionAutoloadPath),
		"--extensions_timeout=10",
		fmt.Sprintf("--config_plugin=%s", opts.configPluginFlag),
```

Though I'm not totally sure on config_plugin

cmd/package-builder/package-builder.go
directionless: I am totally agnostic if we expose it here. Awesome that you did the work

cmd/launcher/options.go
```diff
@@ -22,6 +22,19 @@ const (
 	skipEnvParse         = runtime.GOOS == "windows" // skip environmental variable parsing on windows
 )
 
+// Adapted from
+// https://stackoverflow.com/questions/28322997/how-to-get-a-list-of-values-into-a-flag-in-golang/28323276#28323276
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return strings.Join(*i, " ")
```
directionless, line 30: (nit) Do we call this? And if we do, is `","` more appropriate? (I don't see it called)

zwass: We don't call it, it's just required for the interface. I don't actually know when it would ever be renderd by this method.

pkg/osquery/runtime/runtime.go
zwass: I can drop those in after these and document that those flags cannot be overridden. If it's ever an issue for someone, they can bring it up and propose a solution.

zwass:
@directionless Thank you for the review. I believe I addressed all of your concerns.

directionless:
Agree this looks good. I think I fixed this CI issue in #612 rebase?

PR #602, opened 2020-04-17
Force osquery to verbose mode

directionless:
Force osquery into verbose mode as a mitigation for distributed query correlation.

This is expected to be reverted at some future point.

PR #617, opened 2020-06-19
Filter out excess verbosity

directionless:
Launcher is currently running osquery in verbose mode. This is because we need a single line of the verbose output. While we're doing this, we can filter out the excess verbosity.

This pair with #602


directionless:
Not sure if we'll merge this, or revert the extra verbosity completely. Either way it's good to have it up.

directionless:
Replaced by #621

Issue #614, opened 2020-06-09
Table Request: macOS Screenlock from outside user-context


FritzX6:
## What is being requested?

With the recent addition of the [macOS screenlock table](https://github.com/osquery/osquery/pull/6243) we can now determine the state of the Screenlock setting when a user is logged into a device. This is inherently limited to the current logged in user, meaning the value cannot be reliably returned by osqueryd when it is run as root. I would like to see us use the previous `exec` pattern to launch osqueryi as the current logged in user and output the screenlock settings for example:

```
sudo launchctl asuser ` stat -f '%u' /dev/console` osqueryi "SELECT user AS username, enabled, grace_period FROM logged_in_users, screenlock WHERE tty = 'console';" --json
Password:
[
  {"enabled":"0","grace_period":"-1","username":"fritz-imac"}
]
```

Screenshot of accompanying screen: 
![image](https://user-images.githubusercontent.com/931724/84183728-58187d80-aa5a-11ea-9a7e-5033f11a8fd0.png)

## Desired Output:

```
SELECT * FROM kolide_screenlock;

+------------+---------+--------------+
| username   | enabled | grace_period |
+------------+---------+--------------+
| fritz-imac | 0       | -1           |
+------------+---------+--------------+
```

PR #618, opened 2020-06-22
Screenlock (and cmdwrapper)

directionless:
Expose the `screenlock` table via a simple exec.Command wrapper around osquery. This wrapper supports `runas` as timeout functionality.

It is expected to move most exec.Command functionality to it.

Fixes: #614


pkg/osquery/tables/screenlock/screenlock.go
```diff
@@ -0,0 +1,123 @@
...
+				"enabled":      row.Enabled,
+				"grace_period": row.GracePeriod,
+			}
+			results = append(results, res)
+		}
+	}
+	return results, nil
+}
+
+const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_"
```
FritzX6, line 114: This constraint should include `-` for example my username would throw an error:
```
osquery> SELECT * FROM kolide_screenlock WHERE user = 'fritz-imac';
{"caller":"screenlock.go:65","msg":"Disallowed character in user expression","severity":"info","ts":"2020-06-22T12:13:49.219496Z"}
I0622 08:13:49.219754 435154368 virtual_table.cpp:1008] Invalid response from the extension table. Error 1: error generating table: Disallowed character in user expression
Error: error generating table: Disallowed character in user expression
osquery> 
```

Likewise we should have some consideration for diacritical marks such as: `Björk`

directionless: Yeah... that's a good point. I'll update that.

directionless: Actually.... this table really only needs the user id, not the username. Do you think it makes sense to use that instead? Less lookups

FritzX6: Because this table only works when run as the current logged_in_user I imagined the ideal scenario was as proposed in the issue, where we join on `logged_in_users` and return only the `console` type user, and there is no output if a user is not currently logged in:
```
"SELECT user AS username, enabled, grace_period FROM logged_in_users, screenlock WHERE tty = 'console';"
```

directionless: I think one of the things we've concludes from the last year of use, is that launcher tables are better as small building blocks. Logic should be up a level, and conveyed via sql.

Accordingly, I think this should _do less_ and just take user id as a parameter.

directionless: I'm even skeptical that we want this table, and not something more general. But starting with this seems like an easy trial.

directionless: After thinking a bunch more, and tinkering with code, I've concluded that I want to do the initial implementation as a requires `user`.  I believe our main fear was getting misleading data. Testing, we appear fine. I get the same sorts of misleading data for all users vs logged in users. See comments in the new PR

I've opted to re-do the PR, using something simpler. I don't think the cmdwrapper is a win at this time.

directionless:
Replaced by #619

Issue #544, opened 2019-11-15
Table Request: startup_security


FritzX6:
What is being requested?
---
A new virtual table called `startup_security`

What is the background / purpose of this table?
---
Apple devices (especially those with T2 chips) have a variety of additional security measures that can be configured for pre-boot.

- [ ] EFI Firmware Password (boolean) 
- [x] ~Secure Boot (Full, Medium, No Security, N/A)~ **Addressed by:** https://github.com/osquery/osquery/pull/6053
- [x] ~External Boot (Disallowed, Allowed)~ **Addressed by:** https://github.com/osquery/osquery/pull/6053

Several administrators have asked for a way to enumerate these settings using osquery for their own compliance requirements.

![image](https://user-images.githubusercontent.com/931724/68952598-22ded200-078e-11ea-89c5-0619a1b39025.png)

How can this data be retrieved?
---
While these values are stored partially in nvram, they are inaccessible by querying the existing osquery `nvram` table. The values for each must be called discretely specifying the associated GUID.

For example:

```
➜  ~ nvram -p | grep Fritz               
good-samaritan-message	Hi Fritz
```
```
➜  ~ nvram -p | grep StartupManagerPolicy
➜  ~ nvram 5eeb160f-45fb-4ce9-b4e3-610359abf6f8:StartupManagerPolicy
5eeb160f-45fb-4ce9-b4e3-610359abf6f8:StartupManagerPolicy	%00
```
A clever user on the Jamf forums published a way to retrieve these GUID's: https://www.jamf.com/jamf-nation/discussions/29478/secure-boot-ea

These GUID's are static across devices (all devices use the same GUID's), and they can be found using the following method:
- Open a Terminal and run:
- ```sudo sysdiagnose```
- Expand the archive it creates
- Open the `securebootvariables.txt`

For simplicity's sake I have copied the output of my own `securebootvariables.txt` file below:
```
94b73556-2197-4702-82a8-3e1337dafbfb:ApChipID	%12%80%00%00
94b73556-2197-4702-82a8-3e1337dafbfb:ApBoardID	%0a%00%00%00
94b73556-2197-4702-82a8-3e1337dafbfb:ApSecurityDomain	%01%00%00%00
94b73556-2197-4702-82a8-3e1337dafbfb:ApProductionStatus	%01
94b73556-2197-4702-82a8-3e1337dafbfb:ApSecurityMode	%01
[...]

PR #620, opened 2020-06-26
Add firmwarepassword table

directionless:
Add a table based on exec'ing `/usr/sbin/firmwarepasswd`. This also
plumbs in most of a exec and parse interface for similar work.

Fixes: #544

Needs tests

### DIFF
--- a/pkg/osquery/tables/firmwarepasswd/parser.go
+++ b/pkg/osquery/tables/firmwarepasswd/parser.go
@@ -3,11 +3,9 @@ package firmwarepasswd
 import (
 	"bufio"
 	"bytes"
-	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/pkg/errors"
 )
 
 type Matcher struct {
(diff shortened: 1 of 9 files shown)


pkg/osquery/tables/firmwarepasswd/parser.go
```diff
@@ -0,0 +1,95 @@
...
+		results = append(results, row)
+
+	}
+	if err := scanner.Err(); err != nil {
+		level.Debug(p.logger).Log("msg", "scanner error", "err", err)
+	}
+	return results
+}
+
+func discernValBool(in string) (bool, error) {
```
blaedj, line 86: I was slightly surprised that this func was in `parser.go`, and not in `firmwarepasswd.go`, since that's where it's used. Upon further reflection, it does feel like a 'parser-y' thing, so I can see why it's here, but for the one usage, I would very mildly prefer it closer to where it's being used, since the package is the same.

directionless: Sure. I'll move it. I think it kinda made sense when I thought about a larger refactoring, but it's kinda weird now.

PR #622, opened 2020-06-29
Simpler firmwarepassword table

directionless:


directionless:
After some discussion with the team, we think that #620 is simpler.

PR #623, opened 2020-07-03
Add namespace support to kolide_wmi

directionless:
Adds support for specifying the connection parameters to the underlying `wmi` package, and plumbs namespace through to the table.

To improve code readability, also adds a `tablehelper` function to extract simple arrays of specified constraints. 

The test suite for these packages has been manually compiled and run on a windows machine


pkg/wmi/wmi.go
```diff
@@ -8,12 +8,27 @@
 //
 // To understand the available classes, take a look at the Microsoft
 // documention [3]
-
+//
+// Servers, Namespaces, and connection parameters:
+//
+// WMI has a fairly rich set of connection options. It allows querying
+// on remove servers, via authentiated users names, in different name
```
blaedj, line 15:
```suggestion
// on remote servers, via authenticated users names, in different name
```

PR #625, opened 2020-07-06
MSI User Interface

directionless:
Add a basic user interface to the MSI. This is based on wix's `WixUI_Minimal`, but removes some of the screens we don't need. This is not currently exposed through `package-builder`

An additional notable change is to the MSI's `ProductCode`. This introduces a random  factor, ensuring that each MSI has a unique `ProductCode`. Which means that any MSI will be able to install over other versions. This is an unusual choice from an MSI perspective, but it fit with out minimal usage of it


pkg/packagekit/wix/wix.go
```diff
@@ -76,6 +83,22 @@ func WithDocker(image string) WixOpt {
...
+func WithUI() WixOpt {
+	return func(wo *wixTool) {
+		wo.ui = true
+
+	}
+}
+
+func WithFile(name string, content []byte) WixOpt {
+	return func(wo *wixTool) {
+		if wo.extraFiles == nil {
```
blaedj, line 95: I was a little surprised that this nil check was needed, since it seemed like the ' value for the type `[]extraFile` would be an empty slice, and thus `append`-able. I was playing around with this [here](https://play.golang.org/p/ZtFXCvxF66L), and I _think_ that we don't need the nil check, does that seem correct to you?

directionless: I think you're right. `append` totally works on nil.

PR #626, opened 2020-07-06
Pin actions hash, add timeout

williamtheaker:
* Pins setup-go at v2
* Pins checkout at v2.0.0
* Adds a 25 minute timeout to keep unresponsive builds from eating up your minutes

This PR pins the SHA1 commit hash for external actions, since an action maintainer could break your build process by tagging a later commit as `v1` or a malicious actor could compromise your build process. It's safer to point towards commit hashes instead. Git and [Github have protections against SHA1 collisions](https://github.blog/2017-03-20-sha-1-collision-detection-on-github-com/).

The `setup-go` step might not be necessary since the runners already have [Go installed on them](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/software-installed-on-github-hosted-runners).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   build_and_test:
     name: launcher
+    timeout-minutes: 25
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false # Consider changing this sometime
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -21,12 +22,12 @@ jobs:
     steps:
     - name: Check out code
       id: checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@722adc63f1aa60a57ec37892e133b1d319cae598 # Pinned at v2.0.0
```
directionless, line 25: Seems like the perpetual tradeoff between making it easy to get updates, vs pinning to something known.

I haven't seen much discussion about the risk here, and to be honest, was assuming that the `actions` org was somewhat officially blessed by github. 

I kinda wish there was a signed tag pattern or something

williamtheaker: I think the actions org is Github employees, but that doesn't prevent someone going rogue 🤷

PR #627, opened 2020-07-14
Add macOS ioreg table

directionless:
Add a `kolide_ioreg` table as a wrapper around exec calls to `ioreg`. Uses dataflatten.

Update `tablehelpers.GetConstraints` to make it a bit easier to handle these cases.


pkg/osquery/tables/ioreg/ioreg.go
```diff
@@ -0,0 +1,193 @@
...
+
+		// ioreg options. It feels awkward using the single
+		// character form, but the full names are too generic
+		// and conflict with our established dataflatten columns
+		table.TextColumn("c"),
+		table.IntegerColumn("d"),
+		table.TextColumn("k"),
+		table.TextColumn("n"),
+		table.TextColumn("p"),
+		table.IntegerColumn("r"), // boolean
```
blaedj, line 55: What do you think about something like: 
```suggestion
		table.TextColumn("ioreg_class"),
		table.IntegerColumn("ioreg_depth"),
		table.TextColumn("ioreg_key"),
		table.TextColumn("ioreg_name"),
		table.TextColumn("ioreg_plane"),
		table.IntegerColumn("ioreg_root"), // boolean
```

it's a bit wordy perhaps, but the single char names are really opaque..

we could also do something like: 
```go
		table.TextColumn("class_option"),
		table.IntegerColumn("depth_option"),
		table.TextColumn("key_option"),
		table.TextColumn("name_option"),
		table.TextColumn("plane_option"),
		table.IntegerColumn("root_option"), // boolean
```

directionless: I thought about it. I didn't really like any option, so I went with succinct. On the sql side, it ends up feeling pretty similar to the command line. `select * from ioreg where  c = 'IOHIDSystem and d = 4`

If you or @FritzX6 feels strongly, I don't mind changing it. But today I'm in a succinct mood. 

Either way. I think the giant set of nested loops needs a better idiom, but I don't have one yet. I guess I could take the first value, and log a warning. But I feel like the nesting here is more correct.

blaedj: I'd feel better about the succinct route if we had some documentation I think. 

I'm with you re: loops, that many nested loops is a pretty strong smell, but it seems to produce the correct behavior...

if the constraints were all `AND`-ed together, we could probably just combine them into a single exec of ioreg. If they're `OR`-ed though, we probably need the multiple calls. As I type that out though, the sqlite engine probably takes care of that for us, so we'd still need to get all the possible rows...

directionless: Probably also worth noting these aren't output columns -- they're input ones, ideally they'd be hidden as well, but `osquery-go` doesn't support that yet.

directionless: Semantically, they should be `OR`. I think it's probably wrong to specify more than one, but I don't like breaking the sql assumption of being able to say things like `foo in (a, b)`

What kind of documentation? For better or worse, we don't really document any of these tables.

pkg/osquery/tables/ioreg/ioreg.go
```diff
@@ -0,0 +1,193 @@
...
+	columns := []table.ColumnDefinition{
+		table.TextColumn("fullkey"),
+		table.TextColumn("parent"),
+		table.TextColumn("key"),
+		table.TextColumn("value"),
+		table.TextColumn("query"),
+
+		// ioreg options. It feels awkward using the single
+		// character form, but the full names are too generic
+		// and conflict with our established dataflatten columns
```
directionless, line 49:
```suggestion
		// ioreg input options. These match the ioreg 
		// command line. See the ioreg man page.
```

pkg/osquery/tables/ioreg/ioreg.go
blaedj: seeing as they're input columns, naming matters a lot less to me. I agree that hiding them would be nice, and as mentioned elsewhere, using this table is going to require `man ioreg` - that's fine. 

If/when we upstream this we can revisit documenting them more.

directionless: Made the comment more explicit.

PR #635, opened 2020-08-03
Record installer info in the windows registry

directionless:
This records installer information in the windows registry. It is equivalent to the `installer-info.json` info files left by the macOS and linux installers.

### Why the windows registry?

I spend a long while trying to do this in a similar looking json file. First, I experimented with wix's `XMLFile` and `XMLConfig` elements. I was unable to get those to work reliably.

I then spent awhile setting up a launcher `postinstall` subcommand that would create the `installer-info.json`. This worked, but calling it from wix felt a bit flakey. I had to use `ExeCommand` and go through some weird quoting hassle. It did not feel stable.

Then, I remembered that this is the sort of thing the windows registry is designed for. Both wix and osquery have straightforward support.

### DIFF
--- a/pkg/packagekit/internal/assets.go
+++ b/pkg/packagekit/internal/assets.go
@@ -351,7 +351,7 @@ start on {{ .Opts.StartOn }}
 stop on {{ .Opts.StopOn }}
 {{- end }}
 
-# Respawn upto 15 times within 5 seconds.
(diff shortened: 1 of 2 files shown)


directionless:
> we'll want to test with full packages

For sure. I tested in this PR, but once merged we'll need to give it a solid going over.

PR #636, opened 2020-08-03
DO NOT MERGE: Launcher windows postinstall exec

directionless:
Launcher postinstall work. Replaced by https://github.com/kolide/launcher/pull/635 Need to look through this to see if there's anything else to extract


directionless:
If we ever want to build postinstall functions, we'll revisit

PR #428, opened 2019-02-20
Add JSON RPC implementation of the launcher server

blaedj:
This commit: 
 - adds a new flag, `transport`, that allows launcher to specify a transport other than gRPC.
      -  gRPC is the default
     - some general moving about of client instantiation etc was necessary to support this
 - adds a json rpc transport implementation of the launcher service.

Co-authored-by: Jonathan Nogueira <jonathan@kolide.co> 
Co-authored-by: blaedj <blaedj@gmail.com>
Co-authored-by: seph <seph@directionless.org>

PR #637, opened 2020-08-05
Add jsonrpc go-kit boilerplate

directionless:
This continues the work in #428 and adds the missing endpoint encode/decode functions.

### DIFF
--- a/pkg/service/publish_results.go
+++ b/pkg/service/publish_results.go
@@ -3,6 +3,7 @@ package service
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/go-kit/kit/endpoint"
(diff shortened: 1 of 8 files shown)


pkg/service/publish_results.go
```diff
@@ -57,6 +58,18 @@ func decodeGRPCResultCollection(_ context.Context, grpcReq interface{}) (interfa
 	}, nil
 }
 
+func decodeJSONRPCResultCollection(_ context.Context, msg json.RawMessage) (interface{}, error) {
+	var req resultCollection
+
+	if err := json.Unmarshal(msg, &req); err != nil {
+		return nil, &jsonrpc.Error{
+			Code:    -32000,
```
blaedj, line 66: I was investigating why we use this magic number, and it's probably more correct to use `-32700`, as described in the constants for the [jsonrpc package docs](https://pkg.go.dev/github.com/go-kit/kit@v0.8.0/transport/http/jsonrpc?tab=doc). 

I'm also not sure how much we care about using the correct error codes, since we're writing both the client and the server, and `-32000` is defined in the [jsonrpc spec](https://www.jsonrpc.org/specification#error_object) as a valid application-specific server error code.

directionless: I copy pasted that from the docs, and left it all TBD. Do you think it makes sense to return a `jsonrpc.Error` or just `errors.Wrap` here?

blaedj: I think `jsonrpc.Error` is fine. I enjoy the thought of us returning relevant jsonrpc errors and error codes, such that someone using launcher could use their own jsonrpc-compliant server (unlikely as that may be).

directionless: I'm good with this then. I suspect there's future work in normalizing the GRPC errors and the JSON rpc ones. But it doesn't seem worth thinking hard now.

PR #639, opened 2020-08-13
Additional error handling (and optional debug file logging)

directionless:
Adds several additional logging capabilities. Panic is now caught, and logged. A handful of extra startup log lines have been added.

Adds the capability to mirror debug to a file. This greatly helps when debugging on windows, as extracting structed json from a file is simpler than extracting from the event log. This is something of an early preview -- I'd rather enable this by default, we until we have a log rotation ability, it won't work.

Lastly, this upgrades the go svc version

### DIFF
--- a/pkg/log/teelogger/teelogger.go
+++ b/pkg/log/teelogger/teelogger.go
@@ -1,0 +1,32 @@
+// Package filetee provides a go-kit compatible logger, which will
+// mirror logs into a tmp file. This is designed to aid in debugging.
+package teelogger
+
+import (
+	"github.com/go-kit/kit/log"
+)
+
+type teeLogger struct {
+	loggers []log.Logger
+}
+
+func New(loggers ...log.Logger) log.Logger {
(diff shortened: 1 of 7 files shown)


pkg/log/teelogger/teelogger.go
```diff
@@ -0,0 +1,31 @@
...
+func New(loggers ...log.Logger) log.Logger {
+	l := &teeLogger{
+		loggers: loggers,
+	}
+
+	return l
+}
+
+// Log will log to each logger. If any of them error, it will return a
+// random error.
```
blaedj, line 21: doesn't this actually return the _last_ error returned by a logger, rather than a random error? iiuc the ordering would depend on the order of the loggers inserted...

directionless: I mean... Yes, but I don't want to make any real promises here. The mirroring seems reasonable, but the error handling is janky and subject to change.

PR #650, opened 2020-09-08
Add username context to kolide_pwpolicy

terracatta:
This PR adds the `username` context to the `kolide_pwpolicy_table` so that users of launcher can obtain the password policy for a specified username. If no username is specified the table behaves as it did before 

This also fixes an issue where the first line of output from the `usr/bin/pwpolicy` binary is not valid XML and can occasionally cause the dataflatten library to fail to parse the plist (esp if that first line contains angle brackets)

### DIFF
--- a/pkg/osquery/table/platform_tables_darwin.go
+++ b/pkg/osquery/table/platform_tables_darwin.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kolide/launcher/pkg/osquery/tables/firmwarepasswd"
 	"github.com/kolide/launcher/pkg/osquery/tables/ioreg"
 	"github.com/kolide/launcher/pkg/osquery/tables/munki"
+	"github.com/kolide/launcher/pkg/osquery/tables/pwpolicy"
 	"github.com/kolide/launcher/pkg/osquery/tables/screenlock"
 	"github.com/kolide/launcher/pkg/osquery/tables/systemprofiler"
 	osquery "github.com/kolide/osquery-go"
@@ -40,10 +41,9 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		legacyexec.TablePlugin(),
 		dataflattentable.TablePlugin(client, logger, dataflattentable.PlistType),
 		dataflattentable.TablePluginExec(client, logger,
-			"kolide_pwpolicy", dataflattentable.PlistType, []string{"/usr/bin/pwpolicy", "getaccountpolicies"}),
-		dataflattentable.TablePluginExec(client, logger,
 			"kolide_apfs_users", dataflattentable.PlistType, []string{"/usr/sbin/diskutil", "apfs", "listUsers", "/", "-plist"}),
 		screenlock.TablePlugin(client, logger, currentOsquerydBinaryPath),
+		pwpolicy.TablePlugin(client, logger),
 		systemprofiler.TablePlugin(client, logger),
 		munki.ManagedInstalls(client, logger),
 		munki.MunkiReport(client, logger),
--- a/pkg/osquery/tables/pwpolicy/pwpolicy.go
+++ b/pkg/osquery/tables/pwpolicy/pwpolicy.go
@@ -1,5 +1,12 @@
 //+build darwin
 
+// Package pwpolicy provides a table wrapper around the `pwpolicy` macOS
+// command.
+//
+// As the returned data is a complex nested plist, this uses the
+// dataflatten tooling. (See
+// https://godoc.org/github.com/kolide/launcher/pkg/dataflatten)
+
 package pwpolicy
 
 import (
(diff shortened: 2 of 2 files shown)


pkg/osquery/table/platform_tables_darwin.go
```diff
@@ -39,11 +40,10 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		kextpolicy.TablePlugin(),
 		legacyexec.TablePlugin(),
 		dataflattentable.TablePlugin(client, logger, dataflattentable.PlistType),
-		dataflattentable.TablePluginExec(client, logger,
-			"kolide_pwpolicy", dataflattentable.PlistType, []string{"/usr/bin/pwpolicy", "getaccountpolicies"}),
```
directionless, line 43: I was thinking about whether I could write a generic way to pass args through this `exec` pattern, and alleviate the need to write full tables for each.

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+type Table struct {
+	client       *osquery.ExtensionManagerClient
+	logger       log.Logger
+	tableName    string
```
directionless, line 22: Don't really need this for a one off like this. You can hardcode it on line 46

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,133 @@
...
+const PwpolicyCmd = "getaccountpolicies"
+
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("fullkey"),
+		table.TextColumn("parent"),
+		table.TextColumn("key"),
+		table.TextColumn("value"),
+		table.TextColumn("query"),
+		table.TextColumn("username"),
```
directionless, line 36:
```suggestion

		table.TextColumn("username"),
```

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,133 @@
...
+	return table.NewPlugin(t.tableName, columns, t.generateExec)
+}
+
+func (t *Table) generateExec(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+	var username string
+
+	if q, ok := queryContext.Constraints["username"]; ok && len(q.Constraints) != 0 {
+		username = q.Constraints[0].Expression
+	}
```
directionless, line 55: This will work. but I think there are a couple of issues.

First, it's a singleton -- it won't handle being passed multiple `username` columns. It should instead, be a fore loop.

Second, I recently got annoyed at needing the same `if q, ok ...` stuff, and wrote a helper for this. Check out `tablehelpers.GetConstraints` You can see an example of this over in tablehelpers.GetConstraints Note the `tablehelpers.WithDefaults("")` there.

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd.Stderr = &stderr
+
+	level.Debug(t.logger).Log("msg", "calling %s", "args", PwpolicyPath, cmd.Args)
+
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrapf(err, "calling %s. Got: %s", PwpolicyPath, string(stderr.Bytes()))
+	}
+
+	// Remove first line of output because it always contains non-plist content
+	outputBytes := bytes.SplitAfterN(stdout.Bytes(), []byte("\n"), 2)[1]
```
directionless, line 95: What a special output format

directionless (commented):
Pretty close :)

I think there are 2 paths we could take. 

First, we could patch `dataflattentable.TablePluginExec` to handle the specialness this needs. (optional command line parameter, remove the first line of output). This is my general bias, as it seems useful later. However, these are also slower to get right. 

Second, this could be updated to follow the more recent ioreg nested forloops. I _think_ that's my current preferred pattern for these. 

Happy to work with you either way. Though the first path still feels better

directionless:
> I think there are 2 paths we could take.

I thought about how to implement the first path -- expanding the generalized exec. I think it's pretty ugly. So I would recommend that we ditch that idea (sorry) and instead make this look more like ioreg.

pkg/osquery/tables/pwpolicy/pwpolicy.go
terracatta: Fixed by copying similar pattern in ioreg

pkg/osquery/tables/pwpolicy/pwpolicy.go
terracatta: right??

pkg/osquery/tables/pwpolicy/pwpolicy.go
terracatta: ended up keeping it since I saw this pattern again in ioreg.

terracatta:
> So I would recommend that we ditch that idea (sorry) and instead make this look more like ioreg.

This work is no complete. Can you take another look?

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+const PwpolicyPath = "/usr/bin/pwpolicy"
+const PwpolicyCmd = "getaccountpolicies"
```
directionless, line 29:
```suggestion
const pwpolicyPath = "/usr/bin/pwpolicy"
const pwpolicyCmd = "getaccountpolicies"
```

These don't need to be exported variables

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (t *Table) execPwpolicy(ctx context.Context, args []string) ([]byte, error) {
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, PwpolicyPath, args...)
```
directionless, line 129:
```suggestion
	cmd := exec.CommandContext(ctx, pwpolicyPath, args...)
```

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,143 @@
...
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	gcOpts := []tablehelpers.GetConstraintOpts{
+		tablehelpers.WithDefaults(""),
+		tablehelpers.WithLogger(t.logger),
+	}
+
+	for _, pwpolicyUsername := range tablehelpers.GetConstraints(queryContext, "username", gcOpts...) {
+		pwpolicyArgs := []string{PwpolicyCmd}
```
directionless, line 67:
```suggestion
		pwpolicyArgs := []string{pwpolicyCmd}
```

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	gcOpts := []tablehelpers.GetConstraintOpts{
+		tablehelpers.WithDefaults(""),
+		tablehelpers.WithLogger(t.logger),
+	}
+
+	for _, pwpolicyUsername := range tablehelpers.GetConstraints(queryContext, "username", gcOpts...) {
```
directionless, line 66:
```suggestion
	for _, pwpolicyUsername := range tablehelpers.GetConstraints(queryContext, "username", tablehelpers.WithDefaults("")) {
```

I don't think the gcOpts gets you much here

pkg/osquery/tables/pwpolicy/pwpolicy.go
```diff
@@ -0,0 +1,143 @@
...
+		pwpolicyArgs := []string{PwpolicyCmd}
+
+		if pwpolicyUsername != "" {
+			pwpolicyArgs = append(pwpolicyArgs, "-u")
+			pwpolicyArgs = append(pwpolicyArgs, pwpolicyUsername)
```
directionless, line 71:
```suggestion
			pwpolicyArgs = append(pwpolicyArgs, "-u", pwpolicyUsername)
```

directionless (approved):

PR #652, opened 2020-09-11
Add new table kolide_net_accounts

### LINKED ISSUE #580: Table Request: net_accounts (Windows Security Policy)
## What is being requested?

A new virtual table called `net_accounts`

Several users have requested a way to inspect the authentication requirements for their device in order to meet security benchmarks.

This table would output the values stored in this dialog:
image

## How can we get this information?

These values can be retrieved using an elevated permission command prompt and typing the following command:
```
net accounts
```

Output of this command is demonstrated below:
```
C:\Windows\system32>net accounts
Force user logoff how long after time expires?:       Never
Minimum password age (days):                          0
Maximum password age (days):                          42
Minimum password length:                              6
Length of password history maintained:                None
Lockout threshold:                                    Never
Lockout duration (minutes):                           30
Lockout observation window (minutes):                 30
Computer role:                                        WORKSTATION
The command completed successfully.
```

## What would the virtual table look like?

```shell
osquery> SELECT * FROM net_accounts;

+-------------------+-------------+-------------+----------------+-------------------------------+-------------------+------------------+----------------------------+---------------+
| force_user_logoff | min_pwd_age | max_pwd_age | min_pwd_length | length_pwd_history_maintained | lockout_threshold | lockout_duration | lockout_observation_window | computer_role |
+-------------------+-------------+-------------+----------------+-------------------------------+-------------------+------------------+----------------------------+---------------+
| Never             | 0           | 42          | 6              | None                          | Never             | 30               | 30                         | WORKSTATION   |
+-------------------+-------------+-------------+----------------+-------------------------------+-------------------+------------------+----------------------------+---------------+
```

terracatta:
Add a new table to enumerate password policy information found in the windows command `net accounts` 

Example output
```
+-------------------+------------------+------------------+---------------------+------------------------------------+-------------------+------------------+----------------------------+---------------+
| force_user_logoff | min_password_age | max_password_age | min_password_length | length_password_history_maintained | lockout_threshold | lockout_duration | lockout_observation_window | computer_role |
+-------------------+------------------+------------------+---------------------+------------------------------------+-------------------+------------------+----------------------------+---------------+
| 0                 | 0                | 42               | 0                   | 0                                  | 0                 | 30               | 30                         | WORKSTATION   |
+-------------------+------------------+------------------+---------------------+------------------------------------+-------------------+------------------+----------------------------+---------------+
```

Closes #580

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,8 @@ grpc-extension: .pre-build
 # Convenience tools
 osqueryi-tables: table.ext
 	osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
+osqueryi-tables-windows: table.ext
+	C:\Program Files\osquery\osqueryd\osqueryd.exe -S --allow-unsafe --verbose --extension .\build\windows\tables.exe
 sudo-osqueryi-tables: table.ext
 	sudo osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
 launchas-osqueryi-tables: table.ext
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ grpc-extension: .pre-build
 osqueryi-tables: table.ext
 	osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
 osqueryi-tables-windows: table.ext
-	C:\Program Files\osquery\osqueryd\osqueryd.exe -S --allow-unsafe --verbose --extension .\build\windows\tables.exe
+	osqueryd.exe -S --allow-unsafe --verbose --extension .\build\windows\tables.exe
 sudo-osqueryi-tables: table.ext
 	sudo osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
 launchas-osqueryi-tables: table.ext
--- a/pkg/osquery/table/net_accounts.go
+++ b/pkg/osquery/table/net_accounts.go
@@ -1,3 +1,5 @@
+// +build windows
+
 package table
 
 import (
--- a/pkg/osquery/table/net_accounts.go
+++ b/pkg/osquery/table/net_accounts.go
@@ -1,0 +1,126 @@
+package table
+
+import (
+	"bytes"
+	"context"
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -53,6 +53,8 @@ grpc-extension: .pre-build
 # Convenience tools
 osqueryi-tables: table.ext
 	osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
+osqueryi-tables-windows: table.ext
+	C:\Program Files\osquery\osqueryd\osqueryd.exe -S --allow-unsafe --verbose --extension .\build\windows\tables.exe
```
terracatta, line 57: Hardcoded this path because it's where osquery is normally installed on Windows using the default package...and the package doesn't add it to the path. Felt ok for our purposes...

pkg/osquery/table/net_accounts.go
```diff
@@ -0,0 +1,126 @@
...
+func (t *Table) splitAndReadConfigLine(line string, skipConversionCheck bool) string {
+	val := strings.TrimSpace(strings.Split(line, ":")[1])
+
+	if skipConversionCheck {
+		return val
+	}
+
+	if _, err := strconv.Atoi(val); err == nil {
+		return val
+	}
```
directionless, line 103: This doesn't feel go'ish. 

I'd create `stringFromLine(line string) string` and `intFromLine(line string) string`

I'd also consider having them return a value and err, or maybe a value an ok (boolean) to handle the sentinel nature. Or maybe `stringFromLine(line string, default string) string`

directionless (changes requested):
I am unfond of this parsing style. (I know, we accepted something similar for touchid). I think it's error prone to slight command output differences. The counterpoint, is that it's more agnostic to language locals.

Other possibilities:

1. Find an API (obviously the best)
2. string or regex match each line
If (1) is impossible, I might favor (2), but I could be persuaded we should do it by line number.

If we really do want to do this as an exec, I wonder if there are other things. For example, `secedit.exe /export /cfg <path>` dumps all the security information, and has some options relating to GPO vs local stuff. See https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/secedit-export

https://serverfault.com/questions/91520/are-there-registry-settings-for-password-policies-on-windows-2008 finds a couple of registry keys. But also recommends the above command.

Makefile
terracatta: Sure np.

terracatta:
Closing, as discussed in chat. We are going to look at a broader capability around `secedit.exe`

Issue #232, opened 2017-12-11
unrecoverable DNS issue 


groob:
I'm having a hard time tracking the root cause of this, but noticing the behavior numerous times:

it's possible for launcher to fail DNS resolution once, and then continue failing even though the DNS resolution works outside of that process. The issue happens when resolving both notary.kolide.com and the remote gRPC server. 

Tracking this bug here and will update as I get more info.

directionless:
I suspect I've seen this several times as well.

directionless:
This is still present.

groob:
I suspect it's https://github.com/golang/go/issues/12524 or at least related.

directionless:
I'm pretty sure I've seen this on windows (#563 came out of that)

directionless:
The two cases I have both involve, I think, resolver changes. So it seems like something it failing to re-resolve correct. 

One case is in the `jsonrpc` side, which use `http.NewRequest`

The other case I have is in the autoupdate code path. Also `net/http`. The original bug also cited the grpc transports

directionless:
A likely upstream bugs https://github.com/golang/go/issues/23427

directionless:
I haven't seen this happen in a few years. So I'm guessing something got fixed.

PR #653, opened 2020-09-11
Set RunAtLoad to false in macOS launchd configuration

zwass:
The existing behavior conflicted with the KeepAlive configuration that
only starts the daemon if the enroll secret path is present.

Setting this option to false leaves the general case the same: If a
package includes an enroll secret, the daemon is started immediately on
completion of install.

Setting this option to false allows the less-common case to work as
intended: The daemon will only start after an enroll secret is written
to the expected path.

The previous behavior would be to start the daemon even if no enroll
secret exists, and the daemon would then promptly exit with a nonzero status.

### DIFF
--- a/pkg/packagekit/render_launchd.go
+++ b/pkg/packagekit/render_launchd.go
@@ -53,7 +53,7 @@ func RenderLaunchd(ctx context.Context, w io.Writer, initOptions *InitOptions) e
 		StandardErrorPath: filepath.Join("/var/log", initOptions.Identifier, "launcher-stderr.log"),
 		StandardOutPath:   filepath.Join("/var/log", initOptions.Identifier, "launcher-stdout.log"),
 		KeepAlive:         keepAlive,
-		RunAtLoad:         true,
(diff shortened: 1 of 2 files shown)


directionless:
(I want to test this on a spare machine. I'll try to get to it this week)

directionless (changes requested):
I tested this on my test Big Sur machine and ran into some issues. Or perhaps, this uncovered existing ones.

The logs with this set had some errors along the lines of https://github.com/kolide/launcher/issues/232

zwass:
@directionless Do the errors go away when RunAtLoad is set? I'm not sure how that key could have any interaction with DNS resolution. If you actually saw the service start up (or try to) then I think that demonstrates this PR working as intended.

directionless:
> Do the errors go away when RunAtLoad is set? 

Yes. It works when RunAtLoad is set. I agree this is confusing.

I would theorize that either it changes the startup order. But it's not something I can debug just now.

zwass:
I am not working any further on this.

PR #654, opened 2020-09-11
Add option to install Windows Service in disabled state

zwass:
In some instances it is desirable to install the files and configure the
service but leave it in a disabled state until additional configuration
is performed.

This PR will be followed up with another that adds similar functionality
for systemd. There is no need to implement the same for macOS as the
launchd service is already configured to only start when the enroll
secret file exists.

### DIFF
--- a/pkg/packagekit/package_wix.go
+++ b/pkg/packagekit/package_wix.go
@@ -120,6 +120,11 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
 			wix.ServiceArgs([]string{"svc", "-config", po.FlagFile}),
 			wix.ServiceDescription(fmt.Sprintf("The Kolide Launcher (%s)", po.Identifier)),
 		)
+
+		if po.DisableService {
+			wix.WithDisabledService()(launcherService)
+		}
+
 		wixArgs = append(wixArgs, wix.WithService(launcherService))
 	}
 
--- a/pkg/packagekit/wix/service.go
+++ b/pkg/packagekit/wix/service.go
@@ -142,6 +142,15 @@ func WithDelayedStart() ServiceOpt {
 	}
 }
 
+func WithDisabledService() ServiceOpt {
+	return func(s *Service) {
+		s.serviceInstall.Start = "disabled"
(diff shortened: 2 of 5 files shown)


pkg/packagekit/package_wix.go
```diff
@@ -120,6 +120,11 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
 			wix.ServiceArgs([]string{"svc", "-config", po.FlagFile}),
 			wix.ServiceDescription(fmt.Sprintf("The Kolide Launcher (%s)", po.Identifier)),
 		)
+
+		if po.DisableService {
+			wix.WithDisabledService()(launcherService)
```
directionless, line 125: ha. This is, obviously, a fine way to use the functional arguments. But I don't think I've seen it before. Took me a couple minute to figure out how it worked. <3

pkg/packagekit/wix/service.go
```diff
@@ -142,6 +142,15 @@ func WithDelayedStart() ServiceOpt {
 	}
 }
 
+func WithDisabledService() ServiceOpt {
+	return func(s *Service) {
+		s.serviceInstall.Start = "disabled"
```
directionless, line 147:
```suggestion
		s.serviceInstall.Start = wix.StartDisabled
```

pkg/packagekit/wix/service.go
```diff
@@ -142,6 +142,15 @@ func WithDelayedStart() ServiceOpt {
...
+		// If this is not set to empty the installer hangs trying to start the
+		// disabled service.
+		s.serviceControl.Start = ""
```
directionless, line 150: Huh. Apparently I didn't define a constant for `""`. Do you think we should define the constant, or leave this?

directionless (changes requested):
Seeing all the places things change, I certainly feel like I screwed up an abstraction.... But anyhow, use the constants, and I'll merge!

pkg/packagekit/wix/service.go
zwass: I'll make a constant.

directionless (approved):

PR #656, opened 2020-09-13
Zero Tier Information Tables

directionless:
Add informational reporting tables for zero tier one

### DIFF
--- a/pkg/osquery/table/zerotier_windows.go
+++ b/pkg/osquery/table/zerotier_windows.go
@@ -2,12 +2,15 @@
 
 package table
 
-import "path"
+import (
+	"os"
+	"path"
+)
 
 // zerotierCli returns the path to the CLI executable.
 func zerotierCli(args ...string) []string {
 	cmd := []string{
-		path.Join(os.GetEnv("SYSTEMROOT"), "ProgramData", "ZeroTier", "One", "zerotier-one_x64.exe"),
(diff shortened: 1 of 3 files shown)


pkg/osquery/table/zerotier_windows.go
```diff
@@ -0,0 +1,15 @@
+// +build windows
+
+package table
+
+import "path"
+
+// zerotierCli returns the path to the CLI executable.
+func zerotierCli(args ...string) []string {
+	cmd := []string{
+		path.Join("C:", "ProgramData", "ZeroTier", "One", "zerotier-one_x64.exe"),
```
blaedj, line 10: It's probably not a huge deal, but such a hardcoded path makes me a little uneasy. 

We can likely change it easily enough, but the turnaround time for a launcher change is fairly long compared to changes in a query.. The obvious alternative of supplying an exec path as a query constraint doesn't seem any better to me though.

blaedj (approved):
lgtm

pkg/osquery/table/zerotier_windows.go
terracatta: Can we pass in some environment variables here? Like instead of `C:\` we could be using `SYSTEMROOT`

directionless: I updated this to use `os.GetEnv("SYSTEMROOT")` but until we see more of these, I don't think we there's a good way to be more clever .

Best case, we could hardcode a bunch of places to check. But without knowing more, it's needless boilerplate.

PR #658, opened 2020-09-16
Add where support to kolide_wmi table

directionless:
Add `where` clause support to the `kolide_wmi` table.

As with the rest of the WMI interface, this is ultimately pushing a string around.


FritzX6:
This capability will allow for the usage of the `CIM_Datafile` class which can be used to inspect file version information as demonstrated in @directionless 's PoC:

```
SELECT fullkey, parent, key, value 
FROM kolide_wmi 
WHERE class = 'CIM_DataFile' 
  AND properties = 'name,version' 
  AND whereclause = "name = 'c:\\windows\\system32\\notepad.exe'";
+-----------+--------+---------+---------------------------------+
| fullkey   | parent | key     | value                           |
+-----------+--------+---------+---------------------------------+
| 0/name    | 0      | name    | c:\windows\system32\notepad.exe |
| 0/version | 0      | version | 10.0.18362.693                  |
+-----------+--------+---------+---------------------------------+
```

blaedj (commented):
Should we add a test for multiple where clauses? it appears that the whereclauses are always effectively `AND`-ed together (i.e. a wmi query is run for each whereclause). This behavior is fine, but different enough from typical SQL semantics that it may warrant a test to illustrate expected behavior.

directionless:
> Should we add a test for multiple where clauses? it appears that the whereclauses are always effectively `AND`-ed together (i.e. a wmi query is run for each whereclause). This behavior is fine, but different enough from typical SQL semantics that it may warrant a test to illustrate expected behavior.

Not sure? This is as I'd expect, and is analogous to all the other dataflatten or exec places. (we uniformly have those forloops around options) But that you called out implies there's something missing.

Using a simple example, how would you expect queries to work with`path in (x, y, z)`? Under the hood, those are all invoked separately, and stitched together. sqlite then filters on top of that.

So actually, these may be as expected. If we have: `whereclause = "name= X" or whereclause = "name = Y"` it will be invoked twice, and sqlite will process them.

if you have `whereclause = "name= X" AND whereclause = "name = Y"` sqlite will always filter out all your data.

I guess the case we lose if something like `whereclause = "name = X" and whereclause = "version=Y"` but I think that's very hard to pass through correctly, and would require a parsing out the expression's operator to pass it along.

blaedj:
> Under the hood, those are all invoked separately, and stitched together. **sqlite then filters on top of that.**

(emphasis mine)

I _keep_ forgetting that, I know I've commented to this effect before. Thanks for re-educating me

blaedj (approved):

PR #661, opened 2020-09-17
Fix xml and ini bug in dataflatten table

directionless:
While doing some test case refactoring, I noticed the flatten options weren't being passed to through ini or xml tables. This caused the `query` parameter to get dropped. Fix argument passing.

Also update test cases

And refactor tables a little to use the current helpers.

### DIFF
--- a/pkg/osquery/tables/dataflattentable/tables_test.go
+++ b/pkg/osquery/tables/dataflattentable/tables_test.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"testing"
 
+	"github.com/go-kit/kit/log"
 	"github.com/kolide/launcher/pkg/dataflatten"
 	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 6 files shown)


pkg/osquery/tables/dataflattentable/tables_test.go
```diff
@@ -78,16 +82,76 @@ func TestDataFlattenTable_Animals(t *testing.T) {
...
+			testTables:   map[string]Table{"ini": Table{logger: logger, dataFunc: dataflatten.IniFile}},
+			testFile:     path.Join("testdata", "secdata.ini"),
+			queries:      []string{"Registry Values"},
+			expectedRows: 59,
+		},
+		{
+			testTables:   map[string]Table{"ini": Table{logger: logger, dataFunc: dataflatten.IniFile}},
+			testFile:     path.Join("testdata", "secdata.ini"),
+			queries:      []string{"this/does/not/exist"},
+			expectNoData: true,
```
blaedj, line 132: Nit: could this just be `expectedRows: 0`, and remove the conditional on line 147? (seemed to work as expected locally)

blaedj (approved):

pkg/osquery/tables/dataflattentable/tables_test.go
directionless: It _could_ but the test cases are kinda a mess. Some use `expectedRows` and some use `minRows` and `noData`.  I'm going to leave this as tech debt for a future pass

Issue #660, opened 2020-09-17
Table requests: Several Exec tables to run through dataflatten


FritzX6:
### What is being requested:
Several tables that I would like to see based on the exec'ed output of the `mdmclient` and the `profiles`:

### Profiles

The first is a new call to the `profiles` table to retrieve more comprehensive details of payloads that are being applied to various user accounts. I am not including my output here for privacy reasons but the command can be run on any device with profiles installed to examine structure.

```
sudo profiles -vPo stdout
```

Next is a call to determine the DEP enrollment details of a device. This is currently retrieved via a plist which may or may not be present on the device.
```
sudo profiles -e
```

### MDMclient

Next we have two calls to the `mdmclient` binary:

```
sudo /usr/libexec/mdmclient
```

**QuerySecurityInfo**

```
sudo /usr/libexec/mdmclient QuerySecurityInfo
Daemon response: {
    SecurityInfo =     {
        "FDE_Enabled" = 1;
        "FDE_HasInstitutionalRecoveryKey" = 0;
        "FDE_HasPersonalRecoveryKey" = 0;
        FirewallSettings =         {
            Applications =             (
                                {
                    Allowed = 1;
                    BundleID = "com.astro-hq.AstropadStudioMac";
                    Name = "Astropad Studio 2";
                },
                                {
                    Allowed = 1;
                    BundleID = "com.adobe.cep.CEPHtmlEngine Helper";
                    Name = "CEPHtmlEngine Helper";
                },
                                {
                    Allowed = 1;
                    BundleID = "com.xk72.Charles";
                    Name = Charles;
                }
            );
            BlockAllIncoming = 0;
            FirewallEnabled = 1;
            StealthMode = 0;
        };
        FirmwarePasswordStatus =         {
            AllowOroms = 1;
            ChangePending = 0;
            PasswordExists = 0;
        };
        ManagementStatus =         {
            EnrolledViaDEP = 0;
            IsActivationLockManageable = 0;
            IsUserEnrollment = 0;
            UserApprovedEnrollment = 1;
        };
        RemoteDesktopEnabled = 0;
        SecureBoot =         {
            ExternalBootLevel = disallowed;
            SecureBootLevel = full;
            WindowsBootLevel = disallowed;
        };
        SystemIntegrityProtectionEnabled = 1;
    };
}
Agent response: {
    SecurityInfo =     {
        ManagementStatus =         {
            EnrolledViaDEP = 0;
            IsActivationLockManageable = 0;
[...]

PR #662, opened 2020-09-17
Profiles table

directionless:
This is a new iteration on the `profiles` command.

* It uses the new profiles subcommand form.
* It uses dataflatten
* It replaces the `kolide_mdm_info` table

Fixes: #660

### DIFF
--- a/pkg/osquery/tables/profiles/profiles.go
+++ b/pkg/osquery/tables/profiles/profiles.go
@@ -75,6 +75,11 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 						profileArgs = append(profileArgs, "-type", profileType)
 					}
 
+					// setup the command line. This table overloads the `user`
+					// column so one can select either:
+					//   * All profiles merged, using the special value `_all` (this is the default)
+					//   * The device profiles, using the special value `_device`
+					//   * a user specific one, using the username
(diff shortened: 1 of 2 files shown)


pkg/osquery/tables/profiles/profiles.go
```diff
@@ -0,0 +1,163 @@
...
+
+					if profileType != "" {
+						profileArgs = append(profileArgs, "-type", profileType)
+					}
+
+					switch {
+					case user == "" || user == "_all":
+						profileArgs = append(profileArgs, "-all")
+					case user == "_device":
+						break
```
blaedj, line 82: why the `break` here? I assume that `_device` is a special value for user?

Testing out the `/usr/bin/profiles` command, this might be an invalid value?

blaedj (approved):

pkg/osquery/tables/profiles/profiles.go
directionless: Ah. this has a weird UI. I'm overloading the `users` column to allow selecting:
1. all profiles merged, using `_all` (or empty, as it's the default)
2. device profiles, using `_device` which corresponds to no additional command line flags. Thus break
3. user specific ones.

I added a comment

blaedj (approved):

PR #665, opened 2020-09-29
Fix findOsquery function

directionless:
Fix the findOsquery. It had flipped negatives. And, expand it.

### DIFF
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -314,32 +314,25 @@ func findOsquery() string {
 	)
 
 	for _, dir := range likelyDirectories {
-		maybeOsq := checkDirForOsquery(dir, osqBinaryName)
+		maybeOsq := filepath.Join(filepath.Clean(dir), osqBinaryName)
 
-		if maybeOsq != "" {
-			return maybeOsq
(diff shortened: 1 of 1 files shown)


cmd/launcher/options.go
```diff
@@ -298,15 +298,38 @@ func findOsquery() string {
...
+
+	if exPath, err := os.Executable(); err == nil {
+		likelyDirectories = append(likelyDirectories, filepath.Dir(exPath))
+	}
+
+	// Places to check. We could conditionalize on GOOS, but it doesn't
+	// seem important.
+	likelyDirectories = append(
+		likelyDirectories,
+		"/usr/local/kolide/bin",
```
blaedj, line 311: do we want to add `/usr/local/kolide-k2/bin` here?

directionless: Kinda blurs the line between the OSS side, and our normal SaaS usage, but I think it's reasonable. Good call.

directionless, line 311:
```suggestion
		"/usr/local/kolide/bin",
		"/usr/local/kolide-k2/bin",
```

blaedj (approved):

PR #666, opened 2020-10-01
DRY dataflatten based tables with a couple of helpers

directionless:
Add a couple of helpers to simplify dataflatten derived tables


pkg/dataflatten/to-map.go
```diff
@@ -0,0 +1,26 @@
+package dataflatten
+
+// ToMap is a helper function to convert Flatten output directly for
+// consumption by osquery tables.
+func ToMap(rows []Row, query string, rowData map[string]string) []map[string]string {
```
directionless, line 5: Not totally sure on this. I think it has to be a function that looks like this -- can't just have `Flatten` return a map. Probably can't make it really functional. And this isn't that bad...

Not sure if it should live in `pkg/dataflatten` or `tablehelpers`

blaedj (approved):

PR #667, opened 2020-10-01
Add table for gsettings

### LINKED ISSUE #508: Table Request: Gnome Settings
What is being requested?
---
A new table called `kolide_gnome_settings`, which would allow us to query the schema/key pairs managed by the `gsettings` command in Gnome environments.

Why is this useful?
---
Linux workstations running Gnome (most human-driven Ubuntu installations) manage a variety of settings with the `gsettings` command. These settings vary from things such as login screen options like `enable_user_switching` to screensaver lock settings.

How can this be implemented?
---
Via a plist/preferences like table whose schema consists of various schema/key pairs.

The raw gsettings output looks like this:
https://gist.github.com/FritzX6/9865a601316dd3d06068db0f2a81e294

**Screensaver subset:**
```
org.gnome.desktop.screensaver logout-enabled false
org.gnome.desktop.screensaver lock-enabled true
org.gnome.desktop.screensaver logout-delay uint32 7200
```

I suggest the following structure:

```
osquery> SELECT * from kolide_gnome_settings where domain = 'org.gnome.desktop.screensaver';
+--------------------------------+---------------------------+----------+------------------------------------------------------+
| domain                         | key                       | type     | value                                                |
+--------------------------------+---------------------------+----------+------------------------------------------------------+
| org.gnome.desktop.screensaver  | picture-opacity           | string   | 100                                                  |
| org.gnome.desktop.screensaver  | logout-enabled            | boolean  | 0                                                    |
| org.gnome.desktop.screensaver  | lock-enabled              | boolean  | 1                                                    |
| org.gnome.desktop.screensaver  | logout-delay              | uint32   | 7200                                                 |
| org.gnome.desktop.screensaver  | embedded-keyboard-enabled | boolean  | 0                                                    |
| org.gnome.desktop.screensaver  | primary-color             | string   | #023c88                                              |
| org.gnome.desktop.screensaver  | idle-activation-enabled   | boolean  | 1                                                    |
| org.gnome.desktop.screensaver  | secondary-color           | string   | #5789ca                                              |
| org.gnome.desktop.screensaver  | logout-command            | string   |                                                      |
| org.gnome.desktop.screensaver  | color-shading-type        | string   | solid                                                |
| org.gnome.desktop.screensaver  | embedded-keyboard-command | string   |                                                      |
| org.gnome.desktop.screensaver  | show-notifications        | boolean  | 0                                                    |
| org.gnome.desktop.screensaver  | picture-options           | string   | zoom                                                 |
| org.gnome.desktop.screensaver  | lock-delay                | uint32   | 0                                                    |
| org.gnome.desktop.screensaver  | show-full-name-in-top-bar | boolean  | 1                                                    |
| org.gnome.desktop.screensaver  | picture-uri               | string   | file:///usr/share/backgrounds/warty-final-ubuntu.png |
| org.gnome.desktop.screensaver  | status-message-enabled    | boolean  | 1                                                    |
| org.gnome.desktop.screensaver  | ubuntu-lock-on-suspend    | boolean  | 1                                                    |
| org.gnome.desktop.screensaver  | user-switch-enabled       | boolean  | 1                                                    |
+--------------------------------+---------------------------+----------+------------------------------------------------------+
```

Schema/Key pairs also possess descriptions which could be added to the table, but they can be very long and we should weigh the merit of their inclusion.

An example of a schema/key description:

```
$ gsettings describe org.gnome.desktop.screensaver logout-delay
The number of seconds after the screensaver activation before a logout option will appear in the unlock dialog. This key has effect only if the "logout_enable" key is set to TRUE.

DEPRECATED: This key is deprecated and ignored
```

blaedj:
This commit:
 -  adds `kolide_gsettings` and `kolide_gsettings_metadata` tables 
 - adds an 'osqueryi-tables' makefile target variant for linux

Closes #508

### Examples

```
osquery> select * from kolide_gsettings_metadata where schema = 'org.gnome.Mines' or schema = 'org.gnome.rhythmbox.sources';
+-----------------------------+---------------------+-----------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------+
| schema                      | key                 | description                                                                                                           | type                                                                |
+-----------------------------+---------------------+-----------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------+
| org.gnome.Mines             | nmines              | The number of mines in a custom game                                                                                  | int32                                                               |
| org.gnome.Mines             | window-width        | Width of the window in pixels                                                                                         | int32                                                               |
| org.gnome.Mines             | ysize               | Number of rows in a custom game                                                                                       | int32 (4 to 100)                                                    |
| org.gnome.Mines             | use-question-marks  | Set to true to be able to mark squares as unknown.                                                                    | bool                                                                |
| org.gnome.Mines             | use-autoflag        | Set to true to automatically flag squares as mined when enough squares are revealed                                   | bool                                                                |
| org.gnome.Mines             | use-animations      | Set to false to disable theme-defined transition animations                                                           | bool                                                                |
| org.gnome.Mines             | mode                | Size of the board (0-2 = small-large, 3=custom)                                                                       | int32 (0 to 3)                                                      |
| org.gnome.Mines             | xsize               | Number of columns in a custom game                                                                                    | int32 (4 to 100)                                                    |
| org.gnome.Mines             | theme               | The title of the tile theme to use.                                                                                   | string                                                              |
| org.gnome.Mines             | window-height       | Height of the window in pixels                                                                                        | int32                                                               |
| org.gnome.Mines             | window-is-maximized | true if the window is maximized                                                                                       | bool                                                                |
| org.gnome.rhythmbox.sources | visible-columns     | The list of columns that will be shown. If a given source doesn't support a particular column, it won't be displayed. | array of string                                                     |
| org.gnome.rhythmbox.sources | browser-views       | Views to show in the library browser.                                                                                 | enum: [ 'artists-albums','genres-artists','genres-artists-albums' ] |
+-----------------------------+---------------------+-----------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------+

```

```
osquery> select * from kolide_gsettings where (schema = 'org.gnome.Mines' or schema = 'org.gnome.rhythmbox.sources') and username = 'blaed';
+-----------------------------+---------------------+-----------------------------------------------------------------------+----------+
| schema                      | key                 | value                                                                 | username |
+-----------------------------+---------------------+-----------------------------------------------------------------------+----------+
| org.gnome.Mines             | nmines              | 40                                                                    | blaed    |
| org.gnome.Mines             | window-width        | 600                                                                   | blaed    |
| org.gnome.Mines             | ysize               | 16                                                                    | blaed    |
| org.gnome.Mines             | use-question-marks  | true                                                                  | blaed    |
| org.gnome.Mines             | use-autoflag        | false                                                                 | blaed    |
| org.gnome.Mines             | use-animations      | false                                                                 | blaed    |
| org.gnome.Mines             | mode                | 0                                                                     | blaed    |
| org.gnome.Mines             | xsize               | 16                                                                    | blaed    |
| org.gnome.Mines             | theme               | 'bgcolors'                                                            | blaed    |
| org.gnome.Mines             | window-height       | 400                                                                   | blaed    |
| org.gnome.Mines             | window-is-maximized | false                                                                 | blaed    |
| org.gnome.rhythmbox.sources | visible-columns     | ['track-number', 'artist', 'album', 'genre', 'duration', 'post-time'] | blaed    |
| org.gnome.rhythmbox.sources | browser-views       | 'artists-albums'                                                      | blaed    |
+-----------------------------+---------------------+-----------------------------------------------------------------------+----------+

```

### DIFF
--- a/pkg/osquery/table/platform_tables.go
+++ b/pkg/osquery/table/platform_tables.go
@@ -13,6 +13,7 @@ import (
 // unimplemented platforms.
 func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, currentOsquerydBinaryPath string) []*table.Plugin {
 	return []*table.Plugin{
-		gsettings.TablePlugin(client, logger),
+		gsettings.Settings(client, logger),
+		gsettings.Metadata(client, logger),
 	}
 }
--- a/pkg/osquery/table/platform_tables.go
+++ b/pkg/osquery/table/platform_tables.go
@@ -4,6 +4,7 @@ package table
 
 import (
 	"github.com/go-kit/kit/log"
+	"github.com/kolide/launcher/pkg/osquery/tables/gsettings"
 	osquery "github.com/kolide/osquery-go"
 	"github.com/kolide/osquery-go/plugin/table"
 )
@@ -11,5 +12,7 @@ import (
 // platformTables returns an empty set. It's here as a catchall for
 // unimplemented platforms.
 func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, currentOsquerydBinaryPath string) []*table.Plugin {
-	return []*table.Plugin{}
+	return []*table.Plugin{
+		gsettings.TablePlugin(client, logger),
+	}
 }
--- a/pkg/osquery/table/platform_tables_linux.go
+++ b/pkg/osquery/table/platform_tables_linux.go
@@ -1,9 +1,10 @@
-// +build !darwin,!windows,!linux
+// +build linux
 
 package table
 
 import (
 	"github.com/go-kit/kit/log"
+	"github.com/kolide/launcher/pkg/osquery/tables/gsettings"
 	osquery "github.com/kolide/osquery-go"
 	"github.com/kolide/osquery-go/plugin/table"
 )
@@ -11,5 +12,8 @@ import (
 // platformTables returns an empty set. It's here as a catchall for
 // unimplemented platforms.
 func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, currentOsquerydBinaryPath string) []*table.Plugin {
-	return []*table.Plugin{}
+	return []*table.Plugin{
+		gsettings.Settings(client, logger),
+		gsettings.Metadata(client, logger),
+	}
 }
--- a/pkg/osquery/table/platform_tables_linux.go
+++ b/pkg/osquery/table/platform_tables_linux.go
@@ -9,8 +9,6 @@ import (
 	"github.com/kolide/osquery-go/plugin/table"
 )
 
-// platformTables returns an empty set. It's here as a catchall for
-// unimplemented platforms.
 func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, currentOsquerydBinaryPath string) []*table.Plugin {
 	return []*table.Plugin{
 		gsettings.Settings(client, logger),
--- a/pkg/osquery/tables/gsettings/gsettings.go
+++ b/pkg/osquery/tables/gsettings/gsettings.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"os/user"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
@@ -25,7 +26,7 @@ const gsettingsPath = "/usr/bin/gsettings"
 
 const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-."
 
-type gsettingsExecer func(ctx context.Context, uid int, logger log.Logger, buf *bytes.Buffer) error
+type gsettingsExecer func(ctx context.Context, username string, logger log.Logger, buf *bytes.Buffer) error
 
 type GsettingsValues struct {
 	client   *osquery.ExtensionManagerClient
@@ -57,25 +58,24 @@ func (t *GsettingsValues) generate(ctx context.Context, queryContext table.Query
 
 	users := tablehelpers.GetConstraints(queryContext, "user", tablehelpers.WithAllowedCharacters(allowedCharacters))
 	if len(users) < 1 {
-		return results, errors.New("kolide_gsettings requires at least one user id to be specified")
+		return results, errors.New("kolide_gsettings requires at least one user name to be specified")
 	}
-
-	for _, u := range users {
+	for _, username := range users {
 		var output bytes.Buffer
-		uid, err := strconv.Atoi(u)
+
+		err := t.getBytes(ctx, username, t.logger, &output)
 		if err != nil {
-			level.Info(t.logger).Log("msg", "unable to cast supplied uid as int", "uid", uid)
-		}
-		err = t.getBytes(ctx, uid, t.logger, &output)
-		if err != nil {
-			// assume that getBytes logs the error.. might be better to log it here, but for debugging it's nice to log it in getBytes to append the stdout and stderr
-			// return results, errors.Wrap(err, "calling getbytes")
-			return results, errors.Wrap(err, "getting bytes")
+			level.Info(t.logger).Log(
+				"msg", "error getting bytes for user",
+				"username", username,
+				"err", err,
+			)
+			continue
 		}
 
 		user_results := t.parse(&output)
 		for _, r := range user_results {
-			r["user"] = u
+			r["user"] = username
 			results = append(results, r)
 		}
 	}
@@ -85,19 +85,18 @@ func (t *GsettingsValues) generate(ctx context.Context, queryContext table.Query
 
 // execGsettings writes the output of running 'gsettings' command into the supplied bytes buffer
 // TODO: would maybe be cleaner to make logger a functional option..
-func execGsettings(ctx context.Context, uid int, l log.Logger, buf *bytes.Buffer) error {
+func execGsettings(ctx context.Context, username string, l log.Logger, buf *bytes.Buffer) error {
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
 	defer cancel()
 
-	u, err := user.LookupId(string(uid))
+	u, err := user.Lookup(username)
 	if err != nil {
-		return errors.Wrap(err, "looking up user by uid")
+		level.Info(l).Log("msg", "unable to lookup user by username", "username", username)
+		return errors.Wrap(err, "finding user by username")
 	}
 
-	// cmd := exec.CommandContext(ctx, gsettingsPath, "list-recursively", "--schemadir", u.HomeDir)
 	cmd := exec.CommandContext(ctx, gsettingsPath, "list-recursively")
 
-	// EXPERIMENTAL: uncommenting this seems to cause no results to be returned.
 	// set the HOME for the the cmd so that gsettings is exec'd properly as the
 	// new user.
 	cmd.Env = append(os.Environ(), fmt.Sprintf("HOME=%s", u.HomeDir))
@@ -108,17 +107,19 @@ func execGsettings(ctx context.Context, uid int, l log.Logger, buf *bytes.Buffer
 		return errors.Wrap(err, "checking current user uid")
 	}
 
-	// adding a cmd.SysProcAttr.Credential sets the user the command is run
-	// under, as identified by the uid
-	if strconv.Itoa(uid) != currentUser.Uid {
-		// guid, err := strconv.Atoi(u.Gid)
-		// if err != nil {
-		// 	return errors.Wrap(err, "converting guid from string to int")
-		// }
+	if u.Uid != currentUser.Uid {
+		uid, err := strconv.Atoi(u.Uid)
+		if err != nil {
+			return errors.Wrap(err, "converting uid from string to int")
+		}
+		gid, err := strconv.Atoi(u.Gid)
+		if err != nil {
+			return errors.Wrap(err, "converting gid from string to int")
+		}
 		cmd.SysProcAttr = &syscall.SysProcAttr{}
 		cmd.SysProcAttr.Credential = &syscall.Credential{
 			Uid: uint32(uid),
-			Gid: 20,
+			Gid: uint32(gid),
 		}
 	}
 
@@ -128,9 +129,12 @@ func execGsettings(ctx context.Context, uid int, l log.Logger, buf *bytes.Buffer
 	}
 	defer os.RemoveAll(dir)
 
+	// if we don't chmod the dir, we get errors like:
+	// 'fork/exec /usr/bin/gsettings: permission denied'
 	if err := os.Chmod(dir, 0755); err != nil {
 		return errors.Wrap(err, "chmod")
 	}
+
 	cmd.Dir = dir
 
 	stderr := new(bytes.Buffer)
@@ -138,14 +142,15 @@ func execGsettings(ctx context.Context, uid int, l log.Logger, buf *bytes.Buffer
 	cmd.Stdout = buf
 
 	if err := cmd.Run(); err != nil {
-		// level.Error(l).Log(
-		// 	"msg", "error running gsettings",
-		// 	"stderr", strings.TrimSpace(stderr.String()),
-		// 	"stdout", strings.TrimSpace(buf.String()),
-		// 	"cmd", cmd.Args,
-		// 	"err", err,
-		// )
-		return errors.Wrapf(err, "exec-ing gsettings, cmd was %a, stderr: %s", cmd.Args)
+		level.Error(l).Log(
+			"msg", "error running gsettings",
+			"stderr", strings.TrimSpace(stderr.String()),
+			"stdout", strings.TrimSpace(buf.String()),
+			"cmd", cmd.Args,
+			"err", err,
+		)
+
+		return errors.Wrapf(err, "exec-ing gsettings")
 	}
 
 	return nil
@@ -162,26 +167,24 @@ func (t *GsettingsValues) parse(input *bytes.Buffer) []map[string]string {
 			continue
 		}
 
-		// parts := strings.SplitN(line, " ", 3)
-		// if len(parts) < 3 {
-		// 	level.Error(t.logger).Log(
-		// 		"msg", "unable to process line, not enough segments",
-		// 		"line", line,
-		// 	)
-		// 	continue
-		// }
+		parts := strings.SplitN(line, " ", 3)
+		if len(parts) < 3 {
+			level.Error(t.logger).Log(
+				"msg", "unable to process line, not enough segments",
+				"line", line,
+			)
+			continue
+		}
 		row := make(map[string]string)
-		// row["schema"] = parts[0]
-		// row["key"] = parts[1]
-		// row["value"] = parts[2]
+		row["schema"] = parts[0]
+		row["key"] = parts[1]
+		row["value"] = parts[2]
 
-		row["value"] = line
 		results = append(results, row)
 	}
 
 	if err := scanner.Err(); err != nil {
 		level.Debug(t.logger).Log("msg", "scanner error", "err", err)
-		panic(err)
 	}
 
 	return results
--- a/pkg/osquery/tables/gsettings/gsettings.go
+++ b/pkg/osquery/tables/gsettings/gsettings.go
@@ -107,11 +107,11 @@ func execGsettings(ctx context.Context, username string, buf *bytes.Buffer) erro
 	}
 
 	if u.Uid != currentUser.Uid {
-		uid, err := strconv.Atoi(u.Uid)
+		uid, err := strconv.ParseInt(u.Uid, 10, 32)
 		if err != nil {
 			return errors.Wrap(err, "converting uid from string to int")
 		}
-		gid, err := strconv.Atoi(u.Gid)
+		gid, err := strconv.ParseInt(u.Gid, 10, 32)
 		if err != nil {
 			return errors.Wrap(err, "converting gid from string to int")
 		}
--- a/pkg/osquery/tables/gsettings/gsettings.go
+++ b/pkg/osquery/tables/gsettings/gsettings.go
@@ -1,0 +1,282 @@
+package gsettings
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	osquery "github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+type Table struct {
+	client *osquery.ExtensionManagerClient
+	logger log.Logger
+}
+
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		// TODO: maybe need to add 'path' for relocatable schemas..
+		table.TextColumn("domain"),
+		table.TextColumn("key"),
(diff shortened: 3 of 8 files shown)


pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	domainQ, ok := queryContext.Constraints["domain"]
+	if !ok || len(domainQ.Constraints) == 0 {
+		return nil, errors.New("the kolide_gsettings table requires a domain to be specified")
+	}
+
+	for _, domainConstraint := range domainQ.Constraints {
```
directionless, line 52: Want to use `tablehelpers.GetConstraints` here, or stick with this?

directionless (commented):
Hrm. This is a bit of a different direction than the experiment I started (See https://github.com/kolide/launcher/compare/master...directionless:seph/gsettings)

Do you think it makes sense to use dataflatten? It's maybe overkill, but it's also now a common pattern.

I think this probably needs the equivlent of `sudo` permissions

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,282 @@
...
+type datatype struct {
+	raw string
+}
+
+type keyDescription struct {
+	Description string
+	Type        string
+}
+
+func (t *Table) gsettingsDescribe(ctx context.Context, key, domain string) (keyDescription, error) {
```
directionless, line 115: I think I don't like the N+1. 

I think there are 2 different functions. One is to use `list-recursively` when selecting a large block. The other is to get info about a specific setting.

I could believe doing that through either a single table that behaves different depending on what you call it with, or doing it though two tables.

pkg/osquery/tables/gsettings/gsettings.go
blaedj: oh yea, I'll use the tablehelpers

pkg/osquery/tables/gsettings/gsettings.go
blaedj: Yea I thought of adding a column that's only populated if you explicitly select from it, though it feels odd to break `select *`.  

Or (I think this is what you are referring to) we could only list the description column when it's explicitly mentiond, _and_ when there are constraints on the `key` column..

I'm not entirely against a 2nd table, but I think I prefer a single table. Still need to ponder this more

directionless: I see several models

1. As is. This selects everything. Too many execs
2. two tables. Maybe one to use `list-recursively` to get values. and a `gsettings_descriptions` with a required `key`
3. single table, only return descriptions if the column is selected (this is akin to a hidden column in osquery)
4. single table, only return descriptions if a key is specified. (eg: `select *` gets values, `select * where key in (a, b, c)` would also populate the description

At least this evening, I'm favoring (2) or (4)

blaedj: I kind of like (2) the more I think about it. `gsettings` and `gsettings_metadata` or something, we could have explicit 'type' and 'description' data there.

blaedj:
I'm less sure than I was before that dataflatten is appropriate here. I'm having trouble finding nested data in the the output of gsettings. There are some 'domains'/schema names that appear to have a common parent, e.g. `org.gnome.rythmbox.plugins` and `org.gnome.rythmbox.sharing` , but the data isn't really nested (at least not as gsettings presents it). 

Do you have any examples that would benefit from a dataflattening approach?

directionless:
> Do you have any examples that would benefit from a dataflattening approach?

Do you think it makes sense to split things like `org.gnome.rythmbox.plugins` on `.` or leave them as opaque strings? Maybe leaving them opaque is fine. I don't think I have an opinion here.

But... If we do, then it's inherently a 2 level map -- `domain/key: value`. And some of the values are arrays.  For example:
`org.gnome.rhythmbox.sources visible-columns ['track-number', 'artist', 'album', 'genre', 'duration', 'post-time']`

And once I start thinking about how to represent arrays, I realize I want to in dataflatten and be done with it

blaedj:
I don't think it makes sense to split on the `.`, because to get any information we need the entire domain. If we were the schemas in a tree maybe we could (there is a graphical dconf editor that does that), but even then the grouping seems fairly arbitrary.

blaedj:
as mentioned in the (updated) PR description, I didn't see a clear path to running this data through `dataflatten.Flatten` without implementing support for the gsettings format. Please let me know if I'm missing something obvious, but it didn't seem worth the effort if we aren't planning to use the query and flattening features at the moment.

blaedj:
still need to add 'user' to get the settings per-user

blaedj:
- [X] fetching gsettings keys per-user is implemented 

@directionless ready for review again

pkg/osquery/table/platform_tables.go
```diff
@@ -4,12 +4,16 @@ package table
...
 )
 
 // platformTables returns an empty set. It's here as a catchall for
 // unimplemented platforms.
 func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, currentOsquerydBinaryPath string) []*table.Plugin {
-	return []*table.Plugin{}
+	return []*table.Plugin{
+		gsettings.Settings(client, logger),
+		gsettings.Metadata(client, logger),
+	}
```
directionless, line 18: This table won't build on windows, so this can't be here. We can probably skip this on darwin, so probably best to plumb in a `_linux` flavor here.

(If you think it's worth keeping on darwin, we can make another layer in here)

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,184 @@
...
+	cmd.Stdout = buf
+
+	if err := cmd.Run(); err != nil {
+		return errors.Wrapf(err, "running gsettings, err is: %s", stderr.String())
+	}
+
+	return nil
+}
+
+func (t *GsettingsValues) parse(input *bytes.Buffer) []map[string]string {
```
directionless, line 152: (nit) Shouldn't this should take an `io.Reader`? (bytes.Buffer is one)

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,184 @@
...
+				"err", err,
+			)
+			continue
+		}
+
+		user_results := t.parse(&output)
+		for _, r := range user_results {
+			r["user"] = username
+			results = append(results, r)
+		}
```
directionless, line 82: (nit) Feels weird to loop twice. Consider moving the username _into_ the row creation inside `parse`

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,184 @@
...
+	u, err := user.Lookup(username)
+	if err != nil {
+		return errors.Wrapf(err, "finding user by username '%s'", username)
+	}
+
+	cmd := exec.CommandContext(ctx, gsettingsPath, "list-recursively")
+
+	// set the HOME for the the cmd so that gsettings is exec'd properly as the
+	// new user.
+	cmd.Env = append(os.Environ(), fmt.Sprintf("HOME=%s", u.HomeDir))
```
directionless, line 103: Is there a reason you're passing `os.Environ()` here? That pushes the full environment from launcher, to the exec, generally I'd see that as a bad idea. (Not a nit)

Other places we might do:
```
cmd.Env = append(cmd.Env, fmt.Sprintf("HOME=%s", u.HomeDir))
```

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Settings returns a table plugin for querying setting values from the
+// gsettings command.
+func Settings(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("schema"),
+		table.TextColumn("key"),
+		table.TextColumn("value"),
+		table.TextColumn("user"),
```
directionless, line 46: Other tables, like touchid, we use `uid` for the numeric uid. I don't have a strong feeling about it, but I think we should aim to be consistent. @FritzX6 do you have an opinion?

pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -0,0 +1,184 @@
...
+			results = append(results, r)
+		}
+	}
+
+	return results, nil
+}
+
+// execGsettings writes the output of running 'gsettings' command into the
+// supplied bytes buffer
+func execGsettings(ctx context.Context, username string, buf *bytes.Buffer) error {
```
directionless, line 90: (Nit) You could consider adding an optional `schema` argument here, so as to not always list everything. But it's fast enough I don't think it's required.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+		table.TextColumn("schema"),
+		table.TextColumn("key"),
+		table.TextColumn("description"),
+		table.TextColumn("type"),
+	}
+
+	t := &GsettingsMetadata{
+		client:    client,
+		logger:    logger,
+		cmdRunner: execGsettingsCommand,
```
directionless, line 45: (nit) feels weird to use a different signature here, then above. It's basically the same functionality.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	return table.NewPlugin("kolide_gsettings_metadata", columns, t.generate)
+}
+
+func (t *GsettingsMetadata) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+
+	var results []map[string]string
```
directionless, line 53:
```suggestion
	var results []map[string]string
	
```

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+
+func (k *keyDescription) toMap() map[string]string {
+	return map[string]string{
+		"description": k.Description,
+		"type":        k.Type,
+		"key":         k.Key,
+	}
+}
+
+func (t *GsettingsMetadata) gsettingsDescribeForSchema(ctx context.Context, schema string) ([]keyDescription, error) {
```
directionless, line 102: I can't decide if these are nits, or more real.

This should maybe take an optional `key` argument? Not sure. I don't love the n+1. But, it's probably fine. I guess we'd end up with the same number of execs, they'd just be driven by a join.

I really don't love how many functions this expands to. It makes it very hard to read. I can talk myself into keeping `getType` and `getDescription`, especially since those are obvious places to hang parsing unit tests. But I'm a lot less sold on having `gsettingsDescribeForSchema`, `listKeys`, and `describeKey`.

On reflection, these are probably ignorable nits.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+	return convertType(result), nil
+}
+
+func execGsettingsCommand(ctx context.Context, args []string, output *bytes.Buffer) error {
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+	defer cancel()
+
+	command := args[0]
+	cmd := exec.CommandContext(ctx, gsettingsPath, args...)
+	dir, err := ioutil.TempDir("", fmt.Sprintf("osq-gsettings-%s", command))
```
directionless, line 222: I like the tempdir, and I even like the tempdir down in the exec command (where it's needed). But the effect is that we're creating a huge number of them when this is called. Probably cleaner to create one in a parent and pass it along.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+
+// convertType returns a string describing the GVariantType corresponding to the
+// GVariant-formatted type string. see  https://developer.gnome.org/glib/unstable/glib-GVariantType.html
+// for documentation. Note that not all types listed in the documentation above
+// are supported, for example:
+//  - tuples (e.g. tuple of 2 strings `(ss)`)
+//  - nested types (e.g.// array of tuples: `a(ss)`)
+// and other complex types are not supported.
+func convertType(typ string) string {
+	typ = strings.Replace(typ, "type ", "", 1) // remove any leading 'type ', eg 'type b'
```
directionless, line 263: If you only want to strip the leading one, `strings.TrimPrefix`

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// convertType returns a string describing the GVariantType corresponding to the
+// GVariant-formatted type string. see  https://developer.gnome.org/glib/unstable/glib-GVariantType.html
+// for documentation. Note that not all types listed in the documentation above
+// are supported, for example:
+//  - tuples (e.g. tuple of 2 strings `(ss)`)
+//  - nested types (e.g.// array of tuples: `a(ss)`)
+// and other complex types are not supported.
+func convertType(typ string) string {
```
directionless, line 262: (nit) This is weird enough it should have tests.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,274 @@
...
+
+	err := t.cmdRunner(ctx, []string{"describe", schema, key}, output)
+	if err != nil {
+		return "", errors.Wrap(err, "describing key")
+	}
+
+	return strings.TrimSpace(output.String()), nil
+}
+
+func (t *GsettingsMetadata) getType(ctx context.Context, key, schema string) (string, error) {
```
directionless, line 182: The type conversion is weirdly spread across several things, and I'm not confident it's correct.  I think it's probably fine? I think it would also be fine to return it exactly as `gsettings` does, without this parsing. 

Tests would add clarity.

directionless (changes requested):
Only a couple real issues, should be trivial. 

Several nits, I think you can probably ignore them all. Up to you. 

An uncommented nit about the larger structure. I think you could do this with a single struct, instead of having basically identical ones for metadata and values. And I think there's weird naming inconsistencies around `gsettings`, `settings`, and `values`. But, I'm not sure it's worth refactoring right now.

pkg/osquery/tables/gsettings/metadata.go
blaedj: agreed. this is the result of me copy/pasting from the gsettings code into the metadata, then iterating on the original and not applying the changes here.

pkg/osquery/table/platform_tables.go
blaedj: funnily enough, I actually have `gsettings` installed on my mac (it was some homebrew dependency..). But I don't think we want to bother with this on darwin, it's more likely to lead to confusion than any actual benefit.

pkg/osquery/tables/gsettings/gsettings.go
blaedj: here, "user" is actually the user name, not the uid. Renamed the column to be more clear, but if folks feel strongly we can change the column to be a uid.

pkg/osquery/tables/gsettings/metadata.go
blaedj: Agreed that it's sorta spread out. I added some more tests for convertType, and added more comments to explain what getType does vs what convertType does.

blaedj: as far as returning exactly what gsettings outputs.. let's just say the gsettings cli tool uses very spartan type descriptions that I don't think are useful without the GVariant docs handy.

pkg/osquery/table/platform_tables_linux.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/go-kit/kit/log"
+	"github.com/kolide/launcher/pkg/osquery/tables/gsettings"
+	osquery "github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+)
+
+// platformTables returns an empty set. It's here as a catchall for
+// unimplemented platforms.
```
directionless, line 13:
```suggestion
```

pkg/osquery/tables/gsettings/metadata.go
blaedj: I hear you, as mentioned in conversation, let's clean this up the next time we touch this code.

pkg/osquery/tables/gsettings/gsettings.go
blaedj: I think I had it at one point, but removed it for simplicity. I'm 💯 on board with adding it in the future if we find that the execution time is too slow.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,283 @@
...
+	return results, nil
+}
+
+type keyDescription struct {
+	Description string
+	Type        string
+	Key         string
+}
+
+func (t *GsettingsMetadata) gsettingsDescribeForSchema(ctx context.Context, schema string) ([]keyDescription, error) {
```
directionless, line 90: (nit) the `keyDescription` isn't getting you anything. May as well accept `schema` here and return a `[]map[string]string`. also notable here is that returning an error doesn't matter. It just gets logged and continued. So you could do that here. 

If you combine these, I think it becomes much cleaner:

```
	for _, schema := range schemas {
          append(results, t.gsettingsDescribeForSchema(ctx, schema))
        }
```

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,283 @@
...
+	var descriptions []keyDescription
+
+	dir, err := ioutil.TempDir("", fmt.Sprintf("osq-gsettings-metadata-%s", schema))
```
directionless, line 93: (nit) you don't need all that extra stuff in the temp dir name. These are short lived, and just to catch any kinda weird behavior. `osq-gsettings` seems fine.

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// dir, err := ioutil.TempDir("", fmt.Sprintf("osq-gsettings-%s", command))
+	// if err != nil {
+	// 	return errors.Wrap(err, "mktemp")
+	// }
+	// defer os.RemoveAll(dir)
+
+	// if err := os.Chmod(dir, 0755); err != nil {
+	// 	return errors.Wrap(err, "chmod")
+	// }
```
directionless, line 237:
```suggestion
```

pkg/osquery/tables/gsettings/metadata.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+
+		return fmt.Sprintf("%s%s", typ, scope), nil
+	}
+
+	return convertType(result), nil
+}
+
+// execGsettingsCommand should be called with a tmpdir that will be cleaned up.
+func execGsettingsCommand(ctx context.Context, args []string, tmpdir string, output *bytes.Buffer) error {
```
directionless, line 222: (nit) should be io.Writer

directionless (approved):
A pile of nits, but I'd merge as-is

pkg/osquery/tables/gsettings/gsettings_test.go
```diff
@@ -0,0 +1,305 @@
...
+	t.Parallel()
+
+	currentUser, err := user.Current()
+	require.NoError(t, err)
+	if currentUser.Username != "blaed" {
+		// all these tests will only pass if run as root, and with specific
+		// gnome desktop settings set. The setup to make this test run in CI
+		// is... complex
+		t.Skip("skipping - proper setup not detected")
+	}
```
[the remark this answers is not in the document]
blaedj: yea, could be improved for sure. ENV var is probably better.

PR #669, opened 2020-10-13
Flatten fixes from #666

directionless:
In #666 I moved the if statement  about what happens if no `query` is specified. This inadvertently caused dataflatten to operate with a query of `""`, matching _nothing_ and overriding the default query of `*`. This moves the implicit `*` value to the `tablehelpers.GetConstraints` calls. This is more explicit.

Also add in a test case for pwpolicy


blaedj (approved):
lgtm

PR #682, opened 2020-12-12
Support architectures in the build/release side

directionless:
Update the build/release process to support creating darwin universal binaries. This requires building on an m1 machine, with go installed at the specified paths. 

To make this a bit smoother, it also updates a handful of the abstractions across our build tools.

This will likely get smoother as the m1 ecosystem evolves.


Makefile
```diff
@@ -33,62 +30,127 @@ else
...
-	go run cmd/make/make.go -targets=launcher -linkstamp
+##
+## Build
+##
 
-table.ext: .pre-build
-	go run cmd/make/make.go -targets=table-extension -linkstamp
-table.ext-windows: .pre-build deps
-	go run cmd/make/make.go -targets=table-extension -linkstamp --os windows
+build_%: TARGET =  $(word 2, $(subst _, ,$@))
```
blaedj, line 37: for future reference for myself (and any other make novices): https://www.gnu.org/software/make/manual/html_node/Text-Functions.html

Makefile
```diff
@@ -33,62 +30,127 @@ else
...
-osqueryi-tables-linux: table.ext
-	osqueryd -S --allow-unsafe --verbose --extension ./build/linux/tables.ext
-osqueryi-tables-windows: table.ext
-	osqueryd.exe -S --allow-unsafe --verbose --extension .\build\windows\tables.exe
-sudo-osqueryi-tables: table.ext
-	sudo osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
-launchas-osqueryi-tables: table.ext
-	sudo launchctl asuser 0 osqueryd -S --allow-unsafe --verbose --extension ./build/darwin/tables.ext
+# Actual release targets. Because of the m1 cgo cross stuff, this requires explicit go paths
+rel-amd64: CROSSGOPATH = /Users/seph/go1.15.6.darwin-amd64/bin/go
```
blaedj, line 83: Could this be `~/go1.14.6.darwin-amd64/bin/go` or does the path need to be that specific?

Makefile
directionless: Yep. I do that all over our Makefiles. I think it's more clever than I like, but make.

Makefile
directionless: It could probably use `~`. Once there's a real m1 release, I expect we'll drop these into nicer paths anyhow

blaedj (approved):

PR #684, opened 2020-12-17
Fix path in osqueryi targets

directionless:
With the arch now included in the build directory, the oqueryi paths
needs to change too


blaedj (approved):
nice and simple(r)

PR #686, opened 2020-12-20
Add kolide_filevault table

terracatta:
This PR adds a simple table that simply displays the output of `fdesetup status`. Unfortunately this tool does not provide any structured output, but as far as I an tell (through disassembling the binary in hopper) it will only produce a simple message.

Looking at dissambled binary, when the command is not run in verbose mode it will produce between 1 - 3 lines of output. The first line will always contain the following output.

* FileVault is On
* FileVault is On, but needs to be restarted to finish
* FileVault is Off.
* FileVault is Off, but will be enabled after the next restart.
* FileVault is Off, but needs to be restarted to finish.

The second line or third line, if shown, can contain one of the following strings:

* FileVault master keychain appears to be installed.
* Decryption in progress: Percent completed = 
* Encryption in progress: Percent completed =
* Decryption in progress: Pending
* Encryption in progress: Pending

None of these strings appear to be localized, they are all hardcoded.

I've opted to not get too smart about parsing the output other than trimming the last newline, since we want to leave it up to the caller

### DIFF
--- a/pkg/osquery/tables/filevault/filevault.go
+++ b/pkg/osquery/tables/filevault/filevault.go
@@ -1,0 +1,56 @@
+// +build darwin
+
+package filevault
+
+import (
+	"bytes"
+	"context"
+	"os/exec"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+const fdesetupPath = "/usr/bin/fdesetup"
+
+type Table struct {
+	client *osquery.ExtensionManagerClient
+	logger log.Logger
+}
+
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("status"),
+	}
+
+	t := &Table{
+		client: client,
+		logger: logger,
+	}
+
+	return table.NewPlugin("kolide_filevault", columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	// Read the system's fdesetup configuration
+	var stdout bytes.Buffer
+	stdout.Reset()
+	cmd := exec.CommandContext(ctx, fdesetupPath, "status")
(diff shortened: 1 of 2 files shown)


pkg/osquery/tables/filevault/filevault.go
```diff
@@ -0,0 +1,56 @@
...
+
+	return table.NewPlugin("kolide_filevault", columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	// Read the system's fdesetup configuration
+	var stdout bytes.Buffer
+	stdout.Reset()
```
directionless, line 42: Shouldn't need this.

pkg/osquery/tables/filevault/filevault.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd := exec.CommandContext(ctx, fdesetupPath, "status")
```
directionless, line 43: This should get a timeout. This is done through the `ctx` See https://github.com/kolide/launcher/blob/faa8c090d04ac7a904ede37d34127fa5de906d5f/pkg/osquery/table/mdfind_darwin.go#L15-L16

pkg/osquery/tables/filevault/filevault.go
```diff
@@ -0,0 +1,56 @@
...
+	cmd.Stdout = &stdout
+	if err := cmd.Run(); err != nil {
+		return nil, errors.Wrap(err, "calling fdesetup")
+	}
+	output := string(stdout.Bytes())
+	status := strings.TrimSuffix(output, "\n")
+	result := map[string]string{
+		"status": status,
+	}
+	results = append(results, result)
```
directionless, line 53: This is fine, but since this is a single line you could just return it directly. (FYI)

directionless (changes requested):
Mostly fine. Needs the timeout.

I generally try to have exec it a separate method -- it's more testable. But this is something simple enough I don't think we need parsing test cases. So I'd approve with the timeout.

pkg/osquery/tables/filevault/filevault.go
terracatta: Gonna keep this as is, not familiar enough with golang syntax to do this a different way.

directionless (approved):

PR #688, opened 2020-12-21
Set context timeouts on all exec functions

directionless:
Reviewing the places we call exec, and attempting to get timeouts on all of them

### DIFF
--- a/pkg/osquery/table/mdfind_darwin.go
+++ b/pkg/osquery/table/mdfind_darwin.go
@@ -12,7 +12,7 @@ import (
 )
 
 func mdfind(args ...string) ([]string, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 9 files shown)


pkg/osquery/table/mdfind_darwin.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 func mdfind(args ...string) ([]string, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
blaedj, line 15: we see a lot of timeouts for mdfind, will there significant perf impacts to devices by 10x-ing this timeout? Does mdfind use significant CPU? ( it seems like this is a potential CPU hotspot depending on the query)

directionless: No clue. I just upped the numbers.

blaedj: Fair enough. I don't think we have significant safety nets for other performance sensitive areas, so this is probably in line with current art.

blaedj (approved):

PR #689, opened 2020-12-21
Fix makefile extension target

directionless:


blaedj (approved):
this explains the odd error I was running into last week. I just jumped back a couple commits instead of investigating.

PR #690, opened 2020-12-22
Disable macOS version compatible strings

directionless:
For reasons unknown, apple changes the content of `SystemVersion.plist`
based on the environment variables set in the caller.

Empiracally, it looks like the 10.16 nomenclature is dropping the .x.
So, force versions to use the 11.x.x format.

See:
 * https://eclecticlight.co/2020/08/13/macos-version-numbering-isnt-so-simple/
 * https://github.com/osquery/osquery/pull/6824

### DIFF
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -182,6 +182,14 @@ func (opts *osqueryOptions) createOsquerydCommand(osquerydBinary string, paths *
 		fmt.Sprintf("--config_plugin=%s", opts.configPluginFlag),
 	)
 
+	// On darwin, run osquery using a magic macOS variable to ensure we
+	// get proper versions strings back. I'm not totally sure why apple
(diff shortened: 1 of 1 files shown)


terracatta (approved):

directionless:
Testing a bit with a golang test program, I this needs to be set _prior_ to program start. So this PR is more likely than anything inside osquery. See https://github.com/osquery/osquery/issues/6840

erikng:
I figured I would post this here, but an alternative method is to do something like this: https://github.com/uber/client-platform-engineering/pull/61

I know you've already solved this, but given that you utilize a launchdaemon, this might be a better solution long term.

directionless:
> I know you've already solved this, but given that you utilize a launchdaemon, this might be a better solution long term.

The launchd configs are a good thought, especially for things that should be site configurable.

For launcher, I usually prefer to bake things into the code base. This allows a more seamless upgrade, and settings like this I don't really want people to change

PR #693, opened 2020-12-23
Go Version changes

directionless:
Move the reporting of the go version in `launcher_info` to being based on runtime. As go is static, I'm not sure what this was meant to do.

Update the Makefile's build targets to use the CROSSGO, if set, to run the build. This is not strictly needed, but should make some of the build encoded things more correct.


blaedj (commented):
seems fine

blaedj (approved):

PR #694, opened 2020-12-23
Correct macOS check in kolide_macos_software_update

terracatta:
This PR corrects the version check in the CGO of the software update table. The previous version relied on the old version scheme for macOS (ex: 10.15, 10.16) and failed when os_version reported something like "11.1" for Big Sur.

We now use a more robust method which is to look at the first two digits of the build number.

### DIFF
--- a/pkg/osquery/table/sus_darwin.go
+++ b/pkg/osquery/table/sus_darwin.go
@@ -42,8 +42,8 @@ void softwareUpdate(
 		*doesAppStoreAutoUpdates = 1;
 	}
 
-	// before 10.13 the method was called doesOSXAutoUpdates, since 10.14 it's called doesMacOSAutoUpdate.
-	if (os_version >= 14) {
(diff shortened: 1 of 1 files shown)


pkg/osquery/table/sus_darwin.go
```diff
@@ -42,8 +42,8 @@ void softwareUpdate(
 		*doesAppStoreAutoUpdates = 1;
 	}
 
-	// before 10.13 the method was called doesOSXAutoUpdates, since 10.14 it's called doesMacOSAutoUpdate.
-	if (os_version >= 14) {
+	// before 10.13 (build ver 17 (build ver 18) it's called doesMacOSAutoUpdate.
```
directionless, line 45:
```suggestion
	// before 10.13 (build ver 17) it's called doesMacOSAutoUpdate.
```

directionless (approved):
I'll take it. But I'm vaguely offended that we're using all of sql, go, and C to ultimately determine a boolean. The logic would be cleaner in a single place

PR #695, opened 2020-12-24
Add a startup delay to the autoupdater,  to avoid interfering with critical work

directionless:
When launcher first enrolls with a server, it often begins fetching key data.  However, the autoupdater _also_ starts. And if there's a pending update, this has the effect of pre-empting that early data. 

This introduces a simple option to delay the autoupdater to allow key data to come in first. This option defaults to an hour.

### DIFF
--- a/cmd/launcher/updater.go
+++ b/cmd/launcher/updater.go
@@ -21,6 +21,7 @@ type updaterConfig struct {
 	RootDirectory      string // launcher's root dir. use for holding tuf staging and updates
 	AutoupdateInterval time.Duration
 	UpdateChannel      autoupdate.UpdateChannel
+	InitialDelay       time.Duration // start delay, to avoid whomping critical early data
 
 	NotaryURL    string
 	MirrorURL    string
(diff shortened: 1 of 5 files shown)


cmd/launcher/updater.go
```diff
@@ -64,6 +65,15 @@ func createUpdater(
 
 	return &actor.Actor{
 		Execute: func() error {
+			// When launcher first starts, we'd like the
+			// Saas side to start recieving data
```
blaedj, line 69:
```suggestion
			// server to start receiving data
```

blaedj (approved):
lgtm, though I think we should note that this delay will happen every time launcher starts (reboot etc), right?

directionless:
> lgtm, though I think we should note that this delay will happen every time launcher starts (reboot etc), right?

Yes, this delays applies on eery launcher start. Where do you think we should say that? I think it's implied by the option help text.

blaedj:
> .. Where do you think we should say that? .. 

I just did, above 😄 the option help text is fine as-is, I agree.

PR #698, opened 2021-01-06
Fix output bug in kolide_profiles

directionless:
Apple's `profiles` command does not correctly output stdout in one very narrow case. Work around this by writing output to a temp file. 

Additionally, restrict the commands allowed on profiles to read only commands. This is enforced by a new option to the `GetConstraints` helper.

### DIFF
--- a/pkg/osquery/tables/tablehelpers/getconstraints_test.go
+++ b/pkg/osquery/tables/tablehelpers/getconstraints_test.go
@@ -82,7 +82,7 @@ func TestGetConstraints(t *testing.T) {
 			opts:     []GetConstraintOpts{WithDefaults("a", "b")},
 		},
 
-		// default plus allowed
+		// default plus allowed characters
 		{
 
 			name:     "double",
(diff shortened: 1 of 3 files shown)


pkg/osquery/tables/tablehelpers/getconstraints_test.go
```diff
@@ -101,6 +101,14 @@ func TestGetConstraints(t *testing.T) {
 			expected: []string{"a", "b"},
 			opts:     []GetConstraintOpts{WithDefaults("a", "b"), WithAllowedCharacters("z")},
 		},
+
+		// allowed values
+		{
```
blaedj, line 106: is this supposed to be a test for the `WithAllowedValues` option?

blaedj (commented):
LGTM, tested out locally and can confirm that it produces output as expected for queries that do not return values on commits prior to this.

I did have one question on the tests

pkg/osquery/tables/tablehelpers/getconstraints_test.go
```diff
@@ -101,6 +101,14 @@ func TestGetConstraints(t *testing.T) {
...
+
+			name:     "double",
+			expected: []string{"a"},
+			opts:     []GetConstraintOpts{WithAllowedCharacters("a")},
```
directionless, line 110:
```suggestion
			opts:     []GetConstraintOpts{WithAllowedValues([]string{"a"})},
```

pkg/osquery/tables/tablehelpers/getconstraints_test.go
directionless: good catch. Let's see if the tests pass! Editing in PRs, always fun

blaedj (approved):

PR #699, opened 2021-01-08
Add windows wifi_networks table 

blaedj:
This commit: 
 - adds a new 'kolide_wifi_networks' virtual table for windows

Why?
 - This is somewhat equivalent to the wifi table available on darwin

**Implementation**
This table works by parsing the output of the included powershell script. 
This script imports some C# .Net code (als oincluded), and calling windows
native APIs for wlan.

```
osquery> select * from kolide_wifi_networks;

+-------------------------------+---------+-----------------------+-----------------------------+-------+---------------------+
| fullkey                       | parent  | key                   | value                       | query | ssid                |
+-------------------------------+---------+-----------------------+-----------------------------+-------+---------------------+
| DEFAULT/SSID                  | DEFAULT | SSID                  | ddu23n104                   |       | ddu23n104           |
| DEFAULT/BSSID                 | DEFAULT | BSSID                 | XX-XX-XX-XX-XX-XX           |       | ddu23n104           |
| DEFAULT/dot11BssPhyType       | DEFAULT | dot11BssPhyType       | 7                           |       | ddu23n104           |
| DEFAULT/timestamp             | DEFAULT | timestamp             | 610829842048                |       | ddu23n104           |
| DEFAULT/capabilityInformation | DEFAULT | capabilityInformation | 1073                        |       | ddu23n104           |
| DEFAULT/ieOffset              | DEFAULT | ieOffset              | 1080                        |       | ddu23n104           |
| DEFAULT/linkQuality           | DEFAULT | linkQuality           | 89                          |       | ddu23n104           |
| DEFAULT/wlanRateSet           | DEFAULT | wlanRateSet           | NativeWifi.Wlan+WlanRateSet |       | ddu23n104           |
| DEFAULT/ieSize                | DEFAULT | ieSize                | 182                         |       | ddu23n104           |
| DEFAULT/dot11Bssid            | DEFAULT | dot11Bssid            | {128, 42, ..., 235...}      |       | ddu23n104           |
| DEFAULT/dot11BssType          | DEFAULT | dot11BssType          | Infrastructure              |       | ddu23n104           |
| DEFAULT/beaconPeriod          | DEFAULT | beaconPeriod          | 0                           |       | ddu23n104           |
| DEFAULT/chCenterFrequency     | DEFAULT | chCenterFrequency     | 2412000                     |       | ddu23n104           |
| DEFAULT/hostTimestamp         | DEFAULT | hostTimestamp         | 132550441422028881          |       | ddu23n104           |
| DEFAULT/dot11Ssid             | DEFAULT | dot11Ssid             | NativeWifi.Wlan+Dot11Ssid   |       | ddu23n104           |
| DEFAULT/phyId                 | DEFAULT | phyId                 | 2                           |       | ddu23n104           |
| DEFAULT/rssi                  | DEFAULT | rssi                  | -45                         |       | ddu23n104           |
| DEFAULT/inRegDomain           | DEFAULT | inRegDomain           | true                        |       | ddu23n104           |
| DEFAULT/dot11BssType          | DEFAULT | dot11BssType          | Infrastructure              |       |                     |
| DEFAULT/dot11BssPhyType       | DEFAULT | dot11BssPhyType       | 7                           |       |                     |
| DEFAULT/wlanRateSet           | DEFAULT | wlanRateSet           | NativeWifi.Wlan+WlanRateSet |       |                     |
| DEFAULT/dot11Ssid             | DEFAULT | dot11Ssid             | NativeWifi.Wlan+Dot11Ssid   |       |                     |
| DEFAULT/phyId                 | DEFAULT | phyId                 | 2                           |       |                     |
| DEFAULT/dot11Bssid            | DEFAULT | dot11Bssid            | {130, ..., 168, 235...}     |       |                     |
| DEFAULT/inRegDomain           | DEFAULT | inRegDomain           | true                        |       |                     |
| DEFAULT/beaconPeriod          | DEFAULT | beaconPeriod          | 0                           |       |                     |
| DEFAULT/capabilityInformation | DEFAULT | capabilityInformation | 1073                        |       |                     |
| DEFAULT/ieOffset              | DEFAULT | ieOffset              | 904                         |       |                     |
| DEFAULT/linkQuality           | DEFAULT | linkQuality           | 88                          |       |                     |
| DEFAULT/chCenterFrequency     | DEFAULT | chCenterFrequency     | 2412000                     |       |                     |
| DEFAULT/ieSize                | DEFAULT | ieSize                | 184                         |       |                     |
| DEFAULT/SSID                  | DEFAULT | SSID                  |                             |       |                     |
| DEFAULT/BSSID                 | DEFAULT | BSSID                 | XX-XX-XX-XX-XX-XX           |       |                     |
| DEFAULT/rssi                  | DEFAULT | rssi                  | -46                         |       |                     |
| DEFAULT/timestamp             | DEFAULT | timestamp             | 610829926784                |       |                     |
| DEFAULT/hostTimestamp         | DEFAULT | hostTimestamp         | 132550441422028881          |       |                     |
+-------------------------------+---------+-----------------------+-----------------------------+-------+---------------------+
```


pkg/osquery/tables/wlan/wlan.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return nil
+}
+
+func buildParser(logger log.Logger) *OutputParser {
+	return NewParser(logger,
+		[]Matcher{
+			{
+				Match: func(in string) bool {
```
directionless, line 118: These will probably fail in other localizations.

pkg/osquery/tables/wlan/wlan_test.go
```diff
@@ -0,0 +1,96 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		logger := log.NewNopLogger()
+		table := WlanTable{
+			logger: logger,
+			parser: buildParser(logger),
+			getBytes: func(ctx context.Context, buf *bytes.Buffer) error {
```
directionless, line 74: I'd define this as a function like `func getTestFile(t *testing.T, buf *bytes.Buffer, filename string)` , and not an anonymous one. But that's mostly style. Or use `ioutil.ReadFile`

pkg/osquery/tables/wlan/wlan.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+const netshCmd = "netsh"
+
+var preludeInterfaceLineRegex = regexp.MustCompile(`Interface name :`)
+var preludeCountLineRegex = regexp.MustCompile(`There are [0-9]+ networks currently available`)
+
+type execer func(ctx context.Context, buf *bytes.Buffer) error
```
directionless, line 26: As a general note, the style of passing in a bytes.Buffer is considered higher performance (yay).

But, the underlying call is an os level exec, which is always going to pass a bunch of data around weirdly. I've generally biased to things that returned bytes. Lower performance, but easily swappable with something like `ioutil.ReadFile`

pkg/osquery/tables/wlan/wlan.go
```diff
@@ -0,0 +1,198 @@
...
+					if err != nil {
+						return "", errors.Wrap(err, "converting string to int")
+					}
+					// a signal strength value of 0 implies an RSSI of -100dbm a
+					// signal strength value of 100 implies an RSSI of -50dbm We
+					// can interpolate the values in between. This strategy was
+					// suggested by code comments from here:
+					// https://github.com/metageek-llc/ManagedWifi
+					// and, empirically, seems to work out.
+					rssi := (-100 + (intVal / 2))
```
directionless, line 163: Should this be converted to a float?

pkg/osquery/tables/wlan/wlan.go
```diff
@@ -0,0 +1,198 @@
...
+
+					return fmt.Sprintf("%d", rssi), nil
```
directionless, line 165: I'd bias to `strconv`  over `Sprintf`

directionless (commented):
I know you're still exploring, and this may all be rendered moot. But some comments...

I like the stanza addition to the parser, I think we should make that more real.

pkg/osquery/tables/wlan/wlan.go
```diff
@@ -0,0 +1,198 @@
...
+		table.TextColumn("radio_type"),
+		table.TextColumn("rssi"),
+		table.TextColumn("channel"),
+	}
+
+	parser := buildParser(logger)
+	t := &WlanTable{
+		client:    client,
+		logger:    logger,
+		tableName: "kolide_wifi_networks",
```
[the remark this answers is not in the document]
blaedj: Yea you're right, I renamed the table at the last minute before opening the PR :) I'll make the names match one way or the other.

pkg/osquery/tables/wlan/testdata/multiple_results.txt
```diff
@@ -0,0 +1,46 @@
+Interface name : Wi-Fi
+There are 4 networks currently visible.
+
+SSID 1 : ddu23n104
+    Network type            : Infrastructure
+    Authentication          : WPA2-Personal
+    Encryption              : CCMP
+    BSSID 1                 : 80:2a:9c:eb:bb:65
```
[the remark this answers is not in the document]
blaedj: They aren't real values, I debated making them obviously fake

pkg/osquery/tables/wlan/getbssidps.go
```diff
@@ -0,0 +1,2428 @@
+package wlan
+
+// TODO: move this into a ps1 file for better reading/editing, and load from there.
+const getBSSIDCommandTemplate = `
```
directionless, line 4: Please move this into something with `go-bindata` there are several examples in the packaging code

pkg/osquery/tables/wlan/getbssidps.go
```diff
@@ -0,0 +1,2428 @@
...
+    $scanInProgress = "false" 
+    do {
+      $scanInProgress = "false" 
+      $WlanClient.Interfaces | ForEach-Object { 
+        $ip = $_.scanInProgress
+        if ($ip -eq "True") {
+          $scanInProgress = "true" 
+        }
+      }; 
+      Start-Sleep -Milliseconds 50
```
directionless, line 30: 50ms seems pretty tight in a loop.

pkg/osquery/tables/wlan/psrunner.go
```diff
@@ -0,0 +1,104 @@
...
+		return errors.Wrap(err, "writing native code file")
+	}
+
+	tmpl, err := template.New("command").Parse(getBSSIDCommandTemplate)
+	if err != nil {
+		return errors.Wrap(err, "parsing template")
+	}
+	commandOpts := struct {
+		NativeCodePath string
+	}{NativeCodePath: nativeCodeFile.Name()}
```
directionless, line 55: I wonder if you really need this. It's putting the full path into the powershell `Add-Type`. Any chance it works with a relative path if you set CWD? That would let you skip the whole template parsing

pkg/osquery/tables/wlan/psrunner.go
```diff
@@ -0,0 +1,104 @@
...
+// New create new session
+func New() *PowerShell {
+	ps, _ := exec.LookPath("powershell.exe")
+	return &PowerShell{
+		powerShell: ps,
+	}
+}
+
+func runPos(ctx context.Context, output *bytes.Buffer) error {
+	posh := New()
```
directionless, line 29: The canonical abbreviation for powershell is pwsh. (very small nit)

pkg/osquery/tables/wlan/psrunner.go
```diff
@@ -0,0 +1,104 @@
...
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if err != nil {
+		fmt.Printf("%s", stderr.String())
+	}
+	return err
+}
+
+const simplecommand = `
```
directionless, line 85: This looks like it should be in a `_test.go` ?

blaedj: Oh not a bad idea. I would've just deleted it tbh, just hadn't cleaned this up yet.

pkg/osquery/tables/wlan/psrunner.go
blaedj: I can give it a shot!

pkg/osquery/tables/wlan/getbssidps.go
blaedj: Ok, I was unsure on this. Per MS documentation, the wifi interface should complete the scan in 4 seconds or less. When I just had a 2-second sleep in here (instead of polling for completeness) it seemed to work pretty well, but I'm running windows on apple hardware that's probably not indicative of most windows machines out there.

pkg/osquery/tables/wlan/double_line_scanner.go
```diff
@@ -0,0 +1,28 @@
...
+import "regexp"
+
+var (
+	twoEOLs = regexp.MustCompile(`(\r?\n){2}`)
+)
+
+// blankLineSplitter implements the bufio.SplitFunc type, when used as the Split
+// function for a bufio.Scanner, it will return chunks of bytes separated by an
+// empty newline, e.g. \n\n or \r\n\r\n.
+func blankLineSplitter(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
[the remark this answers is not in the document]
blaedj: Yea, I copy-pasted the signature from the std lib, I'll clean this up.

pkg/osquery/tables/wlan/wlan.go
blaedj: Yea this is a really good call-out. I _think_ that with the new approach (not depending on the output of `netsh wlan show networks`) this isn't an issue. The column names are defined in code.

Makefile
```diff
@@ -198,7 +198,7 @@ deps: deps-go generate
 
 .PHONY: generate
 generate: deps-go
-	go generate ./pkg/packagekit/... ./pkg/packaging/...
+	go generate ./pkg/packagekit/... ./pkg/packaging/... ./pkg/osquery/tables/wifi_networks/...
```
directionless, line 201:
```suggestion
	go generate ./pkg/packagekit/... ./pkg/packaging/... ./pkg/osquery/tables/...
```

I'd just run this across everything in there.

pkg/osquery/tables/tablehelpers/stanza_scanner.go
```diff
@@ -0,0 +1,28 @@
...
+import "regexp"
+
+var (
+	twoEOLs = regexp.MustCompile(`(\r?\n){2}`)
+)
+
+// stanzaSplitter implements the bufio.SplitFunc type, when used as the Split
+// function for a bufio.Scanner, it will return chunks of bytes separated by an
+// empty newline, e.g. \n\n or \r\n\r\n.
+func StanzaSplitter(data []byte, atEOF bool) (int, []byte, error) {
```
directionless, line 12: tests would be nice, but not required.

pkg/osquery/tables/tablehelpers/parser.go
```diff
@@ -0,0 +1,126 @@
...
+		matchers: matchers,
+		logger:   logger,
+	}
+	return p
+}
+
+// Parse looks at a chunk of input. It is assumed that the input contains
+// information to fill in a single result. Do not provide input that contains
+// data for multiple results.
+func (p *OutputParser) Parse(input *bytes.Buffer) map[string]string {
```
directionless, line 33: Yay helpers!

How hard is it to move `pkg/osquery/tables/firmwarepasswd` to using this? If it's more than a couple minutes I'm not sure we should block. Make an issue?

pkg/osquery/tables/tablehelpers/stanza_scanner.go
```diff
@@ -0,0 +1,28 @@
+package tablehelpers
+
+import "regexp"
+
+var (
+	twoEOLs = regexp.MustCompile(`(\r?\n){2}`)
+)
+
+// stanzaSplitter implements the bufio.SplitFunc type, when used as the Split
```
directionless, line 9:
```suggestion
// StanzaSplitter implements the bufio.SplitFunc type, when used as the Split
```

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
```diff
@@ -0,0 +1,34 @@
...
+        $Encoding  = New-Object System.Text.ASCIIEncoding
+        $Encoding.GetString($ByteArray)
+    }
+
+    Add-Type -Path ".\nativewificode.cs"
+    $WlanClient = New-Object NativeWifi.WlanClient
+
+    $WlanClient.Interfaces | ForEach-Object { $_.Scan() }
+
+    #check scan progress for each interface every 50 milliseconds
```
directionless, line 16: Comment is 
```suggestion
    # check scan progress for each interface
```

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
```diff
@@ -0,0 +1,34 @@
...
+    do {
+      $scanInProgress = "false"
+      $WlanClient.Interfaces | ForEach-Object {
+        $ip = $_.scanInProgress
+        if ($ip -eq "True") {
+          $scanInProgress = "true"
+        }
+      };
+      Start-Sleep -Milliseconds 100
+    } while ($scanInProgress -eq "true")
```
directionless, line 27: I can't really parse this. Will this ever timeout? I guess the go code will handle that

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
```diff
@@ -0,0 +1,34 @@
...
+    $scanInProgress = "false"
+    do {
+      $scanInProgress = "false"
+      $WlanClient.Interfaces | ForEach-Object {
+        $ip = $_.scanInProgress
+        if ($ip -eq "True") {
+          $scanInProgress = "true"
+        }
+      };
+      Start-Sleep -Milliseconds 100
```
directionless, line 26: Can't decide if 100 is still tight. 250 maybe? Any discussion online?

pkg/osquery/tables/wifi_networks/testdata/output.txt
```diff
@@ -0,0 +1,18 @@
...
+phyId                 : 2
+dot11Bssid            : {128, 42, ..., 235...}
+dot11BssType          : Infrastructure
+dot11BssPhyType       : 7
+rssi                  : -45
+linkQuality           : 100
+inRegDomain           : True
+beaconPeriod          : 0
+timestamp             : 610829842048
+hostTimestamp         : 132550441422028881
```
directionless, line 11: These timestamps don't look like unix time. Should we do anything to convert them, or simply present data as-is?

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+//go:generate go-bindata -nometadata -nocompress -pkg internal -o internal/assets.go internal/assets/
+
+type execer func(ctx context.Context, buf *bytes.Buffer) error
+
+type WlanTable struct {
```
directionless, line 29: When these are in their own package, I tend to just call them `Table` (nit)

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+	client    *osquery.ExtensionManagerClient
+	logger    log.Logger
+	tableName string
```
directionless, line 32: Only needed if you're meta-programming multiple tables. No point to having the variable here.

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+
+	err := t.getBytes(ctx, &output)
+	if err != nil {
+		return results, err
+	}
+	scanner := bufio.NewScanner(&output)
+	scanner.Split(tablehelpers.StanzaSplitter)
+	for scanner.Scan() {
+		chunk := scanner.Bytes()
+		rows, err := dataflatten.Ini(chunk, dataflatten.WithLogger(t.logger))
```
directionless, line 65: It's in ini format!? Amazing!

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to get asset named %s", "internal/assets/nativewifi.cs")
+		}
+		_, err = nativeCodeFile.Write(nativeCode)
+		if err != nil {
+			return errors.Wrap(err, "writing native code file")
+		}
+		if err := nativeCodeFile.Close(); err != nil {
+			return errors.Wrap(err, "closing native code file")
+		}
```
directionless, line 121: You can clean this up with go-bindata's `RestoreAsset` or `RestoreAssets` (the former is for a single file, the latter for a directory to recurse)

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+
+		pwsh, err := exec.LookPath("powershell.exe")
+		if err != nil {
+			return errors.Wrap(err, "finding powershell.exe path")
+		}
+		psScript, err := internal.Asset("internal/assets/get-networks.ps1")
+		if err != nil {
+			return errors.Wrapf(err, "failed to get asset named %s", "internal/assets/get-networks.ps1")
+		}
+		args := append([]string{"-NoProfile", "-NonInteractive"}, string(psScript))
```
directionless, line 131: Do you think it makes sense to pass this as an ARG, vs writing to disk and exec'ing? (assuming that's possible)

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+		cmd.Stderr = &stderr
+
+		err = cmd.Run()
+		errOutput := stderr.String()
+		// sometimes the powershell script logs errors to stderr, but returns a
+		// successful execution code. It's helpful to log the output of stderr
+		// in this case.
+		if err != nil || errOutput != "" {
+			level.Debug(logger).Log("stderr", errOutput)
+		}
```
directionless, line 145: Should this have a return error here?

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
...
+		ssid := ""
+		for _, r := range rows {
+			if strings.HasSuffix(r.StringPath("/"), "/SSID") {
+				ssid = r.Value
+				break
+			}
+		}
+		rowData := map[string]string{
+			"ssid": ssid,
+		}
```
directionless, line 78: What's the `ssid` pull out for? It's in the EAV data, extracting it here feels superfluous?

pkg/osquery/tables/wifi_networks/wifi_networks_test.go
```diff
@@ -0,0 +1,76 @@
...
+				{"fullkey": "DEFAULT/phyId", "parent": "DEFAULT", "key": "phyId", "value": "2", "query": "", "ssid": "ddu23n104"},
+				{"fullkey": "DEFAULT/rssi", "parent": "DEFAULT", "key": "rssi", "value": "-45", "query": "", "ssid": "ddu23n104"},
+				{"fullkey": "DEFAULT/inRegDomain", "parent": "DEFAULT", "key": "inRegDomain", "value": "true", "query": "", "ssid": "ddu23n104"},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		logger := log.NewNopLogger()
+		table := WlanTable{
```
directionless, line 53: You've got setup stuff that can fail here -- It should be in the `t.Run`

directionless (commented):
Comments, I still need to read the big cs script

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,149 @@
+package wifi_networks
```
directionless, line 1: This should be build tagged to windows (ditto for _test.go)

pkg/osquery/tables/wifi_networks/wifi_networks.go
directionless: Hrm. This data looks suspect. What delineates the networks?

pkg/osquery/tables/wifi_networks/wifi_networks_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestTableGenerate(t *testing.T) {
+	t.Parallel()
+
+	var tests = []struct {
+		filename string
+		expected []map[string]string
+	}{
```
directionless, line 21: You should add a test case for more than one wifi network

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
```diff
@@ -0,0 +1,34 @@
...
+    function Convert-ByteArrayToString {
+        [CmdletBinding()] Param (
+            [Parameter(Mandatory = $True, ValueFromPipeline = $True)] [System.Byte[]] $ByteArray
+            )
+
+        $Encoding  = New-Object System.Text.ASCIIEncoding
+        $Encoding.GetString($ByteArray)
+    }
+
+    Add-Type -Path ".\nativewificode.cs"
```
directionless, line 11: Interestingly, this works as written through this table. But if I try to run this from the command line it needs no quotes

pkg/osquery/tables/wifi_networks/wifi_networks.go
directionless: Ah. I'm starting to see what's happening -- these are objects, which powershell spits out in something close to ini, but each object is empty line delineated. And that means ini parses them as a big honking object which keeps overwriting itself. 

I don't know if there's a simple ini or powershell patch, but it would probably be simple to add section headers in go. Using an incrementing integer would work.

directionless: Found it! `ConvertTo-Json`. See the powershell script

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
```diff
@@ -0,0 +1,34 @@
...
+          $scanInProgress = "true"
+        }
+      };
+      Start-Sleep -Milliseconds 100
+    } while ($scanInProgress -eq "true")
+
+    $WlanClient.Interfaces |
+    ForEach-Object { $_.GetNetworkBssList() } |
+    Select-Object *,@{Name="SSID";Expression={(Convert-ByteArrayToString -ByteArray $_.dot11ssid.SSID)}},
+                    @{Name="BSSID";Expression={[System.BitConverter]::ToString($_.dot11Bssid) }}
```
directionless, line 32: This should get piped to `ConvertTo-Json` at the end, and then we can ingest that. 

I'm also quite unsure on those converters. Looking at the json produced, they're kinda weird. OTOH, they might work well enough?

directionless:
I skimmed the c#. I can't really speak to the quality or correctness, but I don't see anything obviously malicious. I do see some unused functions, but it doesn't seem worth trying to remove them. Might be worth noting where we got it from. (comment in the go table file?)

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
blaedj: Yea, the go code handles it. It doesn't time out, but you're probably right that it should.

pkg/osquery/tables/tablehelpers/parser.go
blaedj: I have a git stash locally that starts to move it, but it requires a slight tweak or two. I'll file an issue for it.

pkg/osquery/tables/wifi_networks/testdata/output.txt
blaedj: tbh I don't know anything about them. I think they're documented by MS, but they weren't part of the original dataset I was collecting, so I haven't looked into it. I can take a peek and at least document it here.

pkg/osquery/tables/wifi_networks/wifi_networks.go
blaedj: Ha I know, I'm not sure if it's 'strict ini' (probably no such thing), but it seems to be all simple key-value pairs, and the output isn't nested. 

The output is just the default that powershell's `Select-Object` prints out, and the objects aren't nested, so it seems to work.

pkg/osquery/tables/wifi_networks/wifi_networks.go
blaedj: I think that execing from disk is a little less straightforward, see https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-7.1. The default policy on windows is to deny executing scripts. 

We might be able to get around that if we run as 'root', but I didn't look into it. I think we could probably change the default execution policy to allow execing, but changing something like that about the system seems undesirable.

pkg/osquery/tables/wifi_networks/wifi_networks.go
blaedj: oh good catch. They were at some point in the past, but I removed it because some tools I use respect them when I don't want them to during development on my mac.

pkg/osquery/tables/wifi_networks/wifi_networks.go
blaedj: > And that means ini parses them as a big honking object which keeps overwriting itself.

I don't think that's the case, because this code passes each empty-line-delineated chunk to the ini parser separately.

pkg/osquery/tables/wifi_networks/wifi_networks.go
blaedj: Oh yea, probably.

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
directionless: Go seems good enough. I don't want to get too fancy in our non-core languages

pkg/osquery/tables/wifi_networks/testdata/output.txt
directionless: We can kick that forward. I think osquery knows how to parse them, so we might have some example c code

pkg/osquery/tables/wifi_networks/wifi_networks_test.go
directionless: (or not, if we move to json)

blaedj:
Yea I should comment where it came from, I _thought_ I had at one point, but it must've been dropped. In a former life, I was fluent in C#, but there's a lot in there. I've read through the code ( I had to figure out how to call the various functions) and at one point I was trying to remove the unneeded functions, but abandoned that effort.

I've changed a bit of the code around, so I have a cursory understanding of how it works, and it all seems ok. I didn't find anything malicious either.

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
blaedj: are you just execing the command from a powershell console? I've been testing the script by running `powershell.exe .\get-networks.ps1` from a **cmd.exe** session, and it works with the quoting.

directionless: Yes. From inside a powershell script. But... my use case is boring, so it's hard to care.

pkg/osquery/tables/wifi_networks/wifi_networks.go
directionless: Sounds like it's not worth digging at. Some indication signing the script would help. But that sounds like future work

pkg/osquery/tables/wifi_networks/internal/assets/get-networks.ps1
blaedj: I can do 250 - MS says that the max time you should wait is 4 seconds, all drivers/interfaces/wifi modules/thingies should respond to a request for networks scans within 4 second as part of the qualification process for a 'made for windows' sticker. (I'm not 💯 on what the sticker thing actually means, but 4s seems like a reasonable bar). 

Empirically, 2s usually seemed like enough time when I was testing. but I don't have a good sense for how that translates to choosing a sleep time, other than I'd like to finish the waiting period as soon as possible.

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	var results []map[string]string
+	var output bytes.Buffer
+
+	err := t.getBytes(ctx, &output)
+	if err != nil {
```
directionless, line 52:
```suggestion
	if err := t.getBytes(ctx, &output); err != nil {
```

pkg/osquery/tables/wifi_networks/wifi_networks.go
```diff
@@ -0,0 +1,125 @@
...
+		return results, errors.Wrap(err, "getting raw data")
+	}
+	rows, err := dataflatten.Json(output.Bytes(), dataflatten.WithLogger(t.logger))
+	if err != nil {
+		return results, errors.Wrap(err, "flattening json output")
+	}
+
+	return append(results, dataflattentable.ToMap(rows, "", map[string]string{})...), nil
```
directionless, line 60: I don't know that I've tested it, but I wonder if ToMap would take a nil instead of `map[string]string{})`. Pretty small nit though

blaedj: yep it does take a nil!

pkg/osquery/tables/wifi_networks/wifi_networks.go
directionless: We can do this later. It's just code cleanup

directionless (approved):

PR #702, opened 2021-01-27
Implement wifi-ish table for linux

blaedj:
This commit: 
 - adds a simple 'string-delimited' strategy for dataflatten, including 
   record-splitting based on the idea that if we see a key 'again', we 
  should start a new record, instead of over-writing the key.
 - adds a new `KeyValue` data source type to dataflattentable
 - adds options to the TablePluginExec function 
 - implements a wifi table via a `dataflattentable.TablePluginExec` call

Note that this table will only work if nmcli is installed, so it's not a true
solution for all linux installations.
```
osquery> select * from kolide_nmcli where key not in ('BSSID', 'SSID', 'SSID-HEX');
+-------------+--------+-----------+-------------------------------------------------+-------+
| fullkey     | parent | key       | value                                           | query |
+-------------+--------+-----------+-------------------------------------------------+-------+
| 7/CHAN      | 7      | CHAN      | 11                                              |       |
| 7/RATE      | 7      | RATE      | 130 Mbit/s                                      |       |
| 7/BARS      | 7      | BARS      | ▂▄▆█                                            |       |
| 7/WPA-FLAGS | 7      | WPA-FLAGS | (none)                                          |       |
| 7/NAME      | 7      | NAME      | AP[8]                                           |       |
| 7/MODE      | 7      | MODE      | Infra                                           |       |
| 7/RSN-FLAGS | 7      | RSN-FLAGS | pair_ccmp group_ccmp psk                        |       |
| 7/SIGNAL    | 7      | SIGNAL    | 90                                              |       |
| 7/DEVICE    | 7      | DEVICE    | wlp2s0                                          |       |
| 7/FREQ      | 7      | FREQ      | 2462 MHz                                        |       |
| 7/SECURITY  | 7      | SECURITY  | WPA2                                            |       |
| 7/ACTIVE    | 7      | ACTIVE    | no                                              |       |
| 7/IN-USE    | 7      | IN-USE    |                                                 |       |
| 7/DBUS-PATH | 7      | DBUS-PATH | /org/freedesktop/NetworkManager/AccessPoint/127 |       |
| 0/SIGNAL    | 0      | SIGNAL    | 100                                             |       |
| 0/SECURITY  | 0      | SECURITY  | WPA2                                            |       |
| 0/DBUS-PATH | 0      | DBUS-PATH | /org/freedesktop/NetworkManager/AccessPoint/123 |       |
| 0/MODE      | 0      | MODE      | Infra                                           |       |
| 0/CHAN      | 0      | CHAN      | 6                                               |       |
| 0/BARS      | 0      | BARS      | ▂▄▆█                                            |       |
| 0/WPA-FLAGS | 0      | WPA-FLAGS | (none)                                          |       |
| 0/DEVICE    | 0      | DEVICE    | wlp2s0                                          |       |
| 0/FREQ      | 0      | FREQ      | 2437 MHz                                        |       |
| 0/RATE      | 0      | RATE      | 195 Mbit/s                                      |       |
| 0/RSN-FLAGS | 0      | RSN-FLAGS | pair_ccmp group_ccmp psk                        |       |
| 0/ACTIVE    | 0      | ACTIVE    | no                                              |       |
| 0/IN-USE    | 0      | IN-USE    |                                                 |       |
| 0/NAME      | 0      | NAME      | AP[1]                                           |       |

```


pkg/osquery/tables/dataflattentable/exec.go
```diff
@@ -15,21 +15,36 @@ import (
 	"github.com/pkg/errors"
 )
 
-func TablePluginExec(client *osquery.ExtensionManagerClient, logger log.Logger, tableName string, dataSourceType DataSourceType, execArgs []string) *table.Plugin {
+type ExecTableOpt func(*Table)
+
+func WithSeperator(separator string) ExecTableOpt {
```
directionless, line 20:
```suggestion
// WithSeparator sets the field separator. If not specified, it defaults to `:`
func WithSeparator(separator string) ExecTableOpt {
```

Mostly wanted to note that it's the _field_ separator, note that it has a default, and fix your spelling error

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+		scanner := bufio.NewScanner(bytes.NewReader(rawdata))
+		row := map[string]interface{}{}
+		i := 0
+		for scanner.Scan() {
+			line := scanner.Text()
+			parts := strings.SplitN(line, delimiter, 2)
+			if len(parts) < 2 {
+				continue
+			}
+			key := parts[0]
```
directionless, line 33: trim space here

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+// properties are grouped into a single record based on 'duplicate key'
+// strategy: If a key/value pair is encountered, and the record being built
+// already has a value for that key, then that record is considered 'complete'.
+// The record is stored in the collection, and a new record is started. This
+// strategy is only suitable if the data output does not exclude k/v pairs with
+// blank/missing values, and assumes that the properties for a single record are
+// grouped together.
+func StringDelimitedUnseparatedFunc(delimiter string) dataFunc {
+	return func(rawdata []byte, opts ...FlattenOpts) ([]Row, error) {
+		v := map[string]interface{}{}
```
directionless, line 23: Why is this a hash? This should be array... I'd call it `results` too
```suggestion
		results :=[]interface{}{}
```

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+			key := parts[0]
+			value := strings.TrimSpace(parts[1])
+			if _, ok := row[key]; ok { // this key already exists, so we want to start a new record.
+				v[strconv.Itoa(i)] = row // store the 'finished' record in the collection
+				i++
+				row = map[string]interface{}{} // reset the record
+			}
+			row[key] = value
+		}
+		v[strconv.Itoa(i)] = row // store the final record
```
directionless, line 42:
```suggestion
		results = append(results, row)
```

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+		for scanner.Scan() {
+			line := scanner.Text()
+			parts := strings.SplitN(line, delimiter, 2)
+			if len(parts) < 2 {
+				continue
+			}
+			key := parts[0]
+			value := strings.TrimSpace(parts[1])
+			if _, ok := row[key]; ok { // this key already exists, so we want to start a new record.
+				v[strconv.Itoa(i)] = row // store the 'finished' record in the collection
```
directionless, line 36:
```suggestion
				results = append(results, row)
```

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// StringDelimited assumes that rawdata only holds key-value pairs for a single
+// record. Additionally, each k/v pair must be on its own line. Useful for
+// output that can be easily separated into separate records before 'flattening'
+func StringDelimited(rawdata []byte, delimiter string, opts ...FlattenOpts) ([]Row, error) {
+	return flattenStringDelimited(rawdata, delimiter, opts...)
+}
+
+func flattenStringDelimited(in []byte, delimiter string, opts ...FlattenOpts) ([]Row, error) {
```
directionless, line 55: Why two functions?  And why these at all, is anything using them?

I would, instead, rename `StringDelimitedUnseparatedFunc` to `StringDelimitedFunc` and have it take an argument for the record separator strategy. I think that generator is a good approach

pkg/osquery/table/platform_tables_linux.go
```diff
@@ -13,5 +14,7 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 	return []*table.Plugin{
 		gsettings.Settings(client, logger),
 		gsettings.Metadata(client, logger),
+		dataflattentable.TablePluginExec(client, logger,
+			"kolide_nmcli", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"}, dataflattentable.WithSeperator(":")),
```
directionless, line 18: Should this be `kolide_nmcli` or `kolide_nmcli_device` ?

directionless (changes requested):
I like this. Some small scale things

pkg/osquery/table/platform_tables_linux.go
blaedj: I think either `kolide_nmcli` or `kolide_nmcli_device_wifi` (maybe even `kolide_nmcli_device_wifi_list`), because `nmcli device` has quite a few other subcommands. 

Actually, how about `kolide_nmcli_wifi`? it leaves us open to add other wifi listing tables for linux **and** other `nmcli` tables, should we need them.

pkg/dataflatten/string_delimited.go
blaedj: correct, these two functions don't have any callers - they were originally mimicking the Ini and Json funcs, with the intent that one could pass in a filename or the raw data itself. I can clean this up.

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,69 @@
...
+
+type dataFunc func(data []byte, opts ...FlattenOpts) ([]Row, error)
+
+// StringDelimitedUnseparatedFunc returns a function that conforms to the
+// interface expected by dataflattentable.Table's execDataFunc property.
+// properties are grouped into a single record based on 'duplicate key'
+// strategy: If a key/value pair is encountered, and the record being built
+// already has a value for that key, then that record is considered 'complete'.
+// The record is stored in the collection, and a new record is started. This
+// strategy is only suitable if the data output does not exclude k/v pairs with
```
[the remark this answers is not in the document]
blaedj: That's kind of apples-to-oranges, the older bssid detection was really just a way to label grouped properties in the generated SQL results, and it depended on _already knowing_ how properties were grouped, whereas this is a method for determining _how_ to group properties. 

In any case, yes, there are definitely nuances, but the main point is that we need consider the implicit assumptions when using this in other contexts.

pkg/osquery/tables/dataflattentable/tables.go
```diff
@@ -33,6 +34,8 @@ type Table struct {
 
 	execDataFunc func([]byte, ...dataflatten.FlattenOpts) ([]dataflatten.Row, error)
 	execArgs     []string
+
+	KeyValueSeparator string
```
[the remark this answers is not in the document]
blaedj: nope I don't think so!

pkg/osquery/table/platform_tables_linux.go
directionless: I don't know the command well enough to feel strongly. Mostly I observed we should think about it. Renaming later is always an option

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,84 @@
...
+	"bytes"
+	"strings"
+)
+
+type dataFunc func(data []byte, opts ...FlattenOpts) ([]Row, error)
+
+type recordSplittingStrategy int
+
+const (
+	None recordSplittingStrategy = iota + 1
```
directionless, line 14: Why +1?

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			row[key] = value
+		}
+		results = append(results, row) // store the final record
+
+		return Flatten(results, opts...)
+	}
+}
+
+// singleRecordFunc returns an execData function that assumes 'rawdata'
```
directionless, line 63: I don't love the duplication between these, but I don't see a better way. Let's keep it, and see if we evolve?

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// duplicateKeyFunc returns a function that conforms to the interface expected
+// by dataflattentable.Table's execDataFunc property. properties are grouped
+// into a single record based on 'duplicate key' strategy: If a key/value pair
+// is encountered, and the record being built already has a value for that key,
+// then that record is considered 'complete'. The record is stored in the
+// collection, and a new record is started. This strategy is only suitable if
+// the data output does not exclude k/v pairs with blank/missing values, and
+// assumes that the properties for a single record are grouped together.
```
directionless, line 37: I still think we should be more clear in this warning. 

Perhaps 
```suggestion
// collection, and a new record is started. This strategy is only suitable if
// properties for a single record are grouped together, and output includes,
// a duplicate field, indicating a new record, before any sparse values.
```

pkg/dataflatten/string_delimited.go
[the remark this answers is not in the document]
directionless: I don't think it's an apples-to-oranges comparison. It had the same assumptions about how properties were grouped that this does. 

Though I guess, by being about `bssid` it was inherently not making any general claims

pkg/dataflatten/string_delimited.go
directionless: Generally the advice is to have the uninitialize value (0) for an int, but the unknown/none case. So in this case, using iota makes a lot of sense. I think the other place in dataflatten that uses `iota + 1` there's no none case, so the 0 value is inherently an error. That might be confusing though

pkg/dataflatten/string_delimited.go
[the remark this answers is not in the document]
blaedj: The bssid was from the time when we were consuming output that included a delimiter between records. The assumptions were different, because we assumed that there was a record separator in the raw output. The raw data was being parsed ('dataflattened') in chunks, where each chunk was known to only contain properties for single wifi network.

pkg/dataflatten/string_delimited.go
blaedj: Ok perfect! I was copying from elsewhere in dataflatten, but neglected to do the research to understand why the `+1` was used there. Thanks!

pkg/dataflatten/string_delimited.go
```diff
@@ -0,0 +1,84 @@
...
+func StringDelimitedFunc(kVDelimiter string, splittingStrategy recordSplittingStrategy) dataFunc {
+	switch splittingStrategy {
+	case None:
+		return singleRecordFunc(kVDelimiter)
+	case DuplicateKeys:
+		return duplicateKeyFunc(kVDelimiter)
+	default:
+		panic("Unknown record splitting strategy")
+	}
+
```
directionless, line 27:
```suggestion
```

directionless (approved):

PR #704, opened 2021-01-29
Table for tmutil destinationinfo

directionless:
Very simple `tmutil destinationinfo` table. Does not present a generalized
`tmutil` interface, as the subcommands, as they do not take `-X`


blaedj (approved):
lgtm pending CI re-run

PR #706, opened 2021-02-04
Add kolide_windows_updates table

blaedj:
This commit:
 - adds a first pass at a windows_updates table, implemented by
 interacting with the COM api for windows updates through powershell.

### DIFF
--- a/pkg/osquery/table/platform_tables_windows.go
+++ b/pkg/osquery/table/platform_tables_windows.go
@@ -3,6 +3,7 @@
 package table
 
 import (
+	"github.com/kolide/launcher/pkg/osquery/tables/dataflattentable"
 	"github.com/kolide/launcher/pkg/osquery/tables/secedit"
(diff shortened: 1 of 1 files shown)


directionless (approved):
Assuming this works, and that launcher can find `powershell.exe` I'm good with this. 

If we do this a lot, I'd like to make a cleaner solution than passing the powershell script as a string. But this seems like a fine beginning

blaedj:
> If we do this a lot, I'd like to make a cleaner solution than passing the powershell script as a string.

Yea, there are other niceties we could add as well. A few that come to mind: 
 - running powershell script from a file (and stored as an asset )
 - writing the output to a file & reading from there (we've needed this in other tables, might be useful in powershell someday)
 - catching the case where powershell returns a `0` exit code but prints an error message to stderr (I've seen this previously, but I'm not sure of all the cases where this can happen)

Issue #685, opened 2020-12-17
New Table Request: fc-list exec


FritzX6:
### What is the purpose of this table?
`fc-list` is a binary which lists all installed fonts on a device. This would allow a user to query whether a given font is installed in Font Book.

### What format is the data?
Comma Separated and New Lines eg.
```
> fc-list
/Users/fritz-imac/Library/Fonts/Knockout-HTF70-FullWelterwt.otf: Knockout,Knockout HTF70\-FullWelterwt:style=HTF70-FullWelterwt,Regular
/Users/fritz-imac/Library/Fonts/Mark Simonson - Proxima Soft ExCn Black.ttf: Proxima Soft Extra Condensed,Proxima Soft ExCn Black:style=Black,Regular
/Users/fritz-imac/Library/Fonts/NeutraText-LightSCItalic.otf: Neutra Text,Neutra Text Light SC:style=Light SC Italic,Italic
/Users/fritz-imac/Library/Fonts/Mark Simonson - Proxima Soft ExCn Extrabold It.ttf: Proxima Soft Extra Condensed,Proxima Soft ExCn Extrabold:style=Extrabold Italic,Italic
/Users/fritz-imac/Library/Fonts/MuseoSans-300Italic.otf: Museo Sans,Museo Sans 300:style=300 Italic,Italic
```

### What options does this exec have that we could pass?
The `-f` format flag can be passed to limit output:
```
> fc-list -f '%{file}\n'
/Users/fritz-imac/Library/Fonts/ITCAvantGardeStd-XLtCnObl.otf
/Users/fritz-imac/Library/Fonts/GoogleSans-Bold.ttf
/Users/fritz-imac/Library/Fonts/NeutraText-BookItalicFrac.otf
/Users/fritz-imac/Library/Fonts/AvenirNextLTPro-UltLtCn.otf
/Users/fritz-imac/Library/Fonts/CrimsonText-Semibold.ttf
```

### Why don't we get this a different way like System Profiler?
Unfortunately the output from `SPFontsDataType` has proven incomplete in terms of what it is able to collect.

directionless:
What information do we want out of this? What would we want in that format?

directionless:
I haven't been able to find a complete list of keys, but at a first pass we can use 
```
fc-list --format 'file: %{file}\nfamily: %{family}\nstyle:%{style}\n'
```

with the nmcli parsers.

fc-list is a cross platform tool as well

directionless:
I think we decided we didn't want this.

PR #707, opened 2021-02-04
Simple fc-list table

directionless:
Only installs if there binary is present. This may not be correct. Possible better would be to return no rows if not present.

`fc-list` is low value. I think without a better answer for the missing binary case, I don't want to merge this. But we might pick this up later.

Fixes: #685

### DIFF
--- a/pkg/osquery/table/table.go
+++ b/pkg/osquery/table/table.go
@@ -45,6 +45,15 @@ func PlatformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 			"kolide_zerotier_peers", dataflattentable.JsonType, zerotierCli("listpeers")),
 	}
 
+	if fcargs := fcListCli(); fcargs != nil {
+		dataflattentable.TablePluginExec(client, logger,
+			"kolide_fclist",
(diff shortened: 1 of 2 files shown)


pkg/osquery/table/table.go
```diff
@@ -45,6 +45,15 @@ func PlatformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 			"kolide_zerotier_peers", dataflattentable.JsonType, zerotierCli("listpeers")),
 	}
 
+	if fcargs := fcListCli(); fcargs != nil {
+		dataflattentable.TablePluginExec(client, logger,
+			"kolide_fclist",
+			dataflattentable.KeyValueType,
+			fcargs,
+			dataflattentable.WithKVSeparator(":"),
+		)
```
blaedj, line 54:
```suggestion
		tables = append(dataflattentable.TablePluginExec(client, logger,
			"kolide_fclist",
			dataflattentable.KeyValueType,
			fcargs,
			dataflattentable.WithKVSeparator(":"),
		))
```

directionless:
Don't want this any more

PR #709, opened 2021-02-08
Return errors from dataflatten exec tables

blaedj:
This commit:
 - adds an 'error' column to the default columns in dataflattentable
 - if the exec function returns an error, that error is used to fill in 
   the 'error' column

Why?
 - this makes it easier to distinguish between no results and no
 possible results when querying launcher

example: 

```
osquery> select * from kolide_nmcli_wifi;
+---------+--------+-----+-------+-------+--------------------------------------------------------------------------------------+
| fullkey | parent | key | value | query | error                                                                                |
+---------+--------+-----+-------+-------+--------------------------------------------------------------------------------------+
|         |        |     |       |       | calling /usr/bin/nmclii. Got: : fork/exec /usr/bin/nmclii: no such file or directory |
+---------+--------+-----+-------+-------+--------------------------------------------------------------------------------------+

```


pkg/osquery/tables/dataflattentable/exec.go
```diff
@@ -53,7 +55,23 @@ func TablePluginExec(client *osquery.ExtensionManagerClient, logger log.Logger,
...
 
-	return table.NewPlugin(t.tableName, columns, t.generateExec)
+	generateFunc := t.generateExec
+
+	// if the binary doesn't exists, no-op the table's generate function.
+	if _, err := os.Stat(t.execArgs[0]); err != nil {
+		level.Info(t.logger).Log("msg", "unable to find binary to exec", "err", err)
+		generateFunc = errorGenerate(fmt.Sprintf("unable to find binary: %s", err))
+	}
+	return table.NewPlugin(t.tableName, columns, generateFunc)
```
directionless, line 65: This seems wrong -- this should be part of `generate` not instantiation. (Binaries may change over the life of the process)

pkg/osquery/tables/dataflattentable/exec.go
```diff
@@ -53,7 +55,23 @@ func TablePluginExec(client *osquery.ExtensionManagerClient, logger log.Logger,
 		panic("Unknown data source type")
 	}
 
-	return table.NewPlugin(t.tableName, columns, t.generateExec)
+	generateFunc := t.generateExec
+
+	// if the binary doesn't exists, no-op the table's generate function.
+	if _, err := os.Stat(t.execArgs[0]); err != nil {
+		level.Info(t.logger).Log("msg", "unable to find binary to exec", "err", err)
```
directionless, line 62: I think there's a cleaner place for this... The exec can fail for any number of reasons. We should populate the `results` with the error after `t.exec(ctx)`

directionless: I'm not sure we even need the stat -- `exec.CommandContext` should handle it. Just push the error along.

pkg/osquery/tables/dataflattentable/helpers.go
```diff
@@ -40,6 +40,7 @@ func Columns(additional ...table.ColumnDefinition) []table.ColumnDefinition {
 		table.TextColumn("key"),
 		table.TextColumn("value"),
 		table.TextColumn("query"),
+		table.TextColumn("error"),
```
directionless, line 43: I find myself unsure of `error` or `_error`. If I think about tables in general, I like `_error` to indicate something internal. But this is all datafatten land. No one is working with these columns directly, so `error` feels clean.

directionless (changes requested):
love the intent, though I want to nitpick the details.

pkg/osquery/tables/dataflattentable/helpers.go
blaedj: On a semi-related note: as I thought about it more, I dislike the idea of adding an 'error' column to a DB table, but as you said this is not a typical sql table, so I think it's fine.

directionless: I'm curious about what you're thinking.

blaedj: I just took a step back and imagined adding an 'error' column to a 'real' database table, and populating that column with internal error messages. It would feel.. weird. But, this isn't a 'real' database table, so I don't think it's an issue.

pkg/osquery/tables/dataflattentable/exec.go
blaedj: 👍 yea I can move it. I had it in `generate` initially, and moved it to the instantiation to avoid unnecessary calls. Probably not worth the need to reboot after installing nmcli for launcher to detect it.

pkg/osquery/tables/dataflattentable/helpers.go
directionless: It's not so obviously wrong to me. It bumps into all the in-band vs out-of-band signaling questions. We have some similar ones in jsonrpc. 

I think the core of the confusion is osquery looks like a database, but it really isn't. It's an API translation layer. There is no pile of data on disk that sqlite is querying. Instead, we use sqlite as a magic api translator. So we need some way to embed errors. 

Arguably it's better to do it via the statuses. But, because of the rest of the system, that just gets weird around being able to batch things everywhere.

I dunno, maybe I'm wrong. Maybe it's better to return a status error, instead of an error column

directionless: But I also think there's no real correct answer, and we probably won't understand all the tradeoffs till we've worked with it for a little

directionless (approved):
Let's try!

directionless (changes requested):
While the code is fine, I am no longer sure this is a good approach. Discussion in office slack.

PR #710, opened 2021-02-10
Table for windows installed updates

directionless:
Simple table using the existing powershell snippet to generate the installed history.

There is interest longer term work in building a richer interface to the windows update stuff, but it's all `com` interfaces, and those aren't quick. https://github.com/ceshihao/windowsupdate might be a good starting point.


blaedj (approved):
LGTM, and in testing it returned results.

PR #713, opened 2021-02-20
Go 1.16

directionless:
Go 1.16 has been [released](https://blog.golang.org/go1.16)

Notable changes:
* core `embed` package, means we can ditch go-binutils
* `ioutils` deprecation
* `vet` [disallows reflect.StringHeader](https://github.com/golang/go/issues/40701). But it looks to be an unneeded usage


blaedj (approved):
this seems fine, we'll probably need to beta test this to suss out any subtle bugs.

PR #714, opened 2021-02-20
Use dsim to get associations on windows

directionless:
Use exec + dataflatten to create a `dsim.exe` based table.

### DIFF
--- a/pkg/osquery/tables/windows_associations/windows_associations.go
+++ b/pkg/osquery/tables/windows_associations/windows_associations.go
@@ -1,0 +1,111 @@
+// +build windows
+
+package windows_associations
+
+import (
+	"bytes"
+	"context"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/dataflattentable"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+const dismCmd = "dism.exe"
+
+type Table struct {
+	client *osquery.ExtensionManagerClient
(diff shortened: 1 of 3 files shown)


pkg/osquery/tables/windows_associations/windows_associations.go
```diff
@@ -0,0 +1,111 @@
...
+
+		results = append(results, dataflattentable.ToMap(rows, dataQuery, nil)...)
+	}
+
+	return results, nil
+}
+
+func (t *Table) execDism(ctx context.Context) ([]byte, error) {
+	// dism.exe outputs xml, but with weird intermingled status. So
+	// instead, we dump it to a temp file.
```
blaedj, line 78: Are the weird 'status headers' omitted from the temp file when the `/Export-DefaultAppAssociations` flag is specified?

directionless: Yes.

directionless: Not bouncing through a tmp file would be nice, it feels like pointless diskio.

But, parsing stdout feels fragile (maybe it'd be fine?)

And handing it a socket to write into feels like non-trivial new work.

So this was the simple route

pkg/osquery/tables/windows_associations/windows_associations.go
```diff
@@ -0,0 +1,111 @@
+// +build windows
+
+package windows_associations
```
blaedj, line 3: I wonder - does it make sense to specify in the package/table name that these are file-to-program/app associations?

blaedj (approved):
lgtm, one naming suggestion.

pkg/osquery/tables/windows_associations/windows_associations.go
directionless: Maybe? I always feel weird when I don't name tables for the data source (in this case, `dsim`). I think I used `windows_associations` because it was simple. Maybe `dism_app_associations` or `dism_default_associations`?

@FritzX6  do you have an opinion?

blaedj: I think either of those is clearer - my concern is primarily that 'associations' is ambiguous.

directionless: Opted for `dsim_default_associations`

PR #715, opened 2021-02-20
Convert to bbolt

directionless:
github.com/boltdb/bolt has been archived for awhile, and modern go has been grumpy about it's pointer use for awhile. (See https://github.com/etcd-io/bbolt/issues/187). Convert over to https://github.com/etcd-io/bbolt

Works on my mac. I'll want to test this more widely, but that's probably easier via the alpha or beta channel.

### DIFF
--- a/cmd/grpc.ext/grpc.go
+++ b/cmd/grpc.ext/grpc.go
@@ -9,7 +9,7 @@ import (
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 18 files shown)


directionless:
I've rebased this, and tested it on both my mac and window box. I think it's good!

blaedj (approved):

PR #717, opened 2021-02-25
Auto Repair the local TUF repo

directionless:
We've seen the local TUF repo get corrupt. This adds some simple local repairs when that happens.

I've seen the TUF repo get corrupt. The cases I've seen are benign, and I believe due to some error handling in the [updater](https://github.com/kolide/updater). I think it's reasonable to detect and repair these. First, we're replacing the bad local copy with an expected good one from the compiled assets. Second, what else do we expect to happen?

Testing this on my machine, it does recover from these states.

### DIFF
--- a/pkg/autoupdate/autoupdate.go
+++ b/pkg/autoupdate/autoupdate.go
@@ -43,6 +43,7 @@
 package autoupdate
 
 import (
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


blaedj (approved):
It took a bit to grok the existing code, but I think I got it. lgtm

PR #719, opened 2021-03-05
Move windows updates to go

directionless:
Move the windows update data from a powershell exec to go-ole.

This leverages the work in https://github.com/ceshihao/windowsupdate

### DIFF
--- a/pkg/osquery/tables/windowsupdate/windowsupdate.go
+++ b/pkg/osquery/tables/windowsupdate/windowsupdate.go
@@ -1,0 +1,177 @@
+// +build windows
+
+package windowsupdate
+
+import (
+	"context"
+	"encoding/json"
+	"strconv"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/dataflattentable"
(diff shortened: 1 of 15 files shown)


pkg/osquery/tables/windowsupdate/windowsupdate.go
```diff
@@ -0,0 +1,184 @@
...
+		flattenOpts = append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))
+	}
+
+	if t.logger != nil {
+		flattenOpts = append(flattenOpts,
+			dataflatten.WithLogger(level.NewFilter(t.logger, level.AllowInfo())),
+		)
+	}
+
+	// Works better if we bounce through json. yuck
```
blaedj, line 135: Works better in what way?

 If it's just for ease of dataflattening (i.e. not needing to implement a new dataflatten routine for the `ISearchResult` etc types), that seems very reasonable to me.

blaedj (commented):
Couple comments. overall this looks pretty good, I suspect that there is the potential for bugs in the `pkg/windows` code, e.g. not matching the windows api types properly, but this seems to work well enough for our purposes.

pkg/osquery/tables/windowsupdate/windowsupdate.go
directionless: Correct. I've updated the comment.

blaedj (approved):
lgtm

PR #720, opened 2021-03-05
Increase timeout for powershell wlanapi execution

blaedj:
This commit:
 - increases the timeout for powershell execution of the wlanapi calls

Why?
 - while network interfaces _should_ complete this in under 4 seconds,
 empirically many do not.


directionless (approved):
Don't totally love that debug line, but let's keep it while we figure this out

PR #721, opened 2021-03-09
Code tidy

directionless:
Batching several small changes

1. Remove CI builder, which is now too new for go 1.13
2. Use the `comshim` library for wmi
3. Update dataflatten to accept a logger, and apply the info filter internally
4. Update dataflatten callers to have fewer if blocks (dataflatten duplicates that internally)

### DIFF
--- a/pkg/dataflatten/flatten.go
+++ b/pkg/dataflatten/flatten.go
@@ -114,7 +114,7 @@ func WithDebugLogging() FlattenOpts {
 // re-writing arrays into maps, and for filtering. See "Query
 // Specification" for docs.
 func WithQuery(q []string) FlattenOpts {
-	if q == nil || len(q) == 0 {
+	if q == nil || len(q) == 0 || (len(q) == 1 && q[0] == "") {
 		return func(_ *Flattener) {}
 	}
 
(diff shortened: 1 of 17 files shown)


pkg/dataflatten/flatten.go
```diff
@@ -90,16 +91,30 @@ func WithNestedPlist() FlattenOpts {
...
+		fl.debugLogging = true
+	}
+}
+
 // WithQuery Specifies a query to flatten with. This is used both for
 // re-writing arrays into maps, and for filtering. See "Query
 // Specification" for docs.
 func WithQuery(q []string) FlattenOpts {
-	if q == nil || len(q) == 0 {
+	if q == nil || len(q) == 0 || (len(q) == 1 && q[0] == "") {
```
blaedj, line 117: `&&` has operator precedence over `||` in go, correct?

blaedj (approved):
lgtm

pkg/dataflatten/flatten.go
directionless: Good call on adding the `(len(q) == 1 && q[0] == "")` stuff. All my callers have a split, so I didn't even think of it. But `""` as a query will just return no results.

directionless: precedence shouldn't matter -- it's got parens. 😎

blaedj: yea it sure does... For some reason I didn't see those ¯\_(ツ)_/¯

PR #722, opened 2021-03-11
Detect and fix improper file ownership on launch

blaedj:
This commit:
 - detects a subset of what osquery will consider improper file
 perms/ownership on the extension binary
 - tries to fix those errors

We could and perhaps should apply this to the entire root directory,
and this doesn't yet work on windows.

### DIFF
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -524,6 +524,14 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "could not calculate osquery file paths")
 	}
 
+	err = ensureProperPermissions(o, paths)
+	if err != nil {
+		level.Info(o.logger).Log(
+			"msg", "unable to ensure proper permissions",
+			"err", err,
+		)
+	}
+
 	// Populate augeas lenses, if requested
 	if o.opts.augeasLensFunc != nil {
 		if err := os.MkdirAll(paths.augeasPath, 0755); err != nil {
@@ -586,17 +594,6 @@ func (r *Runner) launchOsqueryInstance() error {
 		o.opts.binaryPath,
 		autoupdate.DeleteOldUpdates(),
 	)
-
-	// The extensions file should be owned by the process's UID or by root.
-	// Osquery will refuse to load the extension otherwise. Also ensure osqueryd
-	// is owned by root or current uid.
-	err = ensureProperPermissions(o, []string{paths.extensionPath, currentOsquerydBinaryPath})
-	if err != nil {
-		level.Info(o.logger).Log(
-			"msg", "unable to ensure proper permissions",
-			"err", err,
-		)
-	}
 
 	// Now that we have accepted options from the caller and/or determined what
 	// they should be due to them not being set, we are ready to create and start
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -526,8 +526,7 @@ func (r *Runner) launchOsqueryInstance() error {
 
 	// The extensions file should be owned by the process's UID or by root.
 	// Osquery will refuse to load the extension otherwise.
-	err = ensureProperPermissions(o, paths.extensionPath)
-	if err != nil {
+	if err := ensureProperPermissions(o, paths.extensionPath); err != nil {
 		level.Info(o.logger).Log(
 			"msg", "unable to ensure proper permissions on extension path",
 			"err", err,
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -543,7 +543,10 @@ func (r *Runner) launchOsqueryInstance() error {
 				"event", "BCJ_",
 				"msg", "unsafe permissions detected on extension binary")
 
-			// chown the file
+			// chown the file. This could potentially be insecure, since we're
+			// basically chown-ing whatever is there to root, but a certain
+			// level of privilege is needed to place something in the launcher
+			// root directory.
 			err := os.Chown(paths.extensionPath, os.Getuid(), os.Getgid())
 			if err != nil {
 				return errors.Wrap(err, "attempting to chown extension binary")
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +525,32 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "could not calculate osquery file paths")
 	}
 
+	// TODO: should also handle windows. probably
```
directionless, line 528: We run windows osquery with `--allow-unsafe` for a bunch of reasons around this. Or possible that we didn't understand how to permission the socket. Not sure.

pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +525,32 @@ func (r *Runner) launchOsqueryInstance() error {
...
+		if err != nil {
+			return errors.Wrap(err, "stat-ing extension path")
+		}
+		sys := fd.Sys().(*syscall.Stat_t)
+		isRootOwned := (sys.Uid == 0)
+		isProcOwned := (sys.Uid == uint32(os.Geteuid()))
+
+		if !(isRootOwned || isProcOwned) {
+			level.Info(o.logger).Log(
+				"event", "BCJ_",
```
directionless, line 543: o_O

pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +525,32 @@ func (r *Runner) launchOsqueryInstance() error {
...
+				"msg", "unsafe permissions detected on extension binary")
+
+			// chown the file
+			err := os.Chown(paths.extensionPath, os.Getuid(), os.Getgid())
+			if err != nil {
+				return errors.Wrap(err, "attempting to chown extension binary")
```
directionless, line 549: I'd log and attempt to continue? Maybe? I guess I'm not sure

pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +525,32 @@ func (r *Runner) launchOsqueryInstance() error {
...
+		isRootOwned := (sys.Uid == 0)
+		isProcOwned := (sys.Uid == uint32(os.Geteuid()))
+
+		if !(isRootOwned || isProcOwned) {
+			level.Info(o.logger).Log(
+				"event", "BCJ_",
+				"msg", "unsafe permissions detected on extension binary")
+
+			// chown the file
+			err := os.Chown(paths.extensionPath, os.Getuid(), os.Getgid())
```
directionless, line 547: We'll meed to test, but my inclination is that we probably need to chown through all of the root dir

pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +525,32 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "could not calculate osquery file paths")
 	}
 
+	// TODO: should also handle windows. probably
+	if runtime.GOOS != "windows" {
```
directionless, line 529: Could do this in platform specific helpers, but this seems fine.

pkg/osquery/runtime/runtime.go
blaedj: yea it's a draft PR 😉 didn't mean to keep that in though

pkg/osquery/runtime/runtime.go
blaedj: That seems reasonable, it probably won't matter because osquery will almost certainly error out later on.

pkg/osquery/runtime/runtime.go
```diff
@@ -587,6 +587,17 @@ func (r *Runner) launchOsqueryInstance() error {
 		autoupdate.DeleteOldUpdates(),
 	)
 
+	// The extensions file should be owned by the process's UID or by root.
+	// Osquery will refuse to load the extension otherwise. Also ensure osqueryd
+	// is owned by root or current uid.
+	err = ensureProperPermissions(o, []string{paths.extensionPath, currentOsquerydBinaryPath})
```
blaedj, line 593: instead of the current osqueryd binary path, we could export `autoupdate.getUpdateDir`, and pass the result of that into this ensureProperPermissions func. @directionless any objections to that?

pkg/osquery/runtime/runtime_helpers.go
```diff
@@ -33,3 +35,32 @@ func platformArgs() []string {
 func isExitOk(err error) bool {
 	return false
 }
+
+func ensureProperPermissions(o *OsqueryInstance, paths *osqueryFilePaths) error {
+	// The extensions file should be owned by the process's UID or the file
+	// should be owned by root. Osquery will refuse to load the extension
+	// otherwise
+	fd, err := os.Stat(paths.extensionPath)
```
directionless, line 43: This isn't exported, will this work? (If we're only chowning the extension, I'd just pass that directly)

blaedj: It works because they're in the same package, but you're right about just passing the path directly, I'll do that.

directionless: If you want to get fancy, you can use a signature like `func ensure(filepaths ...string)`

pkg/osquery/runtime/runtime_helpers.go
```diff
@@ -33,3 +35,29 @@ func platformArgs() []string {
...
+func ensureProperPermissions(o *OsqueryInstance, path string) error {
+	fd, err := os.Stat(path)
+	if err != nil {
+		return errors.Wrap(err, "stat-ing path")
+	}
+	sys := fd.Sys().(*syscall.Stat_t)
+	isRootOwned := (sys.Uid == 0)
+	isProcOwned := (sys.Uid == uint32(os.Geteuid()))
+
+	if !(isRootOwned || isProcOwned) {
```
directionless, line 48: early return?

pkg/osquery/runtime/runtime.go
```diff
@@ -524,6 +524,16 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "could not calculate osquery file paths")
 	}
 
+	// The extensions file should be owned by the process's UID or by root.
+	// Osquery will refuse to load the extension otherwise.
+	err = ensureProperPermissions(o, paths.extensionPath)
+	if err != nil {
```
directionless, line 530:
```suggestion
	if err := ensureProperPermissions(o, paths.extensionPath); err != nil {
```

pkg/osquery/runtime/runtime_helpers.go
```diff
@@ -33,3 +35,32 @@ func platformArgs() []string {
...
+		"msg", "unsafe permissions detected on path",
+		"path", path,
+	)
+
+	// chown the path. This could potentially be insecure, since
+	// we're basically chown-ing whatever is there to root, but a certain
+	// level of privilege is needed to place something in the launcher root
+	// directory.
+	err = os.Chown(path, os.Getuid(), os.Getgid())
+	if err != nil {
```
directionless, line 62:
```suggestion
	if err = os.Chown(path, os.Getuid(), os.Getgid()); err != nil {
```

directionless (approved):
Tested and appears to work as advertised. Yay!

Digging at a test case, I don't see a homebrew provided command or docs that creates this error. But `chown $USER` does cause this problem, and it shows up in various suggestions

blaedj:
```
            chown blaed ~/launcher_root_bad_perms
            sudo rm ~/launcher_root_bad_perms/kolide/bin/launcher
            sudo rm ~/launcher_root_bad_perms/kolide/bin/osquery-extension.ext

            cp ~/code/go/src/github.com/kolide/launcher/build/launcher ~/launcher_root_bad_perms/kolide/bin/
            cp ~/code/go/src/github.com/kolide/launcher/build/osquery-extension.ext /users/blaed/tmp/launcher_root_bad_perms/kolide/bin/

            sudo launchctl asuser 0 \
                 ~/launcher_root_bad_perms/kolide/bin/launcher \
                 -root_directory "~/launcher_root_bad_perms/kolide" \
                 -hostname localhost:5000 \
                 -enroll_secret_path secret.txt \
                 --osqueryd_path "~/launcher_root_bad_perms/kolide/bin/osqueryd" \
                 --insecure \
                 --insecure_transport \
                 -transport jsonrpc \
                 -autoupdate \
                 -update_channel="beta" \
                 -debug \
                 -with_initial_runner \
                 -disable_control_tls 
```

that's the script I used (or close to it) to test this out.

PR #723, opened 2021-03-16
Simple certificate info

directionless:
Simple certificate parser. Reads pems, uses dataflatten.

Not currently designed to match the osquery `certificates` table, but maybe it should be?


directionless:
I don't think I want to merge the certificates table. That's better in osquery proper. I extracted the parts of this I liked into #736

PR #731, opened 2021-06-02
Add `zfs` related tables

directionless:
Starting to rough in some zfs tables.


pkg/osquery/tables/tablehelpers/exec.go
```diff
@@ -0,0 +1,40 @@
...
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	level.Debug(logger).Log(
+		"msg", "execing",
+		"cmd", cmd.String(),
+	)
+
+	if err := cmd.Run(); err != nil {
+		level.Info(logger).Log(
```
directionless, line 31: TODO: Remove this log. It should be the caller's responsibility

blaedj (approved):
lgtm, didn't have a linux device with any zfs volumes, but with the zfs tools installed the table didn't error out at least.

PR #732, opened 2021-06-02
Add `lsblk` table

directionless:
Very simple `kolide_lsblk` table. I'm not sure that it adds data that isn't in the native osquery tables, but it presents it in a different way. Seems valuable in diagnosing linux disk setups.

To handle different locations on different linux distributions, this also introduces a `dataflattentable.WithBinDirs` option to exec table

### DIFF
--- a/pkg/osquery/table/platform_tables_linux.go
+++ b/pkg/osquery/table/platform_tables_linux.go
@@ -17,7 +17,6 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		dataflattentable.TablePluginExec(client, logger,
 			"kolide_nmcli_wifi", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
 			dataflattentable.WithKVSeparator(":")),
-
 		dataflattentable.TablePluginExec(client, logger, "kolide_lsblk", dataflattentable.JsonType, []string{"/usr/bin/lsblk", "-J"}),
 	}
 }
--- a/pkg/osquery/table/platform_tables_linux.go
+++ b/pkg/osquery/table/platform_tables_linux.go
@@ -15,8 +15,12 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		gsettings.Settings(client, logger),
 		gsettings.Metadata(client, logger),
 		dataflattentable.TablePluginExec(client, logger,
-			"kolide_nmcli_wifi", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
+			"kolide_nmcli_wifi", dataflattentable.KeyValueType,
+			[]string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
 			dataflattentable.WithKVSeparator(":")),
-		dataflattentable.TablePluginExec(client, logger, "kolide_lsblk", dataflattentable.JsonType, []string{"/usr/bin/lsblk", "-J"}),
+		dataflattentable.TablePluginExec(client, logger, "kolide_lsblk", dataflattentable.JsonType,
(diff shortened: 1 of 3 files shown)


pkg/osquery/table/platform_tables_linux.go
```diff
@@ -17,5 +17,7 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		dataflattentable.TablePluginExec(client, logger,
 			"kolide_nmcli_wifi", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
 			dataflattentable.WithKVSeparator(":")),
+
```
directionless, line 20:
```suggestion
```

pkg/osquery/table/platform_tables_linux.go
```diff
@@ -17,5 +17,6 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		dataflattentable.TablePluginExec(client, logger,
 			"kolide_nmcli_wifi", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
 			dataflattentable.WithKVSeparator(":")),
+		dataflattentable.TablePluginExec(client, logger, "kolide_lsblk", dataflattentable.JsonType, []string{"/usr/bin/lsblk", "-J"}),
```
blaedj, line 20: On a test device of mine (a DO server), `lsblk` is in my path but located at `/bin/lsblk`

blaedj: I can't remember - is there an easy way to specify multiple potential locations for the binary?

directionless: Oh, it's great to know that. I don't think, but we can write one.

directionless: Okay, I made one. See if it works on your DO machine. 

Any idea what the path to zfs is on that one?

blaedj (approved):
lgtm, seems to work fine on my test device.

pkg/osquery/table/platform_tables_linux.go
```diff
@@ -15,7 +15,12 @@ func platformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 		gsettings.Settings(client, logger),
 		gsettings.Metadata(client, logger),
 		dataflattentable.TablePluginExec(client, logger,
-			"kolide_nmcli_wifi", dataflattentable.KeyValueType, []string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
+			"kolide_nmcli_wifi", dataflattentable.KeyValueType,
+			[]string{"/usr/bin/nmcli", "--mode=multiline", "--fields=all", "device", "wifi", "list"},
 			dataflattentable.WithKVSeparator(":")),
+		dataflattentable.TablePluginExec(client, logger, "kolide_lsblk", dataflattentable.JsonType,
+			[]string{"lsblk", "-J"},
```
blaedj, line 22: from the man pages for lsblk: 
> Always explicitly define expected columns by using --output columns-list and --list  in  environments
       where a stable output is required.

We're specifying JSON output, so things probably won't break in a crashy way, but might be good to specify the columns we want just for consistency?

directionless: Oh hey, docs! Brilliant.

I think `-l` is wrong for us -- it's a different output format. `-P` is interesting, but as long as we have wide support for `-J` that seems simpler. 

I don't think we should be using `--output-columns` either. We don't need consistency, we're parsing EAV style. TBH I wish there was an option to output all the columns, but I don't see it. And it errors on unknowns. So I think that means we either accept the defaults, or plumb through a query style.

I do see a `-a` which would include "empty" devices. 

I can be convinced to iterate on this, but I think I mostly want to get it out and see what we need

blaedj: Getting this out and iterating sounds good to me!

PR #734, opened 2021-06-03
Add `cryptsetup_status` table

directionless:
This creates a `kolide_cryptsetup_status` as a wrapper over the `cryptsetup status` command. Similar to #732, it's meant to aid in debugging and understanding

This depends on #731

### DIFF
--- a/pkg/osquery/tables/cryptsetup/table.go
+++ b/pkg/osquery/tables/cryptsetup/table.go
@@ -1,0 +1,89 @@
+package cryptsetup
+
+import (
+	"context"
+	"strings"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/dataflattentable"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+const cryptsetupPath = "/usr/sbin/cryptsetup"
+
+const allowedNameCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-/_"
+
+type Table struct {
+	client *osquery.ExtensionManagerClient
+	logger log.Logger
+	name   string
+}
+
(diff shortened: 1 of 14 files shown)


blaedj:
Testing this out manually, it required that I find the name of a LUKS encrypted device (more annoying that it may sound), but seems to work!

directionless:
> Testing this out manually, it required that I find the name of a LUKS encrypted device (more annoying that it may sound), but seems to work!

Yeah, I think that limitation is inherent. And I don't want the table to be clever.

pkg/osquery/tables/cryptsetup/table.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	requestedNames := tablehelpers.GetConstraints(queryContext, "name",
+		tablehelpers.WithAllowedCharacters(allowedNameCharacters),
+		tablehelpers.WithLogger(t.logger),
+	)
+
+	if len(requestedNames) == 0 {
```
blaedj, line 49: In [other tables](https://github.com/kolide/launcher/blob/master/pkg/osquery/table/slack_config.go#L95_L101), we've also checked that the constraint is `=`, and that's functionally all this table handles as well.

directionless: Hrm.... You're not wrong, it feels like it would be better to check that the `constraint.Operator == table.OperatorEquals` But...

1. This is using `tablehelpers.GetConstraints` to get a simple iterable. If we wanted to check the Operator, it should be in there.
2. IIRC osquery doesn't pass anything other than `=` to table anyhow.

I'd be willing to think about how to have the helper be more clever about the Operator, since I think (2) is a bug. But I'm not sure what the desired outcome should be.

blaedj: > 2. IIRC osquery doesn't pass anything other than = to table anyhow.

Ah I didn't realize that, I thought all constraints were passed through, that feels like either a bug or undesirable limitation to me as well.

blaedj: For the record, this isn't necessarily a blocker, but I think consistent handling of the operators is one of those small things that makes the project better overall.

directionless: I agree about consistency. I suspect we should push handling into the helper, and have it return `[]string, error` instead of just `[]string`

blaedj (approved):
There's some discussion about checking for constraint types, but that's not really a blocker. This lgtm

PR #736, opened 2021-06-07
Small cleanup and improvements to dataflattentables

directionless:
A couple of small code tweaks to dataflattentables.

* Adds another string method
* Makes a helper to simply adding tables

### DIFF
--- a/pkg/osquery/tables/dataflattentable/tables.go
+++ b/pkg/osquery/tables/dataflattentable/tables.go
@@ -39,6 +39,16 @@ type Table struct {
 	keyValueSeparator string
 }
 
+// AllTablePlugins is a helper to return all the expected flattening tables.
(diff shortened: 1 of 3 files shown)


pkg/osquery/tables/dataflattentable/tables.go
```diff
@@ -39,6 +39,16 @@ type Table struct {
 	keyValueSeparator string
 }
 
+// AllTablePlugins is a helper to return all the expected flattening tables.
+func AllTablePlugins(client *osquery.ExtensionManagerClient, logger log.Logger) []*table.Plugin {
```
blaedj, line 43: these are the generic dataflatten tables that extract data from a file on disk, correct?

directionless: Yep.

blaedj (approved):

PR #738, opened 2021-06-10
add xrdb table

blaedj:
This commit:
 - adds an xrdb table

Caveats:
 - this probably doesn't work when run as root.
 - the display is hardcoded to `:0`, which is probably a terrible
 hack. We may not need the screen-specific values though, if we're
 trying to fetch information for xautolock. Still have to setup
 xautolock and test how it stores values that might be accessible via
 xrdb

### DIFF
--- a/pkg/osquery/tables/xrdb/testdata/example_query.sql
+++ b/pkg/osquery/tables/xrdb/testdata/example_query.sql
@@ -1,0 +1,8 @@
+WITH human_accounts AS (
+  SELECT username FROM users u WHERE u.uid >= 1000 AND u.uid < 60000
+),
+  user_xrdb_values AS (
+    SELECT * FROM kolide_xrdb kx WHERE kx.username IN (SELECT username from human_accounts)
+    )
+
+SELECT * FROM user_xrdb_values;
--- a/pkg/osquery/tables/xrdb/xrdb.go
+++ b/pkg/osquery/tables/xrdb/xrdb.go
@@ -1,0 +1,179 @@
+// +build linux
+
+package xrdb
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"os/user"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+var potentialPaths = []string{"/usr/bin/xrdb"}
+
+const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-."
+
+type execer func(ctx context.Context, username string, buf *bytes.Buffer) error
+
+type XRDBSettings struct {
+	client   *osquery.ExtensionManagerClient
+	logger   log.Logger
+	getBytes execer
+}
+
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("key"),
+		table.TextColumn("value"),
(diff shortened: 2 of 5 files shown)


pkg/osquery/tables/xrdb/xrdb.go
```diff
@@ -0,0 +1,179 @@
...
+	cmd.Stdout = buf
+
+	if err := cmd.Run(); err != nil {
+		return errors.Wrapf(err, "running xrdb, err is: %s", stderr.String())
+	}
+
+	return nil
+}
+
+func (t *XRDBSettings) parse(username string, input io.Reader) []map[string]string {
```
directionless, line 147: This function needs a test case. (And probably only this one)

pkg/osquery/tables/xrdb/testdata/example_query.sql
```diff
@@ -0,0 +1,8 @@
+WITH human_accounts AS (
```
directionless, line 1: Do we need this file?

pkg/osquery/tables/xrdb/xrdb.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/kolide/osquery-go"
+	"github.com/kolide/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+var xrdbPath = "/usr/bin/xrdb"
+
+const allowedUsernameCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-."
+const allowedDisplayCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:"
```
directionless, line 31:
```suggestion
const allowedDisplayCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:."
```

pkg/osquery/tables/xrdb/xrdb.go
```diff
@@ -0,0 +1,183 @@
...
+
+	return table.NewPlugin("kolide_xrdb", columns, t.generate)
+}
+
+func (t *XRDBSettings) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	users := tablehelpers.GetConstraints(queryContext, "username", tablehelpers.WithAllowedCharacters(allowedUsernameCharacters))
+	if len(users) < 1 {
+		return results, errors.New("kolide_xrdb requires at least one username to be specified")
```
directionless, line 63: I forget how this works. we might also need a `level.Info(t.logger).Log(...)` here

directionless (commented):
A couple of small tweaks, and I think it's good.

pkg/osquery/tables/xrdb/testdata/example_query.sql
directionless: (I'm happy to keep it as an example)

blaedj: Don't need it, I just had it saved to make it easier to run a semi-automatic test during dev. I figured it'd be nice for an example.

pkg/osquery/tables/xrdb/xrdb.go
blaedj: I can certainly add one, but I checked other tables and we don't log alongside the error return. 
The `kolide_cryptsetup_status` table is a recent example that still bubbles up an error (in verbose mode) when a constraint isn't specified: 

```
osquery> select * from kolide_cryptsetup_status;
I0614 12:52:34.403594 39429 virtual_table.cpp:1020] Invalid response from the extension table. Error 1: error generating table: The kolide_cryptsetup_status table requires that you specify a constraint for name
Error: error generating table: The kolide_cryptsetup_status table requires that you specify a constraint for name
```

pkg/osquery/tables/xrdb/testdata/example_query.sql
directionless: WFM

pkg/osquery/tables/xrdb/xrdb.go
directionless: Good enough then. No need.

directionless (approved):

PR #743, opened 2021-06-25
fscrypt table

directionless:
This adds an fscrypt table for linux. This table is based on the go fscrypt code, which needs CGO. This brings in zig to support that

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -123,6 +123,27 @@ func New(opts ...Option) (*Builder, error) {
 	cmdEnv = append(cmdEnv, "GO111MODULE=on")
 	cmdEnv = append(cmdEnv, fmt.Sprintf("GOOS=%s", b.os))
 	cmdEnv = append(cmdEnv, fmt.Sprintf("GOARCH=%s", b.arch))
+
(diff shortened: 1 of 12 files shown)


pkg/make/builder.go
```diff
@@ -123,6 +123,26 @@ func New(opts ...Option) (*Builder, error) {
 	cmdEnv = append(cmdEnv, "GO111MODULE=on")
 	cmdEnv = append(cmdEnv, fmt.Sprintf("GOOS=%s", b.os))
 	cmdEnv = append(cmdEnv, fmt.Sprintf("GOARCH=%s", b.arch))
+
+	// Setup zig as cross compiler
+	if b.os != runtime.GOOS {
+		// This is mostly to support fscrypt on linux, but we can aim for consistency
```
directionless, line 129:
```suggestion
	// Setup zig as cross compiler
	// (This is mostly to support fscrypt on linux)
	if b.os != runtime.GOOS {
```

blaedj (approved):
LGTM, though I would recommend some check/warning in the `deps` make target, or maybe in the `tools/z*` scripts so that we have a more user-friendly error.

PR #744, opened 2021-06-30
For exec tables, don't error for missing binaries

directionless:
When an exec table is missing a binary, this condition should be treated as a warning, not an error. This way the table can be included in larger check queries and simply ignored.


blaedj (approved):
I like this idea

PR #745, opened 2021-06-30
Docker build updates

directionless:
Updates some things in how we build docker containers, as well as add ubuntu20

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,12 +28,6 @@ ARG gitver=master
 RUN git clone https://github.com/kolide/launcher.git
 RUN cd launcher && git checkout "${gitver}"
 
-# copy source into the docker builder. Perhaps this would be cleaner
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -5,33 +5,49 @@
...
+# git clone ensures clean source
+RUN git clone https://github.com/kolide/launcher.git
+RUN cd launcher && git checkout "${gitver}"
 
 # copy source into the docker builder. Perhaps this would be cleaner
 # via a volume mount. But, COPY allows the docker container to have
 # zero impact on the host's build directory. This uses .dockerignore
 # to exclude the build directory
-COPY . ./
+# COPY . ./
```
blaedj, line 35: rm all this?

directionless: I'd left it since I might develop it in the future, but you're right. It's trivial. cleaner to delete and re-add.

PR #746, opened 2021-07-02
Exec table for SecureBoot status

directionless:
This wraps mokutil's status command to return the SecureBoot status.

As an experiment, it inlines the errors. This should allow it to succeed in larger join operations, but still provides a hook for debugging. Not sure I like it yet.

I'm flexible on table and column names.

### DIFF
--- a/pkg/osquery/tables/mokutil/status.go
+++ b/pkg/osquery/tables/mokutil/status.go
@@ -3,6 +3,7 @@ package mokutil
 import (
 	"bytes"
 	"context"
+	"fmt"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


pkg/osquery/tables/mokutil/status.go
```diff
@@ -0,0 +1,66 @@
...
+
+	t := &Table{
+		logger: logger,
+	}
+
+	return table.NewPlugin("kolide_mokutil_status", columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	row := map[string]string{"secureboot": "unknown"}
```
directionless, line 43: I can't decide if the error condition should return `unknown` or `NULL`

directionless:
Pause in this for a little... I want to see if I can read it out of the kernel

directionless:
Closing in favor of #747

PR #747, opened 2021-07-05
Linux SecureBoot table

directionless:
Linux SecureBoot table. Implements a simple efivar to read it from the kernel filesystem.

I opted to implement my own, and not grow the dependency graph. Simple enough here.

Probably worth pushing this to upstream osquery

Closes: #746


blaedj (approved):
lgtm, and pretty much works as expected on my machine.

PR #751, opened 2021-07-29
Go 1.16 (Build only)

directionless:
Move the automated builds to go 1.16. 

This matches how the builds actually happen


blaedj (approved):

directionless:
Simple thing won't work. Need to do it the larger way in #713

Thanks CI! I appreciate the catch

PR #757, opened 2021-09-11
Allow spaces in gsettings table

terracatta:
Linux usernames [are allowed to contain spaces](https://stackoverflow.com/questions/208054/user-names-and-white-spaces so we should not be filtering them in this table. It's possible other allowed characters sets need to be examined for the same assumption, but this one definitely needs to be added.

### DIFF
--- a/pkg/osquery/tables/gsettings/gsettings.go
+++ b/pkg/osquery/tables/gsettings/gsettings.go
@@ -27,7 +27,7 @@ import (
 
 const gsettingsPath = "/usr/bin/gsettings"
 
-const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-."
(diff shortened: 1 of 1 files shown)


pkg/osquery/tables/gsettings/gsettings.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const gsettingsPath = "/usr/bin/gsettings"
 
-const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-."
+const allowedCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_-. "
```
directionless, line 30: This can't be changed in isolation. Pretty sure:
* #L101 will need better quoting
* #L161 may or may not be okay

This is going to need a manual test pass to ensure it's works as designed. Since there's some shell in here, anything with spaces feels risky

directionless (changes requested):

directionless:
This won't work as written, and if it becomes an issue we'll revisit.

PR #761, opened 2021-10-28
Expand Linting

directionless:
`golangci-lint` seems common. This moves our linting to using it. New github job, and cleanup on some of the weird tools pattern.

To support this, there's also a bunch of tiny bits of code cleanup. (And lint ignores)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,6 +215,7 @@ test: generate
 ##
 
 lint:
+	# Requires golangci-lint installed, see https://golangci-lint.run/usage/install#local-installation
 	golangci-lint -j3 run
 
 ##
--- a/Makefile
+++ b/Makefile
@@ -214,37 +214,8 @@ test: generate
 ## Lint
 ##
 
-# These are escape newlines, looks super weird. Allows these to run in
-# parallel with `make -j`
-lint: \
-	lint-go-deadcode \
-	lint-misspell \
-	lint-go-vet \
-	lint-go-nakedret \
-	lint-go-fmt
-
-lint-go-deadcode: deps-go
-	deadcode cmd/ pkg/
-
-lint-misspell: deps-go
-	git ls-files \
-	  | grep -v pkg/simulator/testdata/bad_symlink \
-	  | grep -v assets.go$ \
-	  | xargs misspell -error -f 'misspell: {{ .Filename }}:{{ .Line }}:{{ .Column }}:corrected {{ printf "%q" .Original }} to {{ printf "%q" .Corrected }}'
-
-lint-go-vet:
-	go vet ./cmd/... ./pkg/...
-
-lint-go-nakedret: deps-go
-	nakedret ./pkg/... ./cmd/...
-
-# This is ugly. since go-fmt doesn't have a simple exit code, we use
-# some make trickery to handle failing if there;s output.
-lint-go-fmt: $(foreach c,$(shell gofmt -l ./pkg/ ./cmd/ | grep -vE 'assets.go|bindata.go'),fmt-fail/$(c))
-lint-go-fmt: deps-go
-fmt-fail/%:
-	@echo fmt failure in: $*
-	@false
+lint:
+	golangci-lint -j3 run
 
 ##
 ## Docker Tooling
--- a/cmd/launcher/control_windows.go
+++ b/cmd/launcher/control_windows.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 package main
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# set line endings got go files. Mostly needed for golang-ci
+*.go text eol=lf
+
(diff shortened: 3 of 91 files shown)


Makefile
```diff
@@ -214,37 +214,8 @@ test: generate
...
-
-# This is ugly. since go-fmt doesn't have a simple exit code, we use
-# some make trickery to handle failing if there;s output.
-lint-go-fmt: $(foreach c,$(shell gofmt -l ./pkg/ ./cmd/ | grep -vE 'assets.go|bindata.go'),fmt-fail/$(c))
-lint-go-fmt: deps-go
-fmt-fail/%:
-	@echo fmt failure in: $*
-	@false
+lint:
+	golangci-lint -j3 run
```
blaedj, line 218: `make lint` no longer works (unless I install golangci-lint). This isn't _necessarily_ terrible, but it'd be nice to address it.

Installing is easy for macOS (could just add `brew install golangci-lint` to the `deps` make target), but the recommended way for windows/linux is to curl a binary.

I tried adding `_ "github.com/golangci/golangci-lint/"` to https://github.com/kolide/launcher/blob/seph/modern-lint/pkg/tools/tools.go#L13, but get an error that golangci-lint is a program, not an importable package. 

what _did_ work is: 
```suggestion
	golangci-lint -j3 run
	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
```

I'm not sure how you feel about that pattern to the makefile though.

directionless: Somewhat intentionally, I left it out of things like `tools` and the makefile. It feels like the install should be _outside_ this.  In some ways, I'm stepping back from whatever I advocate from when I wrote this Makefile. But it feels like the world mostly moved to using this tool, and I don't want to fight it.

Might be reasonable to throw an error from `make` though. (Though make's normal command not found is understandable enough. Ish)

blaedj: If it's an intentional choice that's fine, but IMO it's a step backward in the dev experience for the project, I always loved how seamlessly `make deps ; make launcher`  worked. 

On the other hand, I don't see a clearly idiomatic way to install this without `brew` or `curl .. | sh`, so.. ¯\\_(ツ)_/¯
(the `go install` above is not recommended by the golangci-lint project)

directionless: `make deps launcher` should still work, `lint` has always been an optional step.

But, it's also never worked without installing `go` out of band. Feels kinda the same.

blaedj, line 218:
```suggestion
	# Requires golangci-lint installed, see https://golangci-lint.run/usage/install#local-installation
	golangci-lint -j3 run
```

cmd/launcher/control_windows.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
```
blaedj, line 1: asking out of curiosity, are there 2 different build tag formats now? Or is the new one for something specific?

blaedj (commented):
Should the `make lint` emit any errors? Or are we only linting CI on changed code somehow? (I see some errors from pkg/windows/oleconv/oleconv.go` when I run `make lint` locally)

Makefile
blaedj: good enough for me

cmd/launcher/control_windows.go
directionless: To be honest, I'm not sure. I think we're mid change. AFAICT the current recommendation is that all machine parsable stuff start with `//go:`, and I think it might start mattering in 1.17. So a bunch of the linters seem to be complaining about it now. 

I kinda took the easy path here

directionless:
> Should the `make lint` emit any errors? Or are we only linting CI on changed code somehow? (I see some errors from `pkg/windows/oleconv/oleconv.go` when I run `make lint` locally)

Unlike our rails codebase, this one runs the lint everywhere. I would not expect errors.

blaedj:
Here are the errors, if you're interested - these files may be skipped in ci for some reason? Not sure

```
pkg/windows/oleconv/oleconv.go:23:29: undeclared name: `ole` (typecheck)
func ToIDispatchErr(result *ole.VARIANT, err error) (*ole.IDispatch, error) {
                            ^
pkg/windows/oleconv/oleconv.go:30:31: undeclared name: `ole` (typecheck)
func ToStringSliceErr(result *ole.VARIANT, err error) ([]string, error) {
                              ^
pkg/windows/oleconv/oleconv.go:43:25: undeclared name: `ole` (typecheck)
func ToInt64Err(result *ole.VARIANT, err error) (int64, error) {
                        ^
pkg/windows/oleconv/oleconv.go:12:2: "github.com/go-ole/go-ole" imported but not used (typecheck)
        "github.com/go-ole/go-ole"
        ^
pkg/windows/windowsupdate/icategory.go:4:2: "github.com/go-ole/go-ole" imported but not used (typecheck)
        "github.com/go-ole/go-ole"
        ^
pkg/windows/windowsupdate/iimageinformation.go:4:2: "github.com/go-ole/go-ole" imported but not used (typecheck)
        "github.com/go-ole/go-ole"
        ^
```

directionless:
close and open to rebase it

blaedj (approved):
The errors I see that aren't present in CI are likely related to the fact that I'm running go 1.17.2 and CI is on 1.16

PR #768, opened 2021-10-31
Move project workflow things to an action

directionless:
This moves the workflow for project boards to using a new github action. This is to work around needing a checkout in a PR environment.

This should fix the failing job

This time I tested it, at least a little, using a test repo :)


blaedj (approved):
seems fine if you tested it, I don't fully understand the GH workflow.

Issue #763, opened 2021-10-28
Table Request: kolide_remotectl


FritzX6:
## What is being requested?
The `remotectl` binary available on macOS can be utilized to gather additional host information and as a fallback when other datasources do not populate.

Invoking this binary can be done using the following command:
```
/usr/libexec/remotectl dumpstate
```

Output is in a semi-structured and nested key:value format as demonstrated below (items containing the string `**REDACTED**` were redacted by me, not the binary's output):
```
> /usr/libexec/remotectl dumpstate
Local device
	UUID: 4D259086-B8CF-4A99-A2FB-**REDACTED**
	Messaging Protocol Version: 2
	Product Type: iMacPro1,1
	OS Build: 11.6 (20G165)
	Properties: {
		AppleInternal => false
		CPUArchitecture => x86_64h
		EffectiveProductionStatusSEP => false
		EthernetMacAddress => d0:81:**REDACTED**
		HWModel => J137AP
		HasSEP => true
		IsUIBuild => true
		DeviceSupportsLockdown => false
		EffectiveSecurityModeAp => false
		SigningFuse => true
		BuildVersion => 20G165
		OSVersion => 11.6
		BridgeVersion => 18.16.14759.0.1,0
		SensitivePropertiesVisible => true
		ProductType => iMacPro1,1
		SerialNumber => C02V**REDACTED**
		BootSessionUUID => A7C32508-8D2C-**REDACTED**
		DeviceColor => unknown
		EffectiveProductionStatusAp => false
		EffectiveSecurityModeSEP => false
		UniqueChipID => 14531085**REDACTED**
		UniqueDeviceID => e664d52b5fbc5a68**REDACTED**
		RemoteXPCVersionFlags => 7205759**REDACTED**
		CertificateProductionStatus => false
		CertificateSecurityMode => false
		DeviceEnclosureColor => unknown
		SecurityDomain => 1
		OSInstallEnvironment => false
		Image4Supported => false
	}
	Services:
		com.apple.sysdiagnose.remote
		com.apple.mobile.storage_mounter_proxy.bridge
		com.apple.osanalytics.logRelay
		com.apple.corecaptured.remoteservice
		com.apple.mobile.storage_mounter_proxy.bridge.macOS

Found localbridge (bridge)
	State: connected (connectable)
	UUID: 37D16267-478D-4B4D-**REDACTED**
	Product Type: iBridge2,1
	OS Build: 5.5 (18P4759a)
	Messaging Protocol Version: 2
[...]

PR #769, opened 2021-11-01
Add remotectl dataflatten table

FritzX6:
## What is the purpose of this PR?
To add a new exec dataflatten table based on `/usr/libexec/remotectl`


pkg/osquery/tables/remotectl/remotectl.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (t *Table) flattenOutput(dataQuery string, systemOutput []byte) ([]dataflatten.Row, error) {
+	// convert mangled output
+	converted, err := t.transformOutput(systemOutput)
+	if err != nil {
+		level.Info(t.logger).Log("msg", "converting remotectl output", "err", err)
+		return nil, errors.Wrap(err, "converting")
+	}
+	fmt.Println(string(converted))
```
lhmzhou, line 85: Should this be a log instead of standard print?

directionless:
This would have fixed #763, but I think we can do a lot better with the parser

PR #775, opened 2021-11-23
Add `launcher_gc_info` table

directionless:
This adds a table that exposes launcher's GC runs. I don't believe this
is sensitive, and it should help some remote debugging


blaedj (approved):
lgtm

PR #776, opened 2021-11-23
Disable CGO on windows

directionless:
I have a hunch cgo on windows is causing _big_ issues. Disable it.

### DIFF
--- a/cmd/make/make.go
+++ b/cmd/make/make.go
@@ -75,15 +75,23 @@ func main() {
 		opts = append(opts, make.WithGithubActionOutput())
 	}
 
-	optsWithCgo := append(opts, make.WithCgo())
+	// We need to avoid cgo on windows. See
+	// https://github.com/golang/go/issues/22439 which still
+	// appears current, at least with cgo. But, we need to use cgo
+	// on linux (for fscrypt)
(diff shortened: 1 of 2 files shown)


blaedj (approved):

cmd/make/make.go
```diff
@@ -75,15 +75,23 @@ func main() {
 		opts = append(opts, make.WithGithubActionOutput())
 	}
 
-	optsWithCgo := append(opts, make.WithCgo())
+	// We need to avoid cgo on windows. See
+	// https://github.com/golang/go/issues/22439 which still
```
blaedj, line 79: This issue was closed > 4 years ago, did you run the repro steps from the issue with a cgo component, or what made you think the issue still relevant/unfixed?

directionless: I'll flesh in the words in this PR (right now, notes are in a private Kolide ticket) but the fix that issue discusses was only applied to not-CGO. eg: cgo still uses the larger allocations. I think there probably some _other_ issue lurking, but this feels like add on effect.

blaedj: ok gotcha, thanks for clarifying!

PR #780, opened 2021-12-14
Simple table over `airport` command

directionless:
This uses the `airport --xml` to display some debugging information and parse it into an EAV.

One issue here -- the xml output seems crashy. On my test machines it will often stop working mid stream, leaving me with unparsable junk


directionless:
Replaced by https://github.com/kolide/launcher/pull/814

PR #782, opened 2021-12-28
Add support for `uint8` to dataflatten

directionless:
We've now seen `uint8` in flatten input.

### DIFF
--- a/pkg/dataflatten/flatten.go
+++ b/pkg/dataflatten/flatten.go
@@ -462,6 +462,8 @@ func stringify(data interface{}) (string, error) {
 		return base64.StdEncoding.EncodeToString(v), nil
 	case uint8:
 		return strconv.FormatUint(uint64(v), 10), nil
+	case uint16:
+		return strconv.FormatUint(uint64(v), 10), nil
 	case uint32:
 		return strconv.FormatUint(uint64(v), 10), nil
 	case uint64:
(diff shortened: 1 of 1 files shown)


pkg/dataflatten/flatten.go
```diff
@@ -460,6 +460,8 @@ func stringify(data interface{}) (string, error) {
 			return s, nil
 		}
 		return base64.StdEncoding.EncodeToString(v), nil
+	case uint8:
+		return strconv.FormatUint(uint64(v), 10), nil
```
FritzX6, line 464: We also need to add casing for `uint16` (See several properties from [Win32_LogicalDisk](https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-logicaldisk))

```
	case uint16:
		return strconv.FormatUint(uint64(v), 10), nil
```

FritzX6 (changes requested):

pkg/dataflatten/flatten.go
directionless: Good call.

blaedj (approved):
LGTM - the example query[^1] returns nothing before this PR, but returns data in this branch.

[^1]: example query: `WITH
wmi_raw AS (
  SELECT * FROM kolide_wmi 
  WHERE class = 'Win32_UserAccount'
-- Remove SIDType from properties to return data successfully
  AND properties = "SID,SIDType"
  ),
wmi_pivot AS (
  SELECT
  MAX(CASE WHEN key = 'SID' THEN value END) AS SID,
  MAX(CASE WHEN key = 'SIDType' THEN value END) AS SIDType
  FROM wmi_raw
  GROUP BY parent
  )
  SELECT * FROM wmi_pivot;`

FritzX6 (approved):
lgtm

PR #783, opened 2022-01-05
File read tables should not error on bad files

directionless:
We have a variety of file reading tables that we use to parse configuration
data. Sometimes, we encounter badly formatted data. Instead of hard erroring
out, we should instead return what we can.


blaedj (approved):
lgtm, behaves as expected.

PR #784, opened 2022-01-06
Create a local log, to a known path, with regular log messages

directionless:
## Problem

When debugging with launcher users, it is often very hard to get the appropriate logs. This is because they logs go to different places on different platforms. (And on windows, they're especially weird)

## Solution

This replaces the local debug log file option (from #639) with an always-enabled, rolling, log file. The initial implementation suffered from a couple of issues. First, it required being explicitly enabled. Second, it didn't rotate. 

This also adds a periodic log message to aid in finding what host the logs are from. This checkpoint message is an initial PoC, and is expected to evolve over time.


blaedj (approved):
lgtm!

PR #786, opened 2022-01-11
Add support for the debug server to windows

directionless:
## Problem

Launcher has always shipped a debug server. This is enabled by sending `SIGUSR1` to the launcher process. Unfortunately, windows does not support signals, so this has never been enabled on windows. 

## Solution

Use a filesystem based switch instead. To be honest, I don't love polling the filesystem, but from what I can tell, it's that or create a socket for RPC communication. 

I probably need to test this a little bit more.


pkg/osquery/table/table.go
```diff
@@ -43,7 +43,8 @@ func PlatformTables(client *osquery.ExtensionManagerClient, logger log.Logger, c
 			"kolide_zerotier_networks", dataflattentable.JsonType, zerotierCli("listnetworks")),
 		dataflattentable.TablePluginExec(client, logger,
 			"kolide_zerotier_peers", dataflattentable.JsonType, zerotierCli("listpeers")),
-		tdebug.LauncherGcInfo(client, logger),
+		tdebug.LauncherGcStats(client, logger),
+		tdebug.LauncherMemStats(client, logger),
```
lhmzhou, line 47: Does the table `launcher_mem_stats` exist?

PR #792, opened 2022-02-10
Add p12 parsing to `pkg/cryptoinfo`

directionless:
This brings in p12 parsing to `pkg/cryptoinfo`. 

In doing this, I discovered that a bunch of ssl formats are ASN, and thus it is not possible to identify these files by their leading bytes. Instead, this moves to a simply attempting to parse, and moving on if it's wrong.

Note that this _does_ expose a `passphase` option. One should be cautious using it, as it may cause the passphrase to be logged.


pkg/cryptoinfo/identify.go
```diff
@@ -3,31 +3,29 @@
...
 	}
-}
 
-func expandDer(data []byte) *KeyInfo {
-	ki := NewKeyInfo(kiDER, kiCertificate, nil)
-	ki.SetDataName("certificate")
-	ki.SetData(parseCertificate(data))
-	return ki
+	// If we can't parse anything, return nothing. It's not a fatal error, and it's
+	// somewhart obvious from context that nothing was parsed.
```
blaedj, line 29:
```suggestion
	// somewhat obvious from context that nothing was parsed.
```

pkg/cryptoinfo/parse_certificate.go
```diff
@@ -43,12 +42,15 @@ type certExtract struct {
 // parseCertificate parses a certificate from a stream of bytes. We use this, instead of a bare x509.ParseCertificate, to handle some
 // string conversions, and bitfield enumerations.
 func parseCertificate(certBytes []byte) (interface{}, error) {
-
 	c, err := x509.ParseCertificate(certBytes)
 	if err != nil {
-		return nil, errors.Wrap(err, "parsing cert")
+		return nil, fmt.Errorf("parsing certificate: %w", err)
```
blaedj, line 47: why `fmt.Errorf(..)` and not `errors.Wrapf(..)`?

pkg/cryptoinfo/info.go
```diff
@@ -13,18 +13,52 @@ type KeyInfo struct {
...
+	kiCACERTIFICATE = "CA-CERTIFICATE" // What is correct here?
+	kiCaCertificate = "certificate"
+	kiCERTIFICATE   = "CERTIFICATE"
+	kiCertificate   = "certificate"
+	kiKEY           = "KEY"
+	kiKey           = "key"
 )
 
-func NewKeyInfo(typ, encoding string, headers map[string]string) *KeyInfo {
+func NewKIKey(encoding string) *KeyInfo {
```
blaedj, line 28: May I suggest a naming scheme like: `NewKeyKeyInfo/NewCertificateKeyInfo` etc?

Or, we could probably forgo the Hungarian-ish notation, since the return types of the funcs are probably documentation enough. Then we could just name these `NewKey` `NewCertificate`, `NewCaCertificate`, `NewError` etc. 

I think I like the latter option better actually..

pkg/cryptoinfo/info.go
```diff
@@ -13,18 +13,52 @@ type KeyInfo struct {
...
+	kiPEM = "PEM"
+	kiDER = "DER"
+	kiP12 = "P12"
+
+	kiCACERTIFICATE = "CA-CERTIFICATE" // What is correct here?
+	kiCaCertificate = "certificate"
+	kiCERTIFICATE   = "CERTIFICATE"
+	kiCertificate   = "certificate"
+	kiKEY           = "KEY"
+	kiKey           = "key"
```
blaedj, line 25: hm. these variables are.. interesting. maybe fine.

Interesting, because the names of the vars (even when coupled with the values) aren't super informative. Maybe making them types would improve the situation? Not sure.

pkg/cryptoinfo/identify.go
```diff
@@ -3,31 +3,29 @@
 // with dataflatten, and may eventually it may replace pkg/keyidentifier
 package cryptoinfo
 
-import (
-	"bytes"
-)
+type identifierSigfunc func(data []byte, password string) (results []*KeyInfo, err error)
```
blaedj, line 6: What do you think about naming this something like `keyIdentifier`/`cryptoIdentifier`/`keyinfoIdentifier`?

including `func` in the type name feels a bit redundant.

pkg/cryptoinfo/identify.go
```diff
@@ -3,31 +3,29 @@
...
 
 // Identify examines a []byte and attempts to descern what
 // cryptographic material is contained within.
-func Identify(data []byte) ([]*KeyInfo, error) {
-	switch {
-	case bytes.HasPrefix(data, certificateLeadingBytes):
-		return []*KeyInfo{expandDer(data)}, nil
-	default:
-		return decodePem(data)
+func Identify(data []byte, password string) ([]*KeyInfo, error) {
```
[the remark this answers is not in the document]
directionless: Ha. That's true. It's turtles all the way down

pkg/cryptoinfo/identify.go
directionless: I don't think it's worth it. This is an internal type, used only to declare that array. I think the _correct_ way is to use structs and interfaces, and think harder about package boundaries. But I also think that's premature right now. So this was easy enough. 

I guess "Sig" and "func" are kinda redundant. I could change it to `identifierSignature` or something

pkg/cryptoinfo/info.go
directionless: They feel like an uneasy compromise to me. They're not really needed -- I could just put them into the various `New` routines, but it felt nice to use constants instead.

I could convert them to types, it would provide the desired type safety... Hrm

pkg/cryptoinfo/info.go
directionless: Given the package scope, I think your latter suggestion is pretty good. Yeah.

pkg/cryptoinfo/info.go
```diff
@@ -13,18 +13,52 @@ type KeyInfo struct {
 
 // Maybe make these types?
 const (
-	kiPEM         = "PEM"
-	kiDER         = "DER"
-	kiCertificate = "CERTIFICATE"
+	kiPEM = "PEM"
```
levicole, line 16: In Rust, it's common for this to be an `enum`, and then use that enum to pattern match against. Is that essentially what you're using this for?

blaedj (approved):

PR #796, opened 2022-03-02
added notable files log to checkpoint log, added unit tests

James-Pickett:
* Added checkpoint log for "notable" files
* Broke existing CheckPoint code up into functions and added unit tests


pkg/log/checkpoint/checkpoint_test.go
```diff
@@ -0,0 +1,31 @@
...
+	mock.Mock
+}
+
+func (m *mockLogger) Log(keyvals ...interface{}) error {
+	m.Called(keyvals...)
+	return nil
+}
+
+func TestLogCheckPoint(t *testing.T) {
+	mockLogger := new(mockLogger)
```
directionless, line 19:
```suggestion
func TestLogCheckPoint(t *testing.T) {
         t.Parallel()

         mockLogger := new(mockLogger)
```

pkg/log/checkpoint/files.go
```diff
@@ -0,0 +1,32 @@
...
+				results = append(results, fmt.Sprintf("%s/%s", dirname, file.Name()))
+			}
+		}
+	}
+
+	return results
+}
+
+// helper method to generate empty dir message, makes testing easier
+func emptyDirMsg(dirname string) string {
```
directionless, line 30:
```suggestion
// emptyDirMsg is a helper method to generate empty dir message, makes testing easier
func emptyDirMsg(dirname string) string {
```

go doc style

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/kolide/kit/ulid"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFilesFound(t *testing.T) {
+	t.Parallel()
+
+	tempDir, err := os.MkdirTemp("", "log-checkpoint-files-test")
+	require.NoError(t, err, "making temp dir")
```
directionless, line 20:
```suggestion
	tempDir, err := os.MkdirTemp("", "log-checkpoint-files-test")
	require.NoError(t, err, "making temp dir")
	defer os.RemoveAll(tempDir)
```

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+			require.NoError(t, err, "creating test files")
+
+			foundPaths := fileNamesInDirs(dirs...)
+			sort.Strings(foundPaths)
+			require.Equal(t, expectedPaths, foundPaths)
+			require.Equal(t, tt.dirsToCreate*tt.filesPerDir, len(foundPaths))
+		})
+	}
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
```
directionless, line 42: I think better to do this as a defer above. Doing this here means it won't trigger if a test fails. Also, if os.RemoveAll fails, no point in failing the testing.

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func TestDirNotFound(t *testing.T) {
+	t.Parallel()
+
+	tempDir, err := os.MkdirTemp("", "log-checkpoint-files-test")
+	require.NoError(t, err, "making temp dir")
```
directionless, line 49:
```suggestion
	tempDir, err := os.MkdirTemp("", "log-checkpoint-files-test")
	require.NoError(t, err, "making temp dir")
	defer os.RemoveAll(tempDir)
```

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		expectedOutput = append(expectedOutput, pathErr.Error())
+	}
+
+	foundPaths := fileNamesInDirs(nonExistantDirs...)
+
+	require.Equal(t, expectedOutput, foundPaths)
+	require.Equal(t, len(nonExistantDirs), len(foundPaths))
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
```
directionless, line 80: As above

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestDirNotFound(t *testing.T) {
+	t.Parallel()
+
+	tempDir, err := os.MkdirTemp("", "log-checkpoint-files-test")
+	require.NoError(t, err, "making temp dir")
+
+	const pathFmt = "%s/%s"
+	nonExistantDirs := []string{
+		fmt.Sprintf(pathFmt, tempDir, ulid.New()),
+		fmt.Sprintf(pathFmt, tempDir, ulid.New()),
```
directionless, line 54: I don't think this is guaranteed to be cross platform. I'd use `filepath.Join(tempDir, ulid.New())`

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+	require.Equal(t, len(dirs), len(foundPaths))
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
+}
+
+func createTestFiles(baseDir string, dirCount int, filesPerDir int) (dirs []string, files []string, err error) {
+	files = []string{}
+
+	for i := 0; i < dirCount; i++ {
+		dir, err := os.MkdirTemp(baseDir, "notable-files-unit-test")
```
directionless, line 109: If dirCount is > 1, won't they step on eachother?

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+			sort.Strings(foundPaths)
+			require.Equal(t, expectedPaths, foundPaths)
+			require.Equal(t, tt.dirsToCreate*tt.filesPerDir, len(foundPaths))
+		})
+	}
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
+}
+
+func TestDirNotFound(t *testing.T) {
```
directionless, line 45: Is this the same case as `dirsToCreate: 0` above?

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	foundPaths := fileNamesInDirs(nonExistantDirs...)
+
+	require.Equal(t, expectedOutput, foundPaths)
+	require.Equal(t, len(nonExistantDirs), len(foundPaths))
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
+}
+
+func TestDirEmpty(t *testing.T) {
```
directionless, line 83: Is this the same case as `dirsToCreate: 2, filesPerDir: 0` above?

pkg/log/checkpoint/files_test.go
James-Pickett: `os.MkdirTemp()` adds random string at the end of the pattern (second arg)

pkg/log/checkpoint/files_test.go
James-Pickett: no, the test case where `dirsToCreate:0` is passing any empty array to the `fileNamesInDirs` function (it's not searching any dirs) where as this one is looking for a directory that does not exist. left a comment and updated variable name for clarity

pkg/log/checkpoint/files_test.go
James-Pickett: I dropped the `dirsToCreate:2, filesPerDir:0` case since now we are returning something when we search an empty dir, I could re add that case to this test method, but would have to add some logic to check for empty dirs and generate expect output ... felt cleaner to break out to it's own function

pkg/log/checkpoint/checkpoint_test.go
James-Pickett: updated

pkg/log/checkpoint/files.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
directionless: I totally forgot. Thanks!

pkg/log/checkpoint/files.go
```diff
@@ -0,0 +1,32 @@
...
+		files, err := os.ReadDir(dirname)
+
+		switch {
+		case err != nil:
+			results = append(results, err.Error())
+		case len(files) == 0:
+			results = append(results, emptyDirMsg(dirname))
+		default:
+			for _, file := range files {
+				results = append(results, fmt.Sprintf("%s/%s", dirname, file.Name()))
```
directionless, line 21: `filepath.Join()` though for a string log it's a pretty small nit.

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	for _, tt := range tests {
+		t.Run("testFilesFound", func(t *testing.T) {
+			dirsToSearch, expectedPaths, err := createTestFiles(tempDir, tt.dirsToCreate, tt.filesPerDir)
+			require.NoError(t, err, "creating test files")
+
+			foundPaths := fileNamesInDirs(dirsToSearch...)
+			sort.Strings(foundPaths)
+			require.Equal(t, expectedPaths, foundPaths)
+			require.Equal(t, tt.dirsToCreate*tt.filesPerDir, len(foundPaths))
```
directionless, line 41: This feels unneeded with the prior test?

pkg/log/checkpoint/files_test.go
```diff
@@ -0,0 +1,127 @@
...
+	for _, dir := range dirs {
+		expectedOutput = append(expectedOutput, emptyDirMsg(dir))
+	}
+
+	foundPaths := fileNamesInDirs(dirs...)
+
+	require.Equal(t, expectedOutput, foundPaths)
+	require.Equal(t, len(dirs), len(foundPaths))
+
+	require.NoError(t, os.RemoveAll(tempDir), "deleting temp dir")
```
directionless, line 101: Should be in the defer above

pkg/log/checkpoint/files.go
James-Pickett: updated

pkg/log/checkpoint/files_test.go
James-Pickett: updated

directionless (dismissed):

directionless (approved):

PR #797, opened 2022-03-02
Add support for direct osquery transport

directionless:
Add support for osquery communicating directly with a TLS server.

This sidesteps the `kolkide_grpc` plugin. But, because of how the underlying launcher runtime works, it creates a no-op version.


cmd/launcher/extension.go
```diff
@@ -97,25 +99,60 @@ func createExtensionRuntime(ctx context.Context, db *bbolt.DB, launcherClient se
...
 		runtime.WithStdout(osqueryStdoutLogger),
 		runtime.WithStderr(osqueryStderrLogger),
 		runtime.WithLogger(logger),
 		runtime.WithOsqueryVerbose(opts.OsqueryVerbose),
 		runtime.WithOsqueryFlags(opts.OsqueryFlags),
 		runtime.WithAugeasLensFunction(augeas.InstallLenses),
-	)
+	}
+
+	if opts.Transport == "osquery" {
```
James-Pickett, line 114: This function is already pretty large, can we start to break it up for readability? I think the the contents of of the individual if / else blocks here are a good placed to start.

cmd/launcher/internal/record_launcher_version_test.go
```diff
@@ -0,0 +1,64 @@
...
+		{additionalVerFiles: 1, includeMyVersion: true},
+		{additionalVerFiles: 5, includeMyVersion: true},
+		{additionalVerFiles: 5, includeMyVersion: true, unrelatedFiles: 3},
+	}
+
+	for _, tt := range tests {
+		t.Run("", func(t *testing.T) {
+			tempDir, err := os.MkdirTemp("", "record-launcher-version")
+			require.NoError(t, err, "making temp dir")
+			defer os.RemoveAll(tempDir)
```
James-Pickett, line 37: nice temp dir pattern

James-Pickett (changes requested):

cmd/launcher/extension.go
directionless: You're not wrong... As the leading comment says:
```
// TODO: the extension, runtime, and client are all kind of entangled
// here. Untangle the underlying libraries and separate into units
```

I can't quite decide if I want to bring cleanup into scope here. On the other hand, no time like the present.

directionless: Well... I tried splitting that out. I'm not sure. There's a lot of wiring that needs to be threaded through. I think this all mostly points at some much deeper refactoring in the future

cmd/launcher/internal/install_ca_certs.go
```diff
@@ -0,0 +1,35 @@
...
+	_, err := os.Stat(caFile)
+
+	switch {
+	case err == nil:
+		return caFile, nil
+	case !os.IsNotExist(err):
+		return "", err
+	}
+
+	return caFile, os.WriteFile(caFile, defaultCaCerts, 0444)
```
James-Pickett, line 33: found this guy to be confusing to read, how about this?

```suggestion
        if err != nil && !os.IsNotExist(err) {
            return "", err
        }
        
        return caFile, os.WriteFile(caFile, defaultCaCerts, 0444)        
```

James-Pickett (changes requested):

cmd/launcher/extension.go
```diff
@@ -83,39 +85,19 @@ func createExtensionRuntime(ctx context.Context, db *bbolt.DB, launcherClient se
...
-		runtime.WithStdout(osqueryStdoutLogger),
-		runtime.WithStderr(osqueryStderrLogger),
-		runtime.WithLogger(logger),
-		runtime.WithOsqueryVerbose(opts.OsqueryVerbose),
-		runtime.WithOsqueryFlags(opts.OsqueryFlags),
-		runtime.WithAugeasLensFunction(augeas.InstallLenses),
-	)
+	if opts.Transport == "osquery" {
+		var err error
+		runnerOptions, err = osqueryRunnerOptions(logger, db, opts)
```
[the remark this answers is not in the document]
directionless: Heh. I think this is one of the warts of golang. I don't like either option, and I don't know what idiomatic is. 

I have a slight preference for what I have, because it makes the scoping on `runnerOptions` more clear.

cmd/launcher/internal/install_ca_certs.go
directionless: Yeah, that's reasonable. I often try to bias towards the common case first. In this case, `err == nil` is going to be most often, so checking that first should be a slight performance win. But, this is infrequent code, no tight loop here. So readability is a better choice.

James-Pickett (approved):

blaedj (approved):
This looks good to me! I imagine that we'll need to test and iterate, I've started testing this out, but we should probably merge & iterate.

PR #800, opened 2022-03-07
added ip lookups, connectvity tests, & data fetches for required Kolide connectivity to checkpoint log

James-Pickett:
example checkpoint log (formatted for PR readability)
```json
{
   "bbolt db size":98304,
   "caller":"checkpoint.go:54",
   "msg":"log checkpoint",
   "ts":"2022-03-10T21:48:49.004493Z"
}{
   "caller":"checkpoint.go:57",
   "hostname":"Jamess-MBP",
   "msg":"log checkpoint",
   "ts":"2022-03-10T21:48:49.004527Z"
}{
   "caller":"checkpoint.go:58",
   "msg":"log checkpoint",
   "notableFiles":[
      "open /var/osquery: no such file or directory",
      "open /etc/osquery: no such file or directory"
   ],
   "ts":"2022-03-10T21:48:49.004568Z"
}{
   "caller":"checkpoint.go:104",
   "connections":{
      "dl.kolide.co:443":"successful tcp connection",
      "k2control.kolide.com:443":"successful tcp connection",
      "k2device.kolide.com:443":"successful tcp connection",
      "notary.kolide.co:443":"successful tcp connection"
   },
   "msg":"log checkpoint",
   "ts":"2022-03-10T21:48:49.338834Z"
}{
   "caller":"checkpoint.go:115",
   "ip loook ups":{
      "dl.kolide.co":[
         "35.186.193.229"
      ],
      "k2control.kolide.com":[
         "54.161.241.46",
         "52.202.168.65",
         "18.205.222.128",
         "54.237.133.81"
      ],
      "k2device.kolide.com":[
         "18.205.222.128",
         "54.161.241.46",
         "52.202.168.65",
         "54.237.133.81"
      ],
      "notary.kolide.co":[
         "35.196.217.93"
      ]
   },
   "msg":"log checkpoint",
   "ts":"2022-03-10T21:48:49.342507Z"
}{
   "caller":"checkpoint.go:77",
   "kolide server version fetch":{
      "https://k2device.kolide.com:443/version":"200 OK 09eb6df071e432a3d8f9d07ea7949857ca6382da\n"
   },
   "msg":"log checkpoint",
   "ts":"2022-03-10T21:48:53.865082Z"
}{
   "caller":"checkpoint.go:92",
   "msg":"log checkpoint",
   "notary versions":{
      "https://notary.kolide.co:443/v2/kolide/launcher/_trust/tuf/targets/releases.json":"1525"
   },
   "ts":"2022-03-10T21:48:54.340494Z"
}
```


pkg/log/checkpoint/checkpoint.go
```diff
@@ -12,9 +14,20 @@ import (
...
+// log these hosts' IPs to check point
+var hostsToCheckConnectivity = []string{"k2device.kolide.com", "k2control.kolide.com", "notary.kolide.co", "dl.kolide.co"}
+
+// log get data from these urls
+var fetchUrls = []string{"https://k2.kolide.com/version"}
+
+// logger is an interface that allows mocking of logger
+type logger interface {
+	Log(keyvals ...interface{}) error
+}
```
directionless, line 26: Very proper :)

pkg/log/checkpoint/checkpoint.go
```diff
@@ -12,9 +14,20 @@ import (
 // if we find any files in these directories, log them to check point
 var notableFileDirs = []string{"/var/osquery", "/etc/osquery"}
 
+// log these hosts' IPs to check point
+var hostsToCheckConnectivity = []string{"k2device.kolide.com", "k2control.kolide.com", "notary.kolide.co", "dl.kolide.co"}
+
+// log get data from these urls
+var fetchUrls = []string{"https://k2.kolide.com/version"}
```
directionless, line 21:
```suggestion
var fetchUrls = []string{"https://k2device.kolide.com/version"}
```

pkg/log/checkpoint/connectivity.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+	"net"
+)
+
+type dialer interface {
+	Dial(network, address string) (net.Conn, error)
+}
+
+func testConnections(dialer dialer, hosts ...string) []string {
+	results := []string{}
```
directionless, line 13: I could roll with this. But, what about `results := map[string]string` and then let the eventual json marshaling collapse it?

pkg/log/checkpoint/connectivity_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"errors"
+	"net"
+	"reflect"
+	"testing"
+
+	"github.com/kolide/launcher/pkg/log/checkpoint/mocks"
+)
+
+func Test_testConnections(t *testing.T) {
+	type args struct {
```
directionless, line 13:
```suggestion
func TestTestConnections(t *testing.T) {
        t.Parallel()

	type args struct {
```

pkg/log/checkpoint/connectivity_test.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			if got := testConnections(tt.args.dialer, tt.args.hosts...); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("testConnections() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_testConnection(t *testing.T) {
```
directionless, line 81:
```suggestion
func Test_testConnection(t *testing.T) {
        t.Parallel()

```

directionless, line 81: I understand that this is a slightly different unit test, but it feels highly duplicative with above. Do you think it's worth keeping both?

pkg/log/checkpoint/dns.go
```diff
@@ -0,0 +1,51 @@
...
+	"strings"
+)
+
+// ipLookeruper is an interface to allow mocking of ip look ups
+type ipLookuper interface {
+	LookupIP(ctx context.Context, network, host string) ([]net.IP, error)
+}
+
+func lookupHostsIpv4s(ipLookuper ipLookuper, hosts ...string) []string {
+	results := []string{}
```
directionless, line 16: same result question as above

pkg/log/checkpoint/dns.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		results = append(results, sb.String())
+	}
+
+	return results
+}
+
+func lookupIpv4(ipLookuper ipLookuper, host string) ([]string, error) {
+	ips, err := ipLookuper.LookupIP(context.Background(), "ip", host)
```
directionless, line 37: Add a timeout to the context.

pkg/log/checkpoint/dns_test.go
```diff
@@ -0,0 +1,188 @@
...
+			onLookupIPReturns: []func() ([]net.IP, error){
+				func() ([]net.IP, error) {
+					return []net.IP{
+
+						net.ParseIP("192.0.0.0"),
+						net.ParseIP("2001:db8::68"),
+					}, nil
+				},
+			},
+			want: []string{"happy_path_single.com: 192.0.0.0"},
```
directionless, line 41: I always feel weird when we pick things outside of `example.com` but it seems okay in tests. (Less okay when we use `acme.com` in docs)

pkg/log/checkpoint/fetch.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+)
+
+type httpClient interface {
+	Get(url string) (resp *http.Response, err error)
+}
+
+func fetchFromUrls(client httpClient, urls []string) []string {
+
+	results := []string{}
```
directionless, line 18: Same question

pkg/log/checkpoint/fetch.go
```diff
@@ -0,0 +1,72 @@
...
+			sb.WriteString(response)
+		}
+
+		results = append(results, sb.String())
+	}
+
+	return results
+}
+
+type NotaryRelease struct {
```
directionless, line 37: Do we need to export this? (Maybe?)

pkg/log/checkpoint/fetch.go
```diff
@@ -0,0 +1,72 @@
...
+	Signed struct {
+		Version int `json:"version"`
+	} `json:"signed"`
+}
+
+func fetchFromUrl(client httpClient, url string) (string, error) {
+	response, err := client.Get(url)
```
directionless, line 44: We should have a timeout here

pkg/log/checkpoint/fetch_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"errors"
+	"io"
+	"net/http"
+	"reflect"
+	"testing"
+
+	"github.com/kolide/launcher/pkg/log/checkpoint/mocks"
+)
+
+func Test_fetchFromUrls(t *testing.T) {
```
directionless, line 14: parallel

pkg/log/checkpoint/fetch_test.go
```diff
@@ -0,0 +1,206 @@
...
+				tt.args.client.On("Get", url).Return(tt.onGetReturns[i]()) //nolint:bodyclose
+			}
+			if got := fetchFromUrls(tt.args.client, tt.args.urls); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("fetchFromUrls() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_fetchFromUrl(t *testing.T) {
```
directionless, line 95: parallel

directionless (commented):
Seems pretty close.

Small questions, requests:

I think there are a couple of place we should have timeouts. Since several of those come in via context, it might make sense to thread a context through to a parent caller (vs creating it in each of the children here)

The tests feel a bit redundant, testing both the singular and the plural case. I'd take it, but I could also imagine myself removing all the singular tests, and getting the same coverage. I'd go either way though.

In reading this, and thinking about the non-Kolide users, I realize I've mis-spec'ed this. This is pretty good for _us_, but I can imagine it's not what people who don't take to k2 would want. (obvious in retrospect). I think there are 2 reasonable paths, and we could do either in a followup. First would be to conditionalize whether we include this based on an option. Second would be to dynamically pull the endpoint URLs based on config.

pkg/log/checkpoint/connectivity.go
James-Pickett: All of the funcs that are called by `logCheckPoint(logger log.Logger)` return a `[]string`. Are you suggesting that we convert them all to `map[string]string` where the key is either the hostname, url, or filepath? I like that ... looks like this in log
```
"IPs":
  {
    "dl.kolide.co":"35.186.193.229",
    "k2control.kolide.com":"18.205.222.128, 54.237.133.81, 54.161.241.46, 52.202.168.65",
    "k2device.kolide.com":"54.161.241.46, 18.205.222.128, 54.237.133.81, 52.202.168.65",
    "notary.kolide.co":"35.196.217.93"
  }
```

pkg/log/checkpoint/fetch.go
James-Pickett: no we don't, will make it private

pkg/log/checkpoint/fetch.go
James-Pickett: the clients are configured with a time out here, https://github.com/James-Pickett/launcher/blob/checkpoint-log-connectivity/pkg/log/checkpoint/checkpoint.go#L53 it applies to all requests made by the client

pkg/log/checkpoint/connectivity_test.go
James-Pickett: updated

pkg/log/checkpoint/fetch.go
James-Pickett: updated

pkg/log/checkpoint/fetch_test.go
James-Pickett: updated

pkg/log/checkpoint/connectivity_test.go
James-Pickett: yea, I suppose they are pretty duplicative ... just kind of came out naturally, wrote the singular functions first then the plural, I'll drop the singular ones

pkg/log/checkpoint/fetch.go
James-Pickett: however, I did add a context with time out to the IP lookup, unlike the standard lib http client and dialer, the resolver used for IP look up is not configurable with a time out

pkg/log/checkpoint/connectivity.go
directionless: I'm not sure they all need to return the same thing. The underlying logger interface is `keyvals ...interface{}`, and I'm pretty sure that under the hood that's getting converted to json. I think for something like hostname, we should return `string`, but for something like this it feels much more like `map[string]interface{}`, which lets us stick whatever we want in there. 

Of course, I'm assuming logger will marshal it, but I think so. And it seems cleaner than hand rolling what looks like json

pkg/log/checkpoint/fetch.go
directionless: You're right! There is a timeout in the httpClient. I don't know why one might prefer one over the other, and I think this is fine!

pkg/log/checkpoint/dns_test.go
James-Pickett: that will work, I can start using that, thank you

pkg/log/checkpoint/connectivity.go
directionless: Hrm, really? https://go.dev/play/p/QiW45PhqbFy

pkg/log/checkpoint/checkpoint.go
```diff
@@ -47,3 +125,17 @@ func hostName() string {
 
 	return hostname
 }
+
+func logCheckPointError(logger log.Logger, err error) {
```
directionless, line 129: (nit) I don't find this function adds clarity

pkg/log/checkpoint/checkpoint.go
```diff
@@ -47,3 +125,17 @@ func hostName() string {
...
+	logger.Log("checkpoint error", err)
+}
+
+func parseUrl(addr string) (*url.URL, error) {
+	return url.Parse(fmt.Sprintf("https://%s", strings.Split(addr, ":")[0]))
+}
+
+func logMap(logger log.Logger, entry map[string]interface{}) {
```
directionless, line 137: I don't like what this does to the log output. The map should get passed as a map -- flattening it here causes different data to end up on the same keys. I don't think it's needed

pkg/log/checkpoint/connectivity.go
```diff
@@ -0,0 +1,33 @@
...
+		} else {
+			results[host] = "successful tcp connection over 443"
+		}
+	}
+
+	return results
+}
+
+func testConnection(dialer dialer, host string) error {
+	hostPort := net.JoinHostPort(host, "443")
```
directionless, line 26: This is not guaranteed, but maybe fine to roll with.

pkg/log/checkpoint/checkpoint.go
```diff
@@ -23,20 +38,83 @@ func Run(logger log.Logger, db *bbolt.DB) {
...
+		logCheckPointError(logger, errors.Wrap(err, "parsing URL"))
+	}
+
+	logMap(logger, testConnections(dialer, kolideServerUrl.Host))
+	logMap(logger, lookupHostsIpv4s(ipLookuper, kolideServerUrl.Host))
+
+	if opts.KolideHosted {
+		logMap(logger, fetchFromUrls(httpClient, kolideServerUrl.String()+"/version"))
+		logMap(logger, testConnections(dialer, "dl.kolide.co"))
+		logMap(logger, lookupHostsIpv4s(ipLookuper, "dl.kolide.co"))
```
directionless, line 78: These should be in the `if opt.Autoupdate` block, and should use `opts.MirrorURL`

pkg/log/checkpoint/checkpoint.go
```diff
@@ -23,20 +38,83 @@ func Run(logger log.Logger, db *bbolt.DB) {
...
+
+		logMap(logger, testConnections(dialer, notaryServerURL.Host))
+		logMap(logger, lookupHostsIpv4s(ipLookuper, notaryServerURL.Host))
+
+		if opts.KolideHosted {
+			logMap(logger, fetchNotaryVersion(httpClient, notaryServerURL.String()+"/v2/kolide/launcher/_trust/tuf/targets/releases.json"))
+		}
+	}
+
+	if opts.Autoupdate {
```
directionless, line 95: duplicate?

pkg/log/checkpoint/connectivity_test.go
```diff
@@ -0,0 +1,83 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for i, host := range tt.args.hosts {
+				tt.args.dialer.On("Dial", "tcp", net.JoinHostPort(host, "443")).Return(tt.onDialReturns[i]())
+			}
+
+			if got := testConnections(tt.args.dialer, tt.args.hosts...); !reflect.DeepEqual(got, tt.want) {
```
directionless, line 76: I'm used to `if ok`

pkg/log/checkpoint/checkpoint.go
```diff
@@ -23,20 +38,83 @@ func Run(logger log.Logger, db *bbolt.DB) {
...
 
-func logCheckPoint(logger log.Logger) {
-	logger.Log(
-		"msg", "log checkpoint started",
-		"hostname", hostName(),
-		"notableFiles", fileNamesInDirs(notableFileDirs...),
-	)
+func logCheckPoint(logger log.Logger, db *bbolt.DB, opts launcher.Options) {
+
+	logger.Log("msg", "log checkpoint started")
```
directionless, line 51:
```suggestion
	logger = log.With(logger, "msg", "log checkpoint")
```

maybe?

directionless (commented):
The shift from one message to many seems reasonable. If you want to go down that path, pass the logger to all the functions and log directly from them. 

 Though the `logMap`, I think, is getting in the way.

cmd/launcher/launcher.go
```diff
@@ -99,7 +99,7 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	internal.RecordLauncherVersion(rootDirectory)
 
 	// Try to ensure useful info in the logs
-	checkpoint.Run(logger, db)
+	checkpoint.Run(logger, db, *opts)
```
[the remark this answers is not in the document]
James-Pickett: I passed the value instead of a pointer because I didn't think the checkpoint.Run func should have access to edit the options used by the rest of the application. I think the only argument for passing a pointer would be to save some memory? So I guess the questions is ... is the memory we save using a pointer worth giving unnecessary access to app wide settings? .... there is also the possibility that I could be completely missing something =)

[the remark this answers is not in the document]
James-Pickett: I guess another question could be ... is there ever a case where another part of the application edits the options and we would expect that to propagate to the checkpoint log? If so, we would need a pointer.

pkg/log/checkpoint/checkpoint.go
James-Pickett: dropped the function

pkg/log/checkpoint/connectivity.go
James-Pickett: yea, I wrote this part before we started parsing the options, I can see now in the opts we are provided a port, I can update this to accept a port as well

pkg/log/checkpoint/checkpoint.go
James-Pickett: just to make sure I understand ....

the comment for opts.MirrorURL states `MirrorServerURL is the URL for the Notary mirror.`

therefore I can assume that "dl.kolide.co" is the notary mirror when running kolide_hosted?

directionless: `dl.kolide.co` is the default MirrorServerURL. It's often used even if not kolide_hosted. Historically, there's a bunch of people that use our TUF and Mirror servers for their onprem things.

pkg/log/checkpoint/connectivity.go
James-Pickett: Okay, I see what I did here, I was passing just the map with out a key `logger.Log(someMap)` https://go.dev/play/p/CIiuyljG52A and it was giving me the funky output ... but `logger.Log("someKey", someMap)` works as expected.

pkg/log/checkpoint/connectivity_test.go
James-Pickett: this is auto generated using the via one of the packages imported along with the vs code golang extension, I can update if you feel strongly

pkg/log/checkpoint/checkpoint.go
James-Pickett: did this!

cmd/launcher/options.go
```diff
@@ -189,6 +189,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		EnrollSecretPath:                   *flEnrollSecretPath,
 		InsecureTLS:                        *flInsecureTLS,
 		InsecureTransport:                  *flInsecureTransport,
+		KolideHosted:                       *flKolideHosted,
```
James-Pickett, line 192: @directionless , do we need to update the installer to make sure this flag is set to true for Kolide customers?

directionless: Not Yet :)

My plan is that once some of the stuff goes out of beta, match against the server URL and auto set it

directionless (commented):
I think this is pretty close, and I kinda want to just merge it. But...

pkg/log/checkpoint/checkpoint.go
```diff
@@ -2,41 +2,152 @@ package checkpoint
...
+			err = urlErr
+
+		// not first error
+		case urlErr != nil && err != nil:
+			err = errors.Wrap(err, urlErr.Error())
+
+		// no error
+		default:
+			urls = append(urls, url)
+		}
```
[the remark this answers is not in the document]
James-Pickett: opted to just skip URL

directionless (approved):
Let's try!

PR #803, opened 2022-03-14
Make all tests parallel where possible, added linter to look for parallel tests

James-Pickett:


pkg/log/checkpoint/files_test.go
```diff
@@ -31,6 +31,8 @@ func TestFilesFound(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run("testFilesFound", func(t *testing.T) {
+			t.Parallel()
```
directionless, line 34: Do we need to re-init tt here? We probably should, so we're consistent

pkg/make/builder_test.go
```diff
@@ -53,6 +53,8 @@ func TestNamingHelpers(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run("platform="+tt.platform, func(t *testing.T) {
+			t.Parallel()
```
directionless, line 56: re-init tt?

pkg/osquery/tables/munki/munki_test.go
```diff
@@ -46,6 +46,8 @@ func TestGenerateMunkiReport(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
directionless, line 49: re-init?

pkg/simulator/hosts_test.go
```diff
@@ -164,6 +173,8 @@ func TestRunQuery(t *testing.T) {
 
 	for _, tt := range testCases {
 		t.Run("", func(t *testing.T) {
+			t.Parallel()
```
directionless, line 176: re-init?

pkg/log/checkpoint/files_test.go
James-Pickett: This is a good one, the lack of re-init made the test give a false pass.

The all above to `defer os.RemoveAll(tempDir)` is being executed before/during the parallel subtests and was deleting them temp directory ... which makes sense. 

Discovered a nifty little `t.Cleanup(func ())` built into the testing package. 

"Cleanup registers a function to be called when the test and all its subtests complete. Cleanup functions will be called in last added, first called order."

```
	t.Cleanup(func() {
		os.RemoveAll(tempDir)
	})
```

directionless (approved):
Let's try! More parallel is more better!

PR #804, opened 2022-03-15
expand kolide_launcher_info table to include launcher identifier, remove kolide_launcher_identifer table

James-Pickett:


pkg/osquery/table/launcher_info.go
```diff
@@ -18,14 +20,22 @@ func LauncherInfoTable() *table.Plugin {
...
-	return table.NewPlugin("kolide_launcher_info", columns, generateLauncherInfoTable())
+	return table.NewPlugin("kolide_launcher_info", columns, generateLauncherInfoTable(db))
 }
 
-func generateLauncherInfoTable() table.GenerateFunc {
+func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
+
 	return func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+		identifier, err := osquery.IdentifierFromDB(db)
+
```
directionless, line 32:
```suggestion
```

directionless (commented):
I like this, I think it's cleaner than what we had prior. But the CI failure highlights that there was something load bearing there. 

Maybe we need a `kolide_tables_version` table in the place of the old. Or to think about doing that test another way

pkg/osquery/table/launcher_identifier.go
```diff
@@ -1,32 +0,0 @@
-package table
```
blaedj, line 1: This table has been around for a while, do we want to consider a deprecation cycle?

directionless: I don't think we need it. This wouldn't remove it from launcher, just from the standalone tables. And I don't think we've official supported that.

pkg/osquery/table/launcher_info.go
directionless: `kolide_best_practices` doesn't feel like a good choice -- it brings in a lot of weird sub checks, likely filesystem access, and is on the chopping block for future deprecation.

James-Pickett: @directionless , yep good point ... it failed the CI due to not being able to find a dependent table.

How about kolide_ssh_keys?

pkg/osquery/runtime/runtime_test.go
```diff
@@ -372,7 +372,7 @@ func TestExtensionSocketPath(t *testing.T) {
 	require.NoError(t, err)
 	defer client.Close()
 
-	resp, err := client.Query("select * from kolide_launcher_info")
+	resp, err := client.Query("select * from kolide_ssh_keys")
```
directionless, line 375: `launcher_gc_info` is probably a better choice.

directionless (approved):
Yay! It's a usage shift, but I think this makes more sense. Approved pending CI

PR #805, opened 2022-03-18
Updates around socket handling

directionless:
This has several small things to try to fix a socket issue I've been started seeing. 

1. Upgrades osquery-go, to upgrade thrift. This is a bit of a long shot, but seems reasonable
2. Sets better default timeouts to the optional tools we build (Not quite relevant, but seems reasonable in debugging)
3. Improve logging on the code paths that are likely in play
4. Add some simple retry logic to the health check code path

### DIFF
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -849,6 +849,8 @@ func (r *Runner) launchOsqueryInstance() error {
 						failed = false
 						break
 					}
+
+					time.Sleep(1 * time.Second)
 				}
 				if failed {
 					return errors.Wrap(err, "health check failed")
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -833,8 +833,24 @@ func (r *Runner) launchOsqueryInstance() error {
 			case <-o.doneCtx.Done():
 				return o.doneCtx.Err()
 			case <-ticker.C:
-				if err := o.Healthy(); err != nil {
-					level.Info(o.logger).Log("msg", "Health check failed", "err", err)
+				// Health check! Allow a couple
+				// failures before we tear everything
+				// down. This is pretty simple, it
+				// hardcodes the timing. Might be
+				// better for a Limiter
+				failed := false
+				for i := 0; i <= 5; i++ {
+					if err := o.Healthy(); err != nil {
+						level.Info(o.logger).Log("msg", "Health check failed", "attempt", i, "err", err)
+						failed = true
+
+					}
+
+					// err was nil, clear failed
+					failed = false
+					break
+				}
+				if failed {
 					return errors.Wrap(err, "health check failed")
 				}
 			}
--- a/pkg/osquery/runtime/runtime.go
+++ b/pkg/osquery/runtime/runtime.go
@@ -744,11 +744,12 @@ func (r *Runner) launchOsqueryInstance() error {
 	// Kill osquery process on shutdown
 	o.errgroup.Go(func() error {
 		<-o.doneCtx.Done()
+		level.Debug(o.logger).Log("msg", "Starting osquery shutdown")
 		if o.cmd.Process != nil {
 			// kill osqueryd and children
 			if err := killProcessGroup(o.cmd); err != nil {
 				if strings.Contains(err.Error(), "process already finished") || strings.Contains(err.Error(), "no such process") {
-					level.Debug(o.logger).Log("process already gone")
+					level.Debug(o.logger).Log("msg", "tried to stop osquery, but process already gone")
 				} else {
 					level.Info(o.logger).Log("msg", "killing osquery process", "err", err)
 				}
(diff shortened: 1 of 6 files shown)


pkg/osquery/runtime/runtime.go
```diff
@@ -831,7 +833,24 @@ func (r *Runner) launchOsqueryInstance() error {
...
+				failed := false
+				for i := 0; i <= 5; i++ {
+					if err := o.Healthy(); err != nil {
+						level.Info(o.logger).Log("msg", "Health check failed", "attempt", i, "err", err)
+						failed = true
+
+					}
+
+					// err was nil, clear failed
+					failed = false
```
James-Pickett, line 850: I think this is going to be called in all cases, think you need
```suggestion
if err == nil {
  failed = false
}
					
```

James-Pickett: or maybe ...
```suggestion
maxHealthChecks := 5
for i := 1; i <= maxHealthChecks; i++ {
	if err := o.Healthy(); err != nil {
		level.Info(o.logger).Log("msg", "Health check failed", "attempt", i, "err", err)
		
		if i == maxHealthChecks {
			return errors.Wrap(err, "health check failed")
		}
		
	} else {
		break
	}
}
```

directionless: You're right. I knew I wasn't getting that correct. How's this?

directionless: Adding in the sleep is ugly. I like your `maxHealthChecks`

directionless:
I think, there's no reason not to merge this and cut a proper beta. (Assuming I didn't make anymore silly errors in that error handling)

James-Pickett (approved):

pkg/osquery/runtime/runtime.go
```diff
@@ -831,8 +833,27 @@ func (r *Runner) launchOsqueryInstance() error {
...
+						level.Info(o.logger).Log("msg", "Health check failed", "attempt", i, "err", err)
+
+						if i == maxHealthChecks {
+							return errors.Wrap(err, "health check failed")
+						}
+
+						time.Sleep(1 * time.Second)
+
+					} else {
+						// err was nil, clear failed
```
James-Pickett, line 853: just a thought ... should we log health check success here? Just imagining future us looking at logs and seeing a line or 2 about a failed health check ... would be nice to know if it eventually passed.

directionless: Updated all the logging in this block. How's that look?

pkg/osquery/runtime/runtime.go
```diff
@@ -831,8 +833,31 @@ func (r *Runner) launchOsqueryInstance() error {
...
+							return errors.Wrap(err, "health check failed")
+						}
+
+						level.Info(o.logger).Log("msg", "Health check failed. Will retry", "attempt", i, "err", err)
+						time.Sleep(1 * time.Second)
+
+					} else {
+						// err was nil, clear failed
+						if i > 1 {
+							level.Debug(o.logger).Log("msg", "Health check passed. Clearing error", "attempt", i)
```
James-Pickett, line 855:
```suggestion
							level.Info(o.logger).Log("msg", "Health check passed. Clearing error", "attempt", i)
```

James-Pickett (changes requested):

pkg/osquery/runtime/runtime.go
James-Pickett: Looks good other than wondering if it should be Debug or Info level. I kind a feel it should be the same as the health check above ... I don't want to see one without the other ... also I think the final attempt should before erroring out should be level.Error.

pkg/osquery/runtime/runtime.go
```diff
@@ -831,8 +833,31 @@ func (r *Runner) launchOsqueryInstance() error {
...
+				// Health check! Allow a couple
+				// failures before we tear everything
+				// down. This is pretty simple, it
+				// hardcodes the timing. Might be
+				// better for a Limiter
+				maxHealthChecks := 5
+				for i := 1; i <= maxHealthChecks; i++ {
+					if err := o.Healthy(); err != nil {
+						if i == maxHealthChecks {
+							level.Info(o.logger).Log("msg", "Health check failed. Giving up", "attempt", i, "err", err)
```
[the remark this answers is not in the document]
directionless: While I sorta agree in theory, that's not how we use levels in the code base today. 

* `error` is pretty much unused
* `info` is messages that are meant to be displayed in normal usage (they are in logs)
* `debug` is meant for developers, and isn't usually displayed

I don't know that I agree with our current usage, but I'd rather not change them in this one place. So I think this is correct, and that we should revisit all of logging.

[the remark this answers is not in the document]
James-Pickett: that works for me, resolving

James-Pickett (approved):

PR #806, opened 2022-03-18
updated socket mode to include launcher tables, added kolide_launcher_osquery_instance_history table

James-Pickett:
* updated socket mode to include launcher tables
* add kolide_launcher_osquery_instance_history table


cmd/launcher/run_socket.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/kolide/kit/env"
 	"github.com/kolide/kit/fs"
 	"github.com/kolide/launcher/pkg/osquery/runtime"
+	ktable "github.com/kolide/launcher/pkg/osquery/table"
```
directionless, line 14: Why `ktable`? (I'd have picked launchertables, or something)

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,65 @@
...
+
+type History struct {
+	instances []*Instance
+}
+
+// GetHistory returns the last 10 instances of osquery started / restarted by launcher
+// each start / restart cycle is an entry
+func GetHistory() []Instance {
+	results := []Instance{}
+	for _, v := range currentHistory.instances {
```
directionless, line 18: I think there are two ways to reduce allocations -- one would be to use set `results` to `len(currentHistory.instances)` is. But I think better is to just bake `max_instances` into the slice lengths.

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,65 @@
+package osquery_instance_history
+
+import (
+	"errors"
+	"time"
+)
+
+const max_instances = 10
```
directionless, line 8: prefer camel case

pkg/osquery/table/launcher_info.go
```diff
@@ -36,15 +38,16 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
...
+				"branch":      version.Version().Branch,
+				"build_date":  version.Version().BuildDate,
+				"build_user":  version.Version().BuildUser,
+				"go_version":  runtime.Version(),
+				"goarch":      runtime.GOARCH,
+				"goos":        runtime.GOOS,
+				"revision":    version.Version().Revision,
+				"version":     version.Version().Version,
+				"identifier":  identifier,
+				"instance_id": osquery_instance_history.CurrentInstance().InstanceId,
```
directionless, line 50:
```suggestion
				"osquery_instance_id": osquery_instance_history.CurrentInstance().InstanceId,
```

I don't love how long my suggestion is, but I kinda want `osquery` in there, since it's kinda weird that we're returning osquery data on a launcher specific table. (But that's also the point)

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,65 @@
...
+	if h.instances == nil {
+		h.instances = []*Instance{instance}
+		return
+	}
+
+	if len(h.instances) >= max_instances {
+		h.instances = h.instances[1:]
+	}
+
+	h.instances = append(h.instances, instance)
```
directionless, line 60: I think there's a fencepost here

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (h *History) currentInstance() *Instance {
+	if h.instances != nil && len(h.instances) > 0 {
+		return h.instances[len(h.instances)-1]
+	}
+	return nil
+}
+
+func (h *History) setCurrentInstance(instance *Instance) {
```
directionless, line 50: I might quibble over `add` vs `set`. But :shrug: quibbles

pkg/osquery/runtime/osquery_instance_history/update.go
```diff
@@ -0,0 +1,35 @@
+package osquery_instance_history
+
+import "sync"
+
+var mutex sync.Mutex
+var currentHistory = &History{}
```
directionless, line 6: Does it make sense to put the mutex into History and move the locks there? Might not matter

directionless (commented):
Looks pretty straightforward! I really like how the history append style looks

I tend to prefer fewer files, but this isn't unreasonable. Still thinking about the `_`. But the functionality looks like it should be spot on.

pkg/osquery/runtime/osquery_instance_history/history.go
James-Pickett: switching to camel case ... the snake case I think is left over from C# / java in my brain where constants were usually named `LIKE_THIS`

pkg/osquery/table/launcher_info.go
James-Pickett: that makes sense, I will make the update, but I also wonder ... should we be putting the same data in two different tables ... could always just join on the most recent history table row?

pkg/osquery/runtime/osquery_instance_history/history.go
James-Pickett: yes, we should have a quibble session on naming before this goes out

pkg/osquery/runtime/osquery_instance_history/history.go
James-Pickett: I don't see it ... if `max_instances == 1` perhaps? .... I'm going to add some tests though and find out

pkg/osquery/runtime/osquery_instance_history/update.go
James-Pickett: it's cleaner to have the History handle the locks, I'll do that

pkg/osquery/runtime/osquery_instance_history/history.go
directionless: With the append after the truncate, I think it will stabilize at 1 over the max_instances.

James-Pickett: ahh yea, I think I have fixed it now

pkg/osquery/runtime/osquery_instance_history/update.go
James-Pickett: well I put the locks on the history struct, but still letting package exported funcs handling locking, got to confusing to manage the locks in the unexported functions

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (h *history) currentInstance() (*Instance, error) {
+	if h.instances != nil && len(h.instances) > 0 {
+		return h.instances[len(h.instances)-1], nil
+	}
+	return nil, NoCurrentInstanceError{}
+}
+
+func (h *history) nextInstance() error {
```
directionless, line 80: I'd probably name this sort of thing `new` not `next`

directionless (commented):
Some comments, though these may be superceded by the refactoring we were just talking about on slack

pkg/osquery/runtime/osquery_instance_history/history.go
```diff
@@ -0,0 +1,139 @@
...
+// GetHistory returns the last 10 instances of osquery started / restarted by launcher, each start / restart cycle is an entry
+func GetHistory() []Instance {
+	currentHistory.Lock()
+	defer currentHistory.Unlock()
+
+	results := make([]Instance, len(currentHistory.instances))
+	for i, v := range currentHistory.instances {
+		results[i] = *v
+	}
+	return results
```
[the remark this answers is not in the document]
James-Pickett: It's a question of access, I expect the values of these histories to be immutable after they are initially set (except for maybe errors).  By de-referencing we're creating a copy and returning that, any consumer of this package will not have the ability to edit the history, forcing them to use exported functions.

This does in fact come at the cost of extra allocations, but since we are maxing at 10, I feel it's worth it for proper encapsulation.

pkg/osquery/runtime/history/history.go
```diff
@@ -0,0 +1,106 @@
...
+	defer currentHistory.Unlock()
+
+	return currentHistory.newInstance()
+}
+
+func (h *History) newInstance() (*Instance, error) {
+	_, err := h.latestInstance()
+	if err != nil && !errors.Is(err, NoInstancesError{}) {
+		return nil, err
+	}
```
directionless, line 74: What's this for?

pkg/osquery/runtime/history/history.go
```diff
@@ -0,0 +1,106 @@
...
+	results := make([]Instance, len(currentHistory.instances))
+	for i, v := range currentHistory.instances {
+		results[i] = *v
+	}
+
+	return results, nil
+}
+
+// LatestInstance returns the latest osquery instance
+func LatestInstance() (Instance, error) {
```
directionless, line 43: Why return a pointer from `lastInstance`? If you return the object, than `LastestInstance` can just be a wrapper.

pkg/osquery/runtime/runtime.go
```diff
@@ -795,6 +809,11 @@ func (r *Runner) launchOsqueryInstance() error {
 		return errors.Wrap(err, "could not create an extension client")
 	}
 
+	err = o.stats.Connected(o)
+	if err != nil {
+		level.Info(o.logger).Log("msg", fmt.Sprint("osquery instance history error: ", err.Error()))
+	}
```
directionless, line 815:
```suggestion
	if err := o.stats.Connected(o); err != nil {
		level.Info(o.logger).Log("msg", "osquery instance history",  "error", err)
	}
```

A couple of nits... For something like this, I tend to favor merging the if and err assignment. I like the scoping. And for logging I tend to favor structured logging and not string manipulation,

pkg/osquery/runtime/runtime.go
```diff
@@ -28,12 +28,15 @@ import (
...
+
+	"github.com/kolide/launcher/pkg/osquery/runtime/history"
 )
 
 type Runner struct {
-	instance     *OsqueryInstance
-	instanceLock sync.RWMutex
-	shutdown     chan struct{}
+	instance        *OsqueryInstance
+	instanceHistory *history.History
```
directionless, line 37: Is this used?

pkg/osquery/tables/osquery_instance_history/osquery_instance_history.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return table.NewPlugin("kolide_launcher_osquery_instance_history", columns, generate())
+}
+
+func generate() table.GenerateFunc {
+	return func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+		results := []map[string]string{}
+
+		history, err := history.GetHistory()
+
```
directionless, line 28:
```suggestion
```

pkg/osquery/tables/osquery_instance_history/osquery_instance_history.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, v := range history {
+
+			errStr := ""
+			if v.Error != nil {
+				errStr = v.Error.Error()
+			}
```
directionless, line 38: Do you know if we need this? `error` may just marshal correctly.

directionless (commented):
This is feeling very close. Some comments, and I'll stare more next week

pkg/osquery/runtime/runtime.go
James-Pickett: updated

pkg/osquery/tables/osquery_instance_history/osquery_instance_history.go
James-Pickett: done

pkg/osquery/runtime/history/history.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c NoInstancesError) Error() string {
+	return "no osquery instances have been added to history"
+}
+
+// GetHistory returns the last 10 instances of osquery started / restarted by launcher, each start / restart cycle is an entry
+func GetHistory() ([]Instance, error) {
+	if currentHistory.instances == nil {
+		return nil, NoInstancesError{}
+	}
```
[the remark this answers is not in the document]
James-Pickett: on the fence here ... a lack of history.instances feels like an error state ... is there ever a situation where we would launcher and not expect a history?

however, it also feel legit to just return nil and expect the caller to handle that case

I could go either way as long as it's consistent with how we handle a call to `history.LatestInstance()` ... either both `history.LatestInstance()` & 'history.GetHistory()` return just nil when they don't exist or they both return an error.

go.mod
```diff
@@ -58,7 +58,8 @@ require (
 	github.com/sirupsen/logrus v1.4.0 // indirect
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	github.com/spf13/viper v1.2.1 // indirect
-	github.com/stretchr/testify v1.5.1
+	github.com/stretchr/objx v0.3.0 // indirect
+	github.com/stretchr/testify v1.7.1
```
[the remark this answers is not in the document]
James-Pickett: done

pkg/osquery/runtime/history/history.go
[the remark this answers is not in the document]
directionless: Eh. Let's leave it. Until we use it for awhile, it's hard to know what's right or wrong.,

pkg/osquery/runtime/history/history.go
directionless: I get why `LastestInstance` returns an object. But why does `lastInstance` return a pointer? In fact, why have `lastInstance` at all?

pkg/osquery/tables/osquery_instance_history/osquery_instance_history.go
```diff
@@ -0,0 +1,47 @@
...
+func generate() table.GenerateFunc {
+	return func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+		results := []map[string]string{}
+
+		history, err := history.GetHistory()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, v := range history {
```
directionless, line 32: Why `v`?

directionless (approved):
I would take this as-is (thus approval), but I do think there's some oddness in the way the oo style shakes out.

Specifically, I'm not sure we need both the public and the private methods. For example, `NewInstance` and LatestInstance` don't seem like they'll have private callers, so why duplicate things? Probably also implies the mutex locks should be on the inner private methods.

Anyhow, we can merge as-is, or take another pass at the method signatures

pkg/osquery/tables/osquery_instance_history/osquery_instance_history.go
James-Pickett: v for value

`for i, v ... for index,value`

however, I've updated it to `instance`

pkg/osquery/runtime/history/history.go
James-Pickett: You're right, it's redundant, removed.

directionless (approved):

PR #809, opened 2022-03-24
Add helper to run a function using setuid

directionless:
Some macOS APIs take different actions based on the `uid` of the caller. As such, we need to change our uid to use them. But, the normal mechanisms to change uid effect the _entire_ process.

Luckily, macOS provides a syscall to change permissions only on a single execution thread. And go has a `runtime.LockOsThread` that allows us to ensure our thread  is safe to modify. This is a simple wrapper bundling those together.

### DIFF
--- a/pkg/threadas/threadas.go
+++ b/pkg/threadas/threadas.go
@@ -1,10 +1,15 @@
-// Package threadsetuid uses some thread trickery to facilitate
+//go:build darwin
+// +build darwin
+
+// Package threadas uses some thread trickery to facilitate
 // dropping permissions on a single thread. This allows us to call
 // underlying darwin APIs which change behavior based on the userid.
 //
 // Note that this changes permission on a single thread. It cannot
 // change permission on something like osquery (communication is over
-// a socket), and may not do the right thing if you bring in an exec.
+// a socket), and may not do the right thing if you bring in an
+// exec. This also means that if you spawn a new thread, you'll get
+// the parent process permissions.
 //
 // This is based on ideas from:
 //   * https://github.com/golang/go/issues/14592
@@ -13,21 +18,20 @@
 //
 // There's future work in supporting other platforms. Linux may have
 // `syscall(SYS_setresuid, ...)` for this.
-package threadsetuid
+package threadas
 
-import(
-	"time"
-	"syscall"
+import (
 	"fmt"
 	"runtime"
+	"syscall"
+	"time"
 )
-
 
 // ThreadAs will run a function, in a "thread", after using setuid to
 // change permissions on that thread. It uses `LockOSThread` so the
 // thread terminates after with the function, this is to clean up from
 // the permissions change.
-func ThreadAs(fn func()  error, timeout time.Duration, uid uint32, gid uint32)  error {
+func ThreadAs(fn func() error, timeout time.Duration, uid uint32, gid uint32) error {
 	// As we only support calling the function once, so we can
 	// buffer a single size. Makes it a bit easier to
 	// sequence starting the child and our listener.
@@ -56,9 +60,9 @@ func ThreadAs(fn func()  error, timeout time.Duration, uid uint32, gid uint32)  
 	}()
 
 	select {
-	case err := <- errChan:
+	case err := <-errChan:
 		return err
-	case <- time.After(timeout):
+	case <-time.After(timeout):
 		return fmt.Errorf("Timeout after %s", timeout)
 	}
 }
@@ -69,15 +73,14 @@ func ThreadAs(fn func()  error, timeout time.Duration, uid uint32, gid uint32)  
 //
 // There is some discussion in the man page that this is not suitable
 // for security isolation. It's recommended you read and understand that.
-func pthread_setugid_np(uid uint32, gid uint32) (error) {
+func pthread_setugid_np(uid uint32, gid uint32) error {
 	// syscall.RawSyscall blocks, while Syscall does not. Since we
 	// don't think this blocks, we can use the _slightly_ more
 	// performant RawSyscall
 	_, _, errNo := syscall.RawSyscall(syscall.SYS_SETTID, uintptr(uid), uintptr(gid), 0)
 	if errNo != 0 {
 		return errNo
-	} 
+	}
 
 	return nil
 }
-
--- a/pkg/threadas/threadas.go
+++ b/pkg/threadas/threadas.go
@@ -28,7 +28,9 @@ import (
 )
 
 const (
-	// KAUTH_UID_NONE and KAUTH_GID_NONE are special values. See the man page
+	// KAUTH_UID_NONE and KAUTH_GID_NONE are special values. When
+	// pthread_setugid_np is called with them, it will revert to
+	// the process credentials.
 	KAUTH_UID_NONE = ^uint32(0) - 100
 	KAUTH_GID_NONE = ^uint32(0) - 100
 )
@@ -77,8 +79,9 @@ func ThreadAs(fn func() error, timeout time.Duration, uid uint32, gid uint32) er
 // which sets the per-thread userid and groupid. We use this, and not
 // syscall.Setuid, because the latter will impact _all_ threads.
 //
-// There is some discussion in the man page that this is not suitable
-// for security isolation. It's recommended you read and understand that.
+// This is not suitable for security isolation. Something in thread
+// can return to parent permissions be either calling with
+// KAUTH_UID_NONE, or simply spawning a new thread.
 func pthread_setugid_np(uid uint32, gid uint32) error {
 	// syscall.RawSyscall blocks, while Syscall does not. Since we
 	// don't think this blocks, we can use the _slightly_ more
--- a/pkg/threadas/threadas.go
+++ b/pkg/threadas/threadas.go
@@ -27,6 +27,12 @@ import (
 	"time"
 )
 
+const (
+	// KAUTH_UID_NONE and KAUTH_GID_NONE are special values. See the man page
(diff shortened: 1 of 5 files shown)


pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func TestThreadAs(t *testing.T) {
+	t.Parallel()
+
+	if syscall.Getuid() != 0 {
+		t.Skip("Skipping -- test requires root")
+	}
+
+	targetUid := 501
+	targetGid := 20
```
James-Pickett, line 38: Can you comment on where these numbers come from?

pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,152 @@
...
+	myUids := getUidData()
+	fnMyUidsEqual := myUids.GenerateTestFunc(t, "matches my user", assert.Equal)
+	fnMyUidsNotEqual := myUids.GenerateTestFunc(t, "does not matches my user", assert.NotEqual)
+
+	// Be somewhat thoughtful in how parallel works here -- using
+	// t.Parallel will spawn a new thread, which potentially
+	// undermines some of what we're testing. But we do want the
+	// top level to be parallel, so that we have more thread
+	// churn.
+	for i := 1; i < 100; i++ {
```
James-Pickett, line 53: what is the purpose of running 100x? Just to make sure we don't get some kind of deadlocks?

pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func TestTimeout(t *testing.T) {
+	t.Parallel()
+
+	fn := func() error {
+		time.Sleep(10 * time.Second)
+		return nil
+	}
+
+	require.Error(t, ThreadAs(fn, timeout, KAUTH_UID_NONE, KAUTH_GID_NONE))
```
James-Pickett, line 88: if you want to make sure you're getting the timeout error as opposed to _any_ error, you can use https://pkg.go.dev/github.com/stretchr/testify/assert#EqualError

pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	fn := func() error {
+		c := runtime.NumGoroutine()
+		require.Equal(t, startCount+1, c, "go routines should be one more than starting")
+		return nil
+	}
+
+	require.Error(t, ThreadAs(fn, timeout, KAUTH_UID_NONE, KAUTH_GID_NONE))
+
+	require.Equal(t, startCount, runtime.NumGoroutine(), "go routines should return to normal")
```
James-Pickett, line 105: very cool test!

pkg/threadas/threadas.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"fmt"
+	"runtime"
+	"syscall"
+	"time"
+)
+
+const (
+	// KAUTH_UID_NONE and KAUTH_GID_NONE are special values. See the man page
```
blaedj, line 31: this is `man pthread_setugid_np`, correct?

pkg/threadas/threadas.go
```diff
@@ -0,0 +1,92 @@
...
+	go func() {
+		// Calling LockOSThread, without a subsequent Unlock,
+		// will cause the thread to terminate when the
+		// goroutine does. This seems simpler than resetting
+		// the thread permissions.
+		//
+		// An alternate implementation, for darwin, would be
+		// to use `pthread_setugid_np(KAUTH_UID_NONE,
+		// KAUTH_GID_NONE)` to reset permissions
+		// afterwards. See the manpage, or
```
blaedj, line 55: this is `man pthread_setugid_np`, correct?

James-Pickett (changes requested):

pkg/threadas/threadas.go
directionless: Correct

pkg/threadas/threadas.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"fmt"
+	"runtime"
+	"syscall"
+	"time"
+)
+
+const (
+	// KAUTH_UID_NONE and KAUTH_GID_NONE are special values. See the man page
+	KAUTH_UID_NONE = ^uint32(0) - 100
```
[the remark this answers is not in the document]
James-Pickett: okay .... looked at the man page and it's a lot of info ... maybe leave a link to the docs?

pkg/threadas/threadas_test.go
directionless: And no weird cross thread pollution. I updated the comment

pkg/threadas/threadas.go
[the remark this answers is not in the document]
directionless: I was really torn about this. The underlying syscall documents these are magic constants to reset the thread permissions to the parent. I don't really want to use them, but it still somehow feels more correct to have them here.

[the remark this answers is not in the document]
directionless: Updated the comments

pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,192 @@
...
+func TestThreadNoPermission(t *testing.T) {
+	t.Parallel()
+
+	if syscall.Getuid() == 0 {
+		t.Skip("Skipping -- test requires not-root")
+	}
+
+	uids := getUidData()
+	fn := uids.GenerateTestFunc(t, "expected failure", assert.Equal)
+	require.ErrorIs(t, ThreadAs(fn, timeout, uint32(uids.Uid), uint32(uids.Gid)), &NoPermissionsError{}, "Fails when run as non-root")
```
James-Pickett, line 33: nice!

pkg/threadas/threadas_test.go
James-Pickett: I think you are using it correctly ... didn't think to upgrade testify to get access to `require.ErrorIs()` which is better than the `require.EqualError()` I suggested

James-Pickett (approved):

James-Pickett (changes requested):
woops, just noticed something .... dropping my approval, will comment shortly

pkg/threadas/threadas_test.go
directionless: The part I'm unsure is how to do the comparison with the custom errors. Errors are pointers, and I don't really understand how go handles comparing them. Since baking in some context here (like, the seconds in the timeout), go consider `Timeout{t: 1}` different than `Timeout{t: 2}`. But I don't think they should be, so I defined the `Timeout.Is` to use a type assertion.

But this all feels like I'm doing something slightly wrong.

pkg/threadas/threadas_test.go
```diff
@@ -0,0 +1,192 @@
...
+	if syscall.Getuid() != 0 {
+		t.Skip("Skipping -- test requires root")
+	}
+
+	fn := func() error {
+		time.Sleep(10 * time.Second)
+		return nil
+	}
+
+	require.ErrorIs(t, ThreadAs(fn, timeout, uint32(0), uint32(0)), &TimeoutError{})
```
James-Pickett, line 123: this should let you drop the additional Is method on the error
```suggestion
	require.ErrorIs(t, ThreadAs(fn, timeout, uint32(0), uint32(0)), &TimeoutError{t: timeout})
```

pkg/threadas/threadas_test.go
James-Pickett, line 33: this in conjunction with returning `fmt.Errorf()` from `ThreadAs()` will work and you can git rid of custom error
```suggestion
	require.ErrorIs(t, ThreadAs(fn, timeout, uint32(uids.Uid), uint32(uids.Gid)), syscall.Errno(1), "Fails when run as non-root")
```

pkg/threadas/threadas.go
```diff
@@ -0,0 +1,124 @@
...
+//
+// This is not suitable for security isolation. Something in thread
+// can return to parent permissions be either calling with
+// KAUTH_UID_NONE, or simply spawning a new thread.
+func pthread_setugid_np(uid uint32, gid uint32) error {
+	// syscall.RawSyscall blocks, while Syscall does not. Since we
+	// don't think this blocks, we can use the _slightly_ more
+	// performant RawSyscall
+	if _, _, errno := syscall.RawSyscall(syscall.SYS_SETTID, uintptr(uid), uintptr(gid), 0); errno != 0 {
+		return &NoPermissionsError{Errno: errno}
```
James-Pickett, line 120: can probably get away with just wrapping the syscall.Errno and can compare syscall.Errno in tests
```suggestion
		return fmt.Errorf("error changing permission: %w", errno)
```

James-Pickett (changes requested):
Maybe this is slightly more elegant error handling 🤷

James-Pickett (dismissed):

directionless:
We chatted a bit on slack, but I wanted to comment here as well...

I thought about doing something like @James-Pickett suggested -- not defining the `Is` method, and instead wrapping, or having the test also pass along the `timeout`. But that seemed like it wouldn't work well from a caller's perspective. The caller just wants to know if the error is a `TimeoutError`, and the length of the timeout feels wrong. So I see two approaches:
* the `Is` here, with the type assertion. This allows the ultimate error message to be `timeout after 3s`.
* The other would be to define a `TimeoutError` without the duration, and to call it as `fmt.Errorf("after %s: %w", timeout, TimeoutError{})`. Yielding a final error of `after 3s: timeout`

I thought the first of those was slightly cleaner, so use it. I suspect there's something I'm not understanding about the underlying idioms, and there might be a better third way. But we can always change it later

James-Pickett (approved):

PR #812, opened 2022-03-30
converted updater initial delay to select statement, updated updater logging to specify binary being updated

James-Pickett:


directionless (commented):
Looks like it's pretty close.  Nits about logging, and the realization this does need to be outside the `for` loop

cmd/launcher/updater.go
```diff
@@ -63,25 +64,34 @@ func createUpdater(
...
 			// see a simple way to ensure the updater is still running in
 			// the background.
 			for {
-				level.Info(config.Logger).Log("msg", "updater started")
+				// When launcher first starts, we'd like the
+				// server to start receiving data
+				// immediately. But, if updater is trying to
+				// run, this creates an awkward pause for restart.
+				// So, delay starting updates by an hour or two.
+				level.Debug(config.Logger).Log("msg", fmt.Sprintf("%s updater entering initial delay of %s", binaryName, config.InitialDelay))
```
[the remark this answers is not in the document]
lhmzhou: Hi @directionless: What does `log.With` do? I'm not familiar with that method.

cmd/launcher/updater.go
```diff
@@ -63,25 +64,34 @@ func createUpdater(
 	var stop func()
 	stopChan := make(chan bool)
 
+	binaryName := path.Base(binaryPath)
+
 	return &actor.Actor{
 		Execute: func() error {
```
lhmzhou, line 70: Are we missing an error check for this `Execute` call? 

```
if err != nil {
    log.Fatal(err)
}
```

cmd/launcher/updater.go
```diff
@@ -103,20 +113,21 @@ func createUpdater(
...
 			<-ctx.Done()
 
 			return nil
 		},
+
 		Interrupt: func(err error) {
-			level.Info(config.Logger).Log("msg", "updater interrupted", "err", err)
-			level.Debug(config.Logger).Log("msg", "updater interrupted", "err", err, "stack", fmt.Sprintf("%+v", err))
+			level.Info(config.Logger).Log("msg", fmt.Sprintf("%s updater interrupted", binaryName), "err", err)
+			level.Debug(config.Logger).Log("msg", fmt.Sprintf("%s updater interrupted", binaryName), "err", err, "stack", fmt.Sprintf("%+v", err))
```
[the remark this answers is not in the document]
lhmzhou: Agreed. Seems like we're over-logging here.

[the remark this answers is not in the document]
directionless: Actually... I'm using stack traces tonight. So I want to keep these now. 

Useless so much of the time, but occasionally quite handy

cmd/launcher/updater.go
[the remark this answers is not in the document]
directionless: This is using `go-kit/log` as a structured logging tool. (Not stdlib)

[`log.with`](https://pkg.go.dev/github.com/go-kit/log#With) creates a new logger with some set of key/value pairs baked in

cmd/launcher/updater.go
directionless: That should be handled by the caller.

lhmzhou: Gotcha!

cmd/launcher/updater/updater.go
```diff
@@ -0,0 +1,153 @@
...
+func (u *updaterCmd) execute() error {
+	// Failing to start the updater is not a fatal launcher
+	// error. If there's a problem, sleep and try
+	// again. Implementing this is a bit gnarly. In the event of a
+	// success, we get a nil error, and a stop function. But I don't
+	// see a simple way to ensure the updater is still running in
+	// the background.
+
+	// When launcher first starts, we'd like the
+	// server to start receiving data
```
directionless, line 54: nit -- reflow this

cmd/launcher/updater/updater_test.go
```diff
@@ -0,0 +1,226 @@
...
+		updaterRunReturns []func(opts ...tuf.Option) (stop func(), err error)
+		callStopChanAfter time.Duration
+		assertion         assert.ErrorAssertionFunc
+	}{
+		{
+			name: "success",
+			fields: fields{
+				updater: &mocks.Updater{},
+				config: &UpdaterConfig{
+					Logger: log.NewNopLogger(),
```
directionless, line 41: Since there's now a `Create` function, I'd probably take these out of the tests, and do something like:
```
if config.logger == nil {
  config.logger = log.NewNopLogger()
}
```

Probably pairs well with the `log.With` that's there

directionless (commented):
This looks pretty clean. Some nits.

cmd/launcher/updater/updater_test.go
James-Pickett: If I'm understanding correctly, the suggestion is to instantiate the updater via a call to an create/new function ...
I think to do that would require a bit more of a refactor. Currently, the create/new function for `cmd/launcher/updater` is creating a new instance of `pkg/autoupdate`. I think it might be better to pass an instance of `pkg/autoupdate` to the `cmd/launcher/updater` create/new function. This way we can mock that dependency for the `cmd/launcher/update` tests. I suppose somewhere around [here](https://github.com/kolide/launcher/blob/master/cmd/launcher/launcher.go#L188) we could create the new instance of `pkg/autoupdate` and pass it to `cmd/launcher/updater`

cmd/launcher/updater/updater.go
```diff
@@ -0,0 +1,153 @@
...
+		autoupdate.WithNotaryPrefix(config.NotaryPrefix),
+		autoupdate.WithFinalizer(finalizer),
+		autoupdate.WithUpdateChannel(config.UpdateChannel),
+		autoupdate.WithSigChannel(config.SigChannel),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	config.Logger = log.With(config.Logger, "updater", updater.BinaryName())
```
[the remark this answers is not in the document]
James-Pickett: done

cmd/launcher/updater/updater.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if u.stopExecution != nil {
+		u.stopExecution()
+	}
+}
+
+// CreateUpdater returns an Actor suitable for an oklog/run group. It
+// is a light wrapper around autoupdate.NewUpdater to simplify having
+// multiple ones in launcher.
+func CreateUpdater(
```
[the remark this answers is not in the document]
James-Pickett: makes sense, done

cmd/launcher/updater/updater.go
James-Pickett: I think the working is good, just moved the comments closer to the code they were commenting

cmd/launcher/updater/updater_test.go
James-Pickett: updated the create function, but not updating the tests per our conversation as it would require further refactor

cmd/launcher/internal/updater/updater.go
```diff
@@ -0,0 +1,158 @@
...
+		ctx:                     ctx,
+		stopChan:                make(chan bool),
+		config:                  config,
+		runUpdaterRetryInterval: 30 * time.Minute,
+	}
+
+	return &actor.Actor{
+		Execute:   updateCmd.execute,
+		Interrupt: updateCmd.interrupt,
+	}, nil
```
directionless, line 76: Hrm. Would this be cleaner if `actor.Actor` defined an interface, not a struct?

directionless (approved):
LGTM!

We could convert `actor.Actor` to an interface, or leave that for later.

PR #817, opened 2022-04-14
Updates tests to stop occasional failing

James-Pickett:
Updates tests to stop occasional failing

Not sure if the wmitable_test.go failing when parallel is indicative of a race condition somewhere in the code or if it's just affecting the unit test.


directionless (approved):
While I suspect there is some kind of race issue, I'm not sure it's there in practice. This seems like a reasonable approach. It is _no worse_ than prior.

PR #818, opened 2022-04-16
Adjust timeouts and add health checks to startup

directionless:
Continuing to track down the Monterey enrollment issue, I'm pretty sure there's a race condition around the async startup. 

This PR attempts to address that via several mechanisms:
1. Adjusts some timeouts. A bit of a long shot, but I think these were too aggressive before
2. Adds a health check between starting up, and trying to do anything


directionless:
Most of the meat of this was implemented in other PRs, but I want to see which (if any) of the timeout adjustments to persist.

directionless:
don't need this any more

PR #819, opened 2022-04-20
Changes to Enrollment

directionless:
As part of trying to address some subtle race conditions around startup, I've noticed several oddities about enrollment. This is an attempt to clean them up.

1. When we're using osquery transport, we can _avoid_ starting the extension complete. I've added some comments, and an if statement to support that. 

1. Launcher was triggering an enrollment during startup. Because that enrollment depends on osquery, I suspect this is one of the causes of a race condition. Since osquery itself is also doing that enrollment, this seems superfluous. I've gated it behind an environmental variable. 

1. It replaces `backoff.Run` with `backoff.WaitFor`. The former was a naive implementation of some retry logic, but it did not have a timeout mechanism. WaitFor is a small wrapper around `rate.Limiter` and should be more robust.

1. It changes the launcher supplemental information to be optional. If there's a failure communicating with osquery, this is probably indicative of some kind of race condition, and we should let osquery proceed normally.

This does **not** add a pause-for-ready function. I tried a bit to have enrollment defer until the tables were loaded, but I was unable to make it work. It would timeout on the tests. Generally felt like a couple things were deadlocked around the mutex.


pkg/osquery/extension.go
```diff
@@ -786,6 +797,19 @@ func (e *Extension) writeResultsWithReenroll(ctx context.Context, results []dist
 }
 
 func getEnrollDetails(client Querier) (service.EnrollmentDetails, error) {
+	var details service.EnrollmentDetails
+
+	if os.Getenv("LAUNCHER_DEBUG_ENROLL_DETAILS_ERROR") == "true" {
```
James-Pickett, line 802: what is this going to be used for? please leave comment to explain

James-Pickett (changes requested):

pkg/osquery/extension.go
blaedj, line 802: is this just something you're using to simulate errors while querying for enrollment details? Or is there some plan to use it during debugging to reduce the time launcher moves on from trying to get enrollment details?

directionless: Heh. You both flagged this. I've added a comment

James-Pickett (approved):

pkg/osquery/extension.go
```diff
@@ -786,6 +797,21 @@ func (e *Extension) writeResultsWithReenroll(ctx context.Context, results []dist
...
+	// To facilitate manual testing around missing enrollment details,
+	// there is a environmental variable to trigger the failure condition
+	if os.Getenv("LAUNCHER_DEBUG_ENROLL_DETAILS_ERROR") == "true" {
+		return details, errors.New("Skipping enrollment details")
+	}
+
+	// This condition is indicative of a misordering (or race) in
+	// startup. Enrollment has started before `SetQuerier` has
+	// been called.
+	if client == nil {
```
blaedj, line 811: I'm mostly just making sure I'm following things correctly, but...

given https://github.com/kolide/launcher/pull/819/files#diff-222645cc15c22456115e733d8ff807db8ea12a5839e79aff41a0159afa21abfcR122, whenever the transport is 'osquery', we're going to have a nil client (because we return before calling `setQuerier`), but that is fine because if the transport is osquery we aren't ever going to be calling getEnrollmentDetails, correct?

(I see what you mean by this being all very tangled...)

directionless: Well, this _is_ the extension, so if we're using osquery transport, then this is the thing we're not starting. It should never be called. 

But, yes, then the `client == nil` case would be there. though the real purpose is because something I did this week got a panic because this was called before that Querier was set. I think I've since undone that but it seems good to leave it in.

blaedj (dismissed):
These seem like good changes, especially not calling `enroll` multiple times, or failing when enrollment details can't be queried. Not sure if the windows tests are just flaky or...??

James-Pickett (approved):

PR #820, opened 2022-04-20
Fix the duplicate GitHub Actions

directionless:
I noticed that we were running twice the number of GitHub actions as expected. I believe this is because both the `push` and the `pull_request` triggers were active. This shifts them so the `push` is only on the main branches

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,7 +2,7 @@ name: ci
 
 on:
   push:
-    branches: '**'
+    branches: [main, master]
   pull_request:
     branches: '**'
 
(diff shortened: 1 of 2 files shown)


blaedj (approved):

.github/workflows/go.yml
```diff
@@ -2,7 +2,7 @@ name: ci
 
 on:
   push:
-    branches: '**'
```
James-Pickett, line 5: @directionless , this means now CI will only run in my forked repo when I push to a main or master branch ... do we loose anything by having it the way it was?

directionless: It was causing two sets of tests to run on the PRs. One triggered for the push, and one triggered for the PR. 

For my branches, I usually push to a draft PR. But maybe there's a way we can get the best of both worlds. I wonder if we can specify fork names or something

PR #822, opened 2022-04-21
split pkg/osquery/runtime/runtime.go into separate files for osquery instance and runner

James-Pickett:
split pkg/osquery/runtime/runtime.go into separate files for osquery instance and runner


directionless (approved):
I didn't look too carefully, but I assume this is a straightforward moving pieces around

Issue #821, opened 2022-04-21
We should try building universal binaries in GitHub Actions


directionless:
I think recent versions of the macOS SDK include what we need to cross compile targeting m1 machines from x86. This means we should be able to get universal builds out of GitHub actions. This would be a very exciting process improvement.

This looks very hopeful:
```shell
$ uname -m
x86_64

$ CGO_ENABLED=1 GO111MODULE=on GOOS=darwin GOARCH=arm64 go build -o /tmp/launcher.arm64 ./cmd/launcher
$ file /tmp/launcher.arm64
/tmp/launcher.arm64: Mach-O 64-bit executable arm64

$ /tmp/launcher.arm64 --version
-bash: /tmp/launcher.arm64: Bad CPU type in executable
```

PR #824, opened 2022-04-22
Add macOS universal builds to GitHub Actions

directionless:
This configures GitHub actions to build the `amd64` and the `arm64` binaries, and on macOS, stitch them into a universal binary.

We do not build `arm64` on linux, because I can't get the cross compile to work properly. As the main goal here is universal mac binaries, I simply disabled that. It would be nice to loop back in the future.

There's some bonus spring cleaning around the actions file:
* Updates the go build version to match what I'm using locally
* updates some action versions
* updates the linux build machine
* Adds caching to the go dependancies (As described by the [flipt project](https://markphelps.me/posts/speed-up-your-go-builds-with-actions-cache/))

Fixes: #821

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -43,7 +43,12 @@ jobs:
       run: make deps
 
     - name: Build
+      if: ${{ ! contains(matrix.os, 'ubuntu') }}
       run: make -j2 github-build
+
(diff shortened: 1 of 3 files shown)


James-Pickett (approved):
Nice Caching!

PR #827, opened 2022-04-22
Adds files and instructions for live launcher debugging with VS Code

James-Pickett:


.vscode/launch.json
```diff
@@ -0,0 +1,53 @@
...
+            "request": "launch",
+            "mode": "auto",
+            "preLaunchTask": "pre-debug",
+            "program": "${workspaceFolder}/cmd/launcher",
+            "args": [
+                "--debug",
+                "--hostname=localhost:3443",
+                "--control",
+                "--control_hostname=localhost:3443",
+                "--enroll_secret_path=${workspaceFolder}/k2_enroll_secret",
```
directionless, line 37: This all feels a little weird to me. But I'm also not sure what would be cleaner. Maybe we should ship a script to do this as a one-time setup, and then conditionalize the vscode tooling on whether it's been setup?

PR #830, opened 2022-04-26
moved enrollment mutext to lock before node key fetch in enroll func

James-Pickett:


pkg/osquery/extension.go
```diff
@@ -321,6 +321,11 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 		return e.NodeKey, false, nil
 	}
 
+	// Only one thread should ever be allowed to attempt enrollment at the
```
directionless, line 324: I suspect we should move this above `if e.NodeKey != ""` as well.

directionless (approved):

PR #831, opened 2022-04-27
James/persist osquery instance history

James-Pickett:


pkg/osquery/runtime/runner.go
```diff
@@ -82,7 +82,9 @@ func (r *Runner) Start() error {
 			select {
 			case <-r.shutdown:
 				// Intentional shutdown, this loop can exit
-				r.instance.stats.Exited(nil)
+				if err := r.instance.stats.Exited(nil); err != nil {
+					level.Error(r.instance.logger).Log("msg", "error recording osquery instance exit to history", "err", err)
```
directionless, line 86: I'd probably say `Info` for these

cmd/launcher/launcher.go
```diff
@@ -157,6 +158,12 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		}
 	}
 
+	// init osquery instance history
+	err = osqueryInstanceHistory.InitHistory(db)
+	if err != nil {
+		return errors.Wrap(err, "error initializing osquery instance history")
+	}
```
directionless, line 165: Talking aloud... This feels odd here, and not implicit in `NewInstance`, or in the runner setup. But I think I see why... I assume it can't be implicit, because the History struct is global, and we don't pass the DB to the runner.

Honestly, I think this is probably pointing to issues with the existing code. And this is probably a pragmatic choice here.

pkg/osquery/runtime/history/history_test.go
```diff
@@ -189,3 +172,23 @@ func TestLatestInstance(t *testing.T) { // nolint:paralleltest
 		})
 	}
 }
+
+func newTestBoltDb(t *testing.T, instancesToAddToDb ...*Instance) *bbolt.DB {
+	db, err := bbolt.Open(fmt.Sprintf("%s/%s", t.TempDir(), "osquery_instance_history_test.db"), 0600, nil)
```
directionless, line 177: Should provide a timeout here

directionless, line 177: `ioutil.TempFile(t.TempDir(), "osquery_instance_history_test.db")` I think

pkg/osquery/runtime/history/history_test.go
```diff
@@ -189,3 +172,23 @@ func TestLatestInstance(t *testing.T) { // nolint:paralleltest
...
+	require.NoError(t, err, "expect no error opening bolt db")
+
+	json, err := json.Marshal(instancesToAddToDb)
+	require.NoError(t, err, "expect no error marshalling instances")
```
directionless, line 181: Maybe reasonable, but it feels slightly weird to load test data here, and not let the `InitHistory` stuff populate it

pkg/osquery/runtime/history/instance.go
```diff
@@ -49,14 +49,28 @@ func (i *Instance) Connected(querier Querier) error {
 	i.InstanceId = instanceId
 	i.Version = version
 
+	err = currentHistory.save()
+
+	if err != nil {
```
directionless, line 54:
```suggestion
	if err := currentHistory.save(); err != nil {
```

directionless (commented):
I'd thumb this mostly as is, a couple of nits

pkg/osquery/runtime/history/history_test.go
James-Pickett: this that will handle it, here is docs on t.TempDir()

> TempDir returns a temporary directory for the test to use. The directory is automatically removed by Cleanup when the test and all its subtests complete. Each subsequent call to t.TempDir returns a unique directory; if the directory creation fails, TempDir terminates the test by calling Fatal.

pkg/osquery/runtime/history/history_test.go
James-Pickett: This actually seeds or loads it in the db and lets us confirm that the `InitHistory` stuff loads it without error. I've updated the names and added comment.

pkg/osquery/runtime/history/history_test.go
James-Pickett: nvm, I stand corrected, windows threw an error using t.TempDir() =(

James-Pickett: bbolt creates the db if it doesn't exist

directionless: I agree the docs on `t.TempDir()` sound good. I wonder if the windows error was because you're using `Sprintf` to join with `/` instead of `filepath.Join` ?

pkg/osquery/runtime/history/instance.go
```diff
@@ -49,14 +52,29 @@ func (i *Instance) Connected(querier Querier) error {
...
+	}
+
 	return nil
 }
 
 // InstanceExited sets the exit time and appends provided error (if any) to current osquery instance
-func (i *Instance) Exited(exitError error) {
+func (i *Instance) Exited(exitError error) error {
+	currentHistory.Lock()
+	defer currentHistory.Unlock()
```
directionless, line 65: Why lock here, and not in `currentHistory.save()`?

pkg/osquery/runtime/history/history_test.go
```diff
@@ -189,3 +173,45 @@ func TestLatestInstance(t *testing.T) { // nolint:paralleltest
...
+}
+
+// newTestBoltDb creates a new boltdb instance and seeds it with the given instances.
+func newTestBoltDb(t *testing.T, seedInstances ...*Instance) *bbolt.DB {
+
+	dir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	db, err := bbolt.Open(fmt.Sprintf("%s/%s", dir, "osquery_instance_history_test.db"), 0600, &bbolt.Options{
```
directionless, line 197: `filepath.Join` is generally going to be more correct. May not wholly matter

pkg/osquery/runtime/history/instance.go
James-Pickett: There are [places in the code](https://github.com/kolide/launcher/blob/master/pkg/osquery/runtime/history/history.go#L55) where we manipulate this variable and then save to db. We need to have the lock during this whole process. If we put it only in the save method we would introduce a place where it's getting a lock and then waiting to get the same lock.

pkg/osquery/runtime/history/history.go
```diff
@@ -21,15 +25,36 @@ func (c NoInstancesError) Error() string {
...
+	err := createBboltBucketIfNotExists(db)
+	if err != nil {
+		return err
+	}
+
+	currentHistory.db = db
+
+	err = currentHistory.load()
+
+	if err != nil {
```
directionless, line 42:
```suggestion
	if err =: currentHistory.load(); err != nil {
```

Nit... I wonder if we can lint for this.

pkg/osquery/runtime/history/history.go
```diff
@@ -67,6 +92,12 @@ func NewInstance() (*Instance, error) {
 
 	currentHistory.addInstanceToHistory(newInstance)
 
+	err = currentHistory.save()
+
+	if err != nil {
```
directionless, line 97:
```suggestion
	if err := currentHistory.save(); err != nil {
```

pkg/osquery/runtime/history/storage.go
```diff
@@ -0,0 +1,99 @@
...
+
+	instancesBytes, err := json.Marshal(h.instances)
+	if err != nil {
+		return errors.Wrap(err, "error marshalling osquery_instance_history")
+	}
+
+	err = h.db.Update(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(osqueryHistoryInstanceKey))
+		err = b.Put([]byte(osqueryHistoryInstanceKey), instancesBytes)
+		if err != nil {
```
directionless, line 66: scoped err form...

directionless (dismissed):
I think fine! I wonder if we can lint for the scoped if/err blocks. but I'd take this as is.

James-Pickett:
> I think fine! I wonder if we can lint for the scoped if/err blocks. but I'd take this as is.

I couldn't find a linter for this, at least not one already added to golangci

directionless (approved):
Let's try it!

PR #838, opened 2022-06-23
Removing osquery-extension

James-Pickett:
Removes the need for an additional osquery-extension.ext file to run launcher. It may have been required in the past in order to force osquery to wait for other extensions to load. However, now we can add the `extensions_require` flag when launching osquery and it will cause osquery to respect the `extensions_timeout` flag before loading the rest of the configuration.

Also adds new flag for autoloading extensions.


cmd/launcher/options.go
```diff
@@ -45,24 +45,25 @@ func parseOptions(args []string) (*launcher.Options, error) {
...
-		flOsquerydPath        = flagset.String("osqueryd_path", "", "Path to the osqueryd binary to use (Default: find osqueryd in $PATH)")
-		flRootDirectory       = flagset.String("root_directory", "", "The location of the local database, pidfiles, etc.")
-		flRootPEM             = flagset.String("root_pem", "", "Path to PEM file including root certificates to verify against")
-		flVersion             = flagset.Bool("version", false, "Print Launcher version and exit")
-		flLogMaxBytesPerBatch = flagset.Int("log_max_bytes_per_batch", 0, "Maximum size of a batch of logs. Recommend leaving unset, and launcher will determine")
-		flOsqueryFlags        arrayFlags // set below with flagset.Var
-		flCompactDbMaxTx      = flagset.Int64("compactdb-max-tx", 65536, "Maximum transaction size used when compacting the internal DB")
-		_                     = flagset.String("config", "", "config file to parse options from (optional)")
+		flAutoloadedExtensions = flagset.String("autoloaded_extensions", "", "Comma separated filenames of extensions to load, must be in same directory as luncher")
```
directionless, line 48: Exposing this all the way up here is pretty cool. More than I was thinking, but it's been something OSS users have asked for

cmd/launcher/options.go
```diff
@@ -187,6 +188,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		EnableInitialRunner:                *flInitialRunner,
 		EnrollSecret:                       *flEnrollSecret,
 		EnrollSecretPath:                   *flEnrollSecretPath,
+		AutoloadedExtensions:               strings.Split(*flAutoloadedExtensions, ","),
```
directionless, line 191: Might be slightly cleaner to do this with the array style passing. See `flOsqueryFlags` for an example

cmd/launcher/options.go
```diff
@@ -209,6 +211,12 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		Transport:                          *flTransport,
 		UpdateChannel:                      updateChannel,
 	}
+
+	// if it's the default empty, set it as an empty string array
+	if *flAutoloadedExtensions == "" {
+		opts.AutoloadedExtensions = []string{}
+	}
```
directionless, line 218: Do we need this?

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -328,41 +338,66 @@ type osqueryFilePaths struct {
...
+	for index, extension := range opts.autoloadedExtensions {
+		if _, err := os.Stat(extension); err != nil {
+			if os.IsNotExist(err) {
+				return nil, errors.Wrapf(err, "extension path does not exist: %s", extension)
+			} else {
+				return nil, errors.Wrapf(err, "could not stat extension path")
+			}
+		}
+
+		extensionPath := filepath.Join(autoupdate.FindBaseDir(exPath), extension)
```
directionless, line 387: This seems weird? I kinda understand, but there's a file stat 5 lines prior, so the order feels weird?

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -328,41 +338,66 @@ type osqueryFilePaths struct {
...
 	}
 
-	extensionPath := filepath.Join(autoupdate.FindBaseDir(exPath), extensionName)
-	if _, err := os.Stat(extensionPath); err != nil {
-		if os.IsNotExist(err) {
-			return nil, errors.Wrapf(err, "extension path does not exist: %s", extensionPath)
-		} else {
-			return nil, errors.Wrapf(err, "could not stat extension path")
-		}
+	file, err := os.Create(extensionAutoloadPath)
```
directionless, line 370: small nit for a more descriptive var than `file`

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -328,41 +338,66 @@ type osqueryFilePaths struct {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "creating autoload file")
 	}
+	defer file.Close()
 
-	// Determine the path to the extension socket
-	if extensionSocketPath == "" {
-		extensionSocketPath = socketPath(rootDir)
+	writer := bufio.NewWriter(file)
```
directionless, line 376: Why use a `bufio.Writer` here? (and not write straight to `file`?)

directionless (commented):
So what will actually populate `osquery-extension.ext` into the options? I had imagined something around https://github.com/kolide/launcher/blob/fae3d1e4574179df5f7f64765d23201c6934fbe1/cmd/launcher/extension.go#L215

pkg/osquery/runtime/osqueryinstance.go
James-Pickett: yea that is weird ... I guess the stat was only working because we happened to be in the same dir? Moved extension path up and now stating the full path instead which feels more correct.

pkg/osquery/runtime/osqueryinstance.go
James-Pickett: I know it doesn't make any difference at this scale, but bufio should result in fewer sys calls to write to the file. However, I guess since it doesn't make a difference, it's less complex with out bufio, so now writing to the file.

pkg/osquery/runtime/osqueryinstance.go
James-Pickett: updated to `osqueryAutoloadFile`

cmd/launcher/extension.go
```diff
@@ -261,5 +262,10 @@ func grpcRunnerOptions(logger log.Logger, db *bbolt.DB, opts *launcher.Options,
 			distributed.NewPlugin("kolide_grpc", ext.GetQueries, ext.WriteResults),
 			osquerylogger.NewPlugin("kolide_grpc", ext.LogString),
 		),
+		// this ensures osquery-extension.ext is present in the autoloaded extensions
+		// when distributed plugin flag is kolide_grpc, this is a dependency to run kolide_grpc,
+		// this is due to an odd behavior of osquery where we need at least one extension to be autoloaded
+		// inorder for osquery to wait for others to be loaded
```
directionless, line 268:
```suggestion
		// this ensures osquery-extension.ext is present in the autoloaded extensions
		// when using kolide_grpc for the config file. This is due to a bug in osquery.
		// See https://github.com/osquery/osquery/issues/7653
```

cmd/launcher/options.go
```diff
@@ -91,7 +92,9 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		// deprecated options, kept for any kind of config file compatibility
 		_ = flagset.String("debug_log_file", "", "DEPRECATED")
 	)
+
 	flagset.Var(&flOsqueryFlags, "osquery_flag", "Flags to pass to osquery (possibly overriding Launcher defaults)")
+	flagset.Var(&flAutoloadedExtensions, "autoloaded_extension", "Filenames of extensions to autoload, must be in same directory as launcher")
```
directionless, line 97: Should probably support full paths, or same directory semantics. Could leave it for future work? Could also not expose it up here

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -328,41 +337,68 @@ type osqueryFilePaths struct {
...
+		databasePath:          filepath.Join(opts.rootDirectory, "osquery.db"),
+		augeasPath:            filepath.Join(opts.rootDirectory, "augeas-lenses"),
+		extensionSocketPath:   extensionSocketPath,
+		extensionAutoloadPath: extensionAutoloadPath,
+		extensionPaths:        make([]string, len(opts.autoloadedExtensions)),
+	}
+
+	if len(opts.autoloadedExtensions) == 0 {
+
+		// delete the autoload file if there were no provided autoloaded extensions
```
directionless, line 361: Good call! Though won't `os.Create` do that anyhow?

cmd/launcher/extension.go
```diff
@@ -261,5 +262,10 @@ func grpcRunnerOptions(logger log.Logger, db *bbolt.DB, opts *launcher.Options,
...
+		runtime.WithAutoloadedExtensions("osquery-extension.ext"),
```
directionless, line 269: Windows uses `.exe`

pkg/osquery/runtime/osqueryinstance.go
James-Pickett: good point, just I will just move os.Create up in the function so it's always called

directionless (commented):
This is looking pretty nice. Good to remove this stuff.

Could also remove the rosetta install in `preinstall-darwin.sh`, but probably better in a separate PR.

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -456,6 +491,7 @@ func (opts *osqueryOptions) createOsquerydCommand(osquerydBinary string, paths *
 		fmt.Sprintf("--extensions_autoload=%s", paths.extensionAutoloadPath),
 		"--disable_extensions=false",
 		"--extensions_timeout=20",
+		"--extensions_require",
```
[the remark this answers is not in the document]
James-Pickett: Updated, I should have RTFMed that first. Curious that even passing the option blank caused it to respect the extensions_timeout ... ultimately creating a race condition I suppose. Anyways, I think this is an overall improvement.

directionless (approved):
I'm game for testing this as a nightly!

I don't know if we'll need to exempt `tls` from the required extensions (and other osquery buildins) But we should see

pkg/osquery/runtime/osqueryinstance.go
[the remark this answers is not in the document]
directionless: I would bet there are bugs lurking

PR #841, opened 2022-06-30
James/interactive mode

James-Pickett:
WIP of working launcher interactive mode

https://user-images.githubusercontent.com/8415811/176611358-0e520778-8b45-4970-982e-83c4cc3a3ecd.mov


cmd/launcher/interactive.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/kolide/kit/env"
+	"github.com/kolide/kit/fs"
+	"github.com/kolide/launcher/pkg/osquery/table"
+	osquery "github.com/osquery/osquery-go"
+	"github.com/pkg/errors"
```
directionless, line 16: Better to use the native `fmt` based errors for new work

cmd/launcher/interactive.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+func interactive(args []string) error {
```
directionless, line 19:
```suggestion
func runInteractive(args []string) error {
```

slight preference for naming

directionless (commented):
I like this approach, I think i'll be cleaner than what I was doing in runtime.

Some quick early morning comments. 

Do you want to get Augeas in before merge, or update after? (It will need the tmpdir)

cmd/launcher/interactive.go
```diff
@@ -0,0 +1,65 @@
...
+	if osquerydPath == "" {
+		osquerydPath = findOsquery()
+		if osquerydPath == "" {
+			return errors.New("Could not find osqueryd binary")
+		}
+	}
+
+	fmt.Println(">> Starting osquery interactive with launcher tables")
+
+	rootDir, err := os.MkdirTemp("", "")
```
directionless, line 40:
```suggestion
	rootDir, err := os.MkdirTemp("", "launcher-osquery-interactive")
```

pkg/interactive/interactive.go
```diff
@@ -0,0 +1,134 @@
+package interactive
```
directionless, line 1: I'd hang this in `pkg/osquery/interactive`

pkg/interactive/interactive.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, flag := range osqueryFlags {
+		if strings.HasPrefix(flag, "extensions_socket=") {
+			split := strings.Split(flag, "=")
+			if len(split) > 1 && split[1] != "" {
+				path = split[1]
+			} else {
+				return "", errors.New("extensions_socket flag is missing a value")
+			}
+		}
+	}
```
directionless, line 96: I'm not sure there's a use case for passing a custom socket this way, and I think this feels error prone. Curious if you feel otherwise

The way we handled this in launcher's runtime, is to put launcher specific paths _after_ the user flags. That way launcher's is always the last one (which is usually going to take priority)

pkg/interactive/interactive.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if path == "" {
+		path = filepath.Join(rootDir, "sock")
```
directionless, line 99:
```suggestion
		path = filepath.Join(rootDir, "osquery.sock")
```

Unless we're length restricted, slight preference for more descriptive

cmd/launcher/interactive.go
```diff
@@ -0,0 +1,65 @@
...
+
+	osquerydPath := *flOsquerydPath
+	if osquerydPath == "" {
+		osquerydPath = findOsquery()
+		if osquerydPath == "" {
+			return errors.New("Could not find osqueryd binary")
+		}
+	}
+
+	fmt.Println(">> Starting osquery interactive with launcher tables")
```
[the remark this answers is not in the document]
James-Pickett: How about just dropping these info lines? The likes like only place we use the logger is in the default launcher cmd ... seems like too much work to thread this through.

pkg/interactive/interactive.go
James-Pickett: I'm not sure why I got stuck on respecting the socket flag .... probably because I started this effort by looking at the socket cmd. I dropped all the extensions_socket stuff.

pkg/interactive/interactive.go
James-Pickett: `osquery.sock` is a bit tight for running unit tests:

```
error loading extensions: error creating extension manager server: socket path /var/folders/8p/40gfkn2d3zl3mb_m9jk_vsk80000gn/T/TestStartProcessno_flags978291810/001/osquery.sock (99 characters) exceeded the maximum socket path character length of 97
```

how about `osq.sock` ?

Bring us down to 95 characters and from what I've seen the test dir only vary by 1-2 characters

pkg/interactive/interactive.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/kolide/kit/fs"
+	"github.com/kolide/launcher/pkg/osquery/table"
+	osquery "github.com/osquery/osquery-go"
+)
+
+const extensionName = "com.kolide.launcher_interactive"
+
+func StartProcess(rootDir, osquerydPath string, osqueryFlags []string) (*os.Process, error) {
+
+	if err := os.MkdirAll(rootDir, fs.DirMode); err != nil {
```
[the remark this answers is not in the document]
James-Pickett: it was intentional ... either way the caller specifies the directory which implies it's their responsibility to clean up

cmd/launcher/interactive.go
```diff
@@ -0,0 +1,118 @@
...
+	// Wait until user exits the shell
+	state, err := osqueryProc.Wait()
+	if err != nil {
+		return fmt.Errorf("error waiting for osqueryd: %s", err)
+	}
+
+	// Keep on keepin' on.
+	fmt.Printf("<< Exited osquery interactive with launcher tables: %s\n", state.String())
+
+	if err := extensionManagerServer.Shutdown(context.Background()); err != nil {
```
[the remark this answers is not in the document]
James-Pickett: @directionless , I have dropped the logic regarding shutting down the extensions manager server. Since we are not shelling any commands around the extension server and we are cleaning up the dir we created, I don't think we have to worry about this since when the main exits it will kill all other go routines. Do you agree?

pkg/osquery/interactive/interactive.go
```diff
@@ -0,0 +1,98 @@
...
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil, fmt.Errorf("error getting current working directory: %w", err)
+	}
+
+	// Transfer stdin, stdout, and stderr to the new process
+	// and also set target directory for the shell to start in.
+	pa := os.ProcAttr{
+		Files: []*os.File{os.Stdin, os.Stdout, os.Stderr},
+		Dir:   cwd,
```
directionless, line 33: I'll take it, but it seems weird. Why do we need that? Shouldn't it be the default if we're not changing directories?

pkg/osquery/interactive/interactive.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	socketPath := filepath.Join(rootDir, "osq.sock")
+
+	proc, err := os.StartProcess(osquerydPath, buildOsqueryFlags(socketPath, osqueryFlags), &pa)
+	if err != nil {
+		return nil, fmt.Errorf("error starting osqueryd in interactive mode: %w", err)
+	}
+
+	_, err = loadExtensions(socketPath, osquerydPath)
```
directionless, line 43: Do we need to save the extension manager to stop it at the end?

pkg/osquery/interactive/interactive_test.go
```diff
@@ -0,0 +1,102 @@
+//go:build !windows
+// +build !windows
```
directionless, line 2: I realized that you're using a socket, which won't work on windows. Windows probably needs a named pipe. This is in `pkg/osquery/runtime/runtime_helpers_windows.go`

I think it would be fine to ship as is, and then add windows support

directionless (dismissed):
Seems like a reasonable first cut. I think there's followup work in:
* [ ] windows
* [ ] augeas
* [ ] ATC configs

pkg/osquery/interactive/interactive.go
James-Pickett: you're correct, no need for that ... removed

pkg/osquery/interactive/interactive.go
James-Pickett: added call to shutdown extension manager server on exit

directionless (approved):

PR #844, opened 2022-07-20
disabled watifor tests on macos due to flakeyness

James-Pickett:


pkg/backoff/waitfor_test.go
```diff
@@ -16,11 +16,11 @@ func TestWaitFor(t *testing.T) {
 
 	// The github action runners seem very sensitive to timing
 	// wobbles, and I've been unable to get these tests to work
-	// there. As I think it's an issue specific to GitHub's
-	// windows runners, and trying to test timeout values, I'm
+	// consistently there. As I think it's an issue specific to GitHub's
+	// windows and drawin runners, and trying to test timeout values, I'm
```
directionless, line 20:
```suggestion
	// windows and darwin runners, and trying to test timeout values, I'm
```

directionless (approved):
Happy to approve. Could also only skip darwin if it's on github (and let it run on our work machines)

pkg/backoff/waitfor_test.go
James-Pickett: cool, I did that!

directionless (approved):

PR #845, opened 2022-07-25
windows support for interactive mode

James-Pickett:
updates interactive mode to use named pipes for windows

### DIFF
--- a/pkg/osquery/interactive/interactive_test.go
+++ b/pkg/osquery/interactive/interactive_test.go
@@ -1,3 +1,7 @@
+// disabled on windows because go tests doesn't get access to used named pipes
+//go:build !windows
+// +build !windows
+
 package interactive
 
 import (
--- a/pkg/osquery/interactive/interactive_test.go
+++ b/pkg/osquery/interactive/interactive_test.go
@@ -1,6 +1,3 @@
-//go:build !windows
-// +build !windows
-
 package interactive
 
 import (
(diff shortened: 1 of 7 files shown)


pkg/osquery/interactive/interactive_test.go
```diff
@@ -1,3 +1,4 @@
+// disabled on windows because go tests doesn't get access to used named pipes
```
directionless, line 1: Really!?

directionless (commented):
Code looks fine. I assume CI is correct

pkg/osquery/interactive/interactive_test.go
James-Pickett: @directionless , I looked into this deeper and this turned out to be an issue with the tests trying to download the osqueryd binary in parallel to the same location. I moved the initial download and caching of the osqueryd binary to the TestMain func and it cleared up the issue.

directionless (approved):

pkg/osquery/interactive/interactive_test.go
James-Pickett: @directionless , nevermind, I forgot to call m.Run() at the end of test main .... so the tests didnt run ... back to the drawing board.

directionless (dismissed):
Seems fine. Sometime we should solve the windows test issue, but it can be not-today

directionless (approved):

PR #846, opened 2022-07-26
adds augeas lenses to launcher interactive

James-Pickett:


pkg/osquery/interactive/interactive.go
```diff
@@ -73,6 +88,11 @@ func buildOsqueryFlags(socketPath string, osqueryFlags []string) []string {
 		fmt.Sprintf("--extensions_socket=%s", socketPath),
 	}...)
 
+	// only install augeas lenses on non-windows platforms
+	if runtime.GOOS != "windows" {
```
directionless, line 92: Feels a little awkward, but probably fine

directionless (approved):
I didn't test this, but the code looks generally correct.

PR #847, opened 2022-07-27
Correctly drop logs that are too big

directionless:
Squinting at this, I'm pretty sure logs were not correctly being dropped from the too-big conditional.

This re-arranges slightly, to make it a bit easier to see how the code works

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -548,6 +548,26 @@ func (e *Extension) writeLogsLoopRunner() {
 	}
 }
 
+// numberOfBufferedLogs returns the number of logs buffered for a given type.
+func (e *Extension) numberOfBufferedLogs(typ logger.LogType) (int, error) {
+	bucketName, err := bucketNameFromLogType(typ)
+	if err != nil {
+		return 0, err
+	}
+
+	var count int
+	err = e.db.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(bucketName))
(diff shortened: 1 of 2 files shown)


James-Pickett (commented):
Makes sense to me, but unit tests failing now.

directionless:
> Makes sense to me, but unit tests failing now.

Ha. The test was correct, and I added a logic bug. I fixed said logic bug.

And then fixed the test to actually test for dropped logs. It wasn't -- it was only for skipped logs.

pkg/osquery/extension.go
```diff
@@ -548,6 +548,26 @@ func (e *Extension) writeLogsLoopRunner() {
 	}
 }
 
+// numberOfBufferedLogs returns the number of logs buffered for a given type.
+func (e *Extension) numberOfBufferedLogs(typ logger.LogType) (int, error) {
```
James-Pickett, line 552: total nit, but since this func is only used by the the tests, you could put it in the `extension_test.go` file and the build will be ever so slightly smaller =)

James-Pickett (approved):

pkg/osquery/extension.go
directionless: Yeah.... I wasn't sure if we'd want to expose that later. 

But I was thinking about how it would be interesting to put this sort of thing into the `_test` file. Especially with a `must` form, or one that takes `t` and does the error handling internally.

James-Pickett (approved):

PR #850, opened 2022-07-28
adds simple version of systray available in debug mode and for kolide employees

James-Pickett:
* adds systray icon that runs as separate process
* systray exits when parent process exits
* only runs when in debug mode or for kolide employees

image


cmd/launcher/systray/systray.go
```diff
@@ -0,0 +1,49 @@
+package systray
+
+import (
+	_ "embed"
```
directionless, line 4: Is this used?

directionless (approved):
There's a lot here I'm unsure of. But I think this is the right starting point.

cmd/launcher/systray/systray.go
James-Pickett: yes it has to be imported in order to use the `go:embed` directive in
```
//go:embed kolide-mark-only-white.png
var kolideSystrayIcon []byte
```
but if you name the import you'll get an error that it's not in use

directionless: That's not in this file? (I think it moved to icon files?)

James-Pickett: ohh woops, that's me not paying attention to the file the comment was in =), I'll remove it

cmd/launcher/systray/icon_unix.go
```diff
@@ -0,0 +1,11 @@
+//go:build darwin || linux
+// +build darwin linux
```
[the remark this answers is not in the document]
James-Pickett: changed to icon.go

directionless (approved):

James-Pickett:
closing this while I run CI in my forked repo

PR #855, opened 2022-08-12
Add RSA keys to launcher database

directionless:
This adds an RSA keypair to launcher's database. It will be used for some incoming features.

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -3,6 +3,7 @@ package osquery
 import (
 	"bytes"
 	"context"
+	"crypto/rsa"
 	"crypto/x509"
 	"encoding/binary"
 	"encoding/json"
@@ -252,7 +253,7 @@ func SetupLauncherKeys(db *bbolt.DB) error {
 		}
 
 		var pub bytes.Buffer
-		if err := RsaPublicKeyToPem(key, &pub); err != nil {
+		if err := RsaPrivateKeyToPem(key, &pub); err != nil {
 			return errors.Wrap(err, "marshalling pub")
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/osquery/extension.go
```diff
@@ -164,6 +170,10 @@ func NewExtension(client service.KolideService, db *bbolt.DB, opts ExtensionOpts
 		return nil, errors.Wrap(err, "creating DB buckets")
 	}
 
+	if err := SetupLauncherKeys(db); err != nil {
+		return nil, errors.Wrap(err, "setting up initial launcher keys")
```
James-Pickett, line 174: nit / opinion: I feel like we should go ahead and switch to using `fmt.Errorf()` for all new error wrapping. I don't think the inconsistency in the file is really a problem.

pkg/osquery/extension.go
```diff
@@ -219,8 +229,100 @@ func (e *Extension) getHostIdentifier() (string, error) {
 	return IdentifierFromDB(e.db)
 }
 
+func SetupLauncherKeys(db *bbolt.DB) error {
+	err := db.Update(func(tx *bbolt.Tx) error {
+
+		b, err := tx.CreateBucketIfNotExists([]byte(configBucket))
```
James-Pickett, line 235: can we name this variable `bucket`? By the time I made it to the end of the function I forgot what `b` was =)

James-Pickett (approved):

James-Pickett (approved):
Dropped a couple of nits, but looks good!

pkg/osquery/extension.go
directionless: Sure

pkg/osquery/extension.go
directionless: Reasonable. I think I got these. Sometime we'll just need to change a bunch en-mass.

James-Pickett (approved):

PR #859, opened 2022-08-15
added goroutine to wait on systray process so it does not zombie when dies, now only spin up systray where uid less than 501, call to systray.Quit on exit

James-Pickett:


pkg/systray/runtime/runner.go
```diff
@@ -70,4 +73,6 @@ func (r *SystrayUsersProcessesRunner) Interrupt(err error) {
 			)
 		}
 	}
+
+	r.procsWg.Wait()
```
directionless, line 77: If I understand this correctly, what happens if there's a systray process wedged hard? This would block launcher from exiting?

pkg/systray/runtime/runner_darwin.go
```diff
@@ -67,6 +68,22 @@ func (r *SystrayUsersProcessesRunner) runConsoleUserSystray() error {
 		"pid", proc.Pid,
 	)
 
+	r.procsWg.Add(1)
+	go func(uid string, proc *os.Process) {
+		defer r.procsWg.Done()
+		// if the systray proccess dies, the parent must clean up otherwise we get a zombie process
```
directionless, line 74: This comment helps explain why. 👍

directionless (commented):
Given your comment -- that the parent need to stick around for some process cleanup -- this makes sense. But I'm a little leery about some fragility. 

The biggest (potential) issue I see, is that a misbehaving process, will cause the parent to block on exiting. I feel like that sort of thing needs some protections. Which probably means something hooked to the runner's interrupt maybe? Not sure?

The other thing I see, which is more of a feature than bug, is that this adds a go routine to the parent that's watching this child. That would be an obvious place to do signaling. If we want to exit, send a STOP, and if it doesn't die, send a KILL. Etc.

pkg/systray/runtime/runner.go
James-Pickett: added a 5 second time out for this

pkg/systray/runtime/runner.go
```diff
@@ -58,16 +62,42 @@ func (r *SystrayUsersProcessesRunner) Execute() error {
...
+		r.procsWg.Wait()
+	}()
+
+	select {
+	case <-wgDone:
+		return
+	case <-time.NewTimer(time.Second * 5).C:
+		level.Error(r.logger).Log(
+			"msg", "timeout waiting for systray processes to finish",
+		)
```
directionless, line 101: I didn't test this, but it seems reasonable to me.

pkg/systray/runtime/runner.go
```diff
@@ -58,16 +62,42 @@ func (r *SystrayUsersProcessesRunner) Execute() error {
 // Interrupt stops creating launcher systray processes and kills any existing ones.
 func (r *SystrayUsersProcessesRunner) Interrupt(err error) {
 	r.interrupt <- struct{}{}
+
 	for _, proc := range r.uidProcs {
```
directionless, line 66: I think this for loop feel a little weird, but I can't quite say why. I think it's probably fine to merge, and at worst will be noisy and we'll iterate.

directionless (approved):

pkg/systray/runtime/runner.go
James-Pickett: yea, it's weird ....

how about ...

```go
if !processExists(proc.Pid) {
	continue
} else if err := proc.Signal(syscall.SIGTERM); err != nil {
	level.Error(r.logger).Log(
		"msg", "error sending SIGTERM to systray process after interrupt",
		"err", err,
	)
} else if err := proc.Kill(); err != nil {
	level.Error(r.logger).Log(
		"msg", "error killing systray process after interrupt",
		"err", err,
	)
}
```

could use a switch statement, but then you can't capture the err in a variable for logging

```go
switch {
case !processExists(proc.Pid):
	continue
case proc.Signal(syscall.SIGTERM) == nil:
	continue
case proc.Kill() == nil:
	continue
default:
	level.Error(r.logger).Log(
		"msg", "sigterm and kill failed for process",
		"pid", proc.Pid,
	)
}
```

directionless: Yeah, these are both getting at what I mean... Though Ideally you want a small sleep between the kills.

I almost want something like:
```
for _, proc := range r.uidProcs {
   proc.Signal(syscall.SIGTERM)
}

select {
case <-wgDone:
    return
case <-time.NewTimer(time.Second * 5).C:
    for _, proc := range r.uidProcs {
        proc.Kill()
    }
case <-time.NewTimer(time.Second * 5).C:
    level.Error(r.logger).Log("msg", "timeout waiting for systray processes to finish")
}
```

can't quite tell if that's correct, or too complicated, or what

ee/systray/runtime/runner_test.go
```diff
@@ -0,0 +1,85 @@
...
+func TestSystrayUserProcessRunner_Execute(t *testing.T) {
+	t.Parallel()
+
+	// When running this using the golang test harness, it will leave behind proccess if you do not build the binary first.
+	// On mac os you can find these by setting the executable path to an empty string before running the tests, then search
+	// the processes in a terminal using: ps aux -o ppid | runtime.test after the tests have completed, you'll also see the
+	// CPU consumtion go way up.
+
+	// To get around the issue mentioned above, build the binary first and set it's path as the executable path on the runner.
+	executablePath := filepath.Join(t.TempDir(), "systray")
```
directionless, line 29:
```suggestion
	executablePath := filepath.Join(t.TempDir(), "systray.exe")
```

Somewhat optional, but if/when windows, we'll need `.exe`

directionless (approved):

PR #863, opened 2022-08-23
Add firefox_preferences table

### LINKED ISSUE #730: Support firefox's preferences
Firefox stores its preferences in a `prefs.js` file. Despite the name, this is not javascript. It would be nice to be able to parse these.

Some info:
* https://blog.mozilla.org/nnethercote/2018/03/09/a-new-preferences-parser-for-firefox/
* https://searchfox.org/mozilla-central/source/modules/libpref/parser/src/lib.rs
* https://firefox-source-docs.mozilla.org/modules/libpref/index.html
* https://github.com/hansmi/go-mozpref

goronfreeman:
Closes #730

Adds a new `firefox_preferences` EAV table. For now, we're not using [go-mozprefs](https://github.com/hansmi/go-mozpref), but it's a possible future enhancement.


pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	scanner := bufio.NewScanner(file)
+
+	rowData := map[string]string{"path": paths[0]}
+	rawKeyVals := make(map[string]interface{})
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		re := regexp.MustCompile(`user_pref\((.*)\)`)
```
James-Pickett, line 64: can we move this declaration outside of the for loop so it's not reallocated each time?

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,84 @@
...
+		match := re.FindStringSubmatch(line)
+
+		if len(match) > 1 {
+			parts := strings.Split(match[1], ", ")
+			rawKeyVals[parts[0]] = parts[1]
+		}
```
James-Pickett, line 70: just an option .... prefer to reduce nested code and invert if statement
```suggestion
		if len(match) <= 1 {
		    continue
		}
		
		parts := strings.Split(match[1], ", ")
		rawKeyVals[parts[0]] = parts[1]
```

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,84 @@
...
+	if len(paths) != 1 {
+		return nil, errors.Errorf("The %s table requires that you specify a constraint for path", tableName)
+	}
+
+	file, err := os.Open(paths[0])
+
+	if err != nil {
+		// TODO: Investigate what error message looks like. Add filepath possibly
+		return nil, err
+	}
```
James-Pickett, line 55: nit ... the usually formatting I've seen in go does the `if err != nil` on the line immediately after the err assignment

```suggestion
	file, err := os.Open(paths[0])
	if err != nil {
		// TODO: Investigate what error message looks like. Add filepath possibly
		return nil, err
	}
```

James-Pickett (changes requested):

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type Table struct {
+	name   string
+	logger log.Logger
+}
+
+const tableName = "kolide_firefox_preferences"
+
+func TablePlugin(_ *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
```
directionless, line 26: If you don't need the `osquery.ExtensionManagerClient`, you don't need the method signature to include it. (I know most of them do, but that might not be correct)

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	return generateData(queryContext, t.logger)
+}
+
+func generateData(queryContext table.QueryContext, logger log.Logger) ([]map[string]string, error) {
+	paths := tablehelpers.GetConstraints(queryContext, "path")
+
+	if len(paths) != 1 {
```
directionless, line 46: This should be able to take more than one. eg: `WHERE path IN (a, b, c)`

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, errors.Errorf("The %s table requires that you specify a constraint for path", tableName)
+	}
+
+	file, err := os.Open(paths[0])
```
directionless, line 50: > `paths[0]`

See above comment

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		// TODO: Investigate what error message looks like. Add filepath possibly
```
directionless, line 52: Our ability to handle errors is, um, not-ideal.

Generally speaking, we should _not_ return the error (as you did not), returning the error breaks table usage, so that seems good. But  logging it at the debug level can be good too.

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+
+	scanner := bufio.NewScanner(file)
+
+	rowData := map[string]string{"path": paths[0]}
+	rawKeyVals := make(map[string]interface{})
+
+	re := regexp.MustCompile(`user_pref\((.*)\)`)
```
directionless, line 61: This is basically a constant, it should be a package level var.

directionless: Does it need a start of line marker? It might not. No clue really.

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+	for scanner.Scan() {
+		line := scanner.Text()
+		match := re.FindStringSubmatch(line)
+
+		if len(match) <= 1 {
+			continue
+		}
+
+		parts := strings.Split(match[1], ", ")
```
directionless, line 70: What happens if there are multiple commas? Suggest adding some to the test cases

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+		rawKeyVals[parts[0]] = parts[1]
```
directionless, line 71: I don't think `parts[1]` is guaranteed? If not this will panic.

pkg/osquery/tables/firefox_preferences/testdata/prefs.js
```diff
@@ -0,0 +1,13 @@
...
+//
+// If you make changes to this file while the application is running,
+// the changes will be overwritten when the application exits.
+//
+// To change a preference value, you can either:
+// - modify it via the UI (e.g. via about:config in the browser); or
+// - set it within a user.js file in your profile.
+
+user_pref("app.normandy.first_run", false);
+user_pref("app.normandy.migrationsApplied", 12);
```
directionless, line 13: Strongly suggesting adding some more complex test data.

pkg/osquery/tables/firefox_preferences/table.go
directionless: Yes. Early returns FTW

pkg/osquery/tables/firefox_preferences/table.go
goronfreeman: In practice, it seems to do fine without it. Wouldn't hurt anything though, so I may as well add it.

pkg/osquery/tables/firefox_preferences/table.go
goronfreeman: It should be guaranteed once it reaches this point. If the regex matches, then the string returned will always be a key/val pair separated by a comma.

pkg/osquery/tables/firefox_preferences/table.go
goronfreeman: I guess it's technically possible for the `user_pref(` stuff to exist as a value somewhere, so I'm gonna add that 👍

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func generateData(queryContext table.QueryContext, logger log.Logger) ([]map[string]string, error) {
+	paths := tablehelpers.GetConstraints(queryContext, "path")
+
+	if len(paths) == 0 {
+		return nil, errors.Errorf("The %s table requires that you specify a constraint for path", tableName)
+	}
+
+	rowData := map[string]string{"path": paths[0]}
```
goronfreeman, line 51: Just noticed this `paths[0]` leftover from before I allowed multiple. I'll fix this up.

pkg/osquery/tables/firefox_preferences/table.go
directionless: Are any of them user inputted? What happens for things like:
```
user_pref("app.update.background.previous.reasons", "[\"the maintenance service registry, has a comma\"]");
```

basically, you're looking at the impossibility of parsing complex nested data with regexps. There may be some clever things -- you know the larger line structure. But here be dragons

pkg/osquery/tables/firefox_preferences/table.go
directionless: I don't understand how `^user_pref\((.*)\)` guarantees that. Yes, if the file conforms to your expectations. But it may not.

goronfreeman: That makes sense. I was thinking that if the file doesn't conform to what we expect, then we have larger problems. I'll check to make sure `len(parts) == 2` before adding the key/val.

pkg/osquery/tables/firefox_preferences/table.go
goronfreeman: Good point, didn't think about that. I don't think there is any place for a user to input data, but I'll have to look closer. If this is the case, then maybe regex isn't the way to go here, and we want to use the parsing library after all?

pkg/osquery/tables/firefox_preferences/table.go
directionless: This comment should be paired with the above one about split. I think you have several options.

If I understand correctly, you're assuming a structure:
* lines match `user_pref("KEY", VALUE);`
* `KEY` does not include quotes
* `VALUE` might be arbitrarily complex and gnarly
* All other parts of that are assumed as fixed. (eg, ends with `);`

I think your first option is to use a more clever regex. You can extract the KEY and the VALUE directly. With this code, this is the direction I'd go.

I think the second option is to limit the split. In go, that'd be `SplitN`, in ruby it's the 3 argument split. I would bias away from this one.

Last choice is to use (or write) a parsing library

directionless: > if the file doesn't conform to what we expect, then we have larger problems

Maybe. It might have a format we haven't seen, or it might be a broken/truncated/corrupt file, either accidentally or intentionally. While we might not be able to extract data, we should never panic or throw errors.

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,105 @@
...
+func generateData(queryContext table.QueryContext, logger log.Logger) ([]map[string]string, error) {
+	var results []map[string]string
+
+	filePaths := tablehelpers.GetConstraints(queryContext, "path")
+
+	if len(filePaths) == 0 {
+		return nil, errors.Errorf("The %s table requires that you specify a constraint WHERE path IN", tableName)
+	}
+
+	rawKeyVals := make(map[string]interface{})
```
James-Pickett, line 54: I think this should be declared a little further down, just before we start scanning? otherwise it will eventually have all the data from all paths?

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return table.NewPlugin(t.name, columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	return generateData(queryContext, t.logger)
+}
+
+func generateData(queryContext table.QueryContext, logger log.Logger) ([]map[string]string, error) {
+	var results []map[string]string
```
James-Pickett, line 46: can we move this down to closer to where it's used? just before we start the for loops below

directionless: FWIW we often stick results at the top, since all the loops feed into it. I don't have strong feelings.

goronfreeman: I'm cool with moving it, but I was copying a pattern from `kolide_cryptoinfo`.

James-Pickett: since we never error, it may be needed sooner, so it's good as it is =)

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,116 @@
...
+	t := &Table{
+		name:   tableName,
+		logger: logger,
+	}
+
+	return table.NewPlugin(t.name, columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	return generateData(queryContext, t.logger), nil
```
directionless, line 42: Why have both `generate` and `generateData` ?

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, filePath := range filePaths {
+		for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
+			flattenOpts := []dataflatten.FlattenOpts{
+				dataflatten.WithLogger(logger),
+				dataflatten.WithQuery(strings.Split(dataQuery, "/")),
+			}
+
+			file, err := os.Open(filePath)
+			if err != nil {
+				level.Info(logger).Log(
```
directionless, line 69: It's going to be super common to hit non-existent files, I'd probably downgrade this to `Debug`.

I could imagine conditionalizing the level on the type of error, and not logging not-found at all.

directionless (commented):
Looks pretty good

pkg/osquery/tables/firefox_preferences/table.go
```diff
@@ -0,0 +1,116 @@
...
+				// left with escaped quotes that no longer make sense.
+				// i.e. {\"249024122\":[1660860020218]}
+				// Replace those with unescaped quotes.
+				rawKeyVals[match[1]] = strings.ReplaceAll(match[2], "\\\"", "\"")
+			}
+
+			flatData, err := dataflatten.Flatten(rawKeyVals, flattenOpts...)
+			if err != nil {
+				level.Info(logger).Log(
+					"msg", "failed to get data for path",
```
James-Pickett, line 103: this should probably say "failed to flatten data for path", also as seph mentioned above, the logs in this package should probably include the table name

pkg/osquery/tables/firefox_preferences/table.go
goronfreeman: I left it like this for testing purposes, since `generate` is hung off the `Table` struct. Talked with @James-Pickett about this while he was helping me with tests, but if you have a suggestion on how to work around that, I'm happy to change it.

James-Pickett: I felt like having the function with the full signature of `func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error)` would just add noise to the tests and the func itself. What we are really doing / testing here is parsing prefs.js

directionless: Not sure what you mean James...

Usually I'd just instantiate the table `table := Table{logger: log.NewNopLogger()}` and then call `table.generate(...)`.  It's not really different than calling `generateData` here.

James-Pickett: I'll expand a bit .... `generate()` requires the `ctx context.Context` parameter and the `error` return value. Both of these are irrelevant to the actual parsing of prefs.js in `generateData()` since we don't use context or ever return an error there. I feel like it just makes it slightly more difficult for a human to parse. But this point is not a show stopper for me if you feel strongly about it.

directionless: Yeah... You're not wrong. I dunno. I guess my bias it to have fewer functions. But if this was larger I would 100% be with you about splitting it into testable pieces

directionless (approved):
Code looks okay, and it seems to work. Can't promise we won't need to change it, but solid first job

PR #865, opened 2022-08-24
James/systray shutdown on update

James-Pickett:
adds logic to shutdown systray when launcher updates
breaks functions of main runLauncher func
replaces errors.Wrap with fmt.Error


cmd/launcher/launcher.go
```diff
@@ -235,25 +176,143 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
 		runGroup.Add(launcherUpdater.Execute, launcherUpdater.Interrupt)
 	}
 
+	err = runGroup.Run()
+	return fmt.Errorf("running service: %w", err)
+}
+
+func writePidFile(path string) error {
+	err := ioutil.WriteFile(path, []byte(strconv.Itoa(os.Getpid())), 0600)
+	if err != nil {
```
directionless, line 200: 1. These lines can be combined.
2. `ioutil.WriteFile` is deprecated, and should be (I think) `os.WriteFile`

And in doing that, the function is no shorter than the leaving it unsplit. But I guess there's some value in having it be reasonable self documenting. Okay.

cmd/launcher/launcher.go
```diff
@@ -235,25 +176,143 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+		level.Debug(logger).Log("msg", "interrupted", "err", err, "stack", fmt.Sprintf("%+v", err))
+		cancel()
+		close(sigChannel)
+	})
+
+	return sigChannel
+}
+
+func createSystrayRunner(logger log.Logger, runGroup *run.Group, opts *launcher.Options) *systrayruntime.SystrayUsersProcessesRunner {
 	if (opts.KolideServerURL == "k2device-preprod.kolide.com" || opts.KolideServerURL == "localhost:3443") && runtime.GOOS == "darwin" {
```
directionless, line 293: For this one specifically, I kinda want the `opts.KolideServerURL == "k2device-preprod.kolide.com" || opts.KolideServerURL == "localhost:3443"` up in the main func. Since I'm thinking about hanging other things off it.

But doing it here does make for a clean return nil, so I'm not sure.

I think it's going to evolve real fast. There are 2 knobs we're going to need. First one is whether it's talking to Kolide, which can pull in `ee`. Second one is whether it's the Kolide account, which is basically a feature flag.

cmd/launcher/launcher.go
```diff
@@ -235,25 +176,143 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
 
-	err = runGroup.Run()
-	return errors.Wrap(err, "run service")
+	return nil
 }
 
-func writePidFile(path string) error {
-	err := ioutil.WriteFile(path, []byte(strconv.Itoa(os.Getpid())), 0600)
-	return errors.Wrap(err, "writing pidfile")
+func setupControlServer(ctx context.Context, logger log.Logger, db *bbolt.DB, runGroup *run.Group, opts *launcher.Options) error {
```
directionless, line 302: Picking this place to comment, I feel weird about passing in runGroup. Talking aloud, `runLauncher is basically a main function. And I think of main functions as being about wiring together pieces. From that perspective something like:
```
controlServer, err := setupControlServer(...)
if err != nil { ... }
runGroup.Add(...)
```

doesn't feel wrong, but also feels kinda needless verbose, and creates that superfluous `controlServer`.

So I wonder about a helper to add things to the rungroup. Using a simplified example, something like https://go.dev/play/p/wLX_2qkaLsg

```
func main() {

	if err := rgHelper(newBroke()); err != nil {
		fmt.Println("ERR:", err)
	}

	if err := rgHelper(newFoo()); err != nil {
		fmt.Println("ERR:", err)
	}

}

func rungroupAdd(one, two string) {
	fmt.Printf("one: %s, two: %s\n", one, two)
}

func newFoo() (string, string, error) {
	return "foo", "bar", nil
}

func newBroke() (string, string, error) {
	return "", "", errors.New("broke")
}

func rgHelper(one, two string, err error) error {
	if err != nil {
		return err
	}

	rungroupAdd(one, two)
	return nil
}
```

Can be cleaned up a little with an interface around Execute and Interrupt

directionless (commented):
I'm tentatively onboard, but a couple things feel weird still. They might be fine, but I think we should ponder. See comments

PR #867, opened 2022-08-25
shutdown systray during auto update

James-Pickett:
added systray interrupt to launcher shutdown func called during auto updates so systray will shutdown as well


directionless (approved):
Seems like a reasonable quick PR

PR #870, opened 2022-08-27
Improve localserver id fetch timing

directionless:
To account for startup sequencing, we want to do the initial id structure population quickly. But it can slow down after an initial success. This expands the logic around how localserver to handle that case.


James-Pickett (dismissed):

directionless:
closing and re-opening to trigger a rebase and update the workflows

PR #874, opened 2022-08-31
Add request verification to localserver

directionless:
This is an implementation of signature verification on inbound requests. Instead of simply handing them to the `mux`, this creates something closer to an RPC interface. The inbound request is a `box` which is verified and re-written to a new request, before being passed to an authenticated mux.

It also brings in different server certs for production, review apps, and local development. This supports our normal developer pattern of starting a launcher for those environments. The switch between them is a simple check against hostnames, but I think it will work for now. 

Lastly, it cleans up some of the code around the krypto http middleware. Thats now split off from localserver, and in its own struct. I'm not totally sure I love this -- maybe it could be re-integrated with the localserver struct, or made purely functional.


ee/localserver/krypto-unwrap_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+			h := kbm.UnwrapV1Hander(makeTestHandler(t))
+			req := makeRequest(t, tt.boxParam)
+
+			rr := httptest.NewRecorder()
+			h.ServeHTTP(rr, req)
+
+			if tt.loggedErr != "" {
+				assert.Equal(t, http.StatusUnauthorized, rr.Code)
+				assert.Contains(t, logBytes.String(), tt.loggedErr)
```
James-Pickett, line 113: This is a really cool! There a have been times where I had a func that didn't return an error, but wanted to verify that it was logging them. This will be useful.

James-Pickett (approved):
Looks good! Ran this locally with auth disabled and /id worked.

ee/localserver/krypto-unwrap_test.go
directionless: I'm super pleased with myself for it.

PR #875, opened 2022-08-31
Add support for Windows for Desktop

James-Pickett:
* iterates over explorer.exe processes and creates a new desktop process for the owners if they don' t already have one
* consolidates some of the desktop functions


cmd/launcher/desktop/desktop.go
```diff
@@ -39,23 +40,12 @@ func handleSignals() {
...
 func exitWhenParentGone() {
-	ticker := time.NewTicker(2 * time.Second)
-	defer ticker.Stop()
-
-	f := func() {
-		if os.Getppid() <= 1 {
+	for ; true; <-time.NewTicker(2 * time.Second).C {
+		exists, err := process.PidExists(int32(os.Getppid()))
+		if err != nil || !exists {
 			fmt.Println("parent process is gone, exiting")
```
directionless, line 46: Might not be worth it, but we should probably `logger`,  and not just `fmt.Println`

ee/desktop/runtime/runner_linux.go
```diff
@@ -7,10 +7,6 @@ import (
 	"fmt"
 )
 
-func (r *DesktopUsersProcessesRunner) runConsoleUserDesktop() error {
+func (r *DesktopUsersProcessesRunner) runDesktopNative() error {
```
directionless, line 10: I find `native` somewhat misleading -- I usually use it to refer to native code and/or tools. And launcher is basically all native. (There's no emulation or VM things). I think this is about invoking it as a user? `runDesktopAsUser`, or even the old name.

ee/desktop/runtime/runner.go
```diff
@@ -109,3 +110,68 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
 		}
 	}
 }
+
+// waitForProcess adds 1 to DesktopUserProcessRunner.procsWg and runs a goroutine to wait on the process to exit.
+// The go routine will decrement DesktopUserProcessRunner.procsWg when it exits
+func (r *DesktopUsersProcessesRunner) waitOnProcessAsync(uid string, proc *os.Process) {
```
directionless, line 116: Is the intent here to track the processes per user?

Though somewhat elegant, one downside here is that we cannot track which process is stuck. Might still be an okay first pass.

ee/desktop/runtime/runner.go
```diff
@@ -109,3 +110,68 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
+		)
+
+		return false
+	}
+
+	// have running process
+	return true
+}
+
+func processExists(pid int) bool {
```
directionless, line 171: What's the chance of pid reuse?

ee/desktop/runtime/runner_windows.go
```diff
@@ -5,12 +5,138 @@ package runtime
...
+
+	return nil
+}
+
+// explorerProcesses returns a list of explorer processes whose
+// filepath base is "explorer.exe".
+func explorerProcesses() ([]*process.Process, error) {
+	var explorerProcs []*process.Process
+
+	procs, err := process.Processes()
```
directionless, line 73: Could use `ProcessesWithContext` with a timeout

ee/desktop/runtime/runner_test.go
```diff
@@ -20,13 +18,22 @@ import (
...
 	// To get around the issue mentioned above, build the binary first and set it's path as the executable path on the runner.
-	executablePath := filepath.Join(t.TempDir(), "desktop")
+	executablePath := filepath.Join(t.TempDir(), "desktop-test")
+
+	if runtime.GOOS == "windows" {
+		executablePath = fmt.Sprintf("%s.exe", executablePath)
+	}
+
 	err := exec.Command("go", "build", "-o", executablePath, "../../../cmd/launcher").Run()
 	require.NoError(t, err)
```
directionless, line 38: FWIW to get a useful error, I started doing things like:
```go
                                ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
                                defer cancel()
                                cmd := exec.CommandContext(ctx, boxerRB, "encode", rubyInFile, rubyOutFile)
                                out, err := cmd.CombinedOutput()
                                require.NoError(t, err, string(out))
```

cmd/launcher/desktop/desktop.go
```diff
@@ -39,23 +40,12 @@ func handleSignals() {
 
 // continuously monitor for ppid and exit if parent process terminates
 func exitWhenParentGone() {
-	ticker := time.NewTicker(2 * time.Second)
-	defer ticker.Stop()
-
-	f := func() {
-		if os.Getppid() <= 1 {
+	for ; true; <-time.NewTicker(2 * time.Second).C {
```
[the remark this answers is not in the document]
James-Pickett: the mysterious do...while() of go =)

cmd/launcher/desktop/desktop.go
```diff
@@ -39,23 +40,12 @@ func handleSignals() {
...
+		exists, err := process.PidExists(int32(os.Getppid()))
```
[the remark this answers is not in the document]
James-Pickett: nope, nice catch, updated

ee/desktop/runtime/runner_linux.go
James-Pickett: I can see that ... maybe `runDesktopForOs` or `runDesktopForPlatform` ? I kinda wanted something call out that this func has a different definition for macos, windows, & linux, but I can't think of a good name so went back to runConsoleUserDesktop().

cmd/launcher/desktop/desktop.go
```diff
@@ -39,23 +41,22 @@ func handleSignals() {
...
+			break
+		}
 
-	for {
-		select {
-		case <-ticker.C:
-			f()
+		ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		exists, err := process.PidExistsWithContext(ctx, int32(ppid))
+		cancel()
```
James-Pickett, line 53: @directionless , does this look right to you? Can't use `defer` in the for loop since it doesn't get called until the func returns.

ee/desktop/runtime/runner.go
James-Pickett: The intent here is to just wait on the process so that if it does die, the parent can clean it up and we don't get a zombie. If you wanted to test this you could not call this function then terminate the desktop process, a new one will not spin up because this one remains alive or at least undead.

I've updated the function comment with more info

The wait group was to satisfy the race detector.

ee/desktop/runtime/runner.go
James-Pickett: I don't know. I updated the func to take `*os.Process` as a param in hopes that `os.Process` tracks pid, otherwise I'm not sure how one would prevent or test for process reuse .... are you thinking we should go down the route of looking at some of the process attributes (like path?) to make sure it's actually a launcher desktop process?

ee/desktop/runtime/runner_linux.go
directionless: I would't worry about that being clear in the naming.

ee/desktop/runtime/runner.go
directionless: Old skool would be to check the path (or name), and assume it's good enough to head off a reuse issue. Probably won't be an issue for end user devices, but it was common for servers.

James-Pickett: well .... I cannot imagine how difficult it would be to debug a pid reuse issue without having a terminal to the machine so I went ahead and added some logic to track and compare the path as well

ee/desktop/runtime/runner_test.go
James-Pickett: cool, added this to tests as well as some fancy log comparison

ee/desktop/runtime/runner_windows.go
James-Pickett: done

cmd/launcher/desktop/desktop.go
James-Pickett: I prefer to kick the can on this one, there is already a lot going on in this PR and I don't see an easy way to pass any logger configs to the new process.

cmd/launcher/desktop/desktop.go
directionless: Do we care about `cancel()`? From https://pkg.go.dev/context#CancelFunc

> A CancelFunc tells an operation to abandon its work. A CancelFunc does not wait for the work to stop. A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.

I we don't need to trigger it outside the timeout, so: `ctx, _ = context.WithTimeout(...)`

cmd/launcher/desktop/desktop.go
```diff
@@ -39,23 +41,22 @@ func handleSignals() {
...
+		if err != nil || !exists {
+			break
 		}
 	}
+
+	fmt.Println("parent process is gone, exiting")
```
directionless, line 59: println :|

ee/desktop/runtime/runner.go
```diff
@@ -31,12 +33,41 @@ type DesktopUsersProcessesRunner struct {
 	executablePath string
 }
 
+func (r *DesktopUsersProcessesRunner) addProcessForUser(uid string, osProcess *os.Process) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
```
directionless, line 38: Do we need `cancel`?

ee/desktop/runtime/runner.go
```diff
@@ -31,12 +33,41 @@ type DesktopUsersProcessesRunner struct {
 	executablePath string
 }
 
+func (r *DesktopUsersProcessesRunner) addProcessForUser(uid string, osProcess *os.Process) error {
```
directionless, line 36:
```suggestion
// addProcessForUser adds process information to the internal tracking state
func (r *DesktopUsersProcessesRunner) addProcessForUser(uid string, osProcess *os.Process) error {
```

If I read this correct, it's adding the process info to the `DesktopUsersProcessesRunner` internal state? If so, a comment might help

ee/desktop/runtime/runner.go
```diff
@@ -31,12 +33,41 @@ type DesktopUsersProcessesRunner struct {
...
+	if err != nil {
+		return fmt.Errorf("creating process record: %w", err)
+	}
+
+	path, err := psutilProc.ExeWithContext(ctx)
+	if err != nil {
+		return fmt.Errorf("getting process path: %w", err)
+	}
+
+	r.uidProcs[uid] = processRecord{
```
directionless, line 50: Is this racy? I feel like modifying hashes always need a mutex.

ee/desktop/runtime/runner_test.go
```diff
@@ -19,32 +20,56 @@ import (
...
 func TestDesktopUserProcessRunner_Execute(t *testing.T) {
 	t.Parallel()
+	if runtime.GOOS == "linux" {
+		t.Skip("skipping linux test because it's not implemented")
+	}
 
 	// When running this using the golang test harness, it will leave behind process if you do not build the binary first.
 	// On mac os you can find these by setting the executable path to an empty string before running the tests, then search
 	// the processes in a terminal using: ps aux -o ppid | runtime.test after the tests have completed, you'll also see the
 	// CPU consumtion go way up.
```
directionless, line 30:
```suggestion
	// CPU consumption go way up.
```

ee/desktop/runtime/runner_test.go
```diff
@@ -80,6 +111,46 @@ func TestDesktopUserProcessRunner_Execute(t *testing.T) {
...
+						p.process.Wait()
+					}
+				}
+			})
 		})
 	}
 }
+
+// thank you zupa https://stackoverflow.com/a/36226525
+type threadSafeBuffer struct {
```
directionless, line 135: This is good to remember.

ee/desktop/runtime/runner_windows.go
```diff
@@ -4,13 +4,147 @@
...
+	"os/user"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/shirou/gopsutil/process"
 )
 
+// runDesktopNative iterates over all the current explorer processes and
```
directionless, line 20:
```suggestion
// runConsoleUserDesktop iterates over all the current explorer processes and
```

cmd/launcher/desktop/desktop.go
James-Pickett: you prefer `fmt.Print("\nparent process is gone, exiting")` or we need a logger?

directionless: I feel like we should figure out the logger before GA.

directionless: But we can not block this PR merge on it.

ee/desktop/runtime/runner.go
James-Pickett: this is actually not accessed concurrently, we iterate over users sequentially and add them here

directionless (approved):

PR #879, opened 2022-09-07
update go to 1.19

James-Pickett:
* ran go mod tidy
* ran gofmt on all the things
* set linter to latest version
* set action linter to latest version
* tested live query against local instance of k2
* tested using launcher interactive


directionless (approved):
`fmt` is surprisingly aggressive, but it all seems fine.

Thanks for doing this!

PR #880, opened 2022-09-08
James/add govulncheck

James-Pickett:
* adds govulncheck to pipeline https://go.dev/blog/vuln
* updates packages to fix issues found by govulncheck

example CI runs with govulncheck:

Issues Found: https://github.com/James-Pickett/launcher/runs/8252664210?check_suite_focus=true#step:8:14
Issues Fixed: https://github.com/James-Pickett/launcher/runs/8253007225?check_suite_focus=true#step:8:14

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -54,7 +54,7 @@ jobs:
       run: make deps
 
     - name: Run govulncheck
-      run: go install golang.org/x/vuln/cmd/govulncheck@latest; govulncheck -v ./...
+      run: go install golang.org/x/vuln/cmd/govulncheck@latest; govulncheck ./...
 
     - name: Build
       if: ${{ ! contains(matrix.os, 'ubuntu') }}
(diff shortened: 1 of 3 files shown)


seejdev (approved):

.github/workflows/go.yml
```diff
@@ -50,10 +50,12 @@ jobs:
         path: ${{ steps.go-cache-paths.outputs.go-mod }}
         key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
 
-
     - name: Get dependencies
       run: make deps
 
+    - name: Run govulncheck
+      run: go install golang.org/x/vuln/cmd/govulncheck@latest; govulncheck ./...
```
directionless, line 57: Feels cleaner as different lines, or different steps. But I'll take it. 

Side note, do you think in here, or in a new workflow parallel?

directionless (approved):
I'd take it here. I'd also take it in a new workflow. Though I guess I'd expect this to get folded into `golangci-lint`

.github/workflows/go.yml
James-Pickett: I _think_ here is better, it will fail fast, but it means if a new vuln is reported, it will stop all CI ... do we want that? We'll be forced to remediate things immediately. I think we do, but would like to hear any differing opinion.

directionless: I don't have strong feelings.  

Here should have good coverage in the different platforms. But it fails fast. 

Seems fine to call this good enough until it's in the main linter.

Issue #876, opened 2022-09-06
Table Request: `kolide_exec_investigation`


directionless:
When developing tables against third party commands (For example, #788), we are often stymied by not understanding what said command's output is. (is it structured? is it white space delimitated? Etc...) And while I've always been quite resistant to creating a general exec output capturing table, I think I've been convinced there's value in it.

Specifically:
* I think we should have a table that can exec something we're trying to debug
* And capture its output. Perhaps as a single base64 blob?
* We should not allow arbitrary execution -- like the rest of the exec functions, we should enumerate allowed commands. I'm imagining we'd have a column for the command `name`, and it would match something in a `[]string` of possible paths to try. (There's a helper for that)
* I am unsure how to handle command line args. I think we cannot accept an arbitrary one, so we probably just have to hardcode what's acceptable. Other suggestions?

I'm imagining that using this would require a couple launcher releases -- first one adds something to the allowed list, might need to iterate on the args, and then it's written into a table and removed.

PR #881, opened 2022-09-09
Adds dev_table_tooling to capture output of allowed commands

seejdev:
This PR addresses #876.

- `kolide_dev_table_tooling` table has a `name`, `args`, and `output` column
- Each platform has a hard-coded set of named allowed commands, which correspond to a strict list of binaries and arguments
- A test exercises the cases where a command name isn't provided, or is unrecognized
- I've supplied some hard-coded allowed commands purely as an example of how they would be specified in the code, these can be replaced with more suitable options, or removed entirely, prior to merging

### DIFF
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -1,0 +1,18 @@
+//go:build darwin
+// +build darwin
+
+package dev_table_tooling
+
+func GetAllowedCommands() map[string]AllowedCommand {
+	cmds := map[string]AllowedCommand{
+		"system_profiler": {
+			binPaths: []string{"/usr/sbin/system_profiler"},
+			args:     []string{"SPSoftwareDataType", "SPNetworkDataType"},
+		},
+		"diskutil": {
+			binPaths: []string{"diskutil"},
+			args:     []string{"info", "-all"},
+		},
+	}
+	return cmds
+}
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -3,6 +3,6 @@
 
 package dev_table_tooling
 
-var allowedCommands = map[string]AllowedCommand{
+var allowedCommands = map[string]allowedCommand{
 	// Add allowed commands here
 }
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -9,3 +9,6 @@ var allowedCommands = map[string]allowedCommand{
 		args:     []string{"hello"},
 	},
 }
+
+// Platform-specific test data
+var echoHelloOutput = "hello\n"
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -9,6 +9,3 @@ var allowedCommands = map[string]allowedCommand{
 		args:     []string{"hello"},
 	},
 }
-
-// Platform-specific test data
-var echoHelloOutput = "hello\n"
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -3,9 +3,6 @@
 
 package dev_table_tooling
 
-func GetAllowedCommands() map[string]AllowedCommand {
-	cmds := map[string]AllowedCommand{
-		// Add allowed commands here
-	}
-	return cmds
+var allowedCommands = map[string]AllowedCommand{
+	// Add allowed commands here
 }
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -4,5 +4,8 @@
 package dev_table_tooling
 
 var allowedCommands = map[string]allowedCommand{
-	// Add allowed commands here
+	"echo": {
+		binPaths: []string{"echo"},
+		args:     []string{"hello"},
+	},
 }
--- a/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
+++ b/pkg/osquery/tables/dev_table_tooling/commands_darwin.go
@@ -5,14 +5,7 @@ package dev_table_tooling
 
 func GetAllowedCommands() map[string]AllowedCommand {
 	cmds := map[string]AllowedCommand{
-		"system_profiler": {
-			binPaths: []string{"/usr/sbin/system_profiler"},
-			args:     []string{"SPSoftwareDataType", "SPNetworkDataType"},
-		},
-		"diskutil": {
-			binPaths: []string{"diskutil"},
-			args:     []string{"info", "-all"},
-		},
+		// Add allowed commands here
 	}
 	return cmds
 }
--- a/pkg/osquery/tables/dev_table_tooling/table.go
+++ b/pkg/osquery/tables/dev_table_tooling/table.go
@@ -48,7 +48,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 			continue
 		}
 
-		cmd, ok := GetAllowedCommands()[name]
+		cmd, ok := allowedCommands[name]
 
 		if !ok {
 			level.Info(t.logger).Log("msg", "Command not allowed", "name", name)
--- a/pkg/osquery/tables/dev_table_tooling/table.go
+++ b/pkg/osquery/tables/dev_table_tooling/table.go
@@ -3,6 +3,7 @@ package dev_table_tooling
 import (
 	"context"
 	"encoding/base64"
+	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -27,6 +28,7 @@ type Table struct {
 func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
 	columns := []table.ColumnDefinition{
 		table.TextColumn("name"),
+		table.TextColumn("args"),
 		table.TextColumn("output"),
 	}
 
@@ -61,6 +63,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 			} else {
 				results = append(results, map[string]string{
 					"name":   name,
+					"args":   strings.Join(cmd.args, " "),
 					"output": base64.StdEncoding.EncodeToString(output),
 				})
 			}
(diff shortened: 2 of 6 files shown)


pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/osquery/osquery-go"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+// Encapsulates the binary path(s) of a command allowed to execute
+// along with a strict list of arguments.
+type AllowedCommand struct {
```
directionless, line 17:
```suggestion
// AllowedCommand encapsulates the possible binary path(s) of a command allowed to execute
// along with a strict list of arguments.
type AllowedCommand struct {
```


I've never been sure if structs comments are supposed to start with the struct name (analogous to methods). https://tip.golang.org/doc/comment implies not, but does recommend slightly different phrasing. I'd reword slightly

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,74 @@
...
+	args     []string
+}
+
+type Table struct {
+	client    *osquery.ExtensionManagerClient
+	logger    log.Logger
+	tableName string
+}
+
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
```
directionless, line 28: As an FYI we're pretty inconsistent about whether we pass `client` to tables that don't need it. You can keep it, `_` it and not store it in the struct, or not pass it at all. (Some tables do need it, but most do not)

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,74 @@
...
+			level.Info(t.logger).Log("msg", "Command name must not be blank")
+			continue
+		}
+
+		cmd, ok := allowedCommands[name]
+
+		if !ok {
+			level.Info(t.logger).Log("msg", "Command not allowed", "name", name)
+			continue
+		} else {
```
directionless, line 58:
```suggestion
```

With early returns, we can skip the nesting. (deep nesting is hard to read)

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,74 @@
...
+func TablePlugin(client *osquery.ExtensionManagerClient, logger log.Logger) *table.Plugin {
+	columns := []table.ColumnDefinition{
+		table.TextColumn("name"),
+		table.TextColumn("args"),
+		table.TextColumn("output"),
+	}
+
+	t := &Table{
+		client:    client,
+		logger:    logger,
```
directionless, line 37:
```suggestion
		logger:    log.With(logger, "table", "kolide_dev_table_tooling"),
```

Will plumb that through for all subsequent output.

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,74 @@
...
+// along with a strict list of arguments.
+type AllowedCommand struct {
+	binPaths []string
+	args     []string
+}
+
+type Table struct {
+	client    *osquery.ExtensionManagerClient
+	logger    log.Logger
+	tableName string
```
directionless, line 25: You can leave this, or just pass it to NewPlugin.

pkg/osquery/tables/dev_table_tooling/table_test.go
```diff
@@ -0,0 +1,44 @@
...
+		commandName    []string
+		expectedResult []map[string]string
+	}{
+		{
+			name: "no command name",
+		},
+		{
+			name:        "malware",
+			commandName: []string{"ransomware.exe"},
+		},
```
directionless, line 26: Might be worth adding some kind of happy path testing. Which raises the question of what command. I suspect `echo hello` might be okay, can't quickly double check that's not a shell built on linux, but it should work on mac and windows.

directionless (commented):
Table looks really good. A couple of pretty small nits, only thing that I'd really push for is removing the `else` blocks on the early returns.

pkg/osquery/tables/dev_table_tooling/table.go
James-Pickett, line 17: I think this can be a un-exported
```suggestion
type allowedCommand struct {
```

pkg/osquery/tables/dev_table_tooling/commands_darwin.go
```diff
@@ -0,0 +1,14 @@
...
+
+var allowedCommands = map[string]allowedCommand{
+	"echo": {
+		binPaths: []string{"echo"},
+		args:     []string{"hello"},
+	},
+}
+
+// Platform-specific test data
+var echoHelloOutput = "hello\n"
```
James-Pickett, line 14: Feels a little odd to me to have a variable only used for testing outside of a `_test.go` file. I would declare a var in `table_test.go` as a map of os -> output inside the `Test_generate()` like:

```
expectedOutputs := map[string]string{
  "windows": "hello\n",
  "darwin": "hello\n",
  "linux": "hello\r\n",
}
```

then access via

```
expectedOutput := expectedOutputs[runtime.GOOS]
```

inside the test case

James-Pickett (commented):
looking good, very close, left comment about test data organization

pkg/osquery/tables/dev_table_tooling/commands_darwin.go
seejdev: Yeah I didn't like the messiness of my first approach - your suggestion is definitely an improvement.

I reworked the test to avoid needing to know about newline characters in [5cc1059e](https://github.com/kolide/launcher/pull/881/commits/5cc1059e4dc0e29ace806ffed8719c8fa6a8589c)

It does add some complexity to the test verification code, but makes the tests easier to specify and maintain.

pkg/osquery/tables/dev_table_tooling/table_test.go
```diff
@@ -0,0 +1,67 @@
...
+			got, _ := table.generate(context.Background(), tablehelpers.MockQueryContext(constraints))
+
+			if len(tt.expectedResult) > 0 {
+				assert.Equal(t, tt.expectedResult[0]["name"], got[0]["name"])
+				assert.Equal(t, tt.expectedResult[0]["args"], got[0]["args"])
+
+				// To verify output, let's convert back to utf8
+				decodedOutput, _ := base64.StdEncoding.DecodeString(got[0]["output"])
+				scanner := bufio.NewScanner(bytes.NewReader(decodedOutput))
+				for scanner.Scan() {
```
James-Pickett, line 56: that's a neat way to normalize the output

pkg/osquery/tables/dev_table_tooling/table_test.go
```diff
@@ -0,0 +1,67 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			constraints := make(map[string][]string)
+			constraints["name"] = tt.commandName
+
+			got, _ := table.generate(context.Background(), tablehelpers.MockQueryContext(constraints))
+
+			if len(tt.expectedResult) > 0 {
```
James-Pickett, line 49: let's invert this if statement to reduce nesting

```suggestion
			if len(tt.expectedResult) <= 0 {
			    assert.ElementsMatch(t, tt.expectedResults, got)
			    return
			}
			
			// test for expected results
```

James-Pickett (dismissed):
LGTM! Lets give @directionless a chance to take a look before merging.

pkg/osquery/tables/dev_table_tooling/table_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_generate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name           string
+		commandName    []string
```
directionless, line 20: Why is this an array? Is it because (in theory) we could pass multiple commands? If so, `commandNames`

pkg/osquery/tables/dev_table_tooling/table_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+			constraints := make(map[string][]string)
+			constraints["name"] = tt.commandName
+
+			got, _ := table.generate(context.Background(), tablehelpers.MockQueryContext(constraints))
+
+			if len(tt.expectedResult) <= 0 {
+				assert.ElementsMatch(t, tt.expectedResult, got)
+				return
+			}
```
directionless, line 52:
```suggestion
			if len(tt.expectedResult) <= 0 {
				require.Empty(t, got)
				return
			}
```

I think `Empty` conveys the intent. (Though you could leave `assert`)

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,73 @@
...
+		cmd, ok := allowedCommands[name]
+
+		if !ok {
+			level.Info(t.logger).Log("msg", "Command not allowed", "name", name)
+			continue
+		}
+
+		output, err := tablehelpers.Exec(ctx, t.logger, 30, cmd.binPaths, cmd.args)
+		if err != nil {
+			level.Info(t.logger).Log("msg", "dev_table_tooling failed", "name", name, "err", err)
```
directionless, line 61:
```suggestion
			level.Info(t.logger).Log("msg", "execution failed", "name", name, "err", err)
```

directionless (approved):
I'd take this as-is. I left some smallish nits and clarity suggestions

directionless:
closing and reopening to kick the CLA tool

pkg/osquery/tables/dev_table_tooling/table_test.go
seejdev: Good catch, no need for an array here

directionless (approved):

James-Pickett (dismissed):

James-Pickett (approved):

FritzX6:
### Wishlist of commands:

**Azure AD binding state:**
```
dsregcmd /status
```

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	for _, name := range tablehelpers.GetConstraints(queryContext, "name", tablehelpers.WithDefaults("")) {
+		if name == "" {
+			level.Info(t.logger).Log("msg", "Command name must not be blank")
+			continue
+		}
```
directionless, line 48: This is unlikely to trigger -- if name is nil, we won't enter the forloop. But there's no harm in it.

directionless (approved):

James-Pickett (approved):

PR #883, opened 2022-09-12
address performance issue caused by re instantiating new ticker in for loop condition

James-Pickett:
declares exitWhenParentGone() ticker out side of for loop condition, declaring in condition was causing causes ticker to be continuously recreated


directionless (approved):
Nice find.

seejdev (approved):

PR #891, opened 2022-09-15
James/desktop ipc

James-Pickett:
* adds inter process communication over a unix socket or named pipe using http
* adds socket path to debug menu

image


ee/desktop/runtime/runner_posix.go
```diff
@@ -0,0 +1,17 @@
+//go:build darwin || linux
```
James-Pickett, line 1: is the name of this file `runner_posix.go` appropriate? Was trying to find something other than `runner_notwindows.go`

ee/desktop/server/server.go
```diff
@@ -0,0 +1,71 @@
...
+
+	if err := os.RemoveAll(socketPath); err != nil {
+		//TODO: log this
+	}
+
+	if err := server.Shutdown(context.Background()); err != nil {
+		//TODO: log this
+	}
+
+	systray.Quit()
```
James-Pickett, line 62: wonder if this pattern is correct .... so the server be calling `systray.Quit()` or should it accept a channel as a parameter and send on the channel when it gets a shut down signal and let the calling process deal with it? Something else?

cmd/launcher/desktop/desktop.go
```diff
@@ -49,6 +57,7 @@ func RunDesktop(args []string) error {
 		systray.AddSeparator()
 		systray.AddMenuItem("--- DEBUG ---", "").Disable()
 		systray.AddMenuItem(fmt.Sprintf("Hostname: %s", *flhostname), "").Disable()
+		systray.AddMenuItem(fmt.Sprintf("Socket Path: %s", desktop.DesktopSocketPath(os.Getpid())), "").Disable()
```
directionless, line 60: Probably don't want to keep this long term. Seems fine for now

ee/desktop/desktop.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+func DesktopSocketPath(pid int) string {
+	const socketBaseName = "launcher_desktop.sock"
```
directionless, line 11:
```suggestion
	const socketBaseName = "kolide_desktop.sock"
```

ee/desktop/desktop.go
```diff
@@ -0,0 +1,18 @@
+package desktop
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+func DesktopSocketPath(pid int) string {
```
directionless, line 10: This is a bit of a can of worms, and I'm not sure what correct is. To talk aloud....

On windows, we have to be in `\\.pipe`, okay. We'll just need to pick a good name and roll with it.  Maybe add something to detect collisions. Not sure.

On not-windows, it feels weird to use TempDir. Nothing really cleans it up, just feels wrong. I'm not sure what feels right.

In one world, launcher would create the pipe somewhere, chown/chmod it to the user, and then hand it to desktop for startup. Other worlds might be creating a user writable run dir.

ee/desktop/runtime/runner.go
```diff
@@ -119,16 +119,10 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
 
-	signal := os.Interrupt
-	// os.Interrupt is not supported on windows, so use os.Kill instead
-	if runtime.GOOS == "windows" {
-		signal = os.Kill
-	}
-
 	for uid, proc := range r.uidProcs {
-		if err := proc.process.Signal(signal); err != nil {
+		if err := sendShutdownCommand(proc.process.Pid); err != nil {
```
directionless, line 123: Might want to signal _after_ sendShutdownCommand

ee/desktop/runtime/runner_posix.go
directionless: Um, probably. Sure!

But I'd name it `dialer_xxxxx.go` (or `dial_context.go`), since it's just the dial function

consider maybe instead of calling `desktop.DesktopSocketPath(pid)` to pass the pid as an argument. (Or likely, as suggested above, launcher will set the path, and handoff to desktop anyhow)

ee/desktop/server/server.go
```diff
@@ -0,0 +1,71 @@
+package server
```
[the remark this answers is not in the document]
directionless: I don't think it should be `ee/server`, and I am agnostic about whether it's in `desktop/server` or `desktop`. Whatever feels cleaner to you

ee/desktop/server/server.go
```diff
@@ -0,0 +1,71 @@
...
+	mux := http.NewServeMux()
+	mux.HandleFunc("/shutdown", shutdown)
+
+	server := http.Server{
+		Handler: mux,
+	}
+
+	socketPath := desktop.DesktopSocketPath(os.Getpid())
+
+	if err := os.RemoveAll(socketPath); err != nil {
```
directionless, line 36: Feels weird. Might be correct.

ee/desktop/server/server.go
directionless: This feels weird. And maybe dovetails with your package comment. If the server is over here, then it shouldn't be calling `systray.Quit()`. It should have something like a rungroup it triggers shutdown for.

Which makes me wonder about moving a bunch of this into main. Main would create a rungroup or shutdown channel, and then start a server using it. (launcher kinda does that)

ee/desktop/runtime/runner.go
James-Pickett: It will a kill signal after the sendShutdownCommand if the proc is still running after the time out has elapsed. Is this what you meant?

ee/desktop/desktop.go
James-Pickett: what about `$HOME/.kolide/` and put the socket there? launcher could create the folder and give user privs.

ee/desktop/runtime/runner.go
directionless: Yeah, something like that. If the process doesn't exit cleanly when we send it a shutdown command, what's next?

ee/desktop/desktop.go
directionless: If we stored user preferences we could think about it. But it feels like bad form for this.

ee/desktop/runtime/runner.go
James-Pickett: we are doing that, a little lower in the `for select` statement we send a kill after a tmeout

cmd/launcher/desktop.go
```diff
@@ -0,0 +1,156 @@
...
+	level.Info(logger).Log("msg", "starting")
+
+	shutdownChan := make(chan struct{})
+
+	go handleSignals(logger, shutdownChan)
+	go monitorParentProcess(logger, shutdownChan)
+
+	var runGroup run.Group
+
+	server, err := server.New(logger, *flauthtoken, desktop.DesktopSocketPath(os.Getpid()), shutdownChan)
```
directionless, line 52: I think I'm still very unsure about socket path creation. I _suspect_ we should have launcher create them, and pass them via an arg.

I'm not wholly sure why I think this. Just an instinct that this sort of assumption about paths and determinism, usually fails somewhere. (For example, on windows, we moved to using a random path for the thrift socket, because sometimes the old path was weirdly stale)

James-Pickett: finally working now to address this for windows we can just do the same as we do for launcher with some random string on the end of the socket file name

for the users should we create a specific folder for the user in the launcher root dir and give the user privs, then they can create the socket there? 

might look something like:
```
├── launcherRoot
│   ├── desktop_501
│   │   └── desktop.sock
│   └── desktop_502
│       └── desktop.sock
```

ee/desktop/runtime/runner.go
```diff
@@ -1,10 +1,11 @@
+// runtime handles multiuser process management for launcher desktop
```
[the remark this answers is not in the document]
James-Pickett: I have split it into desktop/runner and desktop/client

cmd/launcher/desktop.go
```diff
@@ -0,0 +1,156 @@
...
+				"err", err,
+			)
+			break
+		}
+	}
+
+	parentGoneChan <- struct{}{}
+}
+
+func setFlags(flagSet flag.FlagSet, args []string) error {
```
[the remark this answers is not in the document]
James-Pickett: now using ff

cmd/launcher/desktop/desktop.go
James-Pickett: removed

ee/desktop/server/server.go
James-Pickett: I'm not 100% sure either, but it feels simplest to just let the dialer create the socket

ee/desktop/server/server_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+			assert.Empty(t, logBytes.String())
+			assert.Equal(t, http.StatusOK, rr.Code)
+
+			require.NoError(t, server.Shutdown(context.Background()))
+		})
+	}
+}
+
+func testHandler(t *testing.T) http.Handler {
```
James-Pickett, line 113: `t` parameter is unused

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,386 @@
...
+func (r *DesktopUsersProcessesRunner) Execute() error {
+	f := func() {
+		if err := r.runConsoleUserDesktop(); err != nil {
+			level.Error(r.logger).Log("msg", "running console user desktop", "err", err)
+		}
+	}
+
+	f()
+
+	ticker := time.NewTicker(r.updateInterval)
```
James-Pickett, line 139:
```suggestion
	ticker := time.NewTicker(r.updateInterval)
	defer ticker.Close()
```

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,424 @@
...
+	if len(path) > maxSocketLength {
+		return "", fmt.Errorf("socket path %s (length %d) is too long, max is %d", path, len(path), maxSocketLength)
+	}
+
+	return path, nil
+}
+
+func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socketPath string) (*exec.Cmd, error) {
+
+	cmd := exec.Command(executablePath, "desktop")
```
directionless, line 398: General bias towards exec.CommandContext, with cancelable contexts. But maybe that's not as important here. Not sure?

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,424 @@
...
+	cmd.Env = r.processEnvVars(socketPath)
```
directionless, line 399: Why break this into its own function?

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,424 @@
...
+// WithHostname sets the hostname for the runner
+func WithHostname(hostname string) DesktopUsersProcessesRunnerOption {
+	return func(r *DesktopUsersProcessesRunner) {
+		r.hostname = hostname
+	}
+}
+
+func WithLogger(logger log.Logger) DesktopUsersProcessesRunnerOption {
+	return func(r *DesktopUsersProcessesRunner) {
+		r.logger = logger
```
directionless, line 35: consider `log.With` here.

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,424 @@
...
+		return nil, fmt.Errorf("getting stdout pipe: %w", err)
+	}
+
+	go func() {
+		combined := io.MultiReader(stdErr, stdOut)
+		scanner := bufio.NewScanner(combined)
+
+		for scanner.Scan() {
+			level.Info(r.logger).Log(
+				"uid", uid,
```
directionless, line 417:
```suggestion
			level.Info(r.logger).Log(
				"subprocess", "desktop",
				"uid", uid,
```

ee/desktop/server/server.go
```diff
@@ -0,0 +1,151 @@
...
+		authToken:    authToken,
+		logger:       log.With(logger, "component", "desktop_server"),
+		socketPath:   socketPath,
+	}
+
+	authedMux := http.NewServeMux()
+	authedMux.HandleFunc("/shutdown", desktopServer.shutdownHandler)
+
+	mux := http.NewServeMux()
+	mux.Handle("/", desktopServer.authMiddleware(authedMux))
```
directionless, line 41: Does this need a default route? I forget what mux does as a catch all

ee/desktop/server/server_test.go
```diff
@@ -0,0 +1,146 @@
+// server is a http server that listens to a unix socket or named pipe for windows.
+// Its implementation was driven by the need for "launcher proper" to be able to
+// communicate with launcher desktop running as a separate process.
```
directionless, line 3:
```suggestion
```

directionless (commented):
There's a lot to try to absorb, I think I'm missing some overall understanding and flow. Anyhow, lots of small comments. Looks pretty solid. I'm kinda unsure about how we're handling socket paths, but I feel okay as-is for beta.

ee/desktop/runner/runner_darwin.go
```diff
@@ -0,0 +1,124 @@
...
+			"uid", consoleOwnerUid,
+		)
+
+		return []string{}, nil
+	}
+
+	return []string{consoleOwnerUid}, nil
+}
+
+func consoleOwnerUid() (string, error) {
```
James-Pickett, line 50: @directionless , this is where I changed the way we look for the current console user.

ee/desktop/runner/runner.go
James-Pickett: I think in this situation we do not want to use a context because we are managing the process ourselves. A call to the [cancel func ](https://pkg.go.dev/os/exec#CommandContext)sends an os.kill signal (which we do on interrupt) and we don't want a time out since the process is supposed to run indefinitely.

ee/desktop/server/server.go
James-Pickett: ahh cool, can pass authed mux as handler wrapped in middleware

ee/desktop/runner/runner_darwin.go
```diff
@@ -0,0 +1,124 @@
...
+	const cmd = `echo "show State:/Users/ConsoleUser" | scutil`
+	output, err := exec.Command("sh", "-c", cmd).CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("getting console user: %w", err)
+	}
```
directionless, line 55: Several comments

1. You shouldn't use an `sh` exec. 
2. Instead of `echo ...` just set stdin.
3. You should use a exec.CommandContext with timeout

James-Pickett: updated ... much better

ee/desktop/runner/runner.go
seejdev: 👍 Makes sense to me

seejdev (dismissed):

seejdev:
Nice job! It's a beefy PR but the concerns seem to be well separated, and the process/thread lifetime management is pretty easy to follow.

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,426 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/ulid"
+	"github.com/kolide/launcher/ee/desktop/client"
+	"github.com/shirou/gopsutil/process"
+)
+
+type DesktopUsersProcessesRunnerOption func(*DesktopUsersProcessesRunner)
```
directionless, line 24: You shouldn't need to export this

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,426 @@
...
+
+// WithInterruptTimeout sets the timeout for the runner to wait for processes to exit when interrupted.
+func WithInterruptTimeout(timeout time.Duration) DesktopUsersProcessesRunnerOption {
+	return func(r *DesktopUsersProcessesRunner) {
+		r.interruptTimeout = timeout
+	}
+}
+
+// WithAuth sets the auth token for the runner
+func WithAuthToken(token string) DesktopUsersProcessesRunnerOption {
```
directionless, line 64:
```suggestion
// WithAuthToken sets the auth token for the runner
func WithAuthToken(token string) DesktopUsersProcessesRunnerOption {
```

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+}
+
+func (r *DesktopUsersProcessesRunner) runConsoleUserDesktop() error {
+	executablePath, err := r.determineExecutablePath()
+	if err != nil {
+		return fmt.Errorf("determining executable path: %w", err)
+	}
+
+	consolerUsers, err := r.consoleUsers()
```
directionless, line 218: Is the`r` a typo?

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,426 @@
...
+// the process dies and we do not wait for it, it will live as a zombie and not get cleaned up by the parent.
+// The wait group is needed to prevent races.
+func (r *DesktopUsersProcessesRunner) waitOnProcessAsync(uid string, proc *os.Process) {
+	r.procsWg.Add(1)
+	go func(username string, proc *os.Process) {
+		defer r.procsWg.Done()
+		// waiting here gives the parent a chance to clean up
+		state, err := proc.Wait()
+		if err != nil {
+			level.Error(r.logger).Log(
```
directionless, line 293:
```suggestion
			level.Info(r.logger).Log(
```

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,426 @@
...
+	return path, nil
+}
+
+func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socketPath string) (*exec.Cmd, error) {
+	cmd := exec.Command(executablePath, "desktop")
+
+	const varFmt = "%s=%s"
+	cmd.Env = append(
+		os.Environ(),
+		fmt.Sprintf(varFmt, "HOSTNAME", r.hostname),
```
directionless, line 397:
```suggestion
		fmt.Sprintf("HOSTNAME=%s", r.hostname),
```

Feels cleaner, but 🤷 seems fine too

directionless (commented):
I think I'm basically good with this. It's a solid iteration. 

Some nits, and I think it's worth fixing the spelling things

ee/desktop/server/server.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (s *DesktopServer) Shutdown(ctx context.Context) error {
+	err := s.server.Shutdown(ctx)
+	if err != nil {
+		return err
+	}
+
+	// on windows we need to expliclty close the listener
+	// on non-windows this gives an error
```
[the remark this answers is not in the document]
James-Pickett: Windows won't let you delete a file that is open somewhere. I was getting errors in the test clean up even after shutting down the server, something still held on to the named pipe. I don't see it document anywhere, just had to learn via trial and error.

ee/desktop/runner/runner.go
James-Pickett: not a typo, but there is no reason to hang consoleUsers() on the runner struct, it's now un hanged

directionless: Heh. I didn't mean `r.` I mean `consolerUsers`. What's with the r in the middle?

James-Pickett: oops, fixed =)

directionless: I keep wanting it to be like `sommelier`, someone who can show me exquisitely crafted consoles from around the world. 🤣

ee/desktop/runner/runner.go
```diff
@@ -0,0 +1,429 @@
...
+	}
+
+	return path, nil
+}
+
+func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socketPath string) (*exec.Cmd, error) {
+	cmd := exec.Command(executablePath, "desktop")
+
+	cmd.Env = append(
+		os.Environ(),
```
directionless, line 399: I wonder if we should _not_ pass this

directionless (approved):
This it huge, and looks like a bunch of great work. Let's goooo!

I feel like I could keep finding small things, but I think it's going to be better to merge this and iterate on what proves to be an issue.

directionless (approved):

PR #892, opened 2022-09-16
Improved kolide_airport_util error message when no option provided

### LINKED ISSUE #857: List valid options for kolide_airport_util
When you query `kolide_airport_util` you get this:

```
osquery> select * from kolide_airport_util;
Error: error generating table: The kolide_airport_util table requires that you specify a constraint for option
```

The error message should list the valid options so the user is not left to guess.

The error message is generated here: https://github.com/kolide/launcher/blob/master/pkg/osquery/tables/airport/table_darwin.go#L77

seejdev:
Fixes #857

Small logging improvement for an error case.

### DIFF
--- a/pkg/osquery/tables/airport/table_darwin.go
+++ b/pkg/osquery/tables/airport/table_darwin.go
@@ -74,7 +74,7 @@ func generateAirportData(queryContext table.QueryContext, airportExecutor execut
 	options := tablehelpers.GetConstraints(queryContext, "option", tablehelpers.WithAllowedValues(allowedOptions))
 
 	if len(options) == 0 {
-		return nil, errors.Errorf("The %s table requires that you specify a constraint for option", tableName)
+		return nil, errors.Errorf("The %s table requires that you specify a constraint for WHERE option =. Valid values for option are (%s).", tableName, strings.Join(allowedOptions, ", "))
 	}
 
 	var results []map[string]string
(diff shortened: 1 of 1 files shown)


pkg/osquery/tables/airport/table_darwin.go
```diff
@@ -74,7 +74,7 @@ func generateAirportData(queryContext table.QueryContext, airportExecutor execut
 	options := tablehelpers.GetConstraints(queryContext, "option", tablehelpers.WithAllowedValues(allowedOptions))
 
 	if len(options) == 0 {
-		return nil, errors.Errorf("The %s table requires that you specify a constraint for option", tableName)
+		return nil, errors.Errorf("The %s table requires that you specify a constraint for WHERE option =. Valid values for option are (%s).", tableName, strings.Join(allowedOptions, ", "))
```
directionless, line 77: I'd put this into a log line -- errors don't always make it back to the log.

directionless: Correction, it does bubble up. So let's take it. 

```suggestion
		return nil, errors.Errorf("The %s table requires that you specify a constraint for WHERE option. Valid values for option are (%s).", tableName, strings.Join(allowedOptions, ", "))
```

directionless (approved):

James-Pickett (dismissed):

directionless (approved):

PR #894, opened 2022-09-16
Add Crowdstrike Linux table

goronfreeman:


pkg/osquery/tables/crowdstrike/table.go
```diff
@@ -0,0 +1,68 @@
...
+type Table struct {
+	name   string
+	logger log.Logger
+}
+
+const tableName = "kolide_crowdstrike"
+
+var re = regexp.MustCompile(`\s?=\s?|\s`)
+
+func TablePlugin(logger log.Logger) *table.Plugin {
```
directionless, line 25: If there will be two tables in here, these should be named somewhat more distinctly. 

Also, probably don't need to export the struct.

pkg/osquery/tables/crowdstrike/table.go
```diff
@@ -0,0 +1,68 @@
...
+
+		// TODO: Replace this with the output of the real command
+		output := "aid is not set, aph is not set, app is not set, rfm-state is not set, rfm-reason is not set, feature is not set, metadata-query=enable (unset default), version = 6.38.13501.0"
+		dataArr := strings.Split(output, ", ")
+		rawKeyVals := make(map[string]interface{})
+
+		for _, data := range dataArr {
+			a := re.Split(data, 2)
+			rawKeyVals[a[0]] = a[1]
+		}
```
directionless, line 52: I would recommend putting this parsing into a dedicated function that we can then test.

PR #897, opened 2022-09-19
adds instructions and files for vs code debugging - v0.2

James-Pickett:


tools/vscode-debugging/conf/launch.json
```diff
@@ -0,0 +1,49 @@
...
+                "--osquery_flag=allow-unsafe",
+                "--osquery_flag=verbose",
+                // add atc config to the path below to load atc config
+                "--osquery_flag=config_path=${workspaceFolder}/debug/kolide-atc-config.json",
+            ],
+            "console": "integratedTerminal",
+            "asRoot": true
+        },
+        {
+            "name": "k2",
```
directionless, line 25: I'd probably call this `k2-localhost`

tools/vscode-debugging/conf/pre-debug.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p ./debug
+make deps
+make build
```
directionless, line 7: Maybe a silly question, but why build here? I would have assume the vscode is doing some debug/repl/thing, and not using this?

tools/vscode-debugging/conf/launch.json
```diff
@@ -0,0 +1,49 @@
...
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "preLaunchTask": "pre-debug",
```
directionless, line 29: pre-debug is a shell script. What do you think about having it detect whether the enroll secret and cert are present and error out? (I don't know if it can produce a clear error that bubbles through)

tools/vscode-debugging/conf/pre-debug.sh
James-Pickett: I think this was left over from the need for the extension build

tools/vscode-debugging/conf/launch.json
James-Pickett: It's a little tricky because only need that stuff when running k2-localhost. It's not required for interactive. So then we have to have a condition in there and trying to pass the condition from the `launch.json` to the `task.json` (that runs before the debugger starts) is surprisingly difficult and requires many files and texts. Or I just haven't figured out the easy way. 

Also, the error isn't really in your face, you have to click the task's terminal tab to see it's output.

I would argue that launcher it self should emit a meaningful message if missing these files.

James-Pickett: I was actually able to drop the `pre-debug.sh` after realizing I could just call `make deps` in the task and launcher handles making it's own dirs

directionless: > I would argue that launcher it self should emit a meaningful message if missing these files.

I could believe that. We can improve as we need them.

tools/vscode-debugging/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Now you should be able to set break points in VS Code and hit them by executing queries.
+
+## Debugging Against Local K2 Server (only available to Kolide employees)
+
+### First time setup
+
+1. Log into your local instance of K2 > Inventory > Add Device, the enroll_secret will be displayed in the launcher command
+1. Save your enroll_secret to `./debug/k2_enroll_secret`
+1. Either copy the {k2-repo}/tmp/localhost.crt to `./debug/localhost.crt` or create a sym like from `{k2-repo}/tmp/localhost.crt` -> `./debug/localhost.crt` at the root of your launcher repository
```
directionless, line 26:
```suggestion
1. Either copy the {k2-repo}/tmp/localhost.crt to `./debug/localhost.crt` or create a symlink from `{k2-repo}/tmp/localhost.crt` -> `./debug/localhost.crt`
```

tools/vscode-debugging/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Prerequisites
+
+1. [Install VS Code](https://code.visualstudio.com/download)
+1. [Install VS Code Go Extension](https://code.visualstudio.com/docs/languages/go)
+1. Osqueryd is available in your path or a common location
+1. Open launcher repo with VS Code
+1. Copy ./tools/vscode-debug/conf to ./.vscode (`mkdir -p ./.vscode && cp -r ./tools/vscode-debugging/conf/. ./.vscode`)
+* if this is your first time using the VS Code go extension, you'll be prompted to install various go packages when you start debugging
+
+## Debugging With Osquery Interactive
```
directionless, line 12:
```suggestion
## Debugging With Launcher/Osquery Interactive
```

tools/vscode-debugging/conf/launch.json
```diff
@@ -0,0 +1,49 @@
...
+            "program": "${workspaceFolder}/cmd/launcher",
+            "args": [
+                "interactive",
+                "--osquery_flag=allow-unsafe",
+                "--osquery_flag=verbose",
+                // add atc config to the path below to load atc config
+                "--osquery_flag=config_path=${workspaceFolder}/debug/kolide-atc-config.json",
```
directionless, line 19: Do you think we should include this? I might just omit it, and advocate we fold it into `interactive`

tools/vscode-debugging/conf/launch.json
```diff
@@ -0,0 +1,49 @@
...
+        {
+            "name": "interactive",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "preLaunchTask": "pre-debug",
+            "program": "${workspaceFolder}/cmd/launcher",
+            "args": [
+                "interactive",
+                "--osquery_flag=allow-unsafe",
```
directionless, line 16: Do we still need this? I'd expect launcher to handle it

directionless (approved):
I see a typo, and a couple of nit you could push back on. But fix the typo :)

tools/vscode-debugging/README.md
James-Pickett: fixed

tools/vscode-debugging/conf/launch.json
James-Pickett: removed

tools/vscode-debugging/conf/launch.json
James-Pickett: honestly, I don't remember why I put that there, launcher interactive seems to work without it, removed

directionless: I would assume it dates to when we invoked osquery interactive with tables.ext. But launcher interaction should abstract it.

directionless (approved):

Issue #873, opened 2022-08-30
Table Request: dsregcmd (Azure AD State)


FritzX6:
## What is being requested?
A new table that parses the output of the `dsregcmd` CLI tool on Windows devices.
![image](https://user-images.githubusercontent.com/931724/187509288-881f9b20-0790-4cb8-be09-06450986d6b8.png)
Above Screenshot taken from: https://www.manishbangia.com/how-to-register-windows-10-device-to-azure-ad/

## Why would this be helpful?
Frequently organizations wish to verify the Azure AD join/connection state of a device. The output of this command allows us to determine the connection state:

| AzureAdJoined | EnterpriseJoined | DomainJoined | Device state |
|---|---|---|---|
| YES | NO | NO | Azure AD Joined |
| NO | NO | YES | Domain Joined |
| YES | NO | YES | Hybrid AD Joined |
| NO | YES | YES | On-premises DRS Joined |

https://docs.microsoft.com/en-us/azure/active-directory/devices/troubleshoot-device-dsregcmd?source=recommendations

## How can this be accomplished?
The `dsregcmd` produces semi-structured output which can be parsed. A sample of this output is included below.

I propose that we return the data in a pseudo-dataflatten pattern, splitting on the large section dividers (eg. `----------+`:

eg.
```
+----------------------------------------------------------------------+
| Device State                                                         |
+----------------------------------------------------------------------+

             AzureAdJoined : NO
          EnterpriseJoined : NO
              DomainJoined : NO
               Device Name : lenovo
```
would be rendered as something like:
```
parent = 'Device State'
   key = 'AzureAdJoined"
 value = 'NO'
 
parent = 'Device State'
   key = 'EnterpriseJoined"
 value = 'NO'
 
parent = 'Device State'
   key = 'DomainJoined"
 value = 'NO'
 
parent = 'Device State'
   key = 'Device Name"
 value = 'lenovo'
```

## Unconfigured Device Example Data:
```
PS C:\Windows\system32> dsregcmd /status

+----------------------------------------------------------------------+
| Device State                                                         |
+----------------------------------------------------------------------+

             AzureAdJoined : NO
          EnterpriseJoined : NO
              DomainJoined : NO
               Device Name : lenovo

+----------------------------------------------------------------------+
| User State                                                           |
+----------------------------------------------------------------------+

                    NgcSet : NO
           WorkplaceJoined : NO
             WamDefaultSet : NO

+----------------------------------------------------------------------+
| SSO State                                                            |
[...]

PR #901, opened 2022-09-27
Add a new exec table abstraction, and a `dsregcmd` table

directionless:
While writing an exec-and-parse table, something finally clicked, and I see how to make a clean abstraction.

This introduces `dataflattentable.NewExecAndParseTable`, which is very similar to `TablePluginExec`, but instead of requiring a command to output structured data, it allows us to pass a parser in directly. This should reduce the amount of boiler plate needed to make exec-and-parse tables. I hope to be able to replace several tables with smaller parses used by this. Left unimplemented is an abstraction for mapping column constraints to different exec calls. I suspect that's going to look like some kind of callback. I'm not sure it's needed -- same as today, we can keep writing tables around it. 

To support this, I added a couple of new interfaces for parsing and flattening. And cleaned up the existing exec pattern a little.

Oh, and added a parser and table for `dsregcmd`

Also, discovered that the build wasn't actually happening on go 1.19. Fixed that. 

Fixes: #873 
Closes: #900


pkg/osquery/tables/dataflattentable/exec_and_parse.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/osquery/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+type flattenBytesInt interface {
+	FlattenBytes([]byte, ...dataflatten.FlattenOpts) ([]dataflatten.Row, error)
+}
+
+// execTableV2 is the next iteration of the dataflattentable wrapper. Aim to migrate exec based tables to this.
+type execTableV2 struct {
```
directionless, line 21: I could do this using the existing table struct, as I did in #900, but it feels better to make a new struct and think about trying to clean that up.

Makefile
```diff
@@ -2,10 +2,7 @@ all: build
 
 .PHONY: build
 
-ifndef $GOPATH
-	GOPATH = $(HOME)/go
-endif
-
+GOPATH ?= $(HOME)/go
```
directionless, line 5: This was interesting.

I don't know if it's always been broken, and only just became an issue. Or if it's never worked worked correctly. But looking at the old code, the intent was clearly to set GOPATH if it was not already set, and then to add that to PATH. I think this is because we install a handful of tooling/ci dependancies, and they need to be in the PATH to work.

But, the old logic wasn't working. In CI. GOPATH was blank, the GOPATH setting didn't take, and then we would prepend `/bin` to path. On the CI machine, go 1.17 was in `/bin`, which caused the build to fail. (Since I started using `any`)

pkg/osquery/tables/execparsers/dsregcmd/parse_test.go
```diff
@@ -0,0 +1,96 @@
...
+	return b
+}
+
+func mustReadFile(filepath string) []byte {
+	b, err := os.ReadFile(filepath)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
```
James-Pickett, line 87:
```suggestion
func mustReadFile(t *testing.T, filepath string) []byte {
	b, err := os.ReadFile(filepath)
	require.NoError(t, err)
	return b
}
```

pkg/osquery/tables/execparsers/dsregcmd/parse_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func mustJsonUnmarshal(data []byte) any {
+	var v any
+	err := json.Unmarshal(data, &v)
+	if err != nil {
+		panic(err)
+	}
+	return v
+}
```
James-Pickett, line 96:
```suggestion
func mustJsonUnmarshal(t *testing.T, data []byte) any {
	var v any
	err := json.Unmarshal(data, &v)
	require.NoError(t, err)
	return v
}

```

pkg/osquery/tables/execparsers/dsregcmd/parse_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{
+			name:        "header missing line 3",
+			input:       mustReadFile(path.Join("test-data", "error_header_missing_line3.txt")),
+			expectedErr: true,
+		},
+		{
+			name:        "no section title",
+			input:       mustReadFile(path.Join("test-data", "error_no_section_title.txt")),
+			expectedErr: true,
+		},
```
James-Pickett, line 50:
```suggestion
		{
			name:     "empty input",
			expected: map[string]any{},
		},
		{
			name:     "not configured",
			input:    mustReadFile(path.Join("test-data", "not_configured.txt")),
			expected: mustJsonUnmarshal(t, mustReadFile(t, path.Join("test-data", "not_configured.expected.json"))),
		},

		//
		// Error Cases
		//
		{
			name:        "lines before header",
			input:       mustReadFile(t, path.Join("test-data", "error_lines_before_header.txt")),
			expectedErr: true,
		},
		{
			name:        "header missing line 3",
			input:       mustReadFile(t, path.Join("test-data", "error_header_missing_line3.txt")),
			expectedErr: true,
		},
		{
			name:        "no section title",
			input:       mustReadFile(t, path.Join("test-data", "error_no_section_title.txt")),
			expectedErr: true,
		},
```

pkg/osquery/tables/dataflattentable/exec_and_parse.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func NewExecAndParseTable(logger log.Logger, tableName string, parser parserInt, execCmd []string, opts ...execTableV2Opt) *table.Plugin {
+	t := &execTableV2{
+		logger:         level.NewFilter(log.With(logger, "table", tableName), level.AllowInfo()),
+		tableName:      tableName,
+		flattener:      flattenerFromParser(parser),
+		timeoutSeconds: 30,
+		execPaths:      execCmd[:1],
+		execArgs:       execCmd[1:],
```
James-Pickett, line 58: The extraction of the execPaths & execArgs feels weird to me .... I guess because I'm used to the standard lib command signature `func Command(name string, arg ...string) *Cmd`

Consider just accepting a `name` or `commandName` string param the adding a `func WithArgs(args ...string) execTableV2Opt` option?

James-Pickett (commented):
This is great! Massive reduction in boiler plate. Left a few comments, noting major.

pkg/osquery/tables/dataflattentable/flattener-from-parser.go
```diff
@@ -0,0 +1,39 @@
+package dataflattentable
```
James-Pickett, line 1: any reason why this file is kebab case? would prefer snake case

pkg/osquery/tables/dataflattentable/flattener-from-parser_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	var tests = []struct {
+		name         string
+		input        []byte
+		parser       parserInt
+		expectedRows int
+	}{
+		{
+			name:         "dsreg",
+			input:        mustReadFile("../execparsers/dsregcmd/test-data/not_configured.txt"),
```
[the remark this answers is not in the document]
directionless: For this, and the others, I found it a bit of a toss up. This isn't _really_ test code, it's just setup code. `panic` doesn't feel totally wrong. But I'm happy to change. 

One caveat -- I think of `must` as implying panic. So I'm renaming the functions

pkg/osquery/tables/dataflattentable/flattener-from-parser.go
directionless: sure

pkg/osquery/tables/dataflattentable/flattener-from-parser.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/kolide/launcher/pkg/dataflatten"
+)
+
+type parserInt interface {
```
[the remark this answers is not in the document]
directionless: Yes, exactly. `Int` for interface, sorta hungarian notation style. (Similarly, I use `fh` for filehandle). 

This is the client declared interface, so it's mostly unimportant. But I didn't want the name conflict with `parser` and I can't really bring myself to type `parserer`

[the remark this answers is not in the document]
directionless: If pressed, I guess I'd use `parserer` but I probably line `parserInterface` more.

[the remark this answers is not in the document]
James-Pickett: I can live with `parserInterface`, I'm mostly just against `parserInt` because my brain says integer

pkg/osquery/tables/dataflattentable/exec_and_parse.go
James-Pickett: fair enough

pkg/osquery/tables/dataflattentable/exec_and_parse.go
```diff
@@ -0,0 +1,101 @@
...
+type flattenBytesInt interface {
+	FlattenBytes([]byte, ...dataflatten.FlattenOpts) ([]dataflatten.Row, error)
+}
+
+// execTableV2 is the next iteration of the dataflattentable wrapper. Aim to migrate exec based tables to this.
+type execTableV2 struct {
+	logger         log.Logger
+	tableName      string
+	flattener      flattenBytesInt
+	timeoutSeconds int
```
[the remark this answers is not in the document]
James-Pickett: yep, that has the same effect

pkg/osquery/tables/dataflattentable/exec_and_parse.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/osquery/osquery-go/plugin/table"
+	"github.com/pkg/errors"
+)
+
+type flattenBytesInt interface {
```
James-Pickett, line 16: Apologies if I missed this one. I'm opposed to using the `Int` convention to name interfaces. I automatically assume it has something to do with integer. I think this should be called `bytesFlattener`.

pkg/osquery/tables/dataflattentable/flattener_from_parser.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/kolide/launcher/pkg/dataflatten"
+)
+
+type parserInterface interface {
```
James-Pickett, line 12: Looking at this a little more, is there a naming conflict? Can we just call it `parser`? (Reviewing from phone now, I'll check in the morning)

James-Pickett: checked this morning, looks like `parser` works, my affinity toward using `er` at the end of interfaces comes from the standard lib and [go style guide](https://go.dev/doc/effective_go#interface-names)

directionless: Hrm. Okay, `parser` it is. I don't think I like it, it feels... too short and ambiguous for an internal interface. I think the go docs make more sense for external exported/common interfaces. For internal ones, I think it's much less clear. 

I tend to favor more words for clarity. But :shrug:

James-Pickett (dismissed):

James-Pickett (approved):

PR #904, opened 2022-10-04
Add table for `kolide_macos_recommended_updates`

### LINKED ISSUE #886: Launcher Table Request: macOS Available Software Updates
## What is being requested?
A new table which retrieves available software updates for a macOS device using the SoftwareUpdate framework.

## Why is this necessary?
It is possible that if a user does not have automatic updates configured properly that the catalog will not retrieve the latest available updates. Using this framework should ensure that the list of available updates is more reliable than parsing the on-disk plist cache which could be out of date.

## How can this be accomplished?
The following private framework can be used to retrieve available updates:

**Big Sur:**
https://github.com/FritzX6/ApplePrivateHeaders/blob/main/macOS/11.3/System/Library/PrivateFrameworks/SoftwareUpdate.framework/Versions/A/SoftwareUpdate/SUProduct.h

**Catalina:**
https://github.com/FritzX6/macOS_headers/blob/master/macOS/PrivateFrameworks/SoftwareUpdate/26/SUProduct.h

seejdev:
Fixes #886 

At a high level, these changes:

- Organize the existing macOS software update code within `/tables/macos_software_update/`
- Adds a new table called `kolide_macos_recommended_updates` which pulls "Recommended" software update information directly from the SoftwareUpdate framework.

Example output:
```
osquery> select * from kolide_macos_recommended_updates;
+-------------------------------------------+----------------------+----------------------+------------------------------+-------+
| fullkey                                   | parent               | key                  | value                        | query |
+-------------------------------------------+----------------------+----------------------+------------------------------+-------+
| RecommendedUpdates/0/Identifier           | RecommendedUpdates/0 | Identifier           | MSU_UPDATE_21G115_patch_12.6 | *     |
| RecommendedUpdates/0/Display Name         | RecommendedUpdates/0 | Display Name         | macOS Monterey 12.6          | *     |
| RecommendedUpdates/0/MobileSoftwareUpdate | RecommendedUpdates/0 | MobileSoftwareUpdate | 1                            | *     |
| RecommendedUpdates/0/Product Key          | RecommendedUpdates/0 | Product Key          | MSU_UPDATE_21G115_patch_12.6 | *     |
| RecommendedUpdates/0/Display Version      | RecommendedUpdates/0 | Display Version      | 12.6                         | *     |
| RecommendedUpdates/1/Identifier           | RecommendedUpdates/1 | Identifier           | Safari16.0MontereyAuto       | *     |
| RecommendedUpdates/1/Display Name         | RecommendedUpdates/1 | Display Name         | Safari                       | *     |
| RecommendedUpdates/1/Product Key          | RecommendedUpdates/1 | Product Key          | 012-49737                    | *     |
| RecommendedUpdates/1/Display Version      | RecommendedUpdates/1 | Display Version      | 16.0                         | *     |
| RecommendedUpdates/2/Identifier           | RecommendedUpdates/2 | Identifier           | Command Line Tools for Xcode | *     |
| RecommendedUpdates/2/Display Name         | RecommendedUpdates/2 | Display Name         | Command Line Tools for Xcode | *     |
| RecommendedUpdates/2/Product Key          | RecommendedUpdates/2 | Product Key          | 012-62819                    | *     |
| RecommendedUpdates/2/Display Version      | RecommendedUpdates/2 | Display Version      | 14.0                         | *     |
+-------------------------------------------+----------------------+----------------------+------------------------------+-------+```

### CODE REVIEW

File: pkg/osquery/tables/macos_software_update/recommended_updates_table.go [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+			softwareUpdateArgs = append(softwareUpdateArgs, softwareUpdateNoScanArg)
+		}
+
+		_, err = tablehelpers.Exec(ctx, t.logger, 30, []string{softwareUpdateToolPath}, softwareUpdateArgs)
+		if err != nil {
+			level.Info(t.logger).Log("msg", "softwareupdate failed", "err", err)
+			return nil, nil
+		}
+
+		data := getUpdates()
```
directionless, line 75: I'm a little confused by this. It's doing an exec, tossing the output, then using the frameworks to pull the data? Why have both approaches?

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+const softwareUpdateToolPath = "/usr/sbin/softwareupdate"
+const softwareUpdateListArg = "--list"
+const softwareUpdateNoScanArg = "--no-scan"
+
+var updatesData []map[string]interface{}
+var doNotScan bool
```
directionless, line 32: I'd declare these in `generate`. Might be an extra allocation, but it feels cleaner

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
```diff
@@ -0,0 +1,113 @@
...
+
+		data := getUpdates()
+
+		for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
+			flattened, err := dataflatten.Flatten(data, dataflatten.WithLogger(t.logger), dataflatten.WithQuery(strings.Split(dataQuery, "/")))
+			if err != nil {
+				level.Info(t.logger).Log("msg", "Error flattening data", "err", err)
+				return nil, nil
+			}
+			results = append(results, dataflattentable.ToMap(flattened, dataQuery, nil)...)
```
directionless, line 83: this needs to have the `noscan` column merged in. (That's the intent of the `dataflattentable.ToMap` function. To merge the columns into flatten. That `nil` should be the noscan stuff)

directionless (commented):
The c side looks a lot cleaner this way. Nice.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
seejdev: I left `updatesData` there since I need to reference it in the callbacks from C.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
seejdev: Ah, I was trying to figure that out! Forgot about that parameter. I ended up removing `noscan` all together, but this was still helpful to learn.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
seejdev: So I _thought_ there was some correlation between `softwareupdate --list` invocations and the shared prefs API, but I did some more testing and I can still get the shared prefs API to give me update data regardless if `softwareupdate` is called, the state of `/Library/Updates/` dir, or the state/presence of the SoftwareUpdate cached plist.

The plist will be updated after calling `softwareupdate --list` but that seems to be independent of the framework itself.

I also noticed that even if I disable all macOS software updates behaviors, the preferences for checking & downloading updates gets reset to enabled after reboot 🤷  So even after blowing away all on-disk cached update data, killing the software update daemons, and rebooting, querying the `[SUSharedPrefs recommendedUpdates]` still gives me good data.

This is all to say - I've removed the exec entirely as it seems to have no benefit after all.

James-Pickett (commented):
This looks good, this is the first time I've looked at C being used in go and really understood it. I will book mark this example for later when I need to do some C stuff.

Do you think it might be worth it to add some unit tests? Might not be able to verify the output since the updates are dynamic, but we could run a happy path test to just verify that it doesn't log errors?

James-Pickett (commented):
I don't want to thumb this, because I'm not confident enough on the `C` to give the green light, but I did test it on my machine and nothing exploded!

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/dataflatten"
+	"github.com/kolide/launcher/pkg/osquery/tables/dataflattentable"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+var updatesData []map[string]interface{}
```
directionless, line 24: Does this need to be a package level variable? I can't tell if it needs that for how the c functions work, or not.

If it does, how confident are you that it's reset between generate calls?

pkg/osquery/tables/macos_software_update/sus.m
```diff
@@ -0,0 +1,87 @@
...
+
+	Class SUSharedPrefs = [bundle classNamed:@"SUSharedPrefs"];
+	id manager = [SUSharedPrefs sharedPrefManager];
+
+	NSArray *updates = [manager recommendedUpdates];
+    unsigned int i = 0;
+
+    updatesFound([updates count]);
+
+	for (id update in updates) {
```
directionless, line 78: indentation is kinda weird?

pkg/osquery/tables/macos_software_update/sus.m
```diff
@@ -0,0 +1,87 @@
...
+        for (NSString *key in update) {
+            NSString *value = [update objectForKey:key];
+            updateKeyValueFound(i, (char *)[key UTF8String], (char *)[[value description] UTF8String]);
+        }
+        ++i;
+    }
```
directionless, line 84: Huh. So this is a loop in C, which is calling the (exported) go function, so it updates the go map? That's really neat. 

And presumably this is also the place we'd hang additional information about the update?

directionless (commented):
Code looks really clean. I want to test this on a couple of older macs.

pkg/osquery/tables/macos_software_update/sus.m
seejdev: I've become too reliant on `gofmt` 😆

pkg/osquery/tables/macos_software_update/sus.m
seejdev: Correct, this could be expanded to extract and forward on more record-level fields. I found doing it in this procedural sort of way keeps it fairly simple as far as managing the memory and minimizing copies.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
seejdev: I could have avoided it being package level and moved it to `getUpdates` but then I would need to pass a pointer to it into the C code, then pass it back to Go as an argument in the callbacks, cast it back to it's Go type, etc. Although I prefer to scope var as narrowly as possible, this felt more simple.

When `generate` is first called, this var will be in whatever state is was left in since the last invocation. When `updatesFound` is called, it will reinitialize the var. However, there is an edge case where if the second invocation returns no results in the C code, `updatesFound` won't get called, the var doesn't get reset, and the stale value remains and gets returned. I can add an explicit reset to make sure the stale value is always tossed before subsequent invocations.

seejdev:
> This looks good, this is the first time I've looked at C being used in go and really understood it. I will book mark this example for later when I need to do some C stuff.
> 
> Do you think it might be worth it to add some unit tests? Might not be able to verify the output since the updates are dynamic, but we could run a happy path test to just verify that it doesn't log errors?

Yeah I was on the fence about what I could feasibly and reliably test - but I think you're right it's worth at least verifying it doesn't blow up.

pkg/osquery/tables/macos_software_update/sus.m
seejdev: Applied `clang-format` to the `.h/.m` so it should be consistent now!

directionless:
I build 634b7c0a5622941a4c3bb5944079c32d1793b34a and tested it on the oldest macOS I have active. My Mohave machine does not have pending updates, but this appears to have run (or at least not crashed). So I think from that perspective it's good to merge.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
directionless: I think an explicit reset would be good. Either at the start of the `generate` or in a defer. Whatever feels right.

pkg/osquery/tables/macos_software_update/sus.m
directionless: The back and forth is pretty neat.

pkg/osquery/tables/macos_software_update/recommended_updates_table_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/kolide/launcher/pkg/osquery/tables/tablehelpers"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_generateRecommendedUpdatesHappyPath(t *testing.T) {
+	t.Parallel()
+	table := Table{logger: log.NewNopLogger()}
+	t.Run("HappyPath", func(t *testing.T) {
```
directionless, line 18: You don't need `t.Run` here -- there's no subtest in play. You can run the tests from `Test_generateRecommendedUpdatesHappyPath` directly

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	return table.NewPlugin(tableName, columns, t.generate)
+}
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	updatesData = nil
+	data := getUpdates()
```
directionless, line 46: I understand it, but this looks weird. Maybe put the `updatesData = nil` down in `getUpdates` with an appropriate comment?

directionless (commented):
This looks really clean. If pressed I'd take it as is otherwise, a couple of nits.

pkg/osquery/tables/macos_software_update/recommended_updates_table.go
seejdev: Yep, that's a bit clearer, thanks!

directionless (approved):

PR #905, opened 2022-10-04
added console user info to request id

James-Pickett:
Adds `ConsoleUsers` field to local server request-id

```json
{
  "RequestId": "",
  "UUID": "",
  "InstanceId": "",
  "HardwareSerial": "",
  "Nonce": "01G0FREE0PIZZA0FEGAC5HS88J",
  "Timestamp": "1922-10-04T12:44:58.545871-07:00",
  "ConsoleUsers": [
    {
      "Uid": "501",
      "Gid": "20",
      "Username": "dohnjoe",
      "Name": "Dohn Joe",
      "HomeDir": "/Users/dohnjoe"
    }
  ]
}
```

* when ever [desktop ipc PR](https://github.com/kolide/launcher/pull/891) is mereged, it will begin using consleuser pkg to get console user(s)


pkg/consoleuser/consoleuser_darwin.go
```diff
@@ -0,0 +1,45 @@
+//go:build darwin
+// +build darwin
+
+package consoleuser
```
directionless, line 4: At least for now, I'd drop these in `ee`. They're only used by the ee functions. But if we ever make a table, I'll say we should move them

ee/localserver/request-id.go
```diff
@@ -20,8 +23,9 @@ type identifiers struct {
 type requestIdsResponse struct {
 	RequestId string
 	identifiers
-	Nonce     string
-	Timestamp time.Time
+	Nonce        string
+	Timestamp    time.Time
+	ConsoleUsers []*user.User
```
directionless, line 28: I wonder if this is ever going to be more than one. We'll see!

pkg/consoleuser/consoleuser_windows.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+func CurrentUids(context context.Context) ([]string, error) {
+	explorerProcs, err := explorerProcesses()
+	if err != nil {
+		return nil, fmt.Errorf("getting explorer processes: %w", err)
+	}
+
+	// unclear if windows will ever have more than one explorer process for a single user
+	// guard against this by forcing uniquness
```
directionless, line 23:
```suggestion
	// guard against this by forcing uniqueness
```

pkg/consoleuser/consoleuser_windows.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// first store uids in a map to prevent duplicates
```
directionless, line 25: I am always baffled that golang doesn't have succinct sugar for "get me all the unique items". But I agree this seems fine.

ee/localserver/request-id.go
```diff
@@ -68,6 +72,16 @@ func (ls *localServer) requestIdHandlerFunc(res http.ResponseWriter, req *http.R
 	}
 	response.identifiers = ls.identifiers
 
+	consoleUsers, err := consoleUsers()
+	if err != nil {
```
directionless, line 76: I think this is correct for now, but I wonder if we'll want to think about how to expose errors to k2 through this interface. Not today!

ee/localserver/request-id_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/bbolt"
+)
+
+func Test_localServer_requestIdHandler(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name string
+	}{
```
directionless, line 27: Do you see this as getting other input possibilities? Is the table driven tests adding anything?

ee/localserver/request-id_test.go
```diff
@@ -0,0 +1,101 @@
...
+		require.NoError(t, err)
+
+		key, err := rsa.GenerateKey(rand.Reader, 2048)
+		require.NoError(t, err)
+
+		keyBytes, err := x509.MarshalPKCS8PrivateKey(key)
+		require.NoError(t, err)
+
+		err = bucket.Put([]byte("privateKey"), keyBytes)
+		require.NoError(t, err)
```
directionless, line 91: This seems fine -- I'd take it as is, but the functions here feel a little weird. These aren't things I'd associate with creating a `testBboltDb`, so why have them in this function? Honestly, I'd probably merge `testServer` and `testBboltDb` (I also don't remember if there's a setup function you can call instead of manually creating the keys)

ee/localserver/request-id_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+		return nil
+	})
+
+	t.Cleanup(func() {
```
directionless, line 96: nice

pkg/consoleuser/consoleuser_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"os"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCurrentUids(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
```
directionless, line 14: Same question as above?

Maybe modulated a little about if it's interesting to encode expectations about the platforms into the test table, or create platform specific test files or ???

directionless (commented):
This looks great. Mostly nits and questions. I think none are strictly required

ee/localserver/request-id.go
James-Pickett: This is a little tricky

it's possibly there will be zero `ConsoleUsers`

There appear to [be some windows service](https://learn.microsoft.com/en-us/windows-server/remote/remote-desktop-services/welcome-to-rds) that allow "Session-based virtualization" for multiple users. I have no idea how this works under the hood. Virtualization implies it's a VM on the machine, then what user owns that VM? There would a lot of research here. If they login from the VM then I guess it doesn't matter? It's also for windows server, which feels like an edge case.

I could be persuaded to drop it to just a singular `ConsoleUser`, but this it just felt safer to have it plural so that downstream consumers since I don't fell 100% confident there will always be just 1.

ee/localserver/request-id_test.go
James-Pickett: just consistency .... I'm so used to table driven tests now that when I come across one that is not it throws me off and I have to spend extra time parsing it. Perhaps down the road we'll come up with the need to mock the inputs to the test functions or increase complexity for some reason.

pkg/consoleuser/consoleuser_test.go
James-Pickett: I dropped the table tests ... despite my earlier comment about "consistency", it does look better and is easy to parse

ee/localserver/request-id_test.go
James-Pickett: Yea, it's better to just roll it all together, less setup in the test, also dropped table test

James-Pickett: I could not find a function for it ... but I think this has some value .... shows that local server relies on that key

ee/localserver/request-id.go
James-Pickett: 👍

ee/localserver/request-id.go
directionless: That's reasonable. Let's keep it.

pkg/consoleuser/consoleuser_test.go
directionless: I do the same thing sometimes. I see a test without tables, and I'm like "whoa! Hard to read" then I read it and remember it's what would have been in the `t.Run`.

ee/localserver/request-id_test.go
directionless: https://github.com/kolide/launcher/blob/32a910268b20af5274cf9c6bb847173bf8f31b3f/pkg/osquery/extension.go#L232

James-Pickett: sweet, now using this

directionless (approved):
I think this is good

PR #909, opened 2022-10-13
James/flakey desktop runner tests

James-Pickett:
* uses backoff.WaitFor() to make client request since tests were failing because server hadn't full started before make calls to it
* uses backoff.WaitFor() to remove socket when server shut down

### DIFF
--- a/ee/desktop/server/server.go
+++ b/ee/desktop/server/server.go
@@ -5,7 +5,6 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"net"
 	"net/http"
 	"os"
@@ -15,6 +14,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/backoff"
 )
 
 type DesktopServer struct {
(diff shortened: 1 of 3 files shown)


ee/desktop/server/server.go
```diff
@@ -36,6 +36,7 @@ func New(logger log.Logger, authToken string, socketPath string, shutdownChan ch
 
 	authedMux := http.NewServeMux()
 	authedMux.HandleFunc("/shutdown", desktopServer.shutdownHandler)
+	authedMux.HandleFunc("/ping", pongHandler)
```
directionless, line 39: I'd have called this `pingHandler` -- while it might return pongs, it's handling pings.

pkg/osquery/tables/crowdstrike/falconctl/table_test.go
```diff
@@ -62,26 +82,14 @@ func TestOptionRestrictions(t *testing.T) {
 
 			// We've overridden exec with a noop, but we can check that we get the expected error back
 			_, err := testTable.generate(context.TODO(), mockQC)
-
-			if tt.expectedErr {
-				require.Error(t, err)
-				require.Contains(t, err.Error(), "requested option not allowed")
```
directionless, line 68: I think it's worth detecting this error

directionless (changes requested):

pkg/osquery/tables/crowdstrike/falconctl/table_test.go
James-Pickett: the function no longer returns and error, but we do verify that the it logs the disallowed option

directionless: I was thinking about the log, yeah.

But I missed the addition of `expectedLog`. oops.

`expectedLog` feels slightly at odds with `expectedExecs`. Do you think it makes sense to keep it as is? Remove `expectedExecs`? Add `expectedDisallows`?

James-Pickett: I dropped expected log and now have `expectedExecs int` & `expectedDisallows int`, take a look, I think it makes the test easier to understand

directionless (approved):

PR #911, opened 2022-10-17
updates consoleuser.CurrentUids on macOS to examine kCGSSessionOnConsoleKey when determining console user

James-Pickett:
* consoleuser.CurrentUids on macOS now looks for `kCGSSessionOnConsoleKey : TRUE` to determine console user
* returns nil when `Name : loginwindow` since this is the return value at the login screen

This ensures that another user's desktop icon does not appear on the last logged in user's console. 

Edge Case:

* existing user logs in
* a new user is created
* new user logs in for first time using "fast user switching"
* new user will not get the icon until the second time they log in


ee/consoleuser/consoleuser_darwin.go
```diff
@@ -12,34 +12,105 @@ import (
...
-		return []string{parts[1]}, nil
+		// reset values
+		kCGSSessionOnConsole = ""
+		kCGSSessionUserID = ""
+	}
+
+	return uids, nil
+}
+
+func scutilVal(line, key string) string {
```
directionless, line 109: I'll roll with it if you want, but I'm not sure this is cleaner than 

kv := strings.SplitN(line, ":", 2)
if len(kv) == 2 && kv[0] == kCGSSessionUserIDKey)
...
``` 
(Though that will require some trimming to work around the white space)

**Review by directionless — COMMENTED** [review]
I remember we walked through this in slack. but don't quiet remember how these details worked out.

Could you remind me of the various cases:
1. New user, first login
2. Second user, first login
3. Second user, first login incomplete, second login

File: ee/consoleuser/consoleuser_darwin.go [hunk 2]
```diff
@@ -12,34 +12,105 @@ import (
...
+		// if we don't have both values
+		if kCGSSessionOnConsole == "" || kCGSSessionUserID == "" {
+			continue
+		}
 
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("unexpected output from scutil: %s", line)
+		// have both values
+		if kCGSSessionOnConsole == "TRUE" {
+			uids = append(uids, kCGSSessionUserID)
```
seejdev, line 98: Once we find the console user, do we need to continue scanning?

James-Pickett: I'm not 100% sure if there can be more than one graphical user session on macos. I've tried using the built in screen sharing app but only go control over the "desktop" rather than a user's session. However, research implies that on mac servers there can be multiple graphical user sessions.

ee/consoleuser/consoleuser_darwin.go
James-Pickett: now using the split strategy, I think it make it a little more readable

James-Pickett:
> I remember we walked through this in slack. but don't quiet remember how these details worked out.
> 
> Could you remind me of the various cases:
> 
> 1. New user, first login
> 2. Second user, first login
> 3. Second user, first login incomplete, second login

@directionless , added the information as a commentt

directionless (approved):

PR #914, opened 2022-10-21
kills desktop process if initial ping fails

James-Pickett:
Addresses a bug where if the initial desktop server ping failed, the process was left running and not added to internal tracking. Now killing the process if initial ping fails.


directionless (approved):
While I feel like I do't understand why this happens, I think it's a reasonable addition. As you said, if we can't ping the desktop, we may as well kill it. Feel free to iterate, or merge

PR #918, opened 2022-10-24
james/missing windows desktop icon

James-Pickett:
- adds os.Environ() to desktop process env vars which enables desktop to work properly on windows

### DIFF
--- a/ee/desktop/runner/runner.go
+++ b/ee/desktop/runner/runner.go
@@ -410,11 +410,14 @@ func (r *DesktopUsersProcessesRunner) socketPath(uid string) (string, error) {
 func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socketPath string) (*exec.Cmd, error) {
 	cmd := exec.Command(executablePath, "desktop")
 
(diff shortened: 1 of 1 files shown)


ee/desktop/runner/runner.go
```diff
@@ -410,11 +410,14 @@ func (r *DesktopUsersProcessesRunner) socketPath(uid string) (string, error) {
 func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socketPath string) (*exec.Cmd, error) {
 	cmd := exec.Command(executablePath, "desktop")
 
-	cmd.Env = []string{
+	cmd.Env = append(
+		// if os.Environ() is not included, windows wiil not render the icon, producing the error:
+		// unable to write icon data to temp file: open C:\\windows\\systray_temp_icon_...: Access is denied
+		os.Environ(),
```
directionless, line 416: Hrm. I would be inclined to try to figure out _which_ part of the environment matters. And to consider varying behavior by OS. But if we can isolate it, we can just pass what we need

PR #920, opened 2022-10-25
James/desktop missing icon

### LINKED ISSUE #1013: Windows desktop runner fails to write icon data to temp file, causing desktop icon to be missing
Windows debug logs consistently show this error:

`{"caller":"runner.go:578","component":"desktop_users_processes_runner","level":"info","msg":"2023/02/03 09:22:30 systray error: unable to write icon data to temp file: open C:\\windows\\systray_temp_icon_70049458eb2e766ef057e410f59e8667: Access is denied.","subprocess":"desktop","ts":"2023-02-03T14:22:30.1353616Z","uid":"S-1-5-21-2552723621-1530566307-3129643804-1001"}`

Resulting in the desktop systray icon missing:

Screenshot_20230203_095632
Screenshot_20230203_095658

A [open PR exists for this issue](https://github.com/kolide/launcher/pull/920), but has not been resolved yet.

James-Pickett:
Fixed #1013 

copies temp env var from runner process to desktop process to address windows error where desktop icon was not being rendered


directionless (commented):
Revisiting this...

What is the correct tempdir to use? Should it be the tmpdir of the launcher process, or the user temp dir?

And why do we need one?

James-Pickett:
@directionless 

> What is the correct tempdir to use? Should it be the tmpdir of the launcher process, or the user temp dir?

I don't think the tmpdir of the launcher "proper" process will work, because it's owned by system user. Since we run as the logged in user it has to be a folder they have access to. The user temp dir seems most likely to me.

> And why do we need one?

Judging from the error provided by systray, I'm guessing systray writes the icon data to a temp dir, the reason why it does this is not clear to me. Without a "TEMP" variable set, it tries to write to `c:\windows\...`, since we're running as the user and not system at this point, we get an access denied error.

seejdev:
This change was included in #1015

PR #924, opened 2022-10-27
Launcher needs an uninstall

### LINKED ISSUE #916: Launcher needs an uninstall
Increasingly, it feels time to create an uninstall tools for Launcher. We have [docs](https://help.kolide.com/en/articles/2950178-removing-kolide-k2-launcher), but we should add some tooling.

My general bias for tools, is to make them as subcommand in launcher itself. (`launcher uninstall  -config /etc/kolide-k2/launcher.flags`) This lets us use our normal go tools, and it lets us keep shipping a single binary. As it makes sense, we can UI on top of it later.

I think the tricky bit, is that there's no real consistent path. Today, when Kolide builds packages, it uses those `kolide-k2` paths mentioned in the help docs. But these aren't hard code in launcher, they're functions of the packaging, and usually generated based off the `identifier`. And then some users have totally different paths because they extracted the files from an `rpm` and put them whereever.

So that's where I get stuck:
* Do we make `uninstall` kolide specific, and only check our paths? 
* Do we try to parse launcher.flags, and guess for the k2 one is not provided?
* Iterate the service list?

And maybe all that is an argument for making it part of the packaging, and not launcher itself. The packaging layer knows this. _But_ I can see future work where having it in launcher is a win.

I suspect we're going to iterate on this...

seejdev:
Fixed #916 

Example usage:
```
# Use config file for non-standard locations
launcher uninstall  --config=/etc/kolide-k2/launcher.flags

# Or assume default kolide-k2 installation
launcher uninstall
```

Public Documentation:
https://help.kolide.com/en/articles/2950178-removing-kolide-k2-launcher


cmd/launcher/uninstall_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+	// Now we can delete the plist which controls the launcher daemon
+	if err := os.Remove(launchDaemonPList); err != nil {
+		fmt.Printf("error removing file %s: %s\n", launchDaemonPList, err)
+	}
+
+	dirsToRemove := []string{
+		fmt.Sprintf("/usr/local/%s", identifier),
+		fmt.Sprintf("/etc/%s", identifier),
+		fmt.Sprintf("/var/%s", identifier),
+	}
```
RebeccaMahany, line 38: Would we want to clean up `/var/log/kolide-k2`? Probably not a big deal either way.

cmd/launcher/uninstall_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+	if identifier == "" {
+		// Ensure identifier is non-empty and use the default if none provided
+		identifier = "kolide-k2"
+	}
+
+	launchDaemonPList := fmt.Sprintf("/Library/LaunchDaemons/com.%s.launcher.plist", identifier)
+	launchCtlPath := "/bin/launchctl"
+	launchCtlArgs := []string{"unload", launchDaemonPList}
+
+	if _, err := tablehelpers.Exec(ctx, logger, 30, []string{launchCtlPath}, launchCtlArgs); err != nil {
```
James-Pickett, line 25: feels kind of weird here to use tablehelpers.exec because it requires path in an array and this is un-related to tables specifically 

I would just go with https://pkg.go.dev/os/exec#CommandContext

cmd/launcher/uninstall_linux.go
```diff
@@ -0,0 +1,85 @@
...
+	isDebianOS, err := isDebian(ctx, logger)
+	if err != nil {
+		return err
+	}
+
+	serviceName := fmt.Sprintf("launcher.%s", identifier)
+	packageName := fmt.Sprintf("launcher-%s", identifier)
+
+	// Stop launcher service
+	if _, err := tablehelpers.Exec(ctx, logger, 30, []string{"systemctl"}, []string{"stop", serviceName}); err != nil {
```
James-Pickett, line 33: same in this file, tablehelpers.Exec feels strange here

I would go with

https://pkg.go.dev/os/exec#CommandContext

directionless, line 33: I would have expected the package manager to do this.

cmd/launcher/uninstall_linux.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	// Disable launcher service
+	if _, err := tablehelpers.Exec(ctx, logger, 30, []string{"systemctl"}, []string{"disable", serviceName}); err != nil {
+		return err
+	}
+
+	// Tell the appropriate package manager to remove launcher
+	if isDebianOS {
```
directionless, line 43: I'd recommend a switch/case style approach here. There's more than just dpkg and rpm, so we'll need a default that errors out

cmd/launcher/main.go
```diff
@@ -135,6 +135,8 @@ func runSubcommands() error {
 		run = runInteractive
 	case "desktop":
 		run = runDesktop
+	case "uninstall":
+		run = runUninstall
```
James-Pickett, line 139: regarding the windows questions, for now I would just return "not supported" or something that that for now if we're on windows

cmd/launcher/uninstall.go
```diff
@@ -0,0 +1,72 @@
...
+
+		// Find the root_directory line, parse out the identifier part of the path
+		if strings.HasPrefix(line, prefix) {
+			substrs := strings.Split(strings.TrimSpace(strings.TrimPrefix(line, prefix)), string(os.PathSeparator))
+
+			if len(substrs) >= 3 {
+				identifier = strings.TrimSpace(substrs[2])
+				break
+			}
+		}
```
James-Pickett, line 68: you could return early and invert the if statement to reduce the nesting

```suggestion
		if !strings.HasPrefix(line, prefix) {
		    continue
		}
	        substrs := strings.Split(strings.TrimSpace(strings.TrimPrefix(line, prefix)), string(os.PathSeparator))
        
	        if len(substrs) >= 3 {
		       return strings.TrimSpace(substrs[2])
	      	}
```

cmd/launcher/uninstall.go
```diff
@@ -0,0 +1,72 @@
...
+		return err
+	}
+
+	err = removeLauncher(context.Background(), logger, identifier)
+
+	return err
+}
+
+func getIdentifierFromConfigFile(configFilePath string) (string, error) {
+	var identifier string
```
James-Pickett, line 44: a bit of a nit pick, but I don't think there is any reason to declare the var in advance

cmd/launcher/uninstall_linux.go
```diff
@@ -0,0 +1,85 @@
...
+		fmt.Sprintf("/var/%s", identifier),
+		fmt.Sprintf("/etc/%s", identifier),
+	}
+
+	// Now remove the local dirs used for launcher/osquery binaries and app data
+	for _, dir := range dirsToRemove {
+		if err := os.RemoveAll(dir); err != nil {
+			fmt.Printf("error removing dir %s: %s\n", dir, err)
+		}
+	}
```
directionless, line 63: I am scared of this accidentally removing something it shouldn't. (Like if identifier is nil)

Thinking aloud, what ways can we ensure it's safer. We can check the paths. We can display them to the user with confirmation.

Thoughts?

cmd/launcher/uninstall_darwin.go
directionless: `tablehelpers.Exec` is there to help enforce that we set timeouts on the table execs. Doesn't feel needed here.

cmd/launcher/uninstall_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+		return err
+	}
+
+	// Now we can delete the plist which controls the launcher daemon
+	if err := os.Remove(launchDaemonPList); err != nil {
```
directionless, line 30: Does this error if the file doesn't exist? We should ignore those

cmd/launcher/uninstall_linux.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return nil
+}
+
+func isDebian(ctx context.Context, logger log.Logger) (bool, error) {
+	output, err := tablehelpers.Exec(ctx, logger, 30, []string{"cat"}, []string{"/etc/os-release"})
```
directionless, line 69: If you want this, just read the file.

Though I don't know if it's easier to look for whether `dpkg` or `rpm` is installed.

I could imagine something like:
```
if (path-to-dpkg exists) {
  exec dpkg uninstall
}

if (path-to-rpm exists) {
  exec rpm unionstall
}
```

cmd/launcher/uninstall_darwin.go
seejdev: We don't call it out in https://help.kolide.com/en/articles/2950178-removing-kolide-k2-launcher

...but maybe we should do it here? Unless there's value to those logs post uninstall.

cmd/launcher/uninstall_linux.go
seejdev: Yeah I'm scared too - I put in a protection in the code to prevent identifier being an empty string, but that doesn't feel adequate.

User confirmation feels better - but maybe needs a bypass to allow automated uninstalls?

To avoid this all together, the packaging layer could generate these paths or whole commands in a file, which we just read and process.

cmd/launcher/uninstall_linux.go
seejdev: 😆 I wrote so many execs I forgot how to open a file.

I like that approach!

cmd/launcher/uninstall_darwin.go
seejdev: Actually, I'll just go back to using CommandContext, the timeouts probably don't matter.

James-Pickett: you can pass a context with timeout to command context for the same effect like
```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()

out, err := exec.CommandContext(ctx, "echo", "hello world").CombinedOutput()
```

cmd/launcher/uninstall_linux.go
seejdev: I've changed the identifier extraction to be via regexp for a better guarantee. We can further restrict the regexp based on what are acceptable identifier chars.

cmd/launcher/uninstall_darwin.go
seejdev: It prints the error, but ignores it otherwise.

cmd/launcher/uninstall_linux.go
seejdev: Looks like we can combine these to into `systemctl disable --now <serviceName>`

> --now
>            When used with enable, the units will also be started. When
>            used with disable or mask, the units will also be stopped.
>            The start or stop operation is only carried out when the
>            respective enable or disable operation has been successful.

https://man7.org/linux/man-pages/man1/systemctl.1.html

cmd/launcher/uninstall.go
```diff
@@ -0,0 +1,66 @@
...
+		return fmt.Errorf("parsing flags: %w", err)
+	}
+
+	// Attempt to get the identifier from the config file if one was provided
+	identifier, err := getIdentifierFromConfigFile(*flconfig)
+	if err != nil {
+		return err
+	}
+
+	return removeLauncher(context.Background(), identifier)
```
James-Pickett, line 38: I think its good practice to go ahead and put a time out in there.

```suggestion
        ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
	defer cancel()
	
	return removeLauncher(ctx, identifier)
```

cmd/launcher/uninstall.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func getIdentifierFromConfigFile(configFilePath string) (string, error) {
+	if configFilePath == "" {
+		return "", nil
+	}
+
+	f, err := os.Open(configFilePath)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to open config file '%s'.", configFilePath)
```
James-Pickett, line 48: we started using `fmt.Errorf()` as our standard error wrapping mechanism a while back

```suggestion
		return "", fmt.Errorf("opening config file '%s': %w, configFilePath, err)
```

cmd/launcher/uninstall_linux.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// Disable launcher service
+	cmd = exec.CommandContext(ctx, "systemctl", []string{"disable", serviceName}...)
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+
+	// Tell the appropriate package manager to remove launcher
+	if _, err := os.Stat("/usr/bin/dpkg"); err == nil {
```
James-Pickett, line 34: what happens if both of these if conditions fail?

seejdev: Now this errors out

cmd/launcher/main.go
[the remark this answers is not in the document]
seejdev: For some reason, execing the Apps & Features settings UI had problems finding "start" in the admin `%PATH%` 🤷 So I just opted to remove it and make uninstall unsupported for Windows.

cmd/launcher/uninstall_windows.go
```diff
@@ -0,0 +1,14 @@
...
+package main
+
+import (
+	"context"
+	"github.com/pkg/errors"
+)
+
+func removeLauncher(ctx context.Context, identifier string) error {
+	// Uninstall is not implemented for Windows - users have to use add/remove programs themselves
+	return errors.Errorf("Uninstall subcommand is not supported for Windows platforms.")
```
James-Pickett, line 13: we've moved away from using the `github.com/pkg/errors`
```suggestion
	return fmt.Errorf("Uninstall subcommand is not supported for Windows platforms.")
```

James-Pickett (commented):
LGTM, just left a little nit. Didn't hit approve to make sure seph gets a peek.

RebeccaMahany:
LGTM once James's suggestion is addressed -- no strong feelings either way on the log file removal

cmd/launcher/uninstall.go
```diff
@@ -0,0 +1,69 @@
...
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	defer cancel()
+
+	return removeLauncher(ctx, identifier)
+}
+
+func getIdentifierFromConfigFile(configFilePath string) (string, error) {
```
directionless, line 44: Hrm... This feels a a little weird. 

I'd probably use `ff`'s `WithIgnoreUndefined` option. And parse it the same way we normally parse the config.

cmd/launcher/uninstall_darwin.go
```diff
@@ -0,0 +1,47 @@
...
+		identifier = "kolide-k2"
+	}
+
+	launchDaemonPList := fmt.Sprintf("/Library/LaunchDaemons/com.%s.launcher.plist", identifier)
+	launchCtlPath := "/bin/launchctl"
+	launchCtlArgs := []string{"unload", launchDaemonPList}
+
+	cmd := exec.CommandContext(ctx, launchCtlPath, launchCtlArgs...)
+	if err := cmd.Run(); err != nil {
+		return err
```
directionless, line 25: This is fine, but I think it probably won't expose whatever error message `launchctl` spits out

cmd/launcher/uninstall_darwin.go
directionless: Ha. That, and `/etc/newsyslog.d/kolide-k2.conf` are somewhat undocumented and specific to macOS.

There's no value in keeping them, removing them isn't wrong.

cmd/launcher/uninstall_linux.go
directionless: Still thinking about this one.

cmd/launcher/uninstall_darwin.go
seejdev: 👍 Added these paths to the macOS uninstall

cmd/launcher/uninstall.go
seejdev: Oh wow, I wasn't aware of that feature in ff - that's way better!

cmd/launcher/uninstall_darwin.go
seejdev: Good point, I've added code to log `launchctl`/`systemctl` output.

directionless (approved):
I think this is okay to try.

PR #930, opened 2022-11-04
James/windows-pprof-debug-toggle

James-Pickett:
* Adds socket based debug server (pprof) toggle. 
* Adds command to toggle debug server: `launcher debug --root_directory=<path>`
* Converts existing posix implementation of pprof debug server toggle from signal to socket based
* Moves multi-platform (posix / windows) socket funcs to dedicated package


pkg/debug/toggle.go
```diff
@@ -0,0 +1,211 @@
...
+		return "", fmt.Errorf("reading debug toggle response: %w", err)
+	}
+	return string(buf[:mLen]), nil
+}
+
+// writeSocketPath returns the path to the socket and writes the path to the socketPathFilePath().
+// Since the socket file on windows has a random ulid attached to it, the socket path file can be used by
+// other processes to locate the socket file.
+func writeSocketPath(rootDir string) (string, error) {
+	socketPath := filepath.Join(rootDir, "debug_toggle_socket")
```
directionless, line 179: I suspect this should be a ulid too

pkg/debug/toggle.go
```diff
@@ -0,0 +1,211 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/ulid"
+	"github.com/kolide/launcher/ee/socket"
+	"github.com/kolide/launcher/pkg/backoff"
+)
+
+type DebugToggleListener struct {
```
directionless, line 20: I can roll with this, but reading it, it feels a little weird. 

Seeing it in code, it feels weird to imagine running `toggle`, instead I kinda imagine running `debug start`, and having that return a URL`. and `debug stop`.

With `toggle` the behavior feels much less clear. How does one check state? Or see what URL it's on? Etc.

directionless (changes requested):
I can roll with most of this, but the socket stuff can't be in ee

pkg/debug/toggle.go
James-Pickett: I kinda had that gut feeling too. When I got there it was already a toggle function and for some reason I thought I should keep the pattern. I will update it to have start / stop / status? sub commands

directionless: My inclination is that both `start` and `status` should display the listening port.

James-Pickett:
This work was started to assist in hunting down a memory, leak but started to become overly complex, putting this down for now.

PR #932, opened 2022-11-09
Draft: Point to newest binary inside most recently updated app bundle if available

RebeccaMahany:
TODO


pkg/autoupdate/findnew.go
```diff
@@ -266,6 +263,41 @@ func getUpdateDir(fullBinaryPath string) string {
...
+		possibleUpdates, err := filepath.Glob(fileGlob)
+
+		if err == nil && len(possibleUpdates) > 0 {
+			appBundleNames := make([]string, len(possibleUpdates))
+			for i, binaryPath := range possibleUpdates {
+				appBundleNames[i] = strings.TrimSuffix(binaryPath, "/"+binarySuffix)
+			}
+			sort.Strings(appBundleNames)
+			return appBundleNames, nil
+		}
```
James-Pickett, line 285: I feel like we should return the error here if we get it back, looking at the docs

> Glob returns the names of all files matching pattern or nil if there is no matching file. The syntax of patterns is the same as in Match. The pattern may describe hierarchical names such as /usr/*/bin/ed (assuming the Separator is '/').
> 
> Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed.


```suggestion
		possibleUpdates, err := filepath.Glob(fileGlob)
		if err != nil {
		  return nil, fmt.Errorf("globbing updates: %w, err")
		}

		if len(possibleUpdates) > 0 {
			appBundleNames := make([]string, len(possibleUpdates))
			for i, binaryPath := range possibleUpdates {
				appBundleNames[i] = strings.TrimSuffix(binaryPath, "/"+binarySuffix)
			}
			sort.Strings(appBundleNames)
			return appBundleNames, nil
		}
```

RebeccaMahany: I thought it would be safer to not return the error because then we can fall back (below) to the previous method of looking for possible updates. Wouldn't we want this in the case where the pattern _is_ malformed, so that eventually it could update itself to a newer version where the pattern is fixed? (If yes I should probably update to add a log for the error anyway, though.)

What you suggested is more in line with previous behavior on globbing errors, though, so I'm happy to do it that way too.

James-Pickett: Good point! I agree with logging the error and the going to the fall back, especially while this is new.

pkg/autoupdate/findnew.go
```diff
@@ -266,6 +263,41 @@ func getUpdateDir(fullBinaryPath string) string {
...
+	if runtime.GOOS == "darwin" {
+		binarySuffix := filepath.Join("Contents", "MacOS", binaryName)
+		fileGlob := filepath.Join(updateDir, "*", "*.app", binarySuffix)
+		possibleUpdates, err := filepath.Glob(fileGlob)
+
+		if err == nil && len(possibleUpdates) > 0 {
+			appBundleNames := make([]string, len(possibleUpdates))
+			for i, binaryPath := range possibleUpdates {
+				appBundleNames[i] = strings.TrimSuffix(binaryPath, "/"+binarySuffix)
+			}
```
directionless, line 282: I'm not sure I understand _why_ this strips off the `binarySuffix`. The caller simply re-adds it, so why not return the binary directy?

pkg/autoupdate/findnew.go
```diff
@@ -266,6 +263,41 @@ func getUpdateDir(fullBinaryPath string) string {
...
+// timestamps).
+func getPossibleUpdates(updateDir, binaryName string) ([]string, error) {
+	// If this is launcher running on macOS, then we should have app bundles available instead --
+	// check for those first.
+	if runtime.GOOS == "darwin" {
+		binarySuffix := filepath.Join("Contents", "MacOS", binaryName)
+		fileGlob := filepath.Join(updateDir, "*", "*.app", binarySuffix)
+		possibleUpdates, err := filepath.Glob(fileGlob)
+
+		if err == nil && len(possibleUpdates) > 0 {
```
directionless, line 278: I kinda get it -- this is a narrow carveout that's returning the happy path for macOS app bundles. But it feels a bit weird. It inverts the usual early-return-on-error logic. I'm not sure if there's something simpler. 

Also maybe worth noting that `err` on glob indicates some massive system failure.

pkg/autoupdate/findnew.go
```diff
@@ -266,6 +263,41 @@ func getUpdateDir(fullBinaryPath string) string {
...
+			appBundleNames := make([]string, len(possibleUpdates))
+			for i, binaryPath := range possibleUpdates {
+				appBundleNames[i] = strings.TrimSuffix(binaryPath, "/"+binarySuffix)
+			}
+			sort.Strings(appBundleNames)
+			return appBundleNames, nil
+		}
+	}
+
+	// Either not macOS/launcher or no app bundles found. Fall back to searching for binaries.
```
directionless, line 288: Total speculation, but I wonder if it's any cleaner to just return all the paths (app and not) in the desired sort order.

pkg/autoupdate/findnew.go
```diff
@@ -141,28 +142,24 @@ func FindNewest(ctx context.Context, fullBinaryPath string, opts ...newestOption
...
 	foundCount := 0
 	foundFile := ""
 	for i := len(possibleUpdates) - 1; i >= 0; i-- {
 		file := possibleUpdates[i]
 		basedir := filepath.Dir(file)
 		updateDownloadTime := filepath.Base(basedir)
+		foundExecutable := file
+		if strings.HasSuffix(file, ".app") {
+			foundExecutable = filepath.Join(file, "Contents", "MacOS", binaryName)
+		}
```
directionless, line 162: See comment below -- we add and remove and add this again

directionless (commented):
If I read the tests right, it converts everything to use app bundles. Does it make sense to test the non-bundle path? Or a mixed bundle and not environment?

pkg/autoupdate/findnew.go
RebeccaMahany: It makes it a little tidier for the caller to grab the timestamp out of the path to perform timestamp comparisons (can always be `updateDownloadTime := filepath.Base(basedir)`).

RebeccaMahany: Oh, it's also required so that the cleanup of old/corrupt updates works without any modifications needed -- that was the primary reason I ended up doing it like this.

pkg/autoupdate/findnew.go
RebeccaMahany: So, I'd spent some time doing it this way this morning, and the implementation wasn't significantly cleaner one way or the other. The reason I ended up keeping it like this (returning early if we've found suitable app bundle binaries) is because if we put them all (app and not) in together and sort them, then in the tests, we end up always choosing the top-level binary paired with the app bundle instead of the binary inside the app bundle (`<update-dir>/<timestamp>/binary` is always alphabetically before `<update-dir>/<timestamp>/Test.app/Contents/MacOS/binary`). Fixing that nicely would've required another edge-case-for-app-bundles carveout, and it felt a little brittle to just rely on the name of the app bundle never changing to be after `launcher` in the alphabet.

This also covers why I didn't end up doing mixed app bundle and non-app bundle for the tests like you mention [here](https://github.com/kolide/launcher/pull/932#pullrequestreview-1179716935).

However, this does leave us in the position where if `launcher-updates` has a couple binary-only updates that are fine, but some app bundle updates that are broken, we won't fall back to the binary-only updates because we're only looking for the app bundle updates. I'm not sure how big of a deal this is. I could fix it fairly simply by throwing the `checkExecutable` call in here, but it means more calls to `checkExecutable` than we technically need.

Let me know what you think or if it's easier to chat about it briefly instead.

pkg/autoupdate/findnew.go
RebeccaMahany: James mentioned re the globbing error [here](https://github.com/kolide/launcher/pull/932#discussion_r1021942621) -- I will update to add some logging so that it's clear what the issue is in that case. I could add more documentation here as well if that would make it clearer.

pkg/autoupdate/findnew.go
RebeccaMahany: Replied [here](https://github.com/kolide/launcher/pull/932#discussion_r1021974705)

pkg/autoupdate/findnew.go
directionless: Maybe comment? Something like `We can ignore this error -- we fall back to the prior behavior below`

pkg/autoupdate/findnew.go
directionless: _Ah._ So it's this way to avoider a larger PR. While there are probably clean ways to handle it, I think it's probably reasonable to be narrow here -- but comment about the reasoning please. 
```suggestion
			for i, binaryPath := range possibleUpdates {
			        // We trim and re-add the suffix for compatibility with the prior setup. Things like cleanup and
			        // timezone comparisons work. 
				appBundleNames[i] = strings.TrimSuffix(binaryPath, "/"+binarySuffix)
			}
```

pkg/autoupdate/findnew.go
directionless: I think, like above, this sounds like a tradeoff between a larger refactoring and small PR.

When you explain it, I immediately think we could move `checkExecutable` in here, and skip it in the caller. And then I realize we could put the reverse sort here, and return the first thing with a valid check. But I guess that bumps into the cleanup routine....

_Anyhow_  I think this is fine, thank you for explaining, and the downside you mention feels like an acceptable tradeoff to me.

pkg/autoupdate/findnew.go
RebeccaMahany: Added a comment both here and where we're re-adding the suffix 🙂

pkg/autoupdate/findnew.go
RebeccaMahany: Added a log message and a comment 🙂

Issue #703, opened 2021-01-28
macOS package missing a title


directionless:
The macOS packages are built by running `pkgbuild` and then using `productbuild` to flatten it. Experimenting, this causes the `title` to be lost. It looks like the distribution.xml that productbuild synthesizes, lacks a title attribute. I'm working through what a correct pattern is. It's probably to either ship a distribution.xml, or have a step to extract and modify it

directionless:
As long as I'm monkeying with the distribution.xml, we should set a minimum support os version

Issue #921, opened 2022-10-26
Launcher needs some kind of .app bundle


directionless:
## Introduction

As we experiment with secure enclaves, it seems more and more clear we're going to need to maOS entitlements. Which, as I understand it means launcher needs to, at least partly, be in an app bundle.
 
## What's an app bundle

While app bundles are most commonly the full fledged macOS applications, they can be much simpler. As I understand it, at the very core they can be a directory, with an `Info.plist` and the existing macho binary stashed in the right place. This represents a fairly simple approach, and can live along side the rest of the `pkg` files.

This is the same minimal approach osquery took in https://github.com/osquery/osquery/pull/7263. That PR isn't directly applicable -- it's all cmake, but it has some context. 

## Launcher and app bundles

For launcher, I see two paths. There's an interesting long term idea about replacing the `pkg` with an Application. But I don't think we should do that today. So instead I think theres a simpler path.

Today, we install as (And this includes the updates):
```
$ tree /usr/local/kolide-k2/ /etc/kolide-k2/
/usr/local/kolide-k2/
└── bin
    ├── launcher
    ├── launcher-updates
    │   ├── 1663913195
    │   │   └── launcher
    │   ├── 1665814654
    │   │   └── launcher
    │   ├── 1666157432
    │   │   └── launcher
    │   └── 1666416716
    │       └── launcher
    ├── osquery-extension.ext
    ├── osqueryd
    └── osqueryd-updates
        ├── 1660753700
        │   └── osqueryd
        └── 1660840403
            └── osqueryd
/etc/kolide-k2/
├── installer-info.json
├── installer-info.plist
├── launcher.flags
└── secret
```

Instead, I imagine we'd do something like:
```
$ tree .
.
├── bin
│   └── launcher -> launcher.app/Contents/MacOS/launcher
└── launcher.app
    └── Contents
        ├── Info.plist
        ├── MacOS
        │   └── launcher
        └── embedded.provisionprofile
```

## Autoupdate

I think the trickiest part here, is going to be in solving how this works with autoupdate. 

There are several potential issues:
1. There is a bunch of logic around how launcher [finds the newest](https://github.com/kolide/launcher/blob/abeabd003ec0c1a4499bd4bd58ca6804a0330e41/pkg/autoupdate/findnew.go) version of itself, this will need to be updated
2. The autoupdate code that untars launcher probably assumes it's a single binary, and not a simple directory
[...]

PR #934, opened 2022-11-14
Packaging for app bundle

RebeccaMahany:
Adds packaging support for the launcher app bundle -- https://github.com/kolide/launcher/issues/921.

Adds a step to the CI to confirm that package-builder can create a valid, installable pkg: see example job run [here](https://github.com/kolide/launcher/actions/runs/3498437319/jobs/5858797886).

Additionally closes https://github.com/kolide/launcher/issues/703.

___________

Test steps for packaging changes:
1. Build package-builder: `make package-builder`
2. Make a nightly build: `./build/package-builder make -debug --hostname=<insert here> --enroll_secret=secret --launcher_version=nightly --control_hostname=<insert here> --update_channel=nightly --identifier=<insert here> --transport=jsonrpc`
3. Confirm that all the files expected are present in the pkg: `pkgutil --payload-files /path/to/pkg`
4. Confirm that the `Distribution` file looks correct: `pkgutil --expand /path/to/pkg <new folder>` and view `<new folder>/Distribution`
5. Install without error: `sudo installer -dumplog -pkg /tmp/path/to/launcher.darwin-launchd-pkg.pkg -target /`
6. Verify that https://github.com/kolide/launcher/issues/703 is fixed: installer output should contain `installer: Package name is Kolide Endpoint Agent`
7. Confirm that `/etc/<identifier>/launcher.flags` and `/Library/LaunchDaemons/com.<identifier>.launcher.plist` look correct
8. Confirm that launcher starts up and runs after installation: -- `ps -eaf | grep launcher` should show both `/usr/local/<identifier>/bin/Kolide.app/Contents/MacOS/launcher -config /etc/<identifier>/launcher.flags` and `/usr/local/<identifier>/bin/Kolide.app/Contents/MacOS/launcher desktop`
9. Check app logs for unexpected errors


pkg/packaging/detectLauncherVersion_test.go
```diff
@@ -52,5 +54,37 @@ func TestFormatVersion(t *testing.T) {
...
+	t.Parallel()
+
+	pDarwin := &PackageOptions{target: Target{Platform: Darwin}}
+
+	// First, test that if the app bundle doesn't exist, we fall back to the top-level binary
+	require.Equal(t, filepath.Join("some", "dir", "launcher"), pDarwin.launcherLocation(filepath.Join("some", "dir")))
+
+	// Create a temp directory with an app bundle in it
+	tmpDir := t.TempDir()
+	os.MkdirAll(filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS"), 0755)
```
James-Pickett, line 69:
```suggestion
	require.NoError(t, os.MkdirAll(filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS"), 0755))
```

pkg/packaging/detectLauncherVersion_test.go
```diff
@@ -52,5 +54,37 @@ func TestFormatVersion(t *testing.T) {
...
+	baseDir := filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS")
+	expectedLauncherBinaryPath := filepath.Join(baseDir, "launcher")
+	f, err := os.Create(expectedLauncherBinaryPath)
+	if err != nil {
+		t.Errorf("could not create temp file for test: %v", err)
+		t.FailNow()
+	}
+	defer f.Close()
+	defer os.Remove(expectedLauncherBinaryPath)
```
James-Pickett, line 78:
```suggestion
	defer require.NoError(t, os.Remove(expectedLauncherBinaryPath))
```

pkg/packaging/detectLauncherVersion_test.go
```diff
@@ -52,5 +54,37 @@ func TestFormatVersion(t *testing.T) {
...
+	tmpDir := t.TempDir()
+	os.MkdirAll(filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS"), 0755)
+	baseDir := filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS")
+	expectedLauncherBinaryPath := filepath.Join(baseDir, "launcher")
+	f, err := os.Create(expectedLauncherBinaryPath)
+	if err != nil {
+		t.Errorf("could not create temp file for test: %v", err)
+		t.FailNow()
+	}
+	defer f.Close()
```
James-Pickett, line 77:
```suggestion
	defer require.NoError(t, f.Close())
```

pkg/packaging/packaging_test.go
```diff
@@ -136,6 +139,69 @@ func TestHelperProcess(t *testing.T) {
...
+		t.SkipNow()
+	}
+
+	// Set up cache directory with only a launcher binary in it
+	tmpCacheDir := t.TempDir()
+	binaryName := "launcher"
+	version := "nightly"
+	localBinaryDir := filepath.Join(tmpCacheDir, fmt.Sprintf("launcher-%s-%s", runtime.GOOS, version))
+	err := os.Mkdir(localBinaryDir, 0755)
+	assert.NoError(t, err, "could not make temp cache directory")
```
James-Pickett, line 155:
```suggestion
	assert.NoError(t, os.Mkdir(localBinaryDir, 0755), "could not make temp cache directory")
```

pkg/packaging/packaging_test.go
```diff
@@ -136,6 +139,69 @@ func TestHelperProcess(t *testing.T) {
...
+	cachedBinaryPath := filepath.Join(localBinaryDir, binaryName)
+	f, err := os.Create(cachedBinaryPath)
+	assert.NoError(t, err, "could not create binary")
+	defer f.Close()
+
+	// Set up output directory
+	tmpPkgRoot := t.TempDir()
+	binDir := "bin"
+	err = os.Mkdir(filepath.Join(tmpPkgRoot, binDir), 0755)
+	assert.NoError(t, err, "could not make temp output directory")
```
James-Pickett, line 165:
```suggestion
	assert.NoError(t, os.Mkdir(filepath.Join(tmpPkgRoot, binDir), 0755), "could not make temp output directory")
```

pkg/packaging/packaging_test.go
```diff
@@ -136,6 +139,69 @@ func TestHelperProcess(t *testing.T) {
...
+		err := os.MkdirAll(filepath.Join(appBundleLocation, "Contents", "MacOS"), 0755)
+		assert.NoError(t, err, "could not make temp app bundle directory")
+
+		// Add binary to app bundle
+		f, err := os.Create(filepath.Join(appBundleLocation, "Contents", "MacOS", binaryName))
+		assert.NoError(t, err, "could not create app bundle binary")
+		defer f.Close()
+
+		// Verify we found the app bundle and copied over the entire directory to the expected location
+		err = p.getBinary(context.TODO(), binaryName, binaryName, cachedBinaryPath)
```
James-Pickett, line 192:
```suggestion
		assert.NoError(t, p.getBinary(context.TODO(), binaryName, binaryName, cachedBinaryPath), "expected to find app bundle but did not")
```

James-Pickett (commented):
It's looking good, I just left a bunch of nit that make testing code more concise.

.github/workflows/go.yml
```diff
@@ -131,11 +131,35 @@ jobs:
...
+      id: run-package-builder-macos
+      if: ${{ contains(matrix.os, 'macos') }}
+      shell: bash
+      run: |
+        # Capture last line of output, `Built packages in /var/some/location`
+        OUTPUT=$(${{ steps.build.outputs.binary }} make -debug --hostname=localhost --enroll_secret=secret --launcher_version=nightly | tail -n 1)
+        OUTPUT_PREFIX="Built packages in "
+        echo "launcher-pkg-path=${OUTPUT#"$OUTPUT_PREFIX"}" >> "$GITHUB_OUTPUT"
+
+    - name: Test install macOS
```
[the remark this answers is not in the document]
RebeccaMahany: seph's idea! Glad it was so easy, maybe worth doing for the other OSes in the future too 🙂

.github/workflows/go.yml
```diff
@@ -146,11 +146,35 @@ jobs:
...
 
+    - name: package macOS
+      id: run-package-builder-macos
+      if: ${{ contains(matrix.os, 'macos') }}
+      shell: bash
+      run: |
+        # Capture last line of output, `Built packages in /var/some/location`
+        OUTPUT=$(${{ steps.build.outputs.binary }} make -debug --hostname=localhost --enroll_secret=secret --launcher_version=nightly | tail -n 1)
+        OUTPUT_PREFIX="Built packages in "
+        echo "launcher-pkg-path=${OUTPUT#"$OUTPUT_PREFIX"}" >> "$GITHUB_OUTPUT"
```
directionless, line 166: Hrm. I can live with this, but it feels roundabout. `package-builder` supports a `--output_dir` option if you want to specify where it lands.

.github/workflows/go.yml
```diff
@@ -146,11 +146,35 @@ jobs:
...
+
+    - name: Test install macOS
+      if: ${{ contains(matrix.os, 'macos') }}
+      run: |
+        # Check that we can install
```
directionless, line 171: I love this

pkg/packagekit/assets/distribution.dist
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<installer-gui-script minSpecVersion="2">
+	<title>{{.Title}}</title>
+	<pkg-ref id="com.{{.Identifier}}.launcher">
```
directionless, line 4: I think the `com.{{.Identifier}}.launcher` isn't going to be what we want, and we should treat this the same at `.Title` (eg: put more logic in go)

I think that our builds, should probably come out as `com.kolide.agent` and I have no idea what other people would need.

pkg/packaging/detectLauncherVersion_test.go
```diff
@@ -52,5 +54,37 @@ func TestFormatVersion(t *testing.T) {
...
+	require.NoError(t, os.MkdirAll(filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS"), 0755))
+	baseDir := filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS")
+	expectedLauncherBinaryPath := filepath.Join(baseDir, "launcher")
+	f, err := os.Create(expectedLauncherBinaryPath)
+	if err != nil {
+		t.Errorf("could not create temp file for test: %v", err)
+		t.FailNow()
+	}
+	defer require.NoError(t, f.Close())
+	defer require.NoError(t, os.Remove(expectedLauncherBinaryPath))
```
directionless, line 78: I wouldn't wrap these in `require.NoError`

pkg/packaging/detectLauncherVersion_test.go
```diff
@@ -52,5 +54,37 @@ func TestFormatVersion(t *testing.T) {
...
+	// Create a temp directory with an app bundle in it
+	tmpDir := t.TempDir()
+	require.NoError(t, os.MkdirAll(filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS"), 0755))
+	baseDir := filepath.Join(tmpDir, "Kolide.app", "Contents", "MacOS")
+	expectedLauncherBinaryPath := filepath.Join(baseDir, "launcher")
+	f, err := os.Create(expectedLauncherBinaryPath)
+	if err != nil {
+		t.Errorf("could not create temp file for test: %v", err)
+		t.FailNow()
+	}
```
directionless, line 76:
```suggestion
	f, err := os.Create(expectedLauncherBinaryPath)
	require.NoErrorf(t, err, "could not create temp file for test")
```

is the usual.

pkg/packaging/packaging.go
```diff
@@ -301,6 +301,7 @@ func (p *PackageOptions) Build(ctx context.Context, packageWriter io.Writer, tar
 	p.packagekitops = &packagekit.PackageOptions{
 		Name:                     "launcher",
 		Identifier:               p.Identifier,
+		Title:                    "Kolide Endpoint Agent",
```
directionless, line 304: I don't remember the chain that leads here. Is this the default? I want our official builds to be `Kolide Endpoint Agent`, but I want the other builds to be something like `fmt.Sprintf("Launcher agent for %s", p.Identifier)`

pkg/packaging/packaging_test.go
```diff
@@ -136,6 +139,67 @@ func TestHelperProcess(t *testing.T) {
 
 }
 
+func Test_getBinary(t *testing.T) {
+	t.Parallel()
+
+	if runtime.GOOS == "windows" {
+		t.SkipNow()
+	}
```
directionless, line 147: Why skip this on windows?

pkg/packaging/packaging_test.go
```diff
@@ -136,6 +139,67 @@ func TestHelperProcess(t *testing.T) {
...
+	binDir := "bin"
+	assert.NoError(t, os.Mkdir(filepath.Join(tmpPkgRoot, binDir), 0755), "could not make temp output directory")
+
+	p := &PackageOptions{
+		packageRoot: tmpPkgRoot,
+		binDir:      binDir,
+	}
+
+	// Verify we found the non-app bundle binary and copied it to the expected location
+	err = p.getBinary(context.TODO(), binaryName, binaryName, cachedBinaryPath)
```
directionless, line 171: This will cause the CI system to download launcher. I'm not sure what I think. I kinda want it, but it's potentially slow and error prone.

RebeccaMahany: It doesn't -- because the cached binary path is set appropriately, we don't make the call out to `FetchBinary` and just look at the one that we set up above.

pkg/packaging/detectLauncherVersion_test.go
RebeccaMahany: Oh, fair, makes sense

pkg/packagekit/assets/distribution.dist
RebeccaMahany: I did it this way because this is how our build currently generates the ID. Would we want it to change now?

.github/workflows/go.yml
RebeccaMahany: Ok, better to set `--output_dir` to the current directory and use that in `Test install macOS` below?

pkg/packaging/packaging.go
RebeccaMahany: I didn't do this the way I meant to! I've updated it now so that the default is `Launcher agent for <identifier>`, but that it can be overridden to `Kolide Endpoint Agent` via `PackageOptions.Title` in our own build process.

pkg/packagekit/assets/distribution.dist
directionless: Hrm. I guess I'm not sure. What is this id anyhow?

Our current production one is `kolide-k2`, and `com.kolide-k2.launcher` is kinda weird, and probably not correct for our bundle id. But I'm less sure what the pkg id is used for.

pkg/packaging/packaging_test.go
directionless: Oh, neat! So it's passing a file system path and then it's all just doing the right thing?

pkg/packaging/packaging_test.go
directionless: I wonder if it's something simple, like the fetcher not realizing it's a local path and trying to download from notary

RebeccaMahany: Oh, I remember now. So [the check for the local path](https://github.com/kolide/launcher/blob/main/pkg/packaging/packaging.go#L342) checks for `binaryVersion` (i.e. our cached location) to start with `./` or `/`. On Windows, though, it starts with `C:\`, when then would force the test to go through the `FetchBinary` real download, which I didn't want.

I wasn't sure if it was safe to update that check to also check for a `C:\` prefix or what else might be appropriate, so I just skipped the test on Windows.

directionless: You can update that. 

The intent is to detect whether something is a notary channel (`nightly`, `alpha`, `beta`, `stable`) or a specific path on the file system. When I wrote it, paths always started with `./` or `/`. (even on windows, you can start them that way, powershell and go handle it)

I think it would be totally reasonable to add `c:`, `d:`, and `\` to the prefix list. It might be reasonable to look for `/` or `\` in the string, though harder to understand the points of confusion.

pkg/packagekit/assets/distribution.dist
directionless: I have no idea. 

As you said, this is how it's set to day, maybe better to leave it unchanged. And assume we'll revisit in the future

pkg/packaging/packaging_test.go
RebeccaMahany: All set 🎉

.github/workflows/go.yml
RebeccaMahany: Updated! This is significantly more streamlined now -- thanks for the suggestion

directionless (approved):
Woo! Let's try it!

Might need to iterate, but gotta start somewhere.

PR #937, opened 2022-11-15
Replace errors.Wrap() with fmt.Errorf()

James-Pickett:
This replaces all instances with errors.Wrap() with fmt.Errorf().

This was done with this tool https://github.com/xdg-go/go-rewrap-errors

The major gotcha here is that `errors.Wrap(err, "some msg")` returns nil if the error is nil. There are several places in the code where we were not doing an `if err != nil` or some other check before returning `errors.Wrap(err, "some msg")`. So when converted in some places it would return `fmt.Errorf("some msg: %w", err)` which would result in `some msg: %!w(<nil>)` being returned instead of nil.

When reviewing, please be on the lookout for places where we return fmt.Errorf() without some kind of check before hand.

I think I found all those instances also tested running against local K2 instance.

There is some remaining use of the old `github.com/pkg/errors` package that uses `errors.Cause()`. Since this is kind of a type check and a little tricker, I opted to save removing those for another PR.


pkg/osquery/tables/secedit/secedit.go
```diff
@@ -132,7 +133,7 @@ func (t *Table) execSecedit(ctx context.Context, mergedPolicy bool) ([]byte, err
 	rd := transform.NewReader(file, unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder())
 	data, err := ioutil.ReadAll(rd)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error reading secedit output file: %s", err)
+		return nil, fmt.Errorf("error reading secedit output file: %s: %w", err, err)
```
RebeccaMahany, line 136: Is something else supposed to be for `%s` instead of including `err` twice? Maybe `dst` like on 127 above?

pkg/querytarget/query_target.go
```diff
@@ -34,24 +37,24 @@ func NewQueryTargeter(logger log.Logger, db *bbolt.DB, grpcConn *grpc.ClientConn
...
-		return errors.Wrap(err, "marshaling targets to bytes")
+		return fmt.Errorf("marshaling targets to bytes: %w", err)
 	}
 
 	if err := qtu.db.Update(func(tx *bbolt.Tx) error {
 		b := tx.Bucket([]byte(osquery.ServerProvidedDataBucket))
 		err := b.Put([]byte(table.TargetMembershipKey), targetRespBytes)
 
-		return errors.Wrap(err, "updating target memberships")
+		return fmt.Errorf("updating target memberships: %w", err)
```
RebeccaMahany, line 57: Does this need the `err != nil` check before the return?

pkg/windows/windowsupdate/session.go
```diff
@@ -55,7 +56,7 @@ func (iUpdateSession *IUpdateSession) GetLocal() (uint32, error) {
 
 func (iUpdateSession *IUpdateSession) SetLocal(locale uint32) error {
 	_, err := oleconv.ToUint32Err(oleutil.PutProperty(iUpdateSession.disp, "UserLocale", locale))
-	return errors.Wrap(err, "getproperty userlocale")
+	return fmt.Errorf("getproperty userlocale: %w", err)
```
RebeccaMahany, line 59: Optional -- should be `putproperty`?

RebeccaMahany (approved):
LGTM, just a couple small/optional suggestions -- I think https://github.com/kolide/launcher/pull/937#discussion_r1023177816 is the only one worth double-checking

pkg/autoupdate/findnew_test.go
```diff
@@ -347,11 +346,11 @@ func copyFile(dstPath, srcPath string, truncate bool) error {
 	} else {
 		stat, err := src.Stat()
 		if err != nil {
-			return errors.Wrap(err, "statting srcFile")
+			return fmt.Errorf("statting srcFile: %w", err)
 		}
 
 		if _, err = io.CopyN(dst, src, stat.Size()/2); err != nil {
-			return errors.Wrap(err, "statting srcFile")
+			return fmt.Errorf("statting srcFile: %w", err)
```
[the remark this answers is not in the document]
James-Pickett: fixed

pkg/osquery/tables/secedit/secedit.go
James-Pickett: fixed

pkg/querytarget/query_target.go
James-Pickett: good catch! fixed

pkg/windows/windowsupdate/session.go
James-Pickett: yea I think so, and it need a `err != nil` check, fixed

RebeccaMahany (approved):
👍

Issue #936, opened 2022-11-15
Table for server writable data


directionless:
We keep wanting to be able to have some kind of server writable data inside launcher. This should manifest as a table, which can then be joined against or used in decorators. I'm pretty sure this should be an EAV style table. Likely with columns `entity`, `attribute`, and `value`. 

We have some fragments of this work, I envisioned something like this ages back and it was implemented as [`kolide_targetting`](https://github.com/kolide/launcher/blob/8f2726c262c56849c530b0e6db9c070df169cb57/pkg/osquery/table/kolide_target_membership.go). But, that table has several flaws, and should be revisited. A very rough set of tasks:

* [ ] rename/replace `kolide_target_membership` to something like `kolide_metadata` or `kolide_server_data` or ???
* [ ] convert table from one column, to three columns -- `entity`, `attribute`, `value` (or, if we're stuck on types, maybe `entity`, `attribute`, `string`, and `int`)
* [ ] Remove the GRPC interface to this table
* [ ] This data should come from a new rocksdb bucket. It will be updated by the control server

James-Pickett:
> Remove the GRPC interface to this table

Does removing GRPC imply that we'll remove all the protobuf references as well? I'm not clear on how intertwined the 2 are. If so, the would it just be vanilla json?

directionless:
> > Remove the GRPC interface to this table
> 
> Does removing GRPC imply that we'll remove all the protobuf references as well? I'm not clear on how intertwined the 2 are. If so, the would it just be vanilla json?

We have protobuf interfaces? The jsonrpc ones, if they exist, should also go. I'm not actually sure any of them are still around, you'd need to dig around #390 and see what remains.

Basically, this is getting a lot simpler. 

All the GRPC is moribund, and can be ripped out. And we want a much simpler table that displays whatever is in the bbolt bucket. Some future thing will write to that bucket. 

I guess there's a slight wobble in that buckets are plain KV stores, and this should be an EAV table. But that's just data conversion

PR #940, opened 2022-11-17
kolide server data table

James-Pickett:
* removes kolide_target_membershipt table and associated grpc / protobuf code
* adds kolide_server_data table to expose data in server_provided_data bucket

closes https://github.com/kolide/launcher/issues/936

The issue mentions organizing the data in to an EAV format, however the data in the DB is stored in KV format. I could not think of any major benefit to organizing the KVs into EAVs. I believe this would involve coming up with with some kind of contract on how data would be stored in the db.

For example if we agreed keys in the DB would all be organized as something like `grandparent/parent/child` we could use data flatten, but what does that buy us? The ability to use the `query` option? We could just as easily say `select value where key like "grandparent/parent/%"` to get all the particular children.

If anyone sees some benefits I do not, please let me know, happy to update.


pkg/osquery/tables/kolide_server_data/kolide_server_data.go
```diff
@@ -0,0 +1,50 @@
...
+		return dbKeyValueRows(osquery.ServerProvidedDataBucket, db)
+	}
+}
+
+func dbKeyValueRows(bucketName string, db *bbolt.DB) ([]map[string]string, error) {
+	results := make([]map[string]string, 0)
+
+	if err := db.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(bucketName))
+		b.ForEach(func(k, v []byte) error {
```
directionless, line 37:
```suggestion
		b := tx.Bucket([]byte(bucketName))
		if b.nil {
		    return errors.New("Bucket not found")
		}
		b.ForEach(func(k, v []byte) error {
```

Or something

directionless (approved):

PR #944, opened 2022-11-18
Change file extension to .json for debug log

RebeccaMahany:
Rename `debug.log` to `debug.json`. Log rotation is already handled by [lumberjack](https://pkg.go.dev/gopkg.in/natefinch/lumberjack.v2).


directionless (dismissed):
Seems fine, but this is not enough to remove the old `.log` files.

I suspect we should add something to just rm anything that matches the glob `RootDir/debug*.log` and `RootDir/debug*.log.gz`. That could land in a separate PR

pkg/log/locallogger/locallogger_test.go
```diff
@@ -73,3 +75,46 @@ func TestKitLogging(t *testing.T) {
...
+	legacyDebugLogPath := filepath.Join(tempDir, "debug.log")
+	f1, err := os.Create(legacyDebugLogPath)
+	require.NoError(t, err, "could not create log file for test")
+	defer f1.Close()
+
+	// Make a rotated log file that should be cleaned up
+	legacyDebugLogRotatedPath := filepath.Join(tempDir, "debug-2022-11-18T18-35-48.858.log.gz")
+	f2, err := os.Create(legacyDebugLogRotatedPath)
+	require.NoError(t, err, "could not create log file for test")
+	defer f2.Close()
```
directionless, line 94: As a small nit, you don't need to defer these -- nothing is being written to them. They can get closed immediately. 

Could probably even do it in for loop  with all the names, if you wanted to be fancy WRT variables.

pkg/log/locallogger/locallogger.go
```diff
@@ -40,6 +43,26 @@ func (ll localLogger) Log(keyvals ...interface{}) error {
...
+	// `debug-<date>.log.gz`. We do not return an error if we can't clean up these files -- it's not
+	// a big deal.
+	legacyDebugLogPattern := filepath.Join(cleanupPath, "debug*.log*")
+	filesToCleanUp, err := filepath.Glob(legacyDebugLogPattern)
+	if err != nil {
+		level.Error(logger).Log("msg", "could not glob for legacy debug log files to clean up", "pattern", legacyDebugLogPattern, "err", err)
+	} else {
+		for _, fileToCleanUp := range filesToCleanUp {
+			err := os.Remove(fileToCleanUp)
+			if err != nil {
```
directionless, line 59:
```suggestion
			if err := os.Remove(fileToCleanUp); err != nil {
```

slight preference for not defining `err`

directionless (approved):
I'd take it!

pkg/log/locallogger/locallogger_test.go
RebeccaMahany: Thanks for the suggestion -- tidied this up 🙂

pkg/log/locallogger/locallogger.go
```diff
@@ -40,6 +43,25 @@ func (ll localLogger) Log(keyvals ...interface{}) error {
...
+	if err != nil {
+		level.Error(logger).Log("msg", "could not glob for legacy debug log files to clean up", "pattern", legacyDebugLogPattern, "err", err)
+	} else {
+		for _, fileToCleanUp := range filesToCleanUp {
+			if err := os.Remove(fileToCleanUp); err != nil {
+				level.Error(logger).Log("msg", "could not clean up legacy debug log file", "file", fileToCleanUp, "err", err)
+			}
+		}
+	}
+}
```
James-Pickett, line 63: sorry to jump in an nit pick ... but we can just return if err != nil and reduce nesting

```suggestion
	if err != nil {
		level.Error(logger).Log("msg", "could not glob for legacy debug log files to clean up", "pattern", legacyDebugLogPattern, "err", err)
		return
	}
        for _, fileToCleanUp := range filesToCleanUp {
	        if err := os.Remove(fileToCleanUp); err != nil {
		        level.Error(logger).Log("msg", "could not clean up legacy debug log file", "file", fileToCleanUp, "err", err)
	        }
        }
}
```

RebeccaMahany: Updated, thanks!

James-Pickett (approved):

seejdev (approved):

Issue #928, opened 2022-11-02
Table for `xfconf`


directionless:
While digging into a xfce` based linux device, we noticed that it does not use `gsettings`. This is, of course, not a surprise.

But, it looks like there's a `xfconf-query` command. So we ought be able to make a table exposing that info.

I haven't done much research. Maybe the data is stored somewhere we can access it directly.

RebeccaMahany:
It looks like we could read from the config files: `~/.config/xfce4/xfconf/xfce/perchannel/xml/*.xml`. I don't know whether this is preferable to running `xfconf-query` a bunch -- it looks like we have dataflatten support for xml, so maybe that's simpler.

directionless:
I don't have a well considered answer for `kolide_xm` vs parsing `xfconf-query`. Thinking aloud...

* For `gsettings`, I couldn't figure out how to do it without an exec. Parts seemed to come down to `dbus` and and it looked hard. While xfconf is not gsettings, I assumed it would be similarly hard and didn't dig too deeply
* Generally, exec it's great. Reading data files feels better
* But getting the application specific rules feels hard. We can do simple merges, but is there more nuance? (Like default values)
* And will it change?

I think we could go either way -- use `kolide_xml`, or write a table that execs `xfconf-query`. We could use the former until we see enough of a need for the later. Or we can do the latter because it's fun. 🤷

PR #948, opened 2022-11-22
Add kolide_xfconf table for devices using xfce

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/928.

<details>
<summary>Example output</summary>

Invalid username:

```
osquery> select * from kolide_xfconf where username = "very fake";
Error: error generating table: finding user by username 'very fake': user: unknown user very fake
```

Get lock screen settings:

```
osquery> select * from kolide_xfconf where username = 'headless' and fullkey = 'channel/xfce4-power-manager/xfce4-power-manager/lock-screen-suspend-hibernate';
+-------------------------------------------------------------------------------+-------------------------------------------------+-------------------------------+-------+-------+----------+---------------------------------------------------------------------------------+
| fullkey                                                                       | parent                                          | key                           | value | query | username | path                                                                            |
+-------------------------------------------------------------------------------+-------------------------------------------------+-------------------------------+-------+-------+----------+---------------------------------------------------------------------------------+
| channel/xfce4-power-manager/xfce4-power-manager/lock-screen-suspend-hibernate | channel/xfce4-power-manager/xfce4-power-manager | lock-screen-suspend-hibernate | true  | *     | headless | /home/headless/.config/xfce4/xfconf/xfce-perchannel-xml/xfce4-power-manager.xml |
+-------------------------------------------------------------------------------+-------------------------------------------------+-------------------------------+-------+-------+----------+---------------------------------------------------------------------------------+
```

Get autorun settings:

```
osquery> select * from kolide_xfconf where username = 'headless' and fullkey like 'channel/thunar-volman%';
+------------------------------------------------+----------------------------------------+---------+-------+-------+----------+-------------------------------------------------------------+
| fullkey                                        | parent                                 | key     | value | query | username | path                                                        |
+------------------------------------------------+----------------------------------------+---------+-------+-------+----------+-------------------------------------------------------------+
| channel/thunar-volman/automount-media/enabled  | channel/thunar-volman/automount-media  | enabled | false | *     | headless | /etc/xdg/xfce4/xfconf/xfce-perchannel-xml/thunar-volman.xml |
| channel/thunar-volman/automount-drives/enabled | channel/thunar-volman/automount-drives | enabled | false | *     | headless | /etc/xdg/xfce4/xfconf/xfce-perchannel-xml/thunar-volman.xml |
| channel/thunar-volman/autobrowse/enabled       | channel/thunar-volman/autobrowse       | enabled | false | *     | headless | /etc/xdg/xfce4/xfconf/xfce-perchannel-xml/thunar-volman.xml |
| channel/thunar-volman/autoopen/enabled         | channel/thunar-volman/autoopen         | enabled | false | *     | headless | /etc/xdg/xfce4/xfconf/xfce-perchannel-xml/thunar-volman.xml |
+------------------------------------------------+----------------------------------------+---------+-------+-------+----------+-------------------------------------------------------------+
```

Use query to fetch auto-open settings for removable media:
```
osquery> select * from kolide_xfconf where username = 'headless' and query = '*/autoopen*';
+----------------------------------------+--------------------------------+---------+-------+-------------+----------+-------------------------------------------------------------+
| fullkey                                | parent                         | key     | value | query       | username | path                                                        |
+----------------------------------------+--------------------------------+---------+-------+-------------+----------+-------------------------------------------------------------+
| channel/thunar-volman/autoopen/enabled | channel/thunar-volman/autoopen | enabled | false | */autoopen* | headless | /etc/xdg/xfce4/xfconf/xfce-perchannel-xml/thunar-volman.xml |
+----------------------------------------+--------------------------------+---------+-------+-------------+----------+-------------------------------------------------------------+
```
</details>

<details>

<summary>
Testing steps
</summary>

* Use image https://hub.docker.com/r/accetto/ubuntu-vnc-xfce-g3 -- `docker run --platform linux/amd64 -it --name accettovnc -p 25901:5901 -p 26901:6901 accetto/ubuntu-vnc-xfce-g3:latest`
* Install xfce4-power-manager, thunar-volman, and dbus-x11; run `export $(dbus-launch)`; install osquery; copy built binary into container
* Over novnc (http://localhost:26901/vnc.html) configure some power and autorun settings in the GUI to query later
* In the container shell, run launcher interactive; new table `kolide_xfconf` should be listed in schema
* Query by username and optionally channel/key

</details>


pkg/osquery/tables/xfconf/parser.go
```diff
@@ -0,0 +1,100 @@
...
+		Values     []ArrayValue `xml:"value"`
+	}
+	ChannelXML struct {
+		XMLName     xml.Name   `xml:"channel"`
+		ChannelName string     `xml:"name,attr"`
+		Properties  []Property `xml:"property"`
+	}
+)
+
+// parseXml reads in the given xml file, parses it as an xfconf XML file, and then flattens
```
directionless, line 33:
```suggestion
// parseXfconfXml reads in the given xml file, parses it as an xfconf XML file, and then flattens
```

pkg/osquery/tables/xfconf/parser.go
```diff
@@ -0,0 +1,100 @@
...
+	xmlDecoder := xml.NewDecoder(rdr)
+
+	var result ChannelXML
+	xmlDecoder.Decode(&result)
+
+	return result, nil
+}
+
+// toMap transforms Result r into a map where the top-level key is "channel/<name>".
+func (c ChannelXML) toMap() map[string]interface{} {
```
directionless, line 60: This seems fine, though I'm somewhat surprised we need it. You're starting with a well formed `ChannelXML` object? We _ought_ be able to convert it. On the other hand, I bet this works. So 🤷

pkg/osquery/tables/xfconf/parser_test.go
```diff
@@ -0,0 +1,262 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, tt := range testCases {
+		result, err := readChannelXml(tt.filePath)
```
directionless, line 258:
```suggestion
	for _, tt := range testCases {
	        tt := tt
		result, err := readChannelXml(tt.filePath)
```

Pretty sure we need the tt thing.

Consider, though not required, `t.Run(...)`

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,164 @@
...
+	users := tablehelpers.GetConstraints(queryContext, "username")
+	if len(users) < 1 {
+		return results, errors.New("kolide_xfconf requires at least one username to be specified")
+	}
+
+	for _, username := range users {
+		u, err := user.Lookup(username)
+		if err != nil {
+			return nil, fmt.Errorf("finding user by username '%s': %w", username, err)
+		}
```
directionless, line 52: I'm not sure if this should be a hard error (as you have it) or a soft error (eg: continue with a log message). Hard errors will cause a large blob of SQL o fail, soft errors are more graceful.

For something like the user check, hard error makes sense. It's a syntactic failure. This feels a bit closer to an exec or OS failure, and we probably don't want to through an error. 

But I could go either way

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return []string{filepath.Join("/", "etc", "xdg", xfconfChannelXmlPath)}
+}
+
+// getUserXfconfDir returns to xfconf per-channel xml files that contain user-specific
+// settings. It checks for an override via environment variable, otherwise defaults to
+// ~/.config/...
+// See https://docs.xfce.org/xfce/xfce4-session/advanced#files_and_environment_variables.
+func getUserXfconfDir(u *user.User) string {
+	userConfigDir := os.Getenv("XDG_CONFIG_HOME")
```
directionless, line 141: I'm not not totally sure how I feel about the `os.Getenv` here. It's not wrong, but...
1. Do we think this will be set in a way daemons pick up? I suspect not since it feels like this is meant to override `~/.config`, so if it's set, it won't have a user portion.
2. It's basically a constant. (I don't think the ENV can change once the program is launched). So we should be able to read it once early on.

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// First, get user-specific settings
+	userConfigDir := getUserXfconfDir(u)
+	userConfigRows, err := t.getConfigFromDirectory(userConfigDir, dataQuery, rowData)
+	if err != nil {
+		return nil, fmt.Errorf("error getting config for user %s from directory %s: %w", u.Name, userConfigDir, err)
+	}
+	results = append(results, userConfigRows...)
+
+	// Then, get defaults
```
directionless, line 80: Defaults won't change during the course of a `generate` call, but we will likely invoke for multiple users. This should, ideally, move out of the `getUserConfig` so we don't regen each time.

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return results, nil
+}
+
+// getDefaultXfconfDirs returns the path to xfconf per-channel xml files that contain
+// default settings. It checks for an override, otherwise defaults to /etc/xdg/...
+// See https://docs.xfce.org/xfce/xfce4-session/advanced#files_and_environment_variables.
+func getDefaultXfconfDirs() []string {
+	envDefaultDirsStr := os.Getenv("XDG_CONFIG_DIRS")
```
directionless, line 124: Similar as before. Though this shouldn't be user specific, it feels like it should be a top level call

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, dir := range defaultDirs {
+		defaultConfigRows, err := t.getConfigFromDirectory(dir, dataQuery, rowData)
+		if err != nil {
+			return nil, fmt.Errorf("error getting config for user %s from default directory %s: %w", u.Name, dir, err)
+		}
+		results = append(results, defaultConfigRows...)
+	}
+
+	// Make sure we only include defaults if there isn't already a user-specific setting
+	return deduplicate(results), nil
```
directionless, line 91:
```suggestion
	// deduplicate the user and default configs, by taking the first instance in the results array
	return deduplicate(results), nil
```

This is fairly clever, though I'd suggest a slightly different comment.

One caveat, though this is a note for a different world, this approach has a lot of excess allocation. I don't think that's an issue, it's a bit inherent in how we sling `map[string]any` around. But I do want to mention it, since it matters in some of the other code paths.

pkg/osquery/tables/xfconf/parser.go
RebeccaMahany: I didn't try! Let me see how that works

pkg/osquery/tables/xfconf/xfconf.go
RebeccaMahany: 1 was kind of what I was worried about. I tested by setting the variable in `/etc/environment` and confirmed that launcher picked it up, but 🤷‍♀️

pkg/osquery/tables/xfconf/xfconf.go
RebeccaMahany: Oh yup, good suggestion, I will move this and https://github.com/kolide/launcher/pull/948/#discussion_r1036632898 up so we don't call it for every user

pkg/osquery/tables/xfconf/xfconf.go
RebeccaMahany: I can switch to a soft error -- I see e.g. gsettings does a soft error in a similar situation.

pkg/osquery/tables/xfconf/parser.go
```diff
@@ -0,0 +1,100 @@
...
+	var propertyValue interface{}
+	if len(p.Properties) > 0 {
+		// This is an empty property with nested properties inside -- extract them.
+		childPropertyMaps := make(map[string]interface{}, 0)
+		for _, child := range p.Properties {
+			// Call recursively for each child
+			childPropertyMaps[child.Name] = child.mapValue()
+		}
+
+		propertyValue = childPropertyMaps
```
James-Pickett, line 85: you could just return the value here and git rid of some else statements

pkg/osquery/tables/xfconf/xfconf.go
directionless: We used to use a lot of hard errors, then we had issues, now we're moving to a lot more soft errors. It's not yet clear to me what ideal is

pkg/osquery/tables/xfconf/parser_test.go
James-Pickett: I agree with using `t.Run(...)`. If I understand corrected the `tt := tt` is used to copy the test case before a call to `t.Parallel()` inside `t.Run(...)`. Without out the `tt := tt`, the original `tt` declared in the range statement has the potential to get replaced before a previous `t.Run(...)` has completed.

directionless: _Ah_. Yes, James is right -- we only need `tt := tt` if there's a goroutine (or t.Run) in the mix. Otherwise it's fine.

(Side note, I think this is changing in a future go version)

RebeccaMahany: Updated to do this + `t.Run(...)` -- thank you for the explanation 🙂

pkg/osquery/tables/xfconf/parser.go
RebeccaMahany: Seeing this:

```
flattening at path []: unknown type on [{{ channel} thunar-volman [{automount-media empty  [{enabled bool false [] []}] []} {automount-drives empty  [{enabled bool false [] []}] []} {autobrowse empty  [{enabled bool false [] []}] []} {autoopen empty  [{enabled bool false [] []}] []}]} {{ channel} xfce4-power-manager [{xfce4-power-manager empty  [{power-button-action empty  [] []} {lock-screen-suspend-hibernate bool true [] []}] []}]}]
```

pkg/osquery/tables/xfconf/xfconf.go
RebeccaMahany: Do you think I should leave this as-is, then, or run `echo $XDG_CONFIG_HOME` as the target user? Or another option?

pkg/osquery/tables/xfconf/parser.go
directionless: Yeah... There's something weird in the `Flatten` routine. You can marshal to json, and use FlattenJson, or leave the `toMap` stuff. I suspect we should think a bit harder someday

pkg/osquery/tables/xfconf/xfconf.go
directionless: I wouldn't `echo $XDG_CONFIG_HOME` -- adding in an exec just for that feel premature. 

I think we could leave it, or remove it. Your call.

RebeccaMahany: I guess leave it, then!

pkg/osquery/tables/xfconf/parser.go
RebeccaMahany: I think I'd like to keep `toMap`, then, since it gives us the nice `channel/<name>` top-level key

directionless: 👍

pkg/osquery/tables/xfconf/xfconf.go
```diff
@@ -0,0 +1,230 @@
...
+type xfconfTable struct {
+	logger log.Logger
+}
+
+func TablePlugin(logger log.Logger) *table.Plugin {
+	t := &xfconfTable{
+		logger: logger,
+	}
+
+	return table.NewPlugin("kolide_xfconf", dataflattentable.Columns(table.TextColumn("username")), t.generate)
```
directionless, line 39: Do you think it's worth returning the `path` we found the data at? I think it might be.

directionless (approved):
I think this is okay, I'd take it.

I don't know if we'll want the `path` in this. But :shrug: we can kick that forward.

pkg/osquery/tables/xfconf/xfconf.go
RebeccaMahany: That's a good idea, let me add it 🙂

RebeccaMahany: All set -- examples in the PR description updated also

seejdev (approved):

James-Pickett (approved):

Issue #935, opened 2022-11-15
Revisit and rewrite the control server


directionless:
The control server is coming back! This is a fairly large project that needs a bunch of PRs. The end goal is to have launcher have a control channel to our K2 (our SaaS) which is used for controller and changing launcher's behavior.

Task:
* [x] Enable the control server on windows. It's a no-op, but it'll improve code consistency. (small)
* [x] Design and document a control protocol
* [x] Authentication ??? Should it use the osquery node key (how?) or the launcher RSA keys?
* [x] #936

Some likely control things. Not totally sure how these will shape up:
1. enable/disable desktop
1. Change update channel
1. Re-trigger the updater, not just via callback

directionless:
This is done and live

PR #961, opened 2022-12-05
Control server phase 1

seejdev:
### Rewrite the control server phase 1: Introduce ControlService and a simple communication protocol between client and server.

This MR is the first iterative step towards completing #935, here's the high level changes:

- Enables the control server on Windows
- If enabled, the control service can make requests to K2 to retrieve control server updates, at a specified interval
- Boilerplate code for supporting `consumer` and `subscriber` concepts
- Supporting code to enable handling caching of updates
- Initial draft of documentation covering architecture and control protocol 

This MR aims to simply enable client<->server communications without getting too far into the implementation details of: the update contents themselves, consumer/subscriber interactions, and caching mechanisms. Next phase would iterate and enable more functionality in these areas.


ee/control/client_http.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// controlRequest is the payload sent in control server requests
+type controlRequest struct {
+	// TODO: This is a temporary and simple data format for phase 1
+	Message string `json:"message"`
+}
+
+func NewControlHTTPClient(addr string, opts ...HTTPClientOption) (*HTTPClient, error) {
+	baseURL, err := url.Parse("https://" + addr)
```
James-Pickett, line 33: preference for Sprintf over concat
```suggestion
	baseURL, err := url.Parse(fmt.Sprintf("https://%s", "addr"))
```

ee/control/client_http.go
```diff
@@ -0,0 +1,154 @@
...
+			"err", err,
+		)
+		return "", nil, err
+	}
+	defer response.Body.Close()
+
+	switch response.StatusCode {
+	case http.StatusNotFound:
+		// This could indicate an inconsistency in server data, or a client logic error
+		level.Debug(c.logger).Log(
```
James-Pickett, line 73:
```suggestion
		level.Error(c.logger).Log(
```

ee/control/client_http.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (c *HTTPClient) Get(subsystem, cachedETag string) (etag string, data io.Reader, err error) {
+	verb, path := "GET", "/api/v1/control"
+	params := &controlRequest{
+		Message: "ping",
+	}
+
+	response, err := c.do(verb, path, cachedETag, params)
+	if err != nil {
+		level.Debug(c.logger).Log(
```
James-Pickett, line 62:
```suggestion
		level.Error(c.logger).Log(
```

James-Pickett (commented):
Looking pretty good to me, I was able to follow code and understand whats going on here! Left a few nits.

docs/architecture/control_server.md
```diff
@@ -0,0 +1,44 @@
...
+update for a subsystem, it notifies a consumer registered to handle updates for the subsystem, then pings all subscribers for the subsystem.
+
+The latest update for each subsystem is cached so the control server can avoid re-sending updates
+previously sent to a launcher instance.
+
+
+## Protocol
+
+```mermaid
+sequenceDiagram
```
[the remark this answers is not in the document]
RebeccaMahany: ➕ 1️⃣ -- docs were super helpful

ee/control/control.go
```diff
@@ -0,0 +1,140 @@
...
+	// 	otherwise
+	// 		ask the dataProvider for the latest update for this subsystem (make a new HTTP request)
+
+	// 		if latest update is successfully fetched
+	// 			notify the consumer of this subsystem, and give them the control data
+
+	// 			for each subscriber of this subsystem
+	// 				ping the subscriber
+
+	// 			cache latest update hash retrieved for this subsystem
```
RebeccaMahany, line 113: Where would the caching happen?

seejdev: So far I've kept it simple by using an in-memory cache, `ControlService.lastFetched` would hold the last fetched etag for each subsystem. e.g. after fetching and processing an update for the `desktop` subsystem, it would be cached like:

```
lastFetched[subsystem] = etag
```

Subsequent fetches can use these cached etags in their fetch requests, which will result in either the server sending back a fresh update (stale etag), or an "empty" response indicating that the last update fetched is still the most fresh. Since it is in-memory, launcher will re-fetch everything following a process restart/crash, but if we find that undesirable in the future, we could easily make it persist cache info to disk.

ee/control/client_http.go
```diff
@@ -0,0 +1,154 @@
...
+		addr:    addr,
+	}
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	if c.disableTLS {
+		c.baseURL.Scheme = "http"
+	}
```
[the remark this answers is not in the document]
seejdev: Yeah I like that better!

cmd/launcher/options.go
```diff
@@ -45,25 +45,26 @@ func parseOptions(args []string) (*launcher.Options, error) {
...
-		flRootPEM              = flagset.String("root_pem", "", "Path to PEM file including root certificates to verify against")
-		flVersion              = flagset.Bool("version", false, "Print Launcher version and exit")
-		flLogMaxBytesPerBatch  = flagset.Int("log_max_bytes_per_batch", 0, "Maximum size of a batch of logs. Recommend leaving unset, and launcher will determine")
-		flOsqueryFlags         arrayFlags // set below with flagset.Var
-		flCompactDbMaxTx       = flagset.Int64("compactdb-max-tx", 65536, "Maximum transaction size used when compacting the internal DB")
-		_                      = flagset.String("config", "", "config file to parse options from (optional)")
+		flAutoloadedExtensions   arrayFlags
+		flCertPins               = flagset.String("cert_pins", "", "Comma separated, hex encoded SHA256 hashes of pinned subject public key info")
+		flControl                = flagset.Bool("control", false, "Whether or not the control server is enabled (default: false)")
+		flControlServerURL       = flagset.String("control_hostname", "", "The hostname of the control server")
```
directionless, line 51: I think that we're going to want to keep this enabled for Kolide customers, and that because it's under the EE license, no one else will use it. Eg: We can probably remove these options and hardcode them based on where we're hardcoding from the server names

cmd/launcher/launcher.go
```diff
@@ -179,17 +179,13 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		"build", versionInfo.Revision,
 	)
 
-	// If the control server has been opted-in to, run it
+	// If the control server has been opted-in to, run the control service
 	if opts.Control {
```
directionless, line 183: This block should move into one of the is kolide conditionals

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+	logger          log.Logger
+	cancel          context.CancelFunc
+	requestInterval time.Duration
+	consumers       map[string]consumer
+	subscribers     map[string][]subscriber
+	data            dataProvider
+	lastFetched     map[string]string
+}
+
+// consumer is an interface for something that consumes control server data updates.
```
directionless, line 26:
```suggestion
// consumer is an interface for something that consumes control server data updates. The control server supports at most one consumer per subsection.
```

maybe?

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+type consumer interface {
+	Update(io.Reader)
+}
+
+// subscriber is an interface for something that wants to be notified when a subsystem has been updated.
```
directionless, line 31:
```suggestion
// subscriber is an interface for something that wants to be notified when a subsystem has been updated. Subscribers do not receive data -- they are expected to read the data from where consumers write it. 
```

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+type subscriber interface {
+	Ping()
+}
+
+// dataProvider is an interface for something that can retrieve control data. Authentication, HTTP,
+// file system access, etc. lives below this abstraction layer.
+type dataProvider interface {
+	Get(subsystem, cachedETag string) (etag string, data io.Reader, err error)
```
directionless, line 39: I haven't read all the code yet, but how would this represent that there was nothing changed from the GET request?

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+
+// ControlService is the main object that manages the control service. It is responsible for fetching
+// and caching control data, and updating consumers and subscribers.
+type ControlService struct {
+	logger          log.Logger
+	cancel          context.CancelFunc
+	requestInterval time.Duration
+	consumers       map[string]consumer
+	subscribers     map[string][]subscriber
+	data            dataProvider
```
directionless, line 22:
```suggestion
	dataGetter            dataProvider
```

This doesn't hold data, it just gets it, right? Feels like we should reflect that in the name. Counterpoint, `dataGetter.Get` feels redundant. Maybe `fetcher` instead? `fetcher.Get` doesn't sound bad

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+		opt(cs)
+	}
+
+	return cs
+}
+
+func (cs *ControlService) Start(ctx context.Context) {
+	ctx, cs.cancel = context.WithCancel(ctx)
+	requestTicker := time.NewTicker(cs.requestInterval)
+	for {
```
directionless, line 61: Might not matter, but IIRC this won't do the initial fetch until after requestInterval has elapsed. Probably needs an extra `cs.Fetch()` before the loop.

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+	// 			ping the subscriber
+
+	// 		cache latest update hash retrieved for this subsystem
+
+	return nil
+}
+
+func (cs *ControlService) RegisterConsumer(subsystem string, consumer consumer) error {
+	if _, ok := cs.consumers[subsystem]; ok {
+		return fmt.Errorf("consumer already registered for subsystem %s", subsystem)
```
directionless, line 122: I read this error as saying _this_ consumer is registered. Maybe:
```suggestion
		return fmt.Errorf("subsystem %s already has registered consumer", subsystem)
```

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+// controlResponse is the payload received from the control server
+type controlResponse struct {
+	// TODO: This is a temporary and simple data format for phase 1
+	Message string `json:"message,omitempty"`
+	Err     string `json:"error,omitempty"`
+}
+
+// Performs a retrieval of the latest control server data, and notifies observers of updates.
+func (cs *ControlService) Fetch() error {
+	_, data, err := cs.data.Get("", "")
```
directionless, line 84: None of the etagging is plumbed here, correct?

ee/control/control.go
```diff
@@ -0,0 +1,142 @@
...
+	Ping()
+}
+
+// dataProvider is an interface for something that can retrieve control data. Authentication, HTTP,
+// file system access, etc. lives below this abstraction layer.
+type dataProvider interface {
+	Get(subsystem, cachedETag string) (etag string, data io.Reader, err error)
+}
+
+func NewControlService(logger log.Logger, data dataProvider, opts ...Option) *ControlService {
```
directionless, line 42:
```suggestion
func New(logger log.Logger, data dataProvider, opts ...Option) *ControlService {
```

ee/control/control_test.go
```diff
@@ -0,0 +1,163 @@
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			cs := testCS()
+			err := cs.RegisterConsumer(tt.subsystem, tt.c)
+			require.NoError(t, err)
```
directionless, line 78: Should this test that the consume is actually registered? I might say it's a subset of the `TestControlServiceUpdate` path.

ee/control/control_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+type nopDataProvider struct{}
+
+func (dp nopDataProvider) Get(subsystem, cachedETag string) (etag string, data io.Reader, err error) {
+	return "", nil, nil
+}
+
+func testCS() *ControlService {
+	data := nopDataProvider{}
+	cs := &ControlService{
```
directionless, line 41: Feels a little weird. I think a lot of this would vanish if you called `New()`, might be reduceable to `New(data)`. But I don't know if you have plans to expand the complexity here. 

I guess even if you keep `testCS` I think you should use New and not instantiate directly.

ee/control/control_test.go
```diff
@@ -0,0 +1,163 @@
...
+			err := cs.RegisterConsumer(tt.subsystem, tt.c)
+			require.NoError(t, err)
+			for _, ss := range tt.s {
+				cs.RegisterSubscriber(tt.subsystem, ss)
+			}
+
+			cs.update(tt.subsystem, nil)
+
+			assert.Equal(t, updateCount, 1)
+			assert.Equal(t, pingCount, 2)
```
directionless, line 160: Love this

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func NewControlHTTPClient(addr string, opts ...HTTPClientOption) (*HTTPClient, error) {
+	baseURL, err := url.Parse(fmt.Sprintf("https://%s", addr))
+	if err != nil {
+		return nil, fmt.Errorf("parsing URL: %w", err)
+	}
+	c := &HTTPClient{
+		baseURL: baseURL,
+		client:  http.DefaultClient,
```
directionless, line 39: I don't know best practice, but I suspect we should instantiate the http.Client in the main routine and pass it to this

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+	defer response.Body.Close()
+
+	switch response.StatusCode {
+	case http.StatusNotFound:
+		// This could indicate an inconsistency in server data, or a client logic error
+		level.Error(c.logger).Log(
+			"msg", "got HTTP 404 making control server request",
+			"err", err,
+		)
+		return "", nil, err
```
directionless, line 73:
```suggestion
		)
		return "", nil, nil
```

`err` should be nil here.

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+
+	case http.StatusNotModified:
+		// The control server sends back a 304 Not Modified status, without a body, which tells
+		// the client that the cached version of the response is still good to use
+		level.Debug(c.logger).Log(
+			"msg", "got HTTP 304 making control server request",
+			"err", err,
+		)
+		return "", nil, err
```
directionless, line 82: same err comment

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	if response.StatusCode != http.StatusOK {
+		level.Error(c.logger).Log(
```
directionless, line 86: Should we log the body?

I suspect we should log these at debug, these aren't things I think an end user can action

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+			"msg", "got HTTP 404 making control server request",
+			"err", err,
+		)
+		return "", nil, err
+
+	case http.StatusNotModified:
+		// The control server sends back a 304 Not Modified status, without a body, which tells
+		// the client that the cached version of the response is still good to use
+		level.Debug(c.logger).Log(
```
directionless, line 78: I don't know what the answer is, but this will be noise in the debug log. OTOH, no worse than all the other requestqueries and requestconfig noise. So :shrug: let's keep it

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if response.StatusCode != http.StatusOK {
+		level.Error(c.logger).Log(
+			"msg", "got not-ok status code from control server",
+			"response_code", response.StatusCode,
+		)
+		return "", nil, err
+	}
+
+	body, err := ioutil.ReadAll(response.Body)
```
directionless, line 93: `ioutil` is deprecated

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		level.Error(c.logger).Log(
+			"msg", "error reading response body from control server",
+			"err", err,
+		)
+		return "", nil, err
+	}
+
+	reader := bytes.NewReader(body)
```
directionless, line 102: Why read this into ram, and then return a reader? Can we just return response.Body?

directionless (commented):
The over all shape feels reasonable. I found myself bikeshedding a lot of little nits, which as things go, feels like a good sign.

I'm not totally sure how the http headers/body/etag will all shake out. But I don't think we need to block on that. It'll evolve as we write it.

ee/control/control.go
RebeccaMahany: Thanks, I was curious re behavior on restart. Makes sense to me!

ee/control/client_http.go
```diff
@@ -0,0 +1,150 @@
...
+			"err", err,
+		)
+		return "", nil, err
+
+	case http.StatusNotModified:
+		// The control server sends back a 304 Not Modified status, without a body, which tells
+		// the client that the cached version of the response is still good to use
+		level.Debug(c.logger).Log(
+			"msg", "got HTTP 304 making control server request",
+			"err", err,
```
seejdev, line 80:
```suggestion
```

ee/control/control.go
seejdev: 👍 removed etag stuff overall

ee/control/control.go
seejdev: I've simplified this interface to remove the etag concept and is now just

```
Get(hash string) (data io.Reader, err error)
```

Instead of worrying about etags, the client can keep track of what object hashes it has asked for. When it finds the same hash for a subsystem, it can just choose not to request that same hash again. We also avoid hitting the server endpoint for it just to say "you already have this".

ee/control/client_http.go
seejdev: Good idea, can't hurt!

ee/control/control.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return cs
+}
+
+func (cs *ControlService) Start(ctx context.Context) {
+	ctx, cs.cancel = context.WithCancel(ctx)
+	requestTicker := time.NewTicker(cs.requestInterval)
+	for {
+		// Fetch immediately on each iteration, avoiding the initial ticker delay
+		cs.Fetch()
```
James-Pickett, line 65: 👍 

wow, I don't know how I never saw this as a pattern, this is superior to what I always used in the past:

```go
foo()

for {
  select {
    case <- a:
      return
    case <- b:
      foo()
  }
}
```

ee/control/control_test.go
James-Pickett: I love this too, but won't it turn into a race condition if we add another test case? I guess CI will catch it.

James-Pickett (approved):

ee/control/client_http.go
```diff
@@ -0,0 +1,134 @@
...
+			"response_code", response.StatusCode,
+			"response_body", response.Body,
+		)
+		return nil, err
+	}
+
+	// response.Body will be closed before this function exits, get all the data now
+	body, err := io.ReadAll(response.Body)
+	if err != nil {
+		level.Error(c.logger).Log(
```
directionless, line 94: I'm okay merging as is, but I think we should probably not use `Error` here. This is unlikely an error the customer can do anything about

ee/control/control.go
```diff
@@ -0,0 +1,148 @@
...
+	Get(hash string) (data io.Reader, err error)
+}
+
+func New(logger log.Logger, data dataProvider, opts ...Option) *ControlService {
+	cs := &ControlService{
+		logger:          logger,
+		requestInterval: 60 * time.Second,
+		consumers:       make(map[string]consumer),
+		subscribers:     make(map[string][]subscriber),
+		fetcher:         data,
```
directionless, line 50:
```suggestion
		fetcher:         fetcher,
```

(Or leave till the next iteration

ee/control/control_test.go
directionless: I don't think so. `updateCount` and `pingCount` are inside the tt run. 

I think there is an issue, in that `1` and `2` should be `tt.expectedPings` and `tt.expectedUpdates`. Can roll that to the next phase though

directionless (approved):

PR #966, opened 2022-12-09
Control server phase 2

seejdev:
### Rewrite the control server phase 2: Control protocol and consumer implementations

This MR is the next iterative step towards completing #935, here's the high level changes:

- Encapsulated consumer/subscriber registration into `ControlRegistry` which is embedded into `ControlService`
- Fetch and consumer/subscriber notification logic is implemented
- `desktop` consumer forwards control data on to the user process
- `kolide_server_data` consumer populates the `server_provided_data` bucket


cmd/launcher/launcher.go
```diff
@@ -179,15 +181,22 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		"build", versionInfo.Revision,
 	)
 
+	var controlRegistry *control.Registry
```
James-Pickett, line 184: why use a *control.Registry var here? since *control.Registry is embedded in the ControlService you could just use
```suggestion
	var controlRegistry *control.ControlService
```

I guess it ensures that anyone outside the block below can only access registry

James-Pickett (commented):
Looking good, this is all I have left for Friday afternoon. Will take a look again with fresh eyes on Monday.

cmd/launcher/launcher.go
seejdev: Yeah that was my thinking - clients need only the reference to the Registry outside of the block below. I don't feel strongly either way though, whatever folks feel is most readable I'm cool with.

cmd/launcher/control.go
```diff
@@ -30,12 +30,12 @@ func createHTTPClient(ctx context.Context, logger log.Logger, opts *launcher.Opt
 	return client, nil
 }
 
-func createControlService(ctx context.Context, logger log.Logger, opts *launcher.Options) (*actor.Actor, error) {
+func createControlService(ctx context.Context, logger log.Logger, opts *launcher.Options) (*actor.Actor, *control.ControlService, error) {
```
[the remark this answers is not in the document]
seejdev: I agree, it's not as easily readable. Let me try and clean this up a bit.

ee/control/bucket_consumer_test.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func createDb(t *testing.T, values map[string]string) *bbolt.DB {
+	dir := t.TempDir()
+
+	db, err := bbolt.Open(filepath.Join(dir, "db"), 0600, nil)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		require.NoError(t, db.Close())
```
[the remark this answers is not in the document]
James-Pickett: I think it's worth checking the error here or at least I don't think there is any harm in the test func. Won't it let you know if something is holding on to a db connection when it should be closed?

cmd/launcher/launcher.go
```diff
@@ -179,15 +180,22 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
 	var runner *desktopRunner.DesktopUsersProcessesRunner
 	if (opts.KolideServerURL == "k2device-preprod.kolide.com" || opts.KolideServerURL == "localhost:3443") && runtime.GOOS != "linux" {
 		// If the control server has been opted-in to, run the control service
 		if opts.Control {
-			control, err := createControlService(ctx, logger, opts)
+			// If the control server has been opted-in to, run the control service
```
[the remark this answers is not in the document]
seejdev: Fixed the duplicate comment - I think we still want to gate the control functionality on the `opts.Control` option for now?

cmd/launcher/control.go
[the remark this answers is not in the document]
seejdev: Good suggestions, thanks!

cmd/launcher/launcher.go
seejdev: Opted to simplify and get rid of the `Registry` idea all together, we can always introduce an interface here if it starts feeling not encapsulated enough.

ee/control/bucket_consumer.go
```diff
@@ -0,0 +1,58 @@
...
+			"data", data,
+			"err", err,
+		)
+	}
+
+	bc.db.Update(func(tx *bbolt.Tx) error {
+		// Clear the bucket first
+		tx.DeleteBucket([]byte(bc.bucketName))
+
+		bucket, err := tx.CreateBucketIfNotExists([]byte(bc.bucketName))
```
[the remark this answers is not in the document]
seejdev: There is some additional complexity within `DeleteBucket` - it will handle deleting child buckets, for example. I couldn't find any API for "remove all keys from the bucket, and leave the bucket empty". So the alternate approach is iterating through with `ForEach` and calling `Delete` for each key.

Deleting the bucket does seem a bit heavy-handed, but I also don't think this particular operation would get executed very frequently, so I'm not too concerned.

`DeleteBucket` does return an error - I ignored it because it's expected the first time this happens the bucket won't exist. I could log the error and proceed, though?

Also - I assumed we wanted to blow away the bucket data and replace it entirely with the fresh payload, but is that necessarily correct? I suppose I don't know exactly all the ways this table is intended to be used. I went with this approach to avoid a scenario where the table grows in size forever without ever being purged. But I could also envision use cases where we don't want to lose what has been added to the table in previous updates, and it may be tedious to ensure the newest payload contains all previous key-values we want to persist.

If we want to support both, the update payload could optionally contain a boolean indicating whether or not we want to clear all previous bucket/table data.

James-Pickett (approved):
LGTM! Excited to see the next phase.

ee/control/bucket_consumer.go
```diff
@@ -0,0 +1,59 @@
...
+		tx.DeleteBucket([]byte(bc.bucketName))
+
+		bucket, err := tx.CreateBucket([]byte(bc.bucketName))
+		if err != nil {
+			return fmt.Errorf("creating bucket: %w", err)
+		}
+
+		for key, value := range kvPairs {
+			if err := bucket.Put([]byte(key), []byte(value)); err != nil {
+				return fmt.Errorf("storing key: %w", err)
```
RebeccaMahany, line 53: would we potentially want to log + continue rather than return an error, to at least get the rest of the config values in there?

seejdev: Good suggestion, I don't see any harm in doing that, fixed!

ee/control/bucket_consumer.go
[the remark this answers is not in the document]
seejdev: I revamped this area a bit to avoid bucket deletion and provide a means to delete keys by sending along empty values.

ee/control/bucket_consumer.go
```diff
@@ -0,0 +1,58 @@
...
+		logger:     logger,
+		db:         db,
+		bucketName: bucketName,
+	}
+
+	return bc
+}
+
+func (bc *BucketConsumer) Update(data io.Reader) {
+	var kvPairs map[string]string
```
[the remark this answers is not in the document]
seejdev: Yeah maybe wait and see, my inclination would doing the serialization in launcher.

ee/control/control.go
```diff
@@ -78,71 +87,63 @@ func (cs *ControlService) Stop() {
...
 
-	// 	if latest update is successfully fetched
-	// 		notify the consumer of this subsystem, and give them the control data
+		// Consumer and subscriber(s) notified now
+		cs.update(subsystem, data)
 
-	// 		for each subscriber of this subsystem
-	// 			ping the subscriber
+		// Remember the hash of the last fetched version of this subsystem's data
+		cs.lastFetched[subsystem] = hash
```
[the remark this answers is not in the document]
seejdev: Error propagation and handling has been improved throughout now.

ee/control/bucket_consumer.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"go.etcd.io/bbolt"
+)
+
+// BucketConsumer processes control server updates for a named bucket
+type BucketConsumer struct {
```
directionless, line 14: Do we need to export this struct?

ee/control/bucket_consumer.go
```diff
@@ -0,0 +1,71 @@
...
+
+			if err := bucket.Put([]byte(key), []byte(value)); err != nil {
+				// Log errors but continue processing the remaining key-values
+				level.Error(bc.logger).Log(
+					"msg", "failed to store key-value in bucket",
+					"key", key,
+					"value", value,
+					"err", err,
+				)
+			}
```
directionless, line 66: I think we should `cursor` iterate the bucket and delete anything not in `kvPairs` here

ee/control/control.go
```diff
@@ -17,16 +17,18 @@ type ControlService struct {
...
 	cancel          context.CancelFunc
 	requestInterval time.Duration
-	consumers       map[string]consumer
-	subscribers     map[string][]subscriber
 	fetcher         dataProvider
 	lastFetched     map[string]string
+	consumers       map[string]consumer
+	subscribers     map[string][]subscriber
+	Execute         func() error
+	Interrupt       func(error)
```
directionless, line 25: Why have these as struct variables, and not methods on the struct?

ee/desktop/client/client.go
```diff
@@ -20,6 +22,12 @@ type client struct {
 	base http.Client
 }
 
+// desktopUserStatus is all the device data sent to the desktop user process
```
directionless, line 25: Do you intend this stuff in this PR? (Okay if so)

ee/control/bucket_consumer.go
seejdev: Done! This required a new transaction and a bit more code than I initially expected, but I like it better now. Also updated the tests to exercise this code.

ee/desktop/client/client.go
seejdev: This is a stub for now, I'll replace this with a real data structure in a separate PR that focuses on just desktop stuff.

ee/control/control.go
seejdev: Good idea, I've made this change.

ee/desktop/client/client.go
directionless: okay, I wasn't sure how much to review it

cmd/launcher/launcher.go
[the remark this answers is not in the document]
seejdev: Gotcha - removed the gate

ee/control/bucket_consumer.go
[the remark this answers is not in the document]
seejdev: Took your suggestion and now pruning stale keys not in the latest update.

directionless (dismissed):
I think is good to merge. 

I'm not totally sure on the `desktop` stuff, but I know it'll get changed.

directionless (approved):

PR #969, opened 2022-12-12
Change flag handling to flag.ContinueOnError

directionless:
If we remove a config option, we cause launcher to not startup. This is a bit scary. One approach would be to change the flag handling to `flag.ContinueOnError`.

This has the advantage of letting things just work.

But the downside, is that `launcher --garbage-flag` won't prouce an error.

I'm honestly on the fence about it. Another route would be to update the check in https://github.com/kolide/launcher/blob/main/pkg/autoupdate/findnew.go#L235 to pass the config file along, and ensure that it fails there. 

Curious what folks think

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -81,7 +81,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	opts, err := parseOptions(os.Args[1:])
+	opts, err := parseOptions(logger, os.Args[1:])
 	if err != nil {
 		level.Info(logger).Log("err", err)
 		os.Exit(1)
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/kit/version"
 	"github.com/kolide/launcher/pkg/autoupdate"
 	"github.com/kolide/launcher/pkg/launcher"
@@ -39,7 +41,7 @@ func (i *arrayFlags) Set(value string) error {
 // parseOptions parses the options that may be configured via command-line flags
 // and/or environment variables, determines order of precedence and returns a
 // typed struct of options for further application use
-func parseOptions(args []string) (*launcher.Options, error) {
+func parseOptions(logger log.Logger, args []string) (*launcher.Options, error) {
 	flagset := flag.NewFlagSet("launcher", flag.ContinueOnError)
 	flagset.Usage = func() { usage(flagset) }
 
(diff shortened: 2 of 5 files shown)


James-Pickett (commented):
When I look at this code I become afraid of introducing silent errors. I think we should at least visit all the flags and log any that we don't recognize or have the wrong value type. In addition, I think we should add logic to check the flags when we check the binary before launch.

seejdev:
The error return value of `ff.Parse(flagset, args, ffOpts...)` is being ignored. We'd likely want to either log that error or bubble it up to something that can log it, just so we know if this is happening.

I share James' concern regarding silent errors, and his suggestions may help. Overall I think this is the right change to make. A silent error is not desirable, but it's not as bad "bricking" launcher installations, so I think tolerating invalid flags is acceptable in this instance.

James-Pickett:
> The error return value of `ff.Parse(flagset, args, ffOpts...)` is being ignored. We'd likely want to either log that error or bubble it up to something that can log it, just so we know if this is happening.
> 
> I share James' concern regarding silent errors, and his suggestions may help. Overall I think this is the right change to make. A silent error is not desirable, but it's not as bad "bricking" launcher installations, so I think tolerating invalid flags is acceptable in this instance.

Sounds reasonable to me, I'm good with the continue on error as long as we are logging any errors.

directionless:
Testing, I don't think this is going to work as I'd hoped. `flag.ContinueOnError` does not cause Parser to continue, instead it causes parser to return an error, instead of calling os.Exit. I would probably say one should always use it. Having the flag library call os.Exit feels like an antipattern. (See [docs](https://pkg.go.dev/flag#ErrorHandling))

But, what this appears to mean, is that we cannot ignore errors -- they always cause the flags to stop being parsed. 

eg:
```go
package main

import (
	"flag"
	"fmt"
	"os"

	"github.com/peterbourgon/ff/v3"
)

func main() {
	fs := flag.NewFlagSet("flagtest", flag.ContinueOnError)

	var (
		flOne = fs.String("one", "unset", "")
		flTwo = fs.String("two", "unset", "")

		_ = fs.String("config", "", "config file (optional)")
	)

	if err := ff.Parse(fs, os.Args[1:], ff.WithConfigFileFlag("config"), ff.WithConfigFileParser(ff.PlainParser)); err != nil {
		// Keep going!
		fmt.Printf("Got error parsing flags: %v\n", err)
	}

	fmt.Printf("One: %s\nTwo: %s\n", *flOne, *flTwo)
}
```

Produces:
```
pemberton:launcher2 seph$ go run ./test-cmds/flag-errors/ --one o --garbage  --two 2
flag provided but not defined: -garbage
Usage of flagtest:
  -config string
    	config file (optional)
  -one string
    	 (default "unset")
  -two string
    	 (default "unset")
Got error parsing flags: error parsing commandline args: flag provided but not defined: -garbage
One: o
Two: unset
```

So I think the next best thing is to have `--version` parse flags

PR #970, opened 2022-12-12
James/tpm

James-Pickett:
* exposes tpm public keys in `kolide_launcher_info` table
* please ignore go.mod & go.sum for now until dependent library is updated


pkg/osquery/table/launcher_info.go
```diff
@@ -49,6 +61,15 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
 			fingerprint = ""
 		}
 
+		if runtime.GOOS == "windows" || runtime.GOOS == "linux" {
```
directionless, line 64: Does it make sense to leave it like this, or move the `setHardwareKeys` to a arch specific file.

Actually, I wonder if we need to push this all the way up to main. How do we handle machines _without_ a TPM? I'm sorta imagining that instead of the place we create RSA keys, we create an interface to the TPM/Secure Enclave/Local Storage and then pass it through somewhere

directionless: Similarly, I think we should expose a column for what the source of the keys are.

pkg/osquery/table/launcher_info.go
```diff
@@ -63,9 +84,69 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
...
 		}
 
 		return results, nil
 	}
 }
+
+type keyer interface {
+	PublicSigningKey() (*rsa.PublicKey, error)
+	PublicEncryptionKey() (*rsa.PublicKey, error)
+}
```
directionless, line 103: Thinking about this -- I think this should never change while launcher is running. So we can make them a constant somewhere

PR #974, opened 2022-12-20
Replaced deprecated set-output usage

seejdev:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -575,7 +575,18 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 
 		// Tell github where we're at
 		if b.githubActionOutput {
-			fmt.Printf("::set-output name=binary::%s\n", output)
+			outputFilePath := os.Getenv("GITHUB_OUTPUT")
+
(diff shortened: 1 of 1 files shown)


pkg/make/builder.go
```diff
@@ -575,7 +575,7 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 
 		// Tell github where we're at
 		if b.githubActionOutput {
-			fmt.Printf("::set-output name=binary::%s\n", output)
+			fmt.Printf("name=binary::%s >> $GITHUB_OUTPUT\n", output)
```
directionless, line 578: The old syntax, was parsed from stdout. the new syntax is written into a file. 

So this needs to to have file open/append with `fmt.Fprintf(f, "binary=%s\n", output)`

seejdev: 🤦 hah, yep!

seejdev: Based on the CI checks I think this is fixed now, but I haven't yet figured out where the output is and how to verify it.

directionless (approved):

Issue #955, opened 2022-12-01
We need a `/usr/local/kolide-k2/bin/launcher` symlink (and maybe adjust some paths)


directionless:
Playing with the new app bundle install (sweeeet) I notice that we no longer have `/usr/local/kolide-k2/bin/launcher`. I think this is a path we should be preserving. With the app bundle, it should be a symlink to `Kolide.app/Contents/MacOS/launcher`

I'm also thinking we should move `/usr/local/kolide-k2/bin/Kolide.app` to `/usr/local/kolide-k2/Kolide.app`

PR #975, opened 2022-12-21
Move app bundle install location out of bin directory; create launcher symlink in bin directory

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/955

Followed testing steps as described in [this PR](https://github.com/kolide/launcher/pull/934), confirming the new location of `Kolide.app` and the presence of the launcher symlink, as well as confirming that the symlink actually works.

Renamed a couple variables I'd named `rootDir` that should have been `binDir` to `binDir`.


directionless (approved):
Let's try!

PR #980, opened 2022-12-23
Ensure findnew finds the updates dir even with the new app bundle location

RebeccaMahany:
Now that the app bundle lives in /usr/local/<identifier> and not /usr/local/<identifier>/bin, we need to make sure it can still find updates located in /usr/local/<identifier>/bin/launcher-updates.


directionless (approved):
Seems fine. Does this imply past-me made two functions instead of reusing one?

pkg/packagekit/assets/distribution.dist
```diff
@@ -3,7 +3,7 @@
 	<title>{{.Title}}</title>
 	<pkg-ref id="com.{{.Identifier}}.launcher">
 		<bundle-version>
-			<bundle CFBundleShortVersionString="{{.Version}}" CFBundleVersion="{{.Version}}" id="com.{{.Identifier}}.launcher" path="usr/local/kolide/Kolide.app"/>
+			<bundle CFBundleShortVersionString="{{.Version}}" CFBundleVersion="{{.Version}}" id="com.{{.Identifier}}.launcher" path="usr/local/{{.Identifier}}/Kolide.app"/>
```
RebeccaMahany, line 6: This does not appear to actually affect the output (it gets replaced anyway), but I thought it was better to not have incorrect data here in the first place

directionless (approved):

Issue #943, opened 2022-11-18
On fresh install of windows, with launcher desktop enabled, a cmd prompt stays open


James-Pickett:
On fresh install of windows with launcher desktop enabled, a cmd prompt stays open. This seems to go away after the first auto update.

At the time of writing this, launcher desktop is only enabled for Kolide employees.

RebeccaMahany:
With `windowsgui` set, `launcher.exe version` won't print out its version anymore, which blocks updates because launcher validates new binaries by running the version subcommand and checking output.

Some options for adjusting behavior: https://forum.golangbridge.org/t/no-println-output-with-go-build-ldflags-h-windowsgui/7633/7

PR #982, opened 2022-12-23
Ensure Windows launcher prints version to console on `launcher.exe version`

RebeccaMahany:
Reverts change related to https://github.com/kolide/launcher/issues/943

See https://forum.golangbridge.org/t/no-println-output-with-go-build-ldflags-h-windowsgui/7633/7

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -79,23 +79,23 @@ jobs:
     - name: Test
       run: make test
 
-    # Launcher should always print its version info when called with `version`, which includes the golang version;
-    # we confirm that it contains that info here.
+    # Launcher should always print its version info when called with `version` -- this is a quick
+    # check to ensure that launcher can update to this build.
     - name: Test build - macOS and ubuntu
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -79,6 +79,26 @@ jobs:
     - name: Test
       run: make test
 
+    # Launcher should always print its version info when called with `version`, which includes the golang version;
+    # we confirm that it contains that info here.
+    - name: Test build - macOS and ubuntu
+      if: ${{ contains(matrix.os, 'macos') || contains(matrix.os, 'ubuntu') }}
+      run: |
+        if [[ $(./build/launcher version) != *go${{env.GOVER}}* ]]; then
```
directionless, line 87: Could check for the string `launcher - version` which is probably what update does. But I'd approve this

directionless (dismissed):

.github/workflows/go.yml
RebeccaMahany: I like that better, it's more readable without the environment variable in there 🙂 updated

directionless (approved):

PR #984, opened 2022-12-30
Control server phase 3: Desktop subsystem implementation

seejdev:
This is still WIP... I'll update this with more context.

Still TODO:
- Tests
- Icons
- More validation, error handling, edge cases
- Docs

### DIFF
--- a/cmd/launcher/desktop.go
+++ b/cmd/launcher/desktop.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"os"
@@ -16,6 +17,7 @@ import (
 	"github.com/kolide/kit/logutil"
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/desktop/menu"
+	"github.com/kolide/launcher/ee/desktop/runner"
 	"github.com/kolide/launcher/ee/desktop/server"
 	"github.com/oklog/run"
 	"github.com/peterbourgon/ff/v3"
@@ -39,6 +41,11 @@ func runDesktop(args []string) error {
 			"socket_path",
 			"",
 			"path to create socket",
+		)
+		flstatuspath = flagset.String(
+			"status_path",
+			"",
+			"path to read status data",
 		)
 		fldebug = flagset.Bool(
 			"debug",
@@ -73,6 +80,14 @@ func runDesktop(args []string) error {
 		)
 	}
 
+	if *flstatuspath == "" {
+		*flstatuspath = defaultStatusPath()
+		level.Info(logger).Log(
+			"msg", "using default status path since none was provided",
+			"status_path", *flstatuspath,
+		)
+	}
+
 	var runGroup run.Group
 
 	// listen for signals
@@ -92,6 +107,15 @@ func runDesktop(args []string) error {
 	if err != nil {
 		return err
 	}
+
+	menu := menu.New(logger, *flhostname)
+	buildMenu := func() {
+		menu.Build(getMenuData(logger, *flstatuspath))
+	}
+
+	server.RegisterRefreshListener(func() {
+		buildMenu()
+	})
 
 	// start desktop server
 	runGroup.Add(server.Serve, func(err error) {
@@ -125,7 +149,7 @@ func runDesktop(args []string) error {
 	}()
 
 	// blocks until shutdown called
-	menu.Init(*flhostname)
+	menu.Init(buildMenu)
 
 	return nil
 }
@@ -179,4 +203,25 @@ func defaultSocketPath() string {
 	}
 
 	return filepath.Join(os.TempDir(), fmt.Sprintf("%s_%d", socketBaseName, os.Getpid()))
+}
+
+func defaultStatusPath() string {
+	const statusBaseName = "desktop_status.json"
+
+	return filepath.Join(os.TempDir(), fmt.Sprintf("%s_%d", statusBaseName, os.Getpid()))
+}
+
+// Loads menu data from the desktop status file
+func getMenuData(logger log.Logger, statusPath string) *menu.MenuData {
+	statusFileBytes, err := os.ReadFile(statusPath)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to read desktop status file", "path", statusPath)
+	}
+
+	var desktopStatus runner.DesktopUserStatus
+	if err := json.Unmarshal(statusFileBytes, &desktopStatus); err != nil {
+		level.Error(logger).Log("msg", "failed to unmarshal desktop status json")
+	}
+
+	return &desktopStatus.Menu
 }
--- a/ee/desktop/menu/menu.go
+++ b/ee/desktop/menu/menu.go
@@ -1,36 +1,150 @@
 package menu
 
 import (
-	"fmt"
+	"os/exec"
+	"runtime"
 
 	"fyne.io/systray"
-	"github.com/kolide/kit/version"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/launcher/ee/desktop/assets"
 )
 
-func Init(hostname string) {
-	onReady := func() {
-		systray.SetTemplateIcon(assets.KolideDesktopIcon, assets.KolideDesktopIcon)
-		systray.SetTooltip("Kolide")
-
-		versionItem := systray.AddMenuItem(fmt.Sprintf("Version %s", version.Version().Version), "")
-		versionItem.Disable()
-
-		// if prod environment, return
-		if hostname == "k2device-preprod.kolide.com" || hostname == "k2device.kolide.com" {
-			return
-		}
-
-		// in non prod environment
-		systray.SetTemplateIcon(assets.KolideDebugDesktopIcon, assets.KolideDebugDesktopIcon)
-		systray.AddSeparator()
-		systray.AddMenuItem("--- DEBUG ---", "").Disable()
-		systray.AddMenuItem(fmt.Sprintf("Hostname: %s", hostname), "").Disable()
-	}
-
-	systray.Run(onReady, func() {})
+// MenuData encapsulates a systray icon and accessible menu items
+type MenuData struct {
+	Tooltip string         `json:"tooltip,omitempty"`
+	Items   []MenuItemData `json:"items"`
 }
 
-func Shutdown() {
+// MenuItemData represents a menu item, optionally containing sub menu items
+type MenuItemData struct {
+	Label       string         `json:"label,omitempty"`
+	Tooltip     string         `json:"tooltip,omitempty"`
+	Disabled    bool           `json:"disabled,omitempty"`
+	NonProdOnly bool           `json:"nonProdOnly,omitempty"`
+	IsSeparator bool           `json:"isSeparator,omitempty"`
+	Action      MenuItemAction `json:"action,omitempty"`
+	Items       []MenuItemData `json:"items,omitempty"`
+}
+
+// MenuItemAction encapsulates what action should be performed when a menu item is invoked
+type MenuItemAction struct {
+	URL string `json:"url,omitempty"`
+}
+
+// MenuBuilder is an interface a menu parser can use to specify how the systray menu is built
+type MenuBuilder interface {
+	SetTooltip(tooltip string)
+	AddMenuItem(label, tooltip string, disabled, nonProdOnly bool, action MenuItemAction, parent any) any
+	AddSeparator()
+}
+
+type menu struct {
+	logger   log.Logger
+	hostname string
+}
+
+func New(logger log.Logger, hostname string) *menu {
+	m := &menu{
+		logger:   logger,
+		hostname: hostname,
+	}
+
+	return m
+}
+
+func (m *menu) Init(buildMenu func()) {
+	systray.Run(buildMenu, nil)
+}
+
+func (m *menu) Build(menu *MenuData) {
+	// Remove all menu items each time we rebuild the menu
+	systray.ResetMenu()
+
+	parseMenuData(menu, m)
+}
+
+func (m *menu) SetTooltip(tooltip string) {
+	systray.SetTooltip(tooltip)
+
+	if m.isProd() {
+		systray.SetTemplateIcon(assets.KolideDesktopIcon, assets.KolideDesktopIcon)
+	} else {
+		systray.SetTemplateIcon(assets.KolideDebugDesktopIcon, assets.KolideDebugDesktopIcon)
+	}
+}
+
+func (m *menu) AddMenuItem(label, tooltip string, disabled, nonProdOnly bool, action MenuItemAction, parent any) any {
+	if nonProdOnly && m.isProd() {
+		// This is prod environment, but the menu item is for non-prod only
+		return nil
+	}
+
+	var item, parentItem *systray.MenuItem
+	parentItem, ok := parent.(*systray.MenuItem)
+	if ok {
+		// If a parent menu item was provided, this is meant to be a sub menu item
+		item = parentItem.AddSubMenuItem(label, tooltip)
+	} else {
+		item = systray.AddMenuItem(label, tooltip)
+	}
+
+	if disabled {
+		item.Disable()
+	}
+
+	// Menu items can have actions associated with them
+	m.addAction(item, action)
+
+	return item
+}
+
+func (m *menu) AddSeparator() {
+	systray.AddSeparator()
+}
+
+func (m *menu) Shutdown() {
 	systray.Quit()
+}
+
+func (m *menu) isProd() bool {
+	return m.hostname == "k2device-preprod.kolide.com" || m.hostname == "k2device.kolide.com"
+}
+
+func (m *menu) addAction(item *systray.MenuItem, action MenuItemAction) {
+	if action.URL == "" {
+		return
+	}
+
+	go func() {
+		for {
+			select {
+			case <-item.ClickedCh:
+				err := open(action.URL)
+				if err != nil {
+					level.Error(m.logger).Log("msg", "failed to open URL in browser", "err", err)
+				}
+			}
+		}
+	}()
+}
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	var cmd string
+	var args []string
+
+	switch runtime.GOOS {
+	case "windows":
+		cmd = "cmd"
+		args = []string{"/c", "start"}
+	case "darwin":
+		cmd = "open"
+	default: // "linux", "freebsd", "openbsd", "netbsd"
+		cmd = "xdg-open"
+	}
+	args = append(args, url)
+	return exec.Command(cmd, args...).Start()
 }
--- a/ee/desktop/menu/menu_parse.go
+++ b/ee/desktop/menu/menu_parse.go
@@ -1,0 +1,33 @@
+package menu
+
+func parseMenuData(m *MenuData, builder MenuBuilder) {
+	// Set top-level systray properties
+	builder.SetTooltip(m.Tooltip)
+
+	for _, child := range m.Items {
+		parseMenuItem(&child, builder, nil)
+	}
+}
+
+func parseMenuItem(m *MenuItemData, builder MenuBuilder, parent any) {
+	if m == nil {
+		return
+	}
+
+	if m.IsSeparator {
+		// If the item is a separator, nothing else matters
+		builder.AddSeparator()
+		return
+	}
+
+	var item any
+	if m.Label != "" {
+		// A menu item must have a non-empty label
+		item = builder.AddMenuItem(m.Label, m.Tooltip, m.Disabled, m.NonProdOnly, m.Action, parent)
+	}
+
+	for _, child := range m.Items {
+		// Recursively parse sub menu items
+		parseMenuItem(&child, builder, item)
+	}
+}
--- a/ee/desktop/runner/runner.go
+++ b/ee/desktop/runner/runner.go
@@ -3,6 +3,7 @@ package runner
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -20,6 +21,7 @@ import (
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/consoleuser"
 	"github.com/kolide/launcher/ee/desktop/client"
+	"github.com/kolide/launcher/ee/desktop/menu"
 	"github.com/kolide/launcher/pkg/backoff"
 	"github.com/shirou/gopsutil/process"
 )
@@ -104,9 +106,12 @@ type DesktopUsersProcessesRunner struct {
 	// usersFilesRoot is the launcher root dir with will be the parent dir
 	// for kolide desktop files on a per user basis
 	usersFilesRoot string
+	// processSpawningEnabled controls whether or not desktop user processes are automatically spawned
+	// This effectively represents whether or not the launcher desktop GUI is enabled or not
+	processSpawningEnabled bool
 }
 
-// processRecord is used to track spawned desktop proccesses.
+// processRecord is used to track spawned desktop processes.
 // The path is used to ensure another process has not taken the same pid.
 // The existence of a process record does not mean the process is running.
 // If, for example, a user logs out, the process record will remain until the
@@ -117,16 +122,23 @@ type processRecord struct {
 	socketPath string
 }
 
+// desktopUserStatus is all the device data sent to the desktop user process
+type DesktopUserStatus struct {
+	Enabled bool          `json:"enabled"`
+	Menu    menu.MenuData `json:"menu"`
+}
+
 // New creates and returns a new DesktopUsersProcessesRunner runner and initializes all required fields
 func New(opts ...desktopUsersProcessesRunnerOption) *DesktopUsersProcessesRunner {
 	runner := &DesktopUsersProcessesRunner{
-		logger:           log.NewNopLogger(),
-		interrupt:        make(chan struct{}),
-		uidProcs:         make(map[string]processRecord),
-		updateInterval:   time.Second * 5,
-		procsWg:          &sync.WaitGroup{},
-		interruptTimeout: time.Second * 10,
-		usersFilesRoot:   filepath.Join(os.TempDir(), "kolide-desktop"),
+		logger:                 log.NewNopLogger(),
+		interrupt:              make(chan struct{}),
+		uidProcs:               make(map[string]processRecord),
+		updateInterval:         time.Second * 5,
+		procsWg:                &sync.WaitGroup{},
+		interruptTimeout:       time.Second * 10,
+		usersFilesRoot:         filepath.Join(os.TempDir(), "kolide-desktop"),
+		processSpawningEnabled: true,
 	}
 
 	for _, opt := range opts {
@@ -145,15 +157,16 @@ func (r *DesktopUsersProcessesRunner) Execute() error {
 		}
 	}
 
-	f()
-
 	ticker := time.NewTicker(r.updateInterval)
 	defer ticker.Stop()
 
 	for {
+		// Check immediately on each iteration, avoiding the initial ticker delay
+		f()
+
 		select {
 		case <-ticker.C:
-			f()
+			continue
 		case <-r.interrupt:
 			level.Debug(r.logger).Log("msg", "interrupt received, exiting desktop execute loop")
 			return nil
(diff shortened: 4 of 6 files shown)


ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,150 @@
...
+func (m *menu) isProd() bool {
+	return m.hostname == "k2device-preprod.kolide.com" || m.hostname == "k2device.kolide.com"
+}
+
+func (m *menu) addAction(item *systray.MenuItem, action MenuItemAction) {
+	if action.URL == "" {
+		return
+	}
+
+	go func() {
```
seejdev, line 120: I think these goroutines will leak - I'm thinking I have to make done channels which can be signaled when the either the menu items go away, or the systray is exited.

James-Pickett: done channel makes sense, could signal when you reset he menu

ee/desktop/runner/runner.go
```diff
@@ -145,15 +157,16 @@ func (r *DesktopUsersProcessesRunner) Execute() error {
...
 	}
 
-	f()
-
 	ticker := time.NewTicker(r.updateInterval)
 	defer ticker.Stop()
 
 	for {
+		// Check immediately on each iteration, avoiding the initial ticker delay
+		f()
```
James-Pickett, line 165: this pattern is better

is there any reason to keep the function as a variable?

ee/desktop/runner/runner.go
```diff
@@ -212,40 +233,92 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
+				"err", err,
+			)
+		}
+	}
+
+	return nil
+}
+
+// writeStatusFile copies desktop status data to a shared file for user processes to access
+func (r *DesktopUsersProcessesRunner) writeStatusFile(data io.Reader) error {
```
James-Pickett, line 300: Why write to a file instead of just sending a POST with the data to the server?

James-Pickett (commented):
I like the direction the menu is going with the ability for k2 to send info needed to build the menu.

ee/desktop/runner/runner.go
seejdev: My main motivation was to be able to more easily handle status updates when desktop is disabled, or the user processes are otherwise unavailable.

Particularly, when desktop is disabled and a new update makes it enabled, if the only way to send that data is via the client-server, we have to hold on to that update and wait until the process is spawned and the server becomes available. I could have added code to trigger this quickly and not wait for the ticker, and keep the update in-memory until the server is listening, but it would have made things more complex and introduced new edge cases.

By writing it out to a file, we're able to quickly consume the update, and the existing user process lifecycle management code stays largely the same, other than slurping up that file when spawned and when told to refresh.

I'm still writing tests, but I _think_ this will make it easier to test. Also I'm thinking a non-engineer could easily mock desktop menus by modifying the JSON file on disk and refreshing.

James-Pickett: Thanks for the explanation! That makes sense to me.

ee/desktop/menu/menu.go
seejdev: Added channels and a cleanup method, verified those goroutines are going away now!

ee/desktop/runner/runner.go
seejdev: Nope - simplified this and got rid of the var.

cmd/launcher/desktop.go
```diff
@@ -73,6 +80,14 @@ func runDesktop(args []string) error {
 		)
 	}
 
+	if *flstatuspath == "" {
+		*flstatuspath = defaultStatusPath()
```
directionless, line 84: Why guess? just make it a required option

cmd/launcher/desktop.go
```diff
@@ -180,3 +204,24 @@ func defaultSocketPath() string {
...
+
+	return filepath.Join(os.TempDir(), fmt.Sprintf("%s_%d", statusBaseName, os.Getpid()))
+}
+
+// Loads menu data from the desktop status file
+func getMenuData(logger log.Logger, statusPath string) *menu.MenuData {
+	statusFileBytes, err := os.ReadFile(statusPath)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to read desktop status file", "path", statusPath)
+	}
```
directionless, line 219:
```suggestion
		level.Error(logger).Log("msg", "failed to read desktop status file", "path", statusPath)
		return
	}
```

cmd/launcher/desktop.go
```diff
@@ -180,3 +204,24 @@ func defaultSocketPath() string {
...
+
+	var desktopStatus runner.DesktopUserStatus
+	if err := json.Unmarshal(statusFileBytes, &desktopStatus); err != nil {
+		level.Error(logger).Log("msg", "failed to unmarshal desktop status json")
+	}
```
directionless, line 224:
```suggestion
		level.Error(logger).Log("msg", "failed to unmarshal desktop status json")
		return
	}
```

cmd/launcher/desktop.go
directionless: Oh, this return is wrong, but something like that...

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
-		versionItem.Disable()
+// MenuItemData represents a menu item, optionally containing sub menu items
+type MenuItemData struct {
+	Label       string         `json:"label,omitempty"`
+	Tooltip     string         `json:"tooltip,omitempty"`
+	Disabled    bool           `json:"disabled,omitempty"`
+	NonProdOnly bool           `json:"nonProdOnly,omitempty"`
+	IsSeparator bool           `json:"isSeparator,omitempty"`
+	Action      MenuItemAction `json:"action,omitempty"`
+	Items       []MenuItemData `json:"items,omitempty"`
```
directionless, line 28: I especially like the recursive nature

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+}
 
-		// if prod environment, return
-		if hostname == "k2device-preprod.kolide.com" || hostname == "k2device.kolide.com" {
-			return
-		}
+// MenuItemAction encapsulates what action should be performed when a menu item is invoked
+type MenuItemAction struct {
```
directionless, line 32: Should this be a struct? Will we support things other than URLs? Is it just a type alias for string? (And if it's only going to be URLs, we should just call it URL, and hang it off `MenuItemData`)

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+
+func (m *menu) Init(buildMenu func()) {
+	// buildMenu will be invoked after the systray has been initialized
+	// Before the systray exits, cleanup the goroutines
+	systray.Run(buildMenu, m.cleanup)
+}
+
+func (m *menu) Build(menu *MenuData) {
+	// Remove all menu items each time we rebuild the menu
+	systray.ResetMenu()
```
directionless, line 66: This might be a systray question, but does this cause the menu to blink through empty before being repopluated?

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+
+	// Even though the menu items have been removed, we still have goroutines hanging around
+	m.cleanup()
```
directionless, line 69: cleanup is _before_ parse?

ee/desktop/menu/menu_parse.go
```diff
@@ -0,0 +1,33 @@
...
+
+	var item any
+	if m.Label != "" {
+		// A menu item must have a non-empty label
+		item = builder.AddMenuItem(m.Label, m.Tooltip, m.Disabled, m.NonProdOnly, m.Action, parent)
+	}
+
+	for _, child := range m.Items {
+		// Recursively parse sub menu items
+		parseMenuItem(&child, builder, item)
```
seejdev, line 31: I'm a little weary of using recursion here, but it does the job. I can't imagine we would ever have more than 1 or 2 levels of sub menus, maybe it's worth unwinding this or putting a small limit on the recursion depth.

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+
+	parseMenuData(menu, m)
+}
+
+func (m *menu) SetTooltip(tooltip string) {
+	systray.SetTooltip(tooltip)
+
+	if m.isProd() {
+		systray.SetTemplateIcon(assets.KolideDesktopIcon, assets.KolideDesktopIcon)
```
directionless, line 78: I might just not get it yet, but why are we setting the icon in a function named `SetTooltip`?

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+type MenuBuilder interface {
+	SetTooltip(tooltip string)
+	AddMenuItem(label, tooltip string, disabled, nonProdOnly bool, action MenuItemAction, parent any) any
+	AddSeparator()
+}
+
+type menu struct {
+	logger    log.Logger
+	hostname  string
+	doneChans []chan bool
```
James-Pickett, line 46: absolutely not important, but you can use a `chan struct{}` if you just signalling and don't want to send any data over channel it consumes no memory

ee/desktop/menu/menu.go
directionless: If it's meant to encasulate an action, an interface might be a better choice. Not sure that fits with systray though

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,174 @@
...
+	} else {
 		systray.SetTemplateIcon(assets.KolideDebugDesktopIcon, assets.KolideDebugDesktopIcon)
-		systray.AddSeparator()
-		systray.AddMenuItem("--- DEBUG ---", "").Disable()
-		systray.AddMenuItem(fmt.Sprintf("Hostname: %s", hostname), "").Disable()
 	}
+}
 
-	systray.Run(onReady, func() {})
+func (m *menu) AddMenuItem(label, tooltip string, disabled, nonProdOnly bool, action MenuItemAction, parent any) any {
```
directionless, line 84: I'm confused about what calls this, and why it has the signature it has

ee/desktop/menu/menu.go
James-Pickett, line 46: you can get more specific if you want and make to so menu can only ever send on the channel with `doneChans []chan<- bool` ... leveraging the compiler to make sure you never accidentally receive form the menu struct

directionless (commented):
I think I'm missing something in the overall structure of how this will work. Maybe we can go over it monday.

Some bike shedding about names -- there's a lot of `status` here, but it feels like the over all menu structure.

ee/desktop/menu/menu.go
seejdev: Yeah good point!

ee/desktop/menu/menu.go
seejdev: I'm envisioning there could be different types of actions - for now the only one I know we'll need is "open a browser to a URL". So I imagine this struct would become bigger, and allow for any one of these actions & their associated data to be encapsulated in it.

Some possible actions could be "Open a window", Debug/NonProd operations like "Restart the process", or "Refresh the menu", etc. maybe there's others folks have in mind I'm not aware of.

ee/desktop/menu/menu.go
seejdev: If you have the menu engaged and visible it will quickly flicker. If the menu isn't visible, there's no visual change.

ee/desktop/menu/menu.go
seejdev: Unless this is the very first time the menu is being built, at this point all the menu items will have been removed from the menu, but their action goroutines are still waiting on the clicked channel. Calling cleanup here allows the goroutines to die as we proceed to parse the new menu data and build a brand new menu.

ee/desktop/menu/menu.go
seejdev: I should have put a TODO here - I've glossed over the icon stuff so far. I intend to move the icon setting to a dedicated function and defined by K2.

ee/desktop/menu/menu.go
seejdev: This could definitely use some better docs and function comments.

My intention was to use `MenuBuilder` interface as an abstraction over the underlying systray implementation. Ideally the menu data schema can be parsed independently of whatever underlying framework we use to actually build the menu. The parsing functions use the `MenuBuilder` functions like this one to instruct how the menu should be built without having a dependency on the fyne systray. If we ever swap out systray for say, the v2 one, my hope is that only the menu builder code is changed, and the parsing stays agnostic to the underlying menu dependency.

A non-obvious part of this signature is that in order to support sub menus, we need a reference to the parent menu so we can add menu items to it. This function always returns an `any` type reference to the menu item it created - the parsing code uses this when constructing sub menus by passing it back to the builder, and that's how the builder knows to add the menu item as a child menu item of that parent, instead of the top-level menu item (which doesn't require any menu item reference).

seejdev:
> I think I'm missing something in the overall structure of how this will work. Maybe we can go over it monday.
> 
> Some bike shedding about names -- there's a lot of `status` here, but it feels like the over all menu structure.

I agree the `status` name is overused, I'm cool with finding some better terminology.

ee/desktop/menu/menu_parse.go
directionless: I think recursion is correct. But it would also be reasonable to add a depth check.

ee/desktop/menu/menu.go
directionless: If it's many possible actions it feels better represented by an interface. Like there's a `urlAction`, `windowAction`, etc,  and they each conform to a:
```
type menuItemAction interface {
  Performa(...)
  }
  ```

File: cmd/launcher/desktop.go (hunk 5 above)
Comment (seejdev) [reply cmd/launcher/desktop.go]: This was moved to `menu.getMenuData()` but fixed it there.

File: cmd/launcher/desktop.go (hunk 6 above)
Comment (seejdev) [reply cmd/launcher/desktop.go]: This was moved to `menu.getMenuData()` but fixed it there.

File: ee/desktop/menu/menu.go (hunk 8 above)
Comment (seejdev) [reply ee/desktop/menu/menu.go]: I liked that idea! Now I have a data struct `actionData` which implements the `Action` interface. Definitely feels like a cleaner separation between the action invoker and the action performer code.

File: ee/desktop/menu/menu.go [hunk 15]
```diff
@@ -1,36 +1,186 @@
...
+	// Remove all menu items each time we rebuild the menu
+	systray.ResetMenu()
+
+	// Even though the menu items have been removed, we still have goroutines hanging around
+	m.cleanup()
+
+	// Reparse the menu file & rebuild the menu
+	menuData := m.getMenuData()
+	if menuData == nil {
+		// TODO: Fall back to default menu?
```
seejdev, line 87: What should we do if we can't get the menu from the control server? We can fall back to a simple default menu - maybe just the `Version ...` menu item we have now?

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,196 @@
...
+
+	// Even though the menu items have been removed, we still have goroutines hanging around
+	m.cleanup()
+
+	// Reparse the menu file & rebuild the menu
+	menuData := m.getMenuData()
+	if menuData == nil {
+		var defaultMenuData MenuData
+		menuData = &defaultMenuData
+	}
```
James-Pickett, line 98: Is the intention here just to set menuData to the zero value?

How about:
```suggestion
	if menuData == nil {
		menuData = &MenuData{}
	}
```

ee/desktop/runner/runner.go
```diff
@@ -212,40 +231,91 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
+	var desktopStatus DesktopUserStatus
 	if err := json.NewDecoder(data).Decode(&desktopStatus); err != nil {
 		return fmt.Errorf("failed to decode desktop user control data: %w", err)
 	}
 
-	if desktopStatus.Status == "" {
-		return fmt.Errorf("empty desktop status")
+	// Regardless, we will write the menu data out to a file that can be grabbed by
+	// any desktop user processes, either when they refresh, or when they are spawned.
+	r.writeMenuFile(desktopStatus.Menu)
```
James-Pickett, line 244: this returns an err, should we log it or something?

James-Pickett (commented):
looks good, left a few nits!

ee/desktop/menu/menu.go
```diff
@@ -1,36 +1,196 @@
...
+	parseMenuData(menuData, m)
+}
+
+func (m *menu) getMenuData() *MenuData {
```
RebeccaMahany, line 102: I think it's worth adding some quick documentation here about what writes this file/what's in it/why it's needed

ee/desktop/runner/runner.go
```diff
@@ -212,40 +231,91 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
+		// Next execute loop tick will begin spawning desktop user processes
+		r.processSpawningEnabled = true
+		return nil
+	}
+
+	// If we're here:
+	// A) Desktop is currently disabled, and control server wants it to remain disabled
+	// B) Desktop is currently enabled, and control server wants it to remain enabled
+	//
+	// In case A), there will be no process records and thus no servers to talk to
```
RebeccaMahany, line 271: I think this might read more clearly if we return early in case A, rather than relying on these docs to make it clear why it's fine to proceed?

seejdev: I like that! I can shorten this up.

cmd/launcher/desktop.go
```diff
@@ -73,6 +79,10 @@ func runDesktop(args []string) error {
 		)
 	}
 
+	if *flmenupath == "" {
+		return errors.New("--menu_path must be defined")
```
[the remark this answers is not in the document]
seejdev: Depending on the result of https://github.com/kolide/launcher/pull/984#discussion_r1060146921 maybe we hard-code a static menu definition which is used when either:
- The file doesn't exist
- The file can't be read or is malformed 
- The menu path is unknown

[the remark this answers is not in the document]
RebeccaMahany: Ah I missed that comment -- I think a simple default with `Version...` makes sense!

ee/desktop/menu/menu.go
seejdev: Fixed this, now just falling back to `getDefaultMenu()`

ee/desktop/runner/runner.go
seejdev: Good catch - returning this error now

ee/desktop/runner/runner.go
```diff
@@ -212,40 +231,99 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
 
-	if desktopStatus.Status == "" {
-		return fmt.Errorf("empty desktop status")
+	// Regardless, we will write the menu data out to a file that can be grabbed by
+	// any desktop user processes, either when they refresh, or when they are spawned.
+	if err := r.writeMenuFile(desktopStatus.Menu); err != nil {
+		return err
+	}
+
+	if desktopStatus.Enabled && r.processSpawningEnabled {
```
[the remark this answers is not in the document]
seejdev: `processSpawningEnabled` controls whether or not desktop user processes are created. The process creation doesn't happen until the execute loop ticker fires again - that's why it's a member field of the runner, so it can be persisted until the execute loop fires. I felt like it was simpler to rely on the existing process creation flow rather than interjecting special case process creation in this spot.

However, I am inconsistent and I am killing the desktop processes here rather than waiting for the ticker to fire; I could move the `killDesktopProcesses` call to `runConsoleUserDesktop` to make process creation/killing follow similar flows, where it is done asynchronously according to the execute loop frequency. That would also simply `toggleDesktop` to essentially be a `processSpawningEnabled` setter method only.

ee/desktop/runner/runner.go
```diff
@@ -117,16 +123,23 @@ type processRecord struct {
...
-		interruptTimeout: time.Second * 10,
-		usersFilesRoot:   filepath.Join(os.TempDir(), "kolide-desktop"),
+		logger:                 log.NewNopLogger(),
+		interrupt:              make(chan struct{}),
+		uidProcs:               make(map[string]processRecord),
+		updateInterval:         time.Second * 5,
+		procsWg:                &sync.WaitGroup{},
+		interruptTimeout:       time.Second * 10,
+		usersFilesRoot:         filepath.Join(os.TempDir(), "kolide-desktop"),
+		processSpawningEnabled: true,
```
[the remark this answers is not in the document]
seejdev: `processSpawningEnabled` controls whether or not desktop user processes are created, if the desktop runner is disabled all together (like we have it in prod right now), this doesn't matter since the whole desktop subsystem isn't enabled.

Oh.. I think I know what you mean... when we remove this line https://github.com/kolide/launcher/blob/c844b174dd3bbcf77d36061f6812cef9019dadaf/cmd/launcher/launcher.go#L184 we want `processSpawningEnabled`  to default to `false` (desktop disabled) unless explicitly overriden by the control server. I can make that change, but that would also force anyone not talking to a K2 with control endpoints to not get desktop at all (unless I hardcode in defaulting to `true` if `opts.KolideServerURL == "k2device-preprod.kolide.com" || opts.KolideServerURL == "localhost:3443" `)

Would we want a separate bool to control whether or not the menu is displayed? Not sure if the scenario of having desktop enabled, but menu disabled, is something we care to support. Can't do much with the desktop user process without the menu, so that's why I didn't separate those concepts.

ee/desktop/runner/runner.go
```diff
@@ -212,40 +231,99 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
...
+		// Desktop is currently disabled, control server is asking us to enable
+		level.Debug(r.logger).Log("msg", "runner is enabling desktop users processes")
+
+		// Next execute loop tick will begin spawning desktop user processes
+		r.processSpawningEnabled = true
+	}
+}
+
+// writeMenuFile writes menu data to a shared file for user processes to access
+func (r *DesktopUsersProcessesRunner) writeMenuFile(data menu.MenuData) error {
```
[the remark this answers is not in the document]
seejdev: Actually, good point, I used to have the desktop enabled attribute in the menu file, but now that it's not, it won't get persisted. I'll have to figure out where to put that piece of data so that the runner can ingest it and obey it upon restart.

ee/desktop/menu/menu_action.go
```diff
@@ -0,0 +1,119 @@
...
+		cmd = "/usr/bin/open"
+	default: // "linux", "freebsd", "openbsd", "netbsd"
+		cmd = "xdg-open"
+	}
+	args = append(args, url)
+	return exec.Command(cmd, args...).Start()
+}
+
+// flareCommand invokes the launcher flare executable with the appropriate env vars
+func (m *menu) flareCommand(hostname string) (*exec.Cmd, error) {
```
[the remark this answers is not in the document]
seejdev: No big deal - I was mostly exploring what useful things we could do.

ee/desktop/menu/menu.go
directionless: Yeah, I think we'll ship a simple static menu. I imagine it will say something like "Waiting to connect to Kolide" and have some basic about info

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
directionless: I don't think we need separate control over the desktop process and the visible menu. Today, the desktop process does nothing other than show a menu. And while maybe that changes in the future, everything changes. 

> we want processSpawningEnabled to default to false (desktop disabled) unless explicitly overriden by the control server

Yes, I think that change is correct. With the caveat that the setting should persist across restarts. (I expect the control server to tell clients to enable desktop, and that it is unlikely to ever be disabled)

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
directionless: > I felt like it was simpler to rely on the existing process creation flow rather than interjecting special case process creation in this spot.

That's a good point!

Though I'm still not sure why you need it in this if. If the status is enabled, you can iterate `r.uidProcs`. It should be empty if `r.processSpawningEnabled`. Though maybe that array doesn't actually clean up.

directionless (commented):
This is looking pretty good.

I think the enable/disable isn't persisted, right? We could kick that into another PR if that's easier.

Between this and the notifications PR, I think we need to think harder about assets and tempdirs. But I also think we can merge and revisit.

Something feels a bit weird in the models still -- this is a lot cleaner feeling. But I think the thing-with-file-read probably shouldn't be the same as the systray thing. Maybe? But I also think we can probably roll with this and revisit if we need to. So maybe good enough.

ee/desktop/menu/action_test.go
```diff
@@ -0,0 +1,64 @@
...
+			t.Parallel()
+
+			var a Action
+			err := json.Unmarshal([]byte(tt.data), &a)
+			if tt.expectedErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.action, a)
+			}
```
James-Pickett, line 61: nit
```suggestion
                        if tt.expectedErr {
                          assert.Error(t, err)
                          return
                        }
                        
                        assert.NoError(t, err)
                        assert.Equal(t, tt.action, a)

```

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
seejdev: This is persisted in a `kolide_desktop_flags` bucket now.

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
seejdev: `r.uidProcs` now gets cleared when the runner is interrupted, or told to kill all the user processes (like when desktop is being disabled). This got simpler now with the desktop flags separated.

ee/control/control.go
```diff
@@ -42,6 +42,11 @@ type dataProvider interface {
 	Get(hash string) (data io.Reader, err error)
 }
 
+// StoredDataProvider is an interface for querying data from a persistable data storage layer
+type StoredDataProvider interface {
```
seejdev, line 46: I couldn't think of a better name or a better spot for this - it's meant to be a generic enough interface that any places using `*bbolt.DB` references for simple data queries could be replaced with this. `kit` repo is probably a better place, but I didn't want to expand the changes of an already large PR to start touching another repo.

directionless: Eventually, I'd put this stuff into `pkg/agent/...` but here seems fine for now

cmd/launcher/launcher.go
```diff
@@ -179,33 +179,31 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+	desktopFlagsBucketConsumer := control.NewBucketConsumer(logger, db, "kolide_desktop_flags")
+	controlService.RegisterConsumer("kolide_desktop_flags", desktopFlagsBucketConsumer)
+
+	runner := desktopRunner.New(
+		desktopRunner.WithLogger(logger),
+		desktopRunner.WithUpdateInterval(time.Second*5),
+		desktopRunner.WithHostname(opts.KolideServerURL),
+		desktopRunner.WithAuthToken(ulid.New()),
+		desktopRunner.WithUsersFilesRoot(rootDirectory),
+		desktopRunner.WithProcessSpawningEnabled(opts.KolideServerURL == "k2device-preprod.kolide.com" || opts.KolideServerURL == "localhost:3443"),
```
directionless, line 201: This omits the heroku PRs. Which is fine, we can always followup

ee/control/control.go
directionless: (well, the interface is used by the consumer, so here seems most correct for that. But if we wanted to think about abstractions, I'd think about the pkg/agent space. TBD!)

cmd/launcher/launcher.go
```diff
@@ -179,33 +179,31 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+		return fmt.Errorf("failed to setup control service: %w", err)
 	}
+	runGroup.Add(controlService.Execute, controlService.Interrupt)
+
+	// serverDataBucketConsumer handles server data table updates
+	serverDataBucketConsumer := control.NewBucketConsumer(logger, db, osquery.ServerProvidedDataBucket)
+	controlService.RegisterConsumer("kolide_server_data", serverDataBucketConsumer)
+
+	desktopFlagsBucketConsumer := control.NewBucketConsumer(logger, db, "kolide_desktop_flags")
+	controlService.RegisterConsumer("kolide_desktop_flags", desktopFlagsBucketConsumer)
```
directionless, line 193:
```suggestion
	desktopFlagsBucketConsumer := control.NewBucketConsumer(logger, db, "agent_flags")
	controlService.RegisterConsumer("agent_flags", desktopFlagsBucketConsumer)
```

I think we should make this for all the agent flags.

directionless (approved):
I think we can try!

PR #983, opened 2022-12-29
Enable desktop for linux

RebeccaMahany:
The desktop code for darwin Just Works ™️ for Linux!

#### Ubuntu 22.04, GNOME

Screenshot 2022-12-29 at 10 50 19 AM

#### Ubuntu 22.04, XFCE

Screenshot 2022-12-29 at 11 36 37 AM

#### Ubuntu 22.04, KDE

Screenshot 2022-12-29 at 11 46 52 AM

### Test notes

<details>
<summary>So many details about test setup</summary>

These are my notes for how I tested this on my mac -- saving them because it ended up being a little bit convoluted.

#### Prerequisites

I set up [multipass](https://medium.com/@paulrobu/how-to-run-ubuntu-22-04-vms-on-apple-m1-arm-based-systems-for-free-c8283fb38309) to run Ubuntu 22.04.

I connected to the VM via Microsoft Remote Desktop, which meant my loginctl session wasn't assigned a seat. So I had to comment out the seat check in `ee/consoleuser/consoleuser_linux.go`.

#### Building and installing the launcher package

I updated `pkg/packagekit/package_fpm.go` to use the docker platform `--platform linux/arm64` and image `kong/fpm:latest` for compatibility.

I built launcher: 

```
go run cmd/make/make.go -targets=launcher -linkstamp  --os linux --arch arm64
```

I build package-builder and made a package:

```
make package-builder
./build/package-builder make -debug --hostname=<redacted> --enroll_secret=<redacted> --launcher_version=./build/linux.arm64/launcher --update_channel=nightly --identifier=<redacted> --transport=jsonrpc --targets linux-systemd-deb --package_version 0.13.2-deadbeefcafe-dirty
```

I copied the deb into the VM:

```
multipass transfer /var/whatever/launcher.linux-systemd-deb.deb primary:.
```

I installed the deb in the VM:

```
sudo chown _apt launcher.linux-systemd-deb.deb
sudo apt install ./launcher.linux-systemd-deb.deb
```

My osqueryd installation was also the wrong architecture, so I downloaded `osquery_5.6.0-1.linux_arm64.deb` from the osquery releases, copied it into the VM, installed it systemwide, and copied the install location `/opt/osquery/bin/osqueryd` to the expected location `/usr/local/<identifier>/bin/osqueryd`.

I also had to manually start the service after this: `sudo systemctl start launcher.kolide-k2.service`.

#### Subsequent build updates

[...]

PR #985, opened 2022-12-30
First pass at notifications

RebeccaMahany:
## Description

A first pass at notifications with icons for launcher on Linux, macOS, and Windows, allowing for tracking sent notifications with a TTL to avoid duplicate notifications. Adds a control server consumer to receive notifications.

## How notifications work on each OS

* MacOS: we use the objective-c API to [request authorization to send notifications](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649527-requestauthorizationwithoptions?language=objc) and then to [send them](https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649508-addnotificationrequest). Note: this doesn't work if your app isn't signed, so I haven't been able to test it fully yet.
* Windows: use [go-toast](https://pkg.go.dev/gopkg.in/toast.v1) -- under the hood, this writes a temporary powershell script that loads XML defining the notification, and then sends it. This is also how Fyne does it.
* Linux: send via [dbus's notification service](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html) if available; fall back to sending via [notify-send](https://ss64.com/bash/notify-send.html).

### A note on icons

Both the Windows and Linux implementation options required a filepath to the icon file, so when creating the notifier service, we check to see if the icon exists in the root directory and write out that file if not.

### Lifecycle of a notification

```mermaid
sequenceDiagram
    opt K2
        control server->>+control server: Determines a new notification should be sent,<br>new hash for "desktop_notifier" subsystem
    end
    opt Launcher root process
        control service->>+control server: Request latest data for subsystem "desktop_notifier"<br>since the hash has changed
        control server->>+control service: Return list of notifications
        control service->>+notify consumer: Notify subscriber of updated data
        notify consumer->>+notify consumer: Confirm that notification is valid, has not been sent before
        notify consumer->>+desktop processes runner: Request to send notification
        desktop processes runner->>+desktop server: Send notification
    end
    opt Launcher desktop (user) process
        desktop server->>+notifier service: Send notification to end user
        notifier service->>+desktop server: Return error, or nil if successful
        desktop server->>+desktop processes runner: Return 500 with error if not successful<br>or 200 if successful
    end
    opt Launcher root process
        desktop processes runner->>+notify consumer: Return error, or nil if successful
        notify consumer->>+notify consumer: If successful, store record of sent notification in bucket
    end
```

### Extra

<details>
<summary>Screenshots</summary>

Windows:

windows-notification

Ubuntu, gnome:

Screenshot 2023-01-10 at 3 02 17 PM

Ubuntu, KDE:

Screenshot 2023-01-10 at 4 50 03 PM

Ubuntu, XFCE:

Screenshot 2023-01-10 at 4 58 09 PM

</details>

<details>
<summary>Notes on other options I considered but didn't end up going with</summary>

* [fyne](https://pkg.go.dev/fyne.io/fyne) (both 1.4 and 2.0) -- worked well for macOS, but it would have been non-trivial to get the graphics libraries linking properly for Windows and Linux, and it didn't seem worth it at this time to invest that effort when we aren't using those libraries anyway. If we decide to invest the effort in the future, 2.0 would be preferable over 1.4 because 1.4 is using deprecated macOS APIs.
* [zenity](https://pkg.go.dev/github.com/ncruces/zenity) -- on Windows, notifications displayed with a header `Microsoft.Explorer.Notification.{UUID}` that didn't appear to be configurable and would probably confuse end users; on Linux, it would not display the title.
* A native solution for Windows -- seems like it should be possible but I couldn't figure it out.
* Adding a fallback for Linux that used [kdialog](https://develop.kde.org/deploy/kdialog/#--passivepopup-dialog-box) -- looked correct, except the header for the popup always says `kdialog`, which I didn't think was ideal.
* [osascript display notification](https://developer.apple.com/library/archive/documentation/LanguagesUtilities/Conceptual/MacAutomationScriptingGuide/DisplayNotifications.html) -- not possible to configure the icon, plus when you click the notification it opens up `Script Editor`, which is a confusing end-user experience. (Some docs indicated that running the script from within your bundle should force it to use the bundle's icon, but people noted that this functionality seems broken and it definitely didn't work for me either.) I originally had this as a fallback option for macOS because that's what some other libraries do, but ultimately decided it was not ideal enough that it wasn't worth keeping.

</details>

<details>

Test setup for Linux looks similar to what I documented here: https://github.com/kolide/launcher/pull/983

I tested by creating a [Mockoon](https://mockoon.com/) server to mock out the control server. The server definition looks like this:

```
{
  "uuid": "a05ae8e2-c5bf-4eb0-888d-fceaffa1004b",
  "lastMigration": 24,
  "name": "Control server",
  "endpointPrefix": "",
  "latency": 0,
  "port": 3000,
  "hostname": "0.0.0.0",
  "routes": [
    {
      "uuid": "4223b8d9-3d29-4ec9-ba18-957650cbeff0",
      "documentation": "Get subsystems and hashes",
      "method": "get",
      "endpoint": "api/v1/control/",
      "responses": [
        {
          "uuid": "ad702465-5e32-45df-97fb-271a8e9ff72b",
          "body": "{\n  \"desktop_notifier\": \"{{guid}}\"\n}",
          "latency": 0,
          "statusCode": 200,
          "label": "Get subsystems and hashes",
          "headers": [],
          "bodyType": "INLINE",
          "filePath": "",
          "databucketID": "",
          "sendFileAsBody": false,
          "rules": [],
          "rulesOperator": "OR",
          "disableTemplating": false,
          "fallbackTo404": false,
          "default": false
        }
      ],
      "enabled": true,
      "responseMode": null
    },
    {
      "uuid": "bdca29c0-1cb2-47ee-b677-c98a8b711b91",
      "documentation": "Get content (hardcoded to notifications)",
      "method": "get",
      "endpoint": "api/v1/control/:hash",
      "responses": [
        {
          "uuid": "82bd0a45-5117-4039-8411-0269ac247d7b",
          "body": "[\n  {\n    \"title\": \"Hello {{title}} {{lastName}} -- Kolide here\",\n    \"body\": \"{{lorem}}\",\n    \"id\": \"{{guid}}\",\n    \"valid_until\": \"2026-01-02T15:04:05-0700\"\n  },\n  {\n    \"title\": \"It's Kolide again! How are you, {{title}} {{lastName}}?\",\n    \"body\": \"{{lorem}}\",\n    \"id\": \"{{guid}}\",\n    \"valid_until\": \"2026-01-02T15:04:05-0700\"\n  }\n]",
          "latency": 0,
          "statusCode": 200,
          "label": "",
          "headers": [],
          "bodyType": "INLINE",
          "filePath": "",
          "databucketID": "",
          "sendFileAsBody": false,
          "rules": [],
          "rulesOperator": "OR",
          "disableTemplating": false,
          "fallbackTo404": false,
          "default": true
        }
      ],
      "enabled": true,
      "responseMode": null
    }
  ],
  "proxyMode": false,
  "proxyHost": "",
  "proxyRemovePrefix": false,
  "tlsOptions": {
    "enabled": false,
    "type": "CERT",
    "pfxPath": "",
    "certPath": "",
    "keyPath": "",
    "caPath": "",
    "passphrase": ""
  },
  "cors": true,
  "headers": [
    {
      "key": "Content-Type",
      "value": "application/json"
    }
  ],
  "proxyReqHeaders": [
    {
      "key": "",
      "value": ""
    }
  ],
  "proxyResHeaders": [
    {
      "key": "",
      "value": ""
    }
  ],
  "data": []
}
```

</details>


cmd/launcher/desktop.go
```diff
@@ -180,3 +193,8 @@ func defaultSocketPath() string {
 
 	return filepath.Join(os.TempDir(), fmt.Sprintf("%s_%d", socketBaseName, os.Getpid()))
 }
+
+// TODO pick a more sensible default
+func defaultIconPath() string {
+	return os.TempDir()
+}
```
RebeccaMahany, line 200: Maybe I leave this for now (it works fine) and update to pull from the menu path once https://github.com/kolide/launcher/pull/984/files is merged?

cmd/launcher/desktop.go
```diff
@@ -180,3 +193,8 @@ func defaultSocketPath() string {
 
 	return filepath.Join(os.TempDir(), fmt.Sprintf("%s_%d", socketBaseName, os.Getpid()))
 }
+
+// TODO pick a more sensible default
+func defaultIconPath() string {
+	return os.TempDir()
```
RebeccaMahany, line 199: Maybe I leave this for now (it works fine) and update to pull from the menu path once https://github.com/kolide/launcher/pull/984/files is merged?

cmd/launcher/launcher.go
```diff
@@ -205,6 +205,21 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+			runner,
+			control.WithLogger(logutil.NewServerLogger(true)),
+			control.WithNotificationTtl(time.Second*5),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to set up notifier: %w", err)
+		}
+
+		if controlService != nil {
+			controlService.RegisterConsumer(control.NotificationSubsystem, notificationConsumer)
```
James-Pickett, line 221: we should handle an error if one is returned here

ee/control/notify_consumer_test.go
```diff
@@ -0,0 +1,224 @@
...
+	require.NoError(t, err)
+	testNotificationsData := bytes.NewReader(testNotificationsRaw)
+
+	// Expect that the notifier is called once to send the one notification successfully
+	mockNotifier.On("SendNotification", testTitle, testBody).Return(nil)
+
+	// Call update and assert our expectations about sent notifications
+	err = testNc.Update(testNotificationsData)
+	require.NoError(t, err)
+	mockNotifier.AssertCalled(t, "SendNotification", testTitle, testBody)
```
James-Pickett, line 60: I think this code may only be asserting that mockNotifier was called _at least_ once. I think this will make sure it was only called exactly once:
```suggestion
	mockNotifier.On("SendNotification", testTitle, testBody).Return(nil).Once()

	// Call update and assert our expectations about sent notifications
	err = testNc.Update(testNotificationsData)
	require.NoError(t, err)
	mockNotifier.AssertCalled(t, "SendNotification", testTitle, testBody)
	mockNotifier.AssertExpectations(t)
```

.... and now that I scrolled further down I can see you are validating the number of calls in other tests =). So, you could just keep it consistent and do the same as you did in the other tests and `AssertNumberOfCalls(...)`

ee/desktop/notify/notify.go
```diff
@@ -0,0 +1,52 @@
...
+
+	return notifier
+}
+
+func setIconPath(iconDir string) (string, error) {
+	expectedLocation := filepath.Join(iconDir, assets.KolideIconFilename)
+
+	_, err := os.Stat(expectedLocation)
+
+	if os.IsNotExist(err) {
```
James-Pickett, line 43: I don't this this check is necessary, from the os.WriteFile doc:

> WriteFile writes data to the named file, creating it if necessary.

ee/desktop/notify/notify_darwin.go
```diff
@@ -0,0 +1,43 @@
...
+import "C"
+import (
+	"fmt"
+	"os"
+	"strings"
+	"unsafe"
+)
+
+func (d *DesktopNotifier) SendNotification(title, body string) error {
+	if !isBundle() {
```
James-Pickett, line 21: why does it have to be a bundle, is there entitlements or something needed? mind leaving a comment with a brief explanation?

ee/control/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
+package control
```
James-Pickett, line 1: consider putting the notify portion in its own package, assuming this also the place the desktop menu consumer would go ... maybe?

`/pkg/ee/control/notfiy`
or
`/pkg/ee/control/consumers/notify`

ee/control/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
...
+func (nc *NotificationConsumer) Update(data io.Reader) error {
+	var notificationsToProcess []notification
+	if err := json.NewDecoder(data).Decode(&notificationsToProcess); err != nil {
+		return fmt.Errorf("failed to decode notification data: %w", err)
+	}
+
+	for _, notificationToProcess := range notificationsToProcess {
+		nc.notify(notificationToProcess)
+	}
+
```
James-Pickett, line 92: maybe should consider adding some delete logic, unless it's already there and I missed it ... keep only the last 100 notifications or something? or perhaps delete any that are more than 24 hours old. Perhaps save that for next PR?

James-Pickett (commented):
Nice work! Looks really good.

ee/desktop/notify/notify_darwin.go
RebeccaMahany: Yeah exactly -- it will actually panic if it's not inside a bundle, so it's definitely important to document -- I'll update!

ee/control/notify_consumer_test.go
RebeccaMahany: Good call, I'll update!

ee/control/notify_consumer.go
RebeccaMahany: You didn't miss it -- there's no deletion logic right now. I was thinking we'd probably want to retain the notifications for at least some amount of time, but wasn't sure what was reasonable/what kind of records we'd want to keep about notifications. I imagine we wouldn't want to be pinging the end user more than once or twice a day, so maybe retaining for a week would be fine?

I can save for the next PR while we discuss, since this one is on the larger size anyway!

ee/control/notify_consumer.go
RebeccaMahany: I like both of those options! I'll go with consumers/notify.

(FWIW, the menu consumer lives in `ee/desktop/runner` and the bucket consumer lives in `ee/control`, so I don't think there's a single established location for them yet, unless I'm forgetting something from the phase 3 PR.)

ee/desktop/notify/notify.go
RebeccaMahany: The check is in case we've already written out the icon file to this location before. This maybe made sense before I did the last refactor (because I'd previously put the icon file in the root directory), but if we keep using the temp directory instead (https://github.com/kolide/launcher/pull/985#discussion_r1066399804), then probably it makes sense to remove?

James-Pickett: ohh, I see now ... the check is preventing re-writing the file, so yea, disregard =)

ee/desktop/notify/notify.go
```diff
@@ -0,0 +1,52 @@
+package notify
```
RebeccaMahany, line 1: TODO add documentation in this package or in `docs/architecture` covering:
* Why we chose the options we did per-OS for sending notifications
* Why this has to run in the desktop process

ee/desktop/notify/notify_darwin.m
```diff
@@ -0,0 +1,46 @@
...
+    content.body = body;
+
+    NSString *uuid = [[NSUUID UUID] UUIDString];
+    NSString *identifier = [NSString stringWithFormat:@"kolide-notify-%@", uuid];
+    UNNotificationRequest *request = [UNNotificationRequest requestWithIdentifier:identifier
+        content:content trigger:nil];
+
+    [center addNotificationRequest:request withCompletionHandler:^(NSError * _Nullable error) {
+        if (error != nil) {
+            NSLog(@"Could not send notification: %@", error);
```
directionless, line 22: Will this log go anywhere useful?

ee/desktop/notify/notify_darwin.m
```diff
@@ -0,0 +1,46 @@
...
+    NSString *body = [NSString stringWithUTF8String:cBody];
+
+    UNAuthorizationOptions options = UNAuthorizationOptionAlert;
+    [center requestAuthorizationWithOptions:options
+        completionHandler:^(BOOL granted, NSError *_Nullable error) {
+            if (!granted) {
+                if (error != NULL) {
+                    NSLog(@"Error asking for permission to send notifications %@", error);
+                } else {
+                    NSLog(@"Unable to get permission to send notifications");
```
directionless, line 40: Same! I'm not sure where these logs go.

ee/desktop/notify/notify.go
```diff
@@ -0,0 +1,52 @@
...
+	return notifier
+}
+
+func setIconPath(iconDir string) (string, error) {
+	expectedLocation := filepath.Join(iconDir, assets.KolideIconFilename)
+
+	_, err := os.Stat(expectedLocation)
+
+	if os.IsNotExist(err) {
+		if err := os.WriteFile(expectedLocation, assets.KolideDesktopIcon, 0644); err != nil {
```
directionless, line 44: I forget -- is this running in user or root context? I something about the way we're handling the icons all feels kinda weird

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
...
+// The desktop runner fullfils this interface -- it exists primarily for testing purposes.
+type notifier interface {
+	SendNotification(title, body string) error
+}
+
+// Represents notification received from control server; SentAt is set by this consumer after sending.
+type notification struct {
+	Title  string    `json:"title"`
+	Body   string    `json:"body"`
+	UUID   string    `json:"uuid"`
```
directionless, line 34: This is fine, but consider `id` -- we don't actually need formal uuids, and I'm not sure we want to encourage them. (uuid, ulid, random, whatever...)

directionless (commented):
This looks pretty good, I think the models and organization looks solid. And I really like all the notification libraries. (I know, it's a little sprawling). I especially love the details about which library and why. I'd suggest either dropping that into an [ADR](https://github.com/kolide/launcher/tree/main/docs/architecture), or a comment in `notify.go`

But (of course there's a but) looking at the way the TTL stuff works, I think we can simplify a bunch. The intent, is to allow our server side to generate notifications, which end up in the control data, and are then sent at-most-once to the end user. I do not think we need any kind of retry functionality -- uuids are unique, if we've sent a uuid, it's dead forever. (Well, as long as we feel like caching that we've sent it)

And thinking about _that_ I realize that message have some freshness factor. If we want to send a notification in January, and the client is offline until March, that notification is kinda pointless. (And we should make a new one)

Which leads me to think that the notification data from the server looks like:
```
{ 
  title: "Hi",
  body: "It's a Wednesday!",
  valid_till: 2023-01-11 23:59:59,
  id: randomNonce123,
}
```

And then inside launcher we:
1. check that the notification is still valid to send (uh oh, time skew might be an issue. Oh well?)
2. Send it
3. Record that we've sent `randomNonce123`, and when

And we'd need a separate routine to crawl the bucket and delete IDs that are older than XXX.

I suspect that most K2 messages will have a valid_till period measured in hours-days. And that we should purge the sent records from the bucket on the order of 6-12 months.

ee/desktop/notify/notify.go
RebeccaMahany: You mean, have the launcher non-desktop side write it (to a non-temp location) and pass that as location in a flag to the user side? I like that and had been thinking about doing it that way, but had been waiting for CJ's phase 3 to land since it does something similar-ish with the menu data file

cmd/launcher/launcher.go
```diff
@@ -205,6 +206,23 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		if controlService != nil {
 			controlService.RegisterConsumer("desktop", runner)
 		}
+
+		// Run the notification service
+		notificationConsumer, err := notificationconsumer.NewNotifyConsumer(
+			db,
+			runner,
+			notificationconsumer.WithLogger(logutil.NewServerLogger(true)),
```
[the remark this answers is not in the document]
RebeccaMahany: Oh yup, whoops, I think this is a relic of some the one-off testing I was doing inside `launcher version` originally. I'll update.

ee/desktop/runner/runner.go
```diff
@@ -212,6 +212,29 @@ func (r *DesktopUsersProcessesRunner) Interrupt(interruptError error) {
 	}
 }
 
+func (r *DesktopUsersProcessesRunner) SendNotification(title, body string) error {
```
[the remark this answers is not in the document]
RebeccaMahany: This tied into a larger question I had about whether with this new responsibility we might want to split up the runner a bit, so the runner is only responsible for managing these processes and is separate from the menu consumer. Maybe an update for the next PR. 🙂

ee/desktop/notify/notify_darwin.m
RebeccaMahany: https://github.com/kolide/launcher/pull/985#discussion_r1068153500

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
...
+			return nil
+		}
+
+		var previouslySentNotification notification
+		if err := json.Unmarshal(sentNotificationRaw, &previouslySentNotification); err != nil {
+			return fmt.Errorf("could not unmarshal previously sent notification: %w", err)
+		}
+
+		// Check to see how long ago the previously-sent notification was sent -- if it's within the
+		// configured TTL, then we consider it a duplicate and will not re-send.
```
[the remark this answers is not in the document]
RebeccaMahany: Hm okay, I was thinking about the case where the k2 server wants acknowledgement that a notification was successfully sent, and would re-send that notification if it wasn't. If we think that's overkill, then I can just take it out -- probably better to keep it simple at this point anyway until we have further requirements.

ee/control/consumers/notificationconsumer/notify_consumer_test.go
```diff
@@ -0,0 +1,224 @@
...
+	mockNotifier := newNotifierMock()
+	testNc := &NotificationConsumer{
+		db:              db,
+		runner:          mockNotifier,
+		logger:          log.NewNopLogger(),
+		notificationTtl: time.Hour * 1,
+	}
+
+	// Send one notification that we haven't seen before
+	testTitle := fmt.Sprintf("Test title @ %d - 9779e3b9-75e6-4b59-8ed6-5a2d6d2934ea", time.Now().UnixMicro())
```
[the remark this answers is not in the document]
RebeccaMahany: Sorry, I'm not sure what you want me to change here -- taking out the random UUID?

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, notificationToProcess := range notificationsToProcess {
+		nc.notify(notificationToProcess)
+	}
+
+	return nil
+}
+
+func (nc *NotificationConsumer) notify(notificationToSend notification) {
+	if nc.notificationAlreadySent(notificationToSend) {
+		level.Debug(nc.logger).Log("msg", "received duplicate notification", "title", notificationToSend.Title)
```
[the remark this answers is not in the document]
RebeccaMahany: I think I had this mostly for testing anyway, I'll take it out

ee/desktop/notify/notify.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type Notification struct {
+	Title string `json:"title"`
+	Body  string `json:"body"`
+}
+
+func NewDesktopNotifier(logger log.Logger, iconDir string) *DesktopNotifier {
+	notifier := &DesktopNotifier{
+		logger: logger,
```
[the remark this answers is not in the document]
RebeccaMahany: I'm not sure what else we'd want here, unless we're moving toward including a `component` field with every log?

ee/desktop/notify/notify_darwin.go
```diff
@@ -0,0 +1,45 @@
...
+		return fmt.Errorf("cannot send notification because this application is not bundled")
+	}
+
+	titleCStr := C.CString(title)
+	defer C.free(unsafe.Pointer(titleCStr))
+	bodyCStr := C.CString(body)
+	defer C.free(unsafe.Pointer(bodyCStr))
+
+	C.sendNotification(titleCStr, bodyCStr)
```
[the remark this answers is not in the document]
RebeccaMahany: Yeah let me see what I can figure out here, it's the only one where we aren't passing the error back up

ee/desktop/notify/notify.go
RebeccaMahany: I'm happy to move it wherever -- I don't have a strong preference for where this lives

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -0,0 +1,167 @@
...
+	notificationTtl time.Duration // How long until a notification expires -- i.e. how long until we can re-send that identical notification
+}
+
+// The desktop runner fullfils this interface -- it exists primarily for testing purposes.
+type notifier interface {
+	SendNotification(title, body string) error
+}
+
+// Represents notification received from control server; SentAt is set by this consumer after sending.
+type notification struct {
```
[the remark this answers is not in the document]
RebeccaMahany: Hm, maybe I'll add a comment noting that we will want to update if they need to be per-user?

ee/control/consumers/notificationconsumer/notify_consumer.go
RebeccaMahany: Sure, I'm fine with something instead of UUID -- why don't we want to encourage them, though? Do we just not want to be tied to expecting this string to be formatted a particular way, or is there another concern?

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
RebeccaMahany: Added a comment here

ee/desktop/notify/notify_darwin.go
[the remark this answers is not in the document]
directionless: Error probably doesn't matter? I think all we'd do is log it, and we talked about logs a couple comments up. (I guess maybe we'd not record it as send)

ee/control/consumers/notificationconsumer/notify_consumer.go
directionless: No deep concerns. We just don't use uuids much, so it feels like a weird hint. (No reason for not using uuids, we mostly standardized on ulids, or other IDs)

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
directionless: We might want acknowledgments, but I don't think we'll need resend. Server can just make a new ID

ee/control/consumers/notificationconsumer/notify_consumer_test.go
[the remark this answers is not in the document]
directionless: Yeah, don't hardcode the uuid. We often use `ulid.New()` in the tests.

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
directionless: I might be wrong, but I think:
1. We have one desktop process per user. It handles menu and notifications
2. the runner is responsible for managing the desktop processes
3. the root launcher spawns the runner routine

ee/desktop/notify/notify.go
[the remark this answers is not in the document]
directionless: I like component, but might already be captured by the parent. :shrug:

ee/desktop/notify/notify.go
directionless: Yes, something like that. I'm thinking about two things.

First, I think the desktop process (Which runs as the user) should avoid creating files. It's harder to manage permissions and error cases. There are times it's going to be unavoidable, but when we can avoid it, we should.

Second, we have these long lived fixed assets. It seems reasonable to have launcher drop them into somewhere for consumption by all the pieces.

ee/control/consumers/notificationconsumer/notify_consumer_test.go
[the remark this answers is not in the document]
RebeccaMahany: Done 🙂

ee/desktop/runner/runner.go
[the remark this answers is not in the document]
RebeccaMahany: That's how it currently works, but my point is that the runner is currently also responsible for being the consumer of menu updates from the control server (whereas the notification consumer is responsible for consuming notifications from the control server, and then uses the runner to send them). I'm wondering if it might be a better separation of responsibilities at this point if we created a new, separate menu consumer, and then the menu consumer and the notification consumer both use the runner to send updates to the desktop processes.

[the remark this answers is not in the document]
directionless: I don't know what makes sense. Somewhat it's the same code, running in the same process. So it's mostly a library organization question. (With a side helping of ownership and goroutines). So it feels easy to move it around

I'm happy to go with whatever y'all think makes sense

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
RebeccaMahany: This should be updated as desired now -- we check if we've sent that ID before and check `valid_until`, send, and clean up after 6 months.

ee/control/notify_consumer.go
RebeccaMahany: There's cleanup logic now

ee/desktop/notify/notify_darwin.go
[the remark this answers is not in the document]
RebeccaMahany: Yeah, I think it would be nice to record it as not sent just because that's how it works for the other OSes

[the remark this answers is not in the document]
RebeccaMahany: Updated -- the objective-c passes back a success bool now so we won't record the notification as sent incorrectly

ee/control/consumers/notificationconsumer/notify_consumer_test.go
```diff
@@ -0,0 +1,285 @@
...
+			Body:       "",
+			ID:         ulid.New(),
+			ValidUntil: getValidUntil(),
+		},
+		// Invalid because `ValidUntil` isn't a timestamp
+		{
+			Title:      "Test title 1",
+			Body:       "Test body 1",
+			ID:         ulid.New(),
+			ValidUntil: "some time in the future",
```
directionless, line 92:
```suggestion
			ValidUntil: "not a timestamp",
```

docs/architecture/notifications.md
```diff
@@ -0,0 +1,68 @@
+# Methods chosen for sending notifications to the end user.
```
directionless, line 1: These should be dated files

cmd/launcher/launcher.go
```diff
@@ -205,6 +206,26 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
 			controlService.RegisterConsumer("desktop", runner)
 		}
+
+		// Run the notification service
+		notificationConsumer, err := notificationconsumer.NewNotifyConsumer(
+			db,
+			runner,
+			ctx,
+			notificationconsumer.WithLogger(logger),
+			notificationconsumer.WithNotificationRetentionPeriod(time.Second*5),
```
directionless, line 216: Stale?

directionless (commented):
This looks pretty good. Found a couple nits. 

I'm not around Friday, but if y'all are happy feel free to approve and merge.

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+}
+
+func WithCleanupInterval(cleanupInterval time.Duration) notificationConsumerOption {
+	return func(nc *NotificationConsumer) {
+		nc.cleanupInterval = cleanupInterval
+	}
+}
+
+func NewNotifyConsumer(db *bbolt.DB, runner *desktopRunner.DesktopUsersProcessesRunner, ctx context.Context, opts ...notificationConsumerOption) (*NotificationConsumer, error) {
```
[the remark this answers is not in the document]
RebeccaMahany: No, it needs to be a `desktopRunner.DesktopUsersProcessesRunner` -- the consumer sends notification requests to the desktop user processes (where they're then handled by a `notifier`) via the runner, because the runner owns those processes.

cmd/launcher/desktop.go
RebeccaMahany: Done

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
RebeccaMahany: I've added a diagram to the ADR to hopefully document this a little better -- https://github.com/kolide/launcher/blob/959fa225dcb28b200a0ed763f14526a9c479f2c5/docs/architecture/notifications.md#lifecycle-of-a-notification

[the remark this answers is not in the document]
directionless: Oh, might not have looked hard enough. I thought that's what the `notifier` interface was for?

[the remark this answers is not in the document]
RebeccaMahany: The `notifier` interface exists at this point only so I can mock the desktop runner in notification consumer tests.

I can rename it to something else -- it's named `notifier` and not something more specific mostly by virtue of having refactored this like 4 separate times. 😅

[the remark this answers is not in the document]
directionless: Seems a little weird, but okay. if it's only used in the tests, it could be declared in the tests.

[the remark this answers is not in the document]
RebeccaMahany: Renamed it to `userProcessesRunner` and updated the comment for accuracy

pkg/osquery/extension.go
```diff
@@ -75,6 +75,9 @@ const (
 	// the bucket which we push values into from server-backed tables
 	ServerProvidedDataBucket = "server_provided_data"
 
+	// the bucket where we hold sent notifications
+	SentNotificationsBucket = "sent_notifications"
```
seejdev, line 79: Curious if we intend to expose notification data to osquery, or if this was just a convenient spot to define it?

ee/desktop/notify/notify_darwin.m
James-Pickett: Woops, updating that branch made the links no to to the right parts of code, updated

pkg/osquery/extension.go
directionless: No current plans to expose it. But this is where the bucket name constants live. (Kinda silly really, but :shrug:)

ee/control/consumers/notificationconsumer/notify_consumer_test.go
```diff
@@ -0,0 +1,285 @@
...
+		},
+	}
+	testNotificationsRaw, err := json.Marshal(testNotifications)
+	require.NoError(t, err)
+	testNotificationsData := bytes.NewReader(testNotificationsRaw)
+
+	// Call update and assert our expectations about sent notifications
+	err = testNc.Update(testNotificationsData)
+	require.NoError(t, err)
+	mockNotifier.AssertNumberOfCalls(t, "SendNotification", 0)
```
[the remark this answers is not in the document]
RebeccaMahany: Updated!

ee/desktop/server/server.go
```diff
@@ -24,19 +27,22 @@ type DesktopServer struct {
 	shutdownChan chan<- struct{}
 	authToken    string
 	socketPath   string
+	notifier     *notify.DesktopNotifier
 }
 
-func New(logger log.Logger, authToken string, socketPath string, shutdownChan chan<- struct{}) (*DesktopServer, error) {
+func New(logger log.Logger, authToken string, socketPath string, iconPath string, shutdownChan chan<- struct{}) (*DesktopServer, error) {
```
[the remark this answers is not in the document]
RebeccaMahany: agreed, I'll update in a subsequent PR!

cmd/launcher/launcher.go
```diff
@@ -205,6 +206,25 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+		ctx,
+		notificationconsumer.WithLogger(logger),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to set up notifier: %w", err)
+	}
+	// Runs the cleanup routine for old notification records
+	runGroup.Add(notificationConsumer.Execute, notificationConsumer.Interrupt)
+
+	if controlService != nil {
```
seejdev, line 222: Can remove the if block now that `controlService` is always expected to be non-nil

seejdev (approved):

seejdev:
LGTM - left one nit.

cmd/launcher/launcher.go
RebeccaMahany: Oh yup, thank you!

seejdev (approved):

directionless (approved):

PR #990, opened 2023-01-20
Persist lastFetched control data across launcher restarts

### LINKED ISSUE #979: Improve control server fetching to persist lastFetched data through restarts
https://github.com/kolide/launcher/pull/966 introduces launcher's ability to fetch control server data. The update data is cached in-memory, which avoids re-fetching updates which have already been applied.

If launcher is restarted, the in-memory cache is lost and all update data will be re-fetched and re-applied. Before enabling control server features at scale, we should enable launcher to record and persist update data to disk, and reload them into the in-memory cache after restarts so we can avoid the unnecessary K2 requests involved in re-fetching.

Relates to #935

seejdev:
Fixed #979 

Enables the control service to persist its in-memory cache of control data to disk, so it can be persisted across launcher restarts. The control service will retrieve subsystem hashes from the key/value store if it doesn't have an in-memory cache record for that subsystem.

I also took the opportunity to move the key/value store interfaces to `pkg/agent` and tried to making the naming scheme more succinct.


pkg/agent/keyvalue_store.go
```diff
@@ -0,0 +1,17 @@
...
+	Get(key []byte) (value []byte, err error)
+}
+
+// Storer is an interface for setting data in a persistent key/value store.
+type Storer interface {
+	Set(key, value []byte) error
+}
+
+// RetrieverStorer is an interface that groups the Get and Set methods.
+type RetrieverStorer interface {
```
directionless, line 14: `Getter` and  `Setter` maybe? To match the functions. Not really important though

directionless (dismissed):
Seems okay

pkg/agent/keyvalue_store.go
seejdev: Yeah I've waffled back and forth on a half dozen naming schemes 😂 
Those names are a bit easier, I'll try it

ee/control/bucket_consumer.go
```diff
@@ -28,6 +28,10 @@ func NewBucketConsumer(logger log.Logger, db *bbolt.DB, bucketName string) *buck
 }
 
 func (bc *bucketConsumer) Update(data io.Reader) error {
+	if bc == nil {
```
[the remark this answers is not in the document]
seejdev: Only in testing code. Since some objects can use `bucketConsumer` as a sort of cache, allowing it to be nil gracefully disables the cache, since any cache errors are usually treated as non-fatal.

ee/control/bucket_consumer.go
```diff
@@ -28,6 +28,10 @@ func NewBucketConsumer(logger log.Logger, db *bbolt.DB, bucketName string) *buck
...
+		return fmt.Errorf("bucketConsumer is nil")
```
RebeccaMahany, line 32: Should these (here, line 99, line 119) be errors.New instead since there's no string interpolation happening? Same for the addition to NotificationConsumer.Update

ee/control/control.go
```diff
@@ -119,39 +116,74 @@ func (cs *ControlService) Fetch() error {
...
+func (cs *ControlService) fetchAndUpdate(subsystem, hash string) error {
+	data, err := cs.fetcher.Get(hash)
+	if err != nil {
+		return fmt.Errorf("failed to get control data: %w", err)
+	}
 
-		// Consumer and subscriber(s) notified now
-		err = cs.update(subsystem, data)
+	if data == nil {
+		return fmt.Errorf("control data is nil")
```
RebeccaMahany, line 157: Same errors.New question

RebeccaMahany (dismissed):
Just the one nitpick, LGTM

directionless (dismissed):

ee/control/bucket_consumer.go
seejdev: thanks, fixed!

directionless (approved):

PR #991, opened 2023-01-20
Use dispatch semaphore to signal completion of async request

RebeccaMahany:
Fixes issue where logs would report errors when sending notifications on macOS even though the notifications were sent successfully -- we need to wait for our `completionHandler` to complete and set the success variable before returning.

### DIFF
--- a/ee/desktop/notify/notify_darwin.m
+++ b/ee/desktop/notify/notify_darwin.m
@@ -33,11 +33,11 @@ BOOL doSendNotification(UNUserNotificationCenter *center, NSString *title, NSStr
 
     // Wait for completion handler to complete so that we get a correct value for `success`
     dispatch_time_t timeout = dispatch_time(DISPATCH_TIME_NOW, 30 * NSEC_PER_SEC);
-    dispatch_semaphore_wait(semaphore, timeout);
-
-    // If we potentially timed out, remove the pending request
-    NSArray *requests_to_remove = [NSArray arrayWithObjects:identifier, nil];
-    [center removePendingNotificationRequestsWithIdentifiers:requests_to_remove];
+    intptr_t err = dispatch_semaphore_wait(semaphore, timeout);
+    if (err  != 0) {
+        // Timed out, remove the pending request
+        [center removePendingNotificationRequestsWithIdentifiers:@[identifier]];
+    }
 
     return success;
 }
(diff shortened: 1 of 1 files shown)


ee/desktop/notify/notify_darwin.m
```diff
@@ -18,13 +18,20 @@ BOOL doSendNotification(UNUserNotificationCenter *center, NSString *title, NSStr
...
+            if (error != nil) {
+                NSLog(@"Could not send notification: %@", error);
+                success = NO;
+            }
+            dispatch_semaphore_signal(semaphore);
+        }];
+    });
+
+    // Wait for completion handler to complete so that we get a correct value for `success`
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
seejdev, line 34: Potentially waiting forever scares me, should we use a reasonable timeout? Something like if the notification can't be sent within 30 seconds, we consider it failed and call `removePendingNotificationRequestsWithIdentifiers`?

RebeccaMahany: Yup that's absolutely reasonable, I've updated to 30 seconds per suggestion! We shouldn't need to do anything beyond that, `success` will be `NO` still and the error will get reported back up and handled 🙂

seejdev: The error will be handled, but if the timeout gets hit then we would still have a pending notification in the `UNUserNotificationCenter` queue and it could be delivered after the fact. I think that could allow a weird sequence where each attempt we try to schedule the notification, it doesn't get sent in time, we think it failed, but it did actually show.

RebeccaMahany: Oh, sorry, I misunderstood your initial suggestion! I think I've got it updated appropriately now.

ee/desktop/notify/notify_darwin.m
```diff
@@ -17,14 +17,27 @@ BOOL doSendNotification(UNUserNotificationCenter *center, NSString *title, NSStr
...
-            success = NO;
-        }
-    }];
+    // Wait for completion handler to complete so that we get a correct value for `success`
+    dispatch_time_t timeout = dispatch_time(DISPATCH_TIME_NOW, 30 * NSEC_PER_SEC);
+    dispatch_semaphore_wait(semaphore, timeout);
+
+    // If we potentially timed out, remove the pending request
+    NSArray *requests_to_remove = [NSArray arrayWithObjects:identifier, nil];
+    [center removePendingNotificationRequestsWithIdentifiers:requests_to_remove];
```
seejdev, line 40:
```suggestion
    intptr_t err = dispatch_semaphore_wait(semaphore, timeout);
    if (err  != 0) {
        // Timed out, remove the pending request
        [center removePendingNotificationRequestsWithIdentifiers:@[identifier]];
    }
```

RebeccaMahany: Thanks!!

seejdev (approved):

PR #992, opened 2023-01-20
Improve menu items to allow templated strings

### LINKED ISSUE #988: Improve menu items to allow templated labels
With k2 defining the desktop menu items there will be situations where we want to display a piece of information in a menu item, but k2 will not have that data (or not be able to get it easily).

Things like agent version numbers, local paths, etc. may not be exposed to k2 and we need a way to allow k2 to reference these properties in a templated fashion, where the agent can then supply the remaining data and generate a complete menu item label to be displayed to the user.

We already using templating functionality like this elsewhere in the code base, so this is pretty straightforward. The list of templated properties is less well known but can always be iterated on.

seejdev:
Fixed #988 

This PR introduces functionality to the menu which allows templated strings to be used within menu item text fields. This enables launcher to fill in data that k2 may not be able to get, and further customize the experience to the individual launcher instance.

I've limited the template data to a few simple fields. We can always add more to the list to be used in future menus.
```
type TemplateData struct {
	LauncherVersion  string
	LauncherRevision string
	GoVersion        string
	ServerHostname   string
}
```

The template data is generated by the desktop runner, and applied to either menu updates from the control server, or the default menu. The template expansion is performed on the raw JSON from the control server, so we are able to parse and expand any and all fields, rather than parsing as-needed and per-field. The default menu is created if no there is no existing menu.


cmd/launcher/desktop.go
```diff
@@ -98,7 +99,14 @@ func runDesktop(args []string) error {
 		return err
 	}
 
-	menu := menu.New(logger, *flhostname, *flmenupath)
+	td := menu.NewTemplateData(
+		menu.WithVersion(version.Version()),
+		menu.WithOSQueryVersion("TODO"),
```
seejdev, line 104: I've put TODOs here since I haven't yet explored how difficult it is to get some of these things within launcher desktop process. Can always omit them for this PR and add them in as-needed in the future.

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+	LauncherVersion       string
+	LauncherRevision      string
+	GoVersion             string
+	OSQueryVersion        string
+	Hostname              string
+	LogFilePath           string
+	LauncherFlagsFilePath string
+}
+
+func NewTemplateData(opts ...TemplateDataOption) *templateData {
```
directionless, line 55: I'm not totally sure about the `opts` pattern for this. It feels okay for things that we know during the initial `main` setup, and don't change, but I'm not sure that's true for this.

Though, even things we know at compile time, I might just bake into that `templateData` array, instead of the extra pluming to pass `With...` along.

For things that we don't know... We need to generate them after startup. I did something similar in `localserver`. It has a `Querier`, and a routine that polls it to update the internal data structure. Poke around https://github.com/kolide/launcher/blob/main/ee/localserver/request-id.go#L35

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+	template := &templateData{
+		LauncherVersion:       "unknown",
+		LauncherRevision:      "unknown",
+		OSQueryVersion:        "unknown",
```
directionless, line 59:
```suggestion
		OsqueryVersion:        "unknown",
```

`osquery` is one word.

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+		Hostname:              "k2device.kolide.com",
```
directionless, line 60:
```suggestion
		ServerHostname:              "unknown",
```

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+		LogFilePath:           "/var/kolide-k2/k2device-preprod.kolide.com/debug.json",
+		LauncherFlagsFilePath: "/etc/kolide-k2/launcher.flags",
```
directionless, line 62: I'm not sure this is one we want to expose.

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+		return "", fmt.Errorf("templateData is nil")
+	}
+
+	t, err := template.New("menu_template").Parse(text)
+	if err != nil {
+		return "", fmt.Errorf("could not parse menu template: %w", err)
+	}
+
+	var b strings.Builder
+	err = t.ExecuteTemplate(&b, "menu_template", td)
```
directionless, line 83:
```suggestion
	err = t.Execute(&b, td)
```

I think.

There's another route, which is to use multiple named templates inside the singular template object. But 🤷 I'd say simple is better...

ee/desktop/menu/menu_template.go
directionless: When I say that, it all sounds really complicated. So maybe we just hardcode the things we know at compile time and expand later?

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,89 @@
...
+	LauncherFlagsFilePath string
+}
+
+func NewTemplateData(opts ...TemplateDataOption) *templateData {
+	template := &templateData{
+		LauncherVersion:       "unknown",
+		LauncherRevision:      "unknown",
+		OSQueryVersion:        "unknown",
+		Hostname:              "k2device.kolide.com",
+		LogFilePath:           "/var/kolide-k2/k2device-preprod.kolide.com/debug.json",
```
directionless, line 61: Not sure how I feel about hard coding this. 
```suggestion
		LogFilePath:           "/var/kolide-k2/k2device.kolide.com/debug.json",
```

ee/desktop/menu/menu_template.go
seejdev: Yeah I was on the fence with this pattern too - I have to make a larger change to how template data is fed to desktop, and I think that will make this go away.

seejdev:
Moving this PR back to draft status while I refactor how template data is injected in to desktop.

ee/desktop/menu/menu_template.go
seejdev: This has been simplified and I ripped this stuff out!

ee/desktop/menu/menu_template.go
seejdev: Fixed

ee/desktop/menu/action_open_url.go
```diff
@@ -12,12 +12,21 @@ type actionOpenURL struct {
 	URL string `json:"url"`
 }
 
-func (a actionOpenURL) Perform(m *menu) {
-	if err := open(a.URL); err != nil {
+func (a actionOpenURL) Perform(m *menu, parser textParser) {
```
RebeccaMahany, line 15: Why do we have to pass in the parser if we already have `menu.parser`?

RebeccaMahany (approved):

ee/desktop/menu/action_open_url.go
seejdev: Good catch! This was something left over from my first iteration, removed now!

RebeccaMahany (approved):
LGTM!

directionless (dismissed):
I'm happy to try this, though I think I have 2 questions.

First, I think this is doing the template expansion in the Desktop process, correct? That feels correct, in that it gets the desktop version, but I think it's going to be harder to pass highly dynamic data through. (Which is mostly only in the daemon) OTOH, maybe we'll drop a bunch of variables, and a template file. 🤷 

Second, I think this is doing the templating expansion at each string? Does it make sense to expand once at the initial file read time?

seejdev:
> I'm happy to try this, though I think I have 2 questions.
> 
> First, I think this is doing the template expansion in the Desktop process, correct? That feels correct, in that it gets the desktop version, but I think it's going to be harder to pass highly dynamic data through. (Which is mostly only in the daemon) OTOH, maybe we'll drop a bunch of variables, and a template file. 🤷
> 
> Second, I think this is doing the templating expansion at each string? Does it make sense to expand once at the initial file read time?

Yes, you're correct the template expansion is handled by desktop process. I hadn't thought about doing that in launcher proper, but it could work and may be a better approach. Launcher proper can easily access more data and handle more dynamic data queries/templates, but the user process could be better suited to handle things like i18n and user-centric data.

I like the idea of expanding once - it simplifies some of the design. In order to do that, I'd have to take the JSON from k2 control server, feed the raw JSON bytes to the template parsing, and then unmarshal the expanded output back into JSON. One potential drawback is that any templating error in any field would cause all of the menu data to be treated as invalid, rather than the one field.

But I still kind of like desktop being more "dumb" and not having to worry about template expansion, etc. I'll see how easy it is to move some of this to launcher proper.

directionless:
> I like the idea of expanding once - it simplifies some of the design. In order to do that, I'd have to take the JSON from k2 control server, feed the raw JSON bytes to the template parsing, and then unmarshal the expanded output back into JSON. One potential drawback is that any templating error in any field would cause all of the menu data to be treated as invalid, rather than the one field.

Yes, I think that's right. Template expand before the json unmarshal. Template errors would produce bad content, but then again, so does bad json. So 🤷 server has to be careful either way

directionless:
> 
I like the idea of expanding once - it simplifies some of the design. In order to do that, I'd have to take the JSON from k2 control server, feed the raw JSON bytes to the template parsing, and then unmarshal the expanded output back into JSON. One potential drawback is that any templating error in any field would cause all of the menu data to be treated as invalid, rather than the one field.

Yes, I think that's right. Template expand before the json unmarshal. Template errors would produce bad content, but then again, so does bad json. So 🤷 server has to be careful either way

So I tried going down this path, but it has some wrinkles. My biggest concern was this: performing the template parsing/expansion on raw JSON strings means that the replacement strings in the template data must be JSON-compliant, and any special characters must be escaped. I felt that was a weird restriction to put on the template data, and it made me worried that adding this sort of non-standard manipulation of JSON exposes us to various ways we could produce malformed JSON as a result.

Having launcher desktop to do the template expansion also enables the default/fallback menu to use templates too.

So if we're okay with doing the template expansion in launcher desktop, the second question is should we do it on-demand per-field (how it's working now), or all at once. It probably doesn't make a big difference either way, other than that doing it up front entirely simplifies some design parts where the textParser dependency doesn't need to be passed to as many things.

Of the 4 choices enumerated (launcher vs desktop and single shot vs per-field) I'm good with whatever you chose. 

I suspect doing the parsing as a single shot is going to be both simpler, and more powerful, so I'd bias that way. But, it's pretty easy to iterate.

seejdev:
> > So if we're okay with doing the template expansion in launcher desktop, the second question is should we do it on-demand per-field (how it's working now), or all at once. It probably doesn't make a big difference either way, other than that doing it up front entirely simplifies some design parts where the textParser dependency doesn't need to be passed to as many things.
> 
> Of the 4 choices enumerated (launcher vs desktop and single shot vs per-field) I'm good with whatever you chose.
> 
> I suspect doing the parsing as a single shot is going to be both simpler, and more powerful, so I'd bias that way. But, it's pretty easy to iterate.

I made the change to expand templates up-front and as a single-pass of the entire menu data. I'm happy with this approach, as it reduces complexity in the desktop process, reduces dependencies of the menu itself, and we get the nice property that any templated data added to menus in the future will automatically get parsed with out additional launcher changes (other than adding _new_ templated fields).

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,47 @@
...
+	t, err := template.New("menu_template").Parse(text)
+	if err != nil {
+		return "", fmt.Errorf("could not parse template: %w", err)
+	}
+
+	var b strings.Builder
+	err = t.Execute(&b, tp.td)
+	if err != nil {
+		return "", fmt.Errorf("could not write template output: %w", err)
+	}
```
James-Pickett, line 44:
```suggestion
	if err := t.Execute(&b, tp.td) && err != nil {
		return "", fmt.Errorf("could not write template output: %w", err)
	}
```

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Parse parses text as a template body for the menu template data
+// if an error occurs while parsing, an empty string is returned along with the error
+func (tp *templateParser) Parse(text string) (string, error) {
+	if tp == nil || tp.td == nil {
+		return "", fmt.Errorf("templateData is nil")
+	}
+
+	t, err := template.New("menu_template").Parse(text)
```
James-Pickett, line 35: template dohicky is really cool and from stdlib ❤️

ee/desktop/runner/runner.go
```diff
@@ -313,32 +309,128 @@ func (r *DesktopUsersProcessesRunner) Ping() {
...
+		return fmt.Errorf("failed to decode menu data: %w", err)
+	}
+
+	// Regardless, we will write the menu data out to a file that can be grabbed by
+	// any desktop user processes, either when they refresh, or when they are spawned.
+	if err := r.writeSharedFile(r.menuPath(), menu); err != nil {
+		return err
 	}
 
 	return nil
```
James-Pickett, line 374: I don't think current tests don't make it to this line. Not sure if that should be concerning, perhaps we have an unhandled error some where. I did a bit of testing and found that if I replaced
`WithUsersFilesRoot(launcherRootDir(t)),`
with
`WithUsersFilesRoot(t.TmpDir()),`

in the `TestDesktopUserProcessRunner_Execute` test in runner_test.go

it does make it to this line, guessing it's a permissions issue.

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return "", fmt.Errorf("could not parse template: %w", err)
+	}
+
+	var b strings.Builder
```
directionless, line 40: I'm fine with this. But... Since Execute takes an `io.Writer`, I'd probably move that up to the method call. Then you get to do things like `return t.Execute(w, tp.td)`.  I suspect callers will get simpler too

ee/desktop/runner/runner.go
```diff
@@ -313,32 +309,128 @@ func (r *DesktopUsersProcessesRunner) Ping() {
...
 
-	if err := os.Chmod(menuPath, 0644); err != nil {
+	if err := os.Chmod(path, 0644); err != nil {
 		return fmt.Errorf("os.Chmod: %w", err)
 	}
 
-	defer statusFile.Close()
-	_, err = io.Copy(statusFile, bytes.NewReader(menuBytes))
+	defer file.Close()
+	_, err = io.Copy(file, bytes.NewReader(menuBytes))
```
directionless, line 329: why `io.Copy`, and not `file.Write`

ee/desktop/runner/runner.go
```diff
@@ -313,32 +309,128 @@ func (r *DesktopUsersProcessesRunner) Ping() {
...
 	return nil
 }
 
+// writeDefaultMenuFile will create the menu file, if it does not already exist
+func (r *DesktopUsersProcessesRunner) writeDefaultMenuFile() {
+	menuPath := r.menuPath()
+	_, err := os.Stat(menuPath)
+
+	if os.IsNotExist(err) {
+		defaultMenuJSON := `{
```
directionless, line 383: Slight preference for having this as an embedded asset.

directionless (commented):
Looks really clean. Some nits.

ee/desktop/menu/menu_template.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+type TemplateData struct {
+	LauncherVersion  string `json:",omitempty"`
+	LauncherRevision string `json:",omitempty"`
+	GoVersion        string `json:",omitempty"`
+	ServerHostname   string `json:",omitempty"`
+}
+
+type templateParser struct {
```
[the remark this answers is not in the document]
seejdev: If feels very light at the moment, but yeah I anticipate it to grow a bit more as we explore more templated data and options.

ee/desktop/runner/runner.go
seejdev: 👍 Changes to `file.Write` so I can avoid having to create a reader.

ee/desktop/runner/runner.go
seejdev: Fixed, I matched how you did it in your other PR

ee/desktop/runner/runner.go
seejdev: Good catch! The nested temp dir wasn't getting created so I added a call to `os.MkdirAll` to require it to be created using the `t.TempDir()` and now this line is getting executed in the tests.

ee/desktop/runner/runner.go
directionless: Yeah, I did kinda just do this. 😆

ee/desktop/runner/runner.go
```diff
@@ -324,32 +326,85 @@ func (r *DesktopUsersProcessesRunner) Ping() {
...
 	return nil
 }
 
+// writeDefaultMenuFile will create the menu file, if it does not already exist
+func (r *DesktopUsersProcessesRunner) writeDefaultMenuFile() {
+	menuPath := r.menuPath()
+	_, err := os.Stat(menuPath)
+
+	if os.IsNotExist(err) {
+		if err := r.generateMenuFile(bytes.NewReader(menu.InitialMenu)); err != nil {
```
directionless, line 400: Not totally sure what I think. I'll go with it, but it feels a little tangled using `menu.InitialMenu` here. I think to use that, we need to add in the defaults. Which was the purpose of calls like `newInitialMenuData`.

So this is going to write a partial menu, which then has to get read and have the defaults added in. Is it better to leave this blank and fall through later? I don't know. (And thus feel happy to roll with this)

ee/desktop/runner/runner.go
```diff
@@ -324,32 +326,85 @@ func (r *DesktopUsersProcessesRunner) Ping() {
...
+	parsedMenuDataStr, err := parser.Parse(string(menuDataBytes))
+	if err != nil {
+		return fmt.Errorf("failed to parse menu data: %w", err)
+	}
+
+	// Convert the parsed string back to bytes, which can now be decoded per usual
+	parsedMenuDataBytes := []byte(parsedMenuDataStr)
+
+	var menu menu.MenuData
+	if err := json.NewDecoder(bytes.NewReader(parsedMenuDataBytes)).Decode(&menu); err != nil {
```
directionless, line 381: I'd say lets followup, but I think this is all cleaner if `Parse` works off io readers and writers. There's a lot of extra buffering

directionless (approved):

ee/desktop/runner/runner.go
seejdev: Cool, I'll add this to a follow up PR!

PR #993, opened 2023-01-23
Add ECC Key Support

directionless:


pkg/agent/keys.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		if err := b.Put([]byte(publicEccData), pub); err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	return nil
```
RebeccaMahany, line 88: I think this is an extra return

pkg/agent/keys_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+			return fmt.Errorf("creating key: %w", err)
+		}
+
+		if err := storeKeyData(db, nil, ecdsaToRaw(pub)); err != nil {
+			return fmt.Errorf("storing key: %w", err)
+		}
+
+		k, err := secureenclave.New(*pub)
+		if err != nil {
+			return fmt.Errorf("creating secureenclave signer:, from new key %w", err)
```
RebeccaMahany, line 36: `:,` -- typo or missing another arg we wanted to include in this error message?

James-Pickett (commented):
general shape looks good to me, we'll shake things out after krypto merges

pkg/agent/keys_darwin.go
directionless: discussing deletion below

pkg/osquery/table/launcher_info.go
```diff
@@ -66,6 +74,15 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
 			},
 		}
 
+		// No logger, so just ignore errors. generate the pem encoding if we can.
+		if eccKey := agent.Keys.Public(); eccKey != nil {
+			var pem bytes.Buffer
+			if err := osquery.PublicKeyToPem(eccKey, &pem); err == nil {
```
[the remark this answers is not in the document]
directionless: Don't have a logger here. At least today, I don't think it's worth the refactoring to get one.

pkg/agent/keys_darwin.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if pubData == nil {
+		level.Info(logger).Log("Generating new keys")
+		pub, err := secureenclave.CreateKey()
+		if err != nil {
+			return fmt.Errorf("creating key: %w", err)
+		}
+
+		if err := storeKeyData(db, nil, ecdsaToRaw(pub)); err != nil {
+			return fmt.Errorf("storing key: %w", err)
```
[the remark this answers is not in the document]
James-Pickett: > If this fails, do we need to destroy the key we created? Or maybe it doesn't matter?

If you're referring to the key inside the TPM, nothing is actually stored in there. The data we use to fetch the key is probably just a key encrypted by the primary key in the TPM ... but I'm not 100% sure.

Regarding secure enclave ... I honestly don't know if we need to. I found a delete function, but I had to hunt for it. I know the amount of data is trivial. 

from our favorite apple developer advocate:

> Wouldn't that mean that the secure enclave will somehow have lots of keys that are propably not used anymore?
> 
> Quite possibly. Fortunately, these items aren’t actually stored on the SE. Rather, they are stored on the device itself, but in a way that the item is only available to the SE. And the device has tonnes of space (-:

https://developer.apple.com/forums/thread/129402

pkg/osquery/table/launcher_info.go
```diff
@@ -66,6 +75,22 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
 			},
 		}
 
+		// No logger, so just ignore errors. generate the pem encoding if we can.
+		if eccKey := agent.Keys.Public(); eccKey != nil {
```
James-Pickett, line 79: I would prefer it here if the agent package encapsulated the package level `Keys` var and we accessed `Public()` or `PublicKey()` via a package level func.

pkg/osquery/table/launcher_info.go
```diff
@@ -66,6 +75,22 @@ func generateLauncherInfoTable(db *bbolt.DB) table.GenerateFunc {
...
+			var pem bytes.Buffer
+			if err := osquery.PublicKeyToPem(eccKey, &pem); err == nil {
+				results[0]["signing_key"] = pem.String()
+				results[0]["signing_key_source"] = "TBD"
+			}
+		}
+
+		if localKey := agent.LocalDbKeys.Public(); localKey != nil {
```
James-Pickett, line 87: same here, prefer to key package level var unexported and get data via a func.

pkg/agent/keys.go
```diff
@@ -0,0 +1,105 @@
...
+type keyInt interface {
+	crypto.Signer
+	//Type() string // Not Yet Supported by Krypto
+}
+
+var Keys keyInt = keys.Noop
+var LocalDbKeys keyInt = keys.Noop
+
+func SetupKeys(logger log.Logger, db *bbolt.DB) error {
+	var err error
```
James-Pickett, line 22: we the err var declaration? should be able to initialize on like 25?

pkg/agent/keys.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func storeKeyData(db *bbolt.DB, pri, pub []byte) error {
+	return db.Update(func(tx *bbolt.Tx) error {
+		b, err := tx.CreateBucketIfNotExists([]byte(bucketName))
+		if err != nil {
+			return fmt.Errorf("creating bucket: %w", err)
+		}
+
+		if err := b.Put([]byte(privateEccData), pri); err != nil {
```
James-Pickett, line 78: it's possible pri will be nil here form a secure enclave key... does bbolt accept that with out choking?

pkg/agent/keys.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.etcd.io/bbolt"
+)
+
+type keyInt interface {
+	crypto.Signer
+	//Type() string // Not Yet Supported by Krypto
+}
+
+var Keys keyInt = keys.Noop
+var LocalDbKeys keyInt = keys.Noop
```
James-Pickett, line 19: mentioned in other comment, I would prefer to have these unexported and exposed an expose via a getter.

James-Pickett (commented):
looking good! left a few comments

pkg/agent/keys.go
directionless: Line 25 is assigning directly to the package level `LocalDbKeys`. So it seems cleaner to pre-declare `err`

pkg/osquery/table/launcher_info.go
directionless: I can't decide. 

I instinctively don't like the global nature of the package vars. But they feel vaguely acceptable here?

We'll need to pass them to the challenge/response thing, so I think we need the full Signer interface, not just the exposed `Public()`.  And I wanted to expose both the hardware one, and the DB one. 

I feel like _correct_ would be to create some underlying struct to hold these free floating things that get used across a bunch of launcher. Like, we have `db`, which is created over in main, and then threaded through everything. I sorta imagine the keys similar. Which makes me think about threading an object (or, interface really) through a bunch of places. But then I balk at it.

directionless: I updated to 
```
var hardwareKeys keyInt = keys.Noop
func Keys() keyInt {
	return hardwareKeys
}
```

In some ways, it's the same. But in other ways, it feels a little closer to `agent` being an interface. So maybe?

Anyhow, happy to keep talking about it

pkg/agent/keys/local_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/bbolt"
+)
+
+func TestSetupLocalDbKey(t *testing.T) {
+	t.Parallel()
+
+	db := setupDb(t)
+	logger := log.NewJSONLogger(os.Stderr) //log.NewNopLogger()
```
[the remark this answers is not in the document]
directionless: Indeed. I needed a logger for testing, and forgot :)

pkg/agent/keys.go
directionless: (resolving this, because we only need one place to talk about it)

pkg/osquery/table/launcher_info.go
directionless: (resolving this, because we only need one place to talk about it)

pkg/agent/keys.go
directionless: Good question. I think even ignoring bbolt, I'm not sure what _correct_ is. I added some "only write if not nil" ifs, because I'm not sure we'd want to overwrite the data.

pkg/agent/keys_tpm.go
```diff
@@ -0,0 +1,43 @@
...
+
+		if err := storeKeyData(db, priData, pubData); err != nil {
+			clearKeyData(logger, db)
+			return nil, fmt.Errorf("storing key: %w", err)
+		}
+	}
+
+	k, err := tpm.New(priData, pubData)
+	if err != nil {
+		return nil, fmt.Errorf("creating tpm signer:, from new key %w", err)
```
James-Pickett, line 39:
```suggestion
		return nil, fmt.Errorf("creating tpm signer from new key: %w", err)
```

James-Pickett (dismissed):

RebeccaMahany (approved):

James-Pickett (approved):

PR #996, opened 2023-01-24
Entitlements File

directionless:


RebeccaMahany:
I think we need to copy this into the app bundle in the makefile as well: https://github.com/kolide/launcher/blob/main/Makefile#L61 -- sounds like James had success with it living in `Contents/Resources`?

RebeccaMahany (approved):

James-Pickett (approved):

PR #998, opened 2023-01-25
Update Entitlements

directionless:
The last one didn't work, maybe this one will.

### DIFF
--- a/tools/packaging/entitlements.plist
+++ b/tools/packaging/entitlements.plist
@@ -1,8 +1,6 @@
 <plist version="1.0">
 <dict>
-    <key>keychain-access-groups</key>
-	<array>
-		<string>X98UFR7HA3.com.kolide.agent</string>
-	</array>
+    <key>com.apple.developer.team-identifier</key>
+    <string>X98UFR7HA3</string>
 </dict>
 </plist>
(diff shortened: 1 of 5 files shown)


James-Pickett (approved):

tools/packaging/entitlements.plist
```diff
@@ -1,8 +1,6 @@
 <plist version="1.0">
 <dict>
-    <key>keychain-access-groups</key>
-	<array>
-		<string>X98UFR7HA3.com.kolide.agent</string>
-	</array>
+    <key>com.apple.developer.team-identifier</key>
+    <string>X98UFR7HA3</string>
 </dict>
 </plist>
```
James-Pickett, line 6: I tried to use this entitlement file for the [krypto secure enclave test](https://github.com/kolide/krypto/blob/main/pkg/secureenclave/test_app_resources/entitlements). But got the error:

> The operation couldn’t be completed. (OSStatus error -34018 - failed to generate asymmetric keypair) 

However, this worked:

```xml
<plist version="1.0">
<dict>
    <key>com.apple.application-identifier</key>
    <string>X98UFR7HA3.com.kolide.agent</string>
</dict>
</plist>
```

still approved because it won't hurt and the signing for development is different than signing for release

directionless: I bet you're right about the entitlements, but I'm going to merge and test this first.

PR #1000, opened 2023-01-27
updates local server the handle new challenge protocol with ecc keys

James-Pickett:


ee/localserver/krypto-ec-middleware.go
```diff
@@ -0,0 +1,178 @@
...
+
+		// Extract the box from the URL query parameters
+		boxRaw := r.URL.Query().Get("box")
+		if boxRaw == "" {
+			level.Debug(e.logger).Log("msg", "no data in box query parameter")
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		outerChallenge, err := marshalOuterChallenge(boxRaw)
```
directionless, line 47: I haven't absorbed this, but I'm somewhat surprised? I expected this would be using `challenge.Respond`

James-Pickett: well we have to figure out what the actual command request is, to do that we have to open the box. When we actually get to the wrap function, we can call `challenge.Respond`

directionless (approved):
LGTM

PR #1003, opened 2023-01-27
Run desktop process with launchctl asuser so that notifications work; additionally, some notifications cleanup

RebeccaMahany:
What the title says 🙂 : 
* We run the desktop process with `launchctl asuser` in order to run the desktop process in the user context, where the process will have access to the user bootstrap service and can therefore find the notification service to send notifications. This fixes the issue where notifications would not send when running launcher as a daemon.

As a drive-by:
* Fixes a timeout issue where notifications would send successfully but the runner client would time out before receiving the success
* Also fixes an issue where the notification code was not correctly waiting for the completion handlers to complete, returning failure inappropriately
* Removes troubleshooting code that logged user notification settings; it is no longer needed

____________

Testing notes:

* I tested notifications with a signed app (should send, no error logs) and an unsigned app (should not send, should have error logs)
* I returned data from my mock control server for `kolide_desktop_menu` and confirmed it was reflected in the menu

### DIFF
--- a/ee/desktop/client/client.go
+++ b/ee/desktop/client/client.go
@@ -35,7 +35,7 @@ func New(authToken, socketPath string) client {
 	client := client{
 		base: http.Client{
 			Transport: transport,
-			Timeout:   1 * time.Second,
+			Timeout:   30 * time.Second,
 		},
 	}
 
@@ -69,9 +69,7 @@ func (c *client) Notify(title, body string) error {
 		return fmt.Errorf("could not send notification: %w", err)
 	}
 
-	if resp.Body != nil {
-		resp.Body.Close()
-	}
+	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
 		return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
--- a/ee/desktop/runner/runner_darwin.go
+++ b/ee/desktop/runner/runner_darwin.go
@@ -4,19 +4,43 @@
 package runner
 
 import (
+	"fmt"
 	"os/exec"
+	"os/user"
 )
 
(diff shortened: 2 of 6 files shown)


ee/desktop/runner/runner_darwin.go
```diff
@@ -0,0 +1,22 @@
...
+package runner
+
+import (
+	"os/exec"
+)
+
+// For notifications to work, we must run in the user context with launchctl asuser.
+func runAsUser(uid string, cmd *exec.Cmd) error {
+
+	// the remaining code in this function is not covered by unit test since it requires root privileges
```
directionless, line 13: If called with your own UID, I think this will work. But changing UIDs requires root.

directionless (commented):
Code seems fine. Not totally sure how to test this across the weird slew of first boot behaviors.

ee/desktop/runner/runner_darwin.go
RebeccaMahany: Added back the check from runner_linux to handle the current user not being root

ee/desktop/client/client.go
```diff
@@ -35,7 +35,7 @@ func New(authToken, socketPath string) client {
 	client := client{
 		base: http.Client{
 			Transport: transport,
-			Timeout:   1 * time.Second,
+			Timeout:   30 * time.Second,
```
RebeccaMahany, line 38: We allow notifications 20sec to complete now, so a 30sec timeout seemed safe

seejdev (approved):

directionless (approved):

James-Pickett (approved):

PR #1006, opened 2023-01-30
makes ec middleware test valid, adds some sanity assertions to test

James-Pickett:
While combing through launcher code trying to resolve an issue, realized a test was invalid. The krypto-ec-middleware test handler was returning the data we expected the middle ware to return. Fixed test and added a few sanity assertions, but the bug was only in the test. Actual code still good.


RebeccaMahany (approved):
🔥

PR #1014, opened 2023-02-03
Let launcher choose debug icon when running in non-prod

seejdev:
This enables k2 to not have to worry about if launcher is running in prod or not. Launcher chooses which icon to use, so the default menu will now go back to the debug icon.

### DIFF
--- a/ee/desktop/menu/menu.go
+++ b/ee/desktop/menu/menu.go
@@ -14,8 +14,7 @@ import (
 type menuIcon string
 
 const (
-	KolideDesktopIcon      = "kolide-desktop"
-	KolideDebugDesktopIcon = "kolide-debug-desktop"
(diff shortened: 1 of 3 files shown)


James-Pickett (approved):

ee/desktop/menu/menu.go
```diff
@@ -14,8 +14,7 @@ import (
 type menuIcon string
 
 const (
-	KolideDesktopIcon      = "kolide-desktop"
-	KolideDebugDesktopIcon = "kolide-debug-desktop"
```
RebeccaMahany, line 18: Looks like removing this const is making the menu parse tests unhappy, may need to update those tests too?

James-Pickett (approved):

PR #1015, opened 2023-02-04
Consolidating agent temp path/dir creation to a single function

seejdev:
This is a tangent off of https://github.com/kolide/launcher/pull/920 intended to streamline temp file/dir paths so that:

- The locations where we use temp paths can be found easily
- We can treat temp path creation uniformly, and avoid one-off path construction
- Temp dir creation is wrapped and ensured to always use the preferred temp location, and secure directory permissions

This PR does not address potential vulnerabilities regarding privilege escalation where:

- Sometimes we use paths that can be predicted by attackers
- There are directories that are expected to exist before launcher starts, but we do not ensure their permissions are correct

### DIFF
--- a/cmd/launcher/interactive.go
+++ b/cmd/launcher/interactive.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/kolide/launcher/pkg/agent"
 	"github.com/kolide/launcher/pkg/autoupdate"
 	"github.com/kolide/launcher/pkg/osquery/interactive"
 )
@@ -40,7 +41,7 @@ func runInteractive(args []string) error {
 	}
 
 	// have to keep tempdir name short so we don't exceed socket length
-	rootDir, err := os.MkdirTemp("", "launcher-interactive")
+	rootDir, err := os.MkdirTemp(agent.TempPath(""), "launcher-interactive")
 	if err != nil {
 		return fmt.Errorf("creating temp dir for interactive mode: %w", err)
 	}
--- a/cmd/launcher/interactive.go
+++ b/cmd/launcher/interactive.go
@@ -41,7 +41,7 @@ func runInteractive(args []string) error {
 	}
 
 	// have to keep tempdir name short so we don't exceed socket length
-	rootDir, err := os.MkdirTemp(agent.TempPath(""), "launcher-interactive")
+	rootDir, err := agent.MkdirTemp("launcher-interactive")
 	if err != nil {
 		return fmt.Errorf("creating temp dir for interactive mode: %w", err)
 	}
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kolide/launcher/ee/control/consumers/notificationconsumer"
 	desktopRunner "github.com/kolide/launcher/ee/desktop/runner"
 	"github.com/kolide/launcher/ee/localserver"
+	"github.com/kolide/launcher/pkg/agent"
 	"github.com/kolide/launcher/pkg/contexts/ctxlog"
 	"github.com/kolide/launcher/pkg/debug"
 	"github.com/kolide/launcher/pkg/launcher"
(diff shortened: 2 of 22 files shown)


cmd/launcher/interactive.go
```diff
@@ -40,7 +41,7 @@ func runInteractive(args []string) error {
 	}
 
 	// have to keep tempdir name short so we don't exceed socket length
```
RebeccaMahany, line 43: Optional -- would we want to add a wrapper `agent.MkdirTempForSocket` (or whatever) that could hold a check for this?

ee/desktop/runner/runner.go
```diff
@@ -553,6 +553,9 @@ func (r *DesktopUsersProcessesRunner) desktopCommand(executablePath, uid, socket
 	cmd := exec.Command(executablePath, "desktop")
 
 	cmd.Env = []string{
+		// without passing the temp var, the desktop icon will not appear on windows and emit the error:
+		// unable to write icon data to temp file: open C:\\windows\\systray_temp_icon_...: Access is denied
```
RebeccaMahany, line 557: Nice find 🔥

RebeccaMahany (approved):
LGTM, let's try it!

cmd/launcher/launcher.go
```diff
@@ -48,11 +48,9 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	// determine the root directory, create one if it's not provided
 	rootDirectory := opts.RootDirectory
 	if rootDirectory == "" {
-		rootDirectory = filepath.Join(os.TempDir(), defaultRootDirectory)
-		if _, err := os.Stat(rootDirectory); os.IsNotExist(err) {
-			if err := os.Mkdir(rootDirectory, fsutil.DirMode); err != nil {
-				return fmt.Errorf("creating temporary root directory: %w", err)
-			}
+		rootDirectory, err := agent.MkdirTemp(defaultRootDirectory)
```
seejdev, line 51: Worth noting the previous code created `rootDirectory` with `0755` permission, whereas `MkdirTemp` defaults to `0700`. However, we then call `os.MkdirAll` with `0700`. 🤷

ee/desktop/runner/runner.go
```diff
@@ -150,7 +150,7 @@ func New(opts ...desktopUsersProcessesRunnerOption) *DesktopUsersProcessesRunner
 		updateInterval:         time.Second * 5,
 		procsWg:                &sync.WaitGroup{},
 		interruptTimeout:       time.Second * 10,
-		usersFilesRoot:         filepath.Join(os.TempDir(), "kolide-desktop"),
+		usersFilesRoot:         agent.TempPath("kolide-desktop"),
```
seejdev, line 153: Ideally, we should create this temp directory and ensure proper permissions, instead of only defining a path.

cmd/launcher/launcher.go
seejdev: Related - the `os.MkdirAll` does not ensure the root directory's permissions are `0700`. Only if the directory does not exist beforehand.

ee/desktop/runner/runner.go
seejdev: Credit goes to @James-Pickett !

cmd/launcher/interactive.go
seejdev: Don't see an issue with that, as long as folks are onboard with the wrapper idea in general

James-Pickett (approved):

cmd/launcher/launcher.go
directionless: This path is only taken when `rootDirectory == ""` which is such an uncommon use case I just don't want to angst about it.

ee/desktop/runner/runner.go
directionless: This one I think we need to think harder about. But this is clearly bug-compatible, so we could take this as improvement and iterate.

directionless (commented):
Generally seems reasonable. Let me consult my notes about windows temp directories.

I wonder if we can add some linting rules to keep us using these wrappers.

cmd/launcher/launcher.go
seejdev: Correct. Just to be clear if `rootDirectory` is non-empty and the dir exists, we will trust the dir even if it's permissions are wide open.

ee/desktop/runner/runner.go
seejdev: Maybe we make `usersFilesRoot` a required param to avoid the temp path all together. Permissions aside.

cmd/launcher/launcher.go
directionless: I'm not sure that's wrong -- most of what we put into `rootDirectory` isn't private. I think when we create the private things we do set the mode.

seejdev: I think you're right. So far I haven't found anything which appears vulnerable to some sort of hijacking here.

cmd/launcher/interactive.go
RebeccaMahany: 🤷‍♀️ I don't feel strongly about it, might as well leave as is for now

directionless (approved):
Okay, I think we can merge as-is, and we'll iterate.

PR #1017, opened 2023-02-07
Activation actions for notifications

RebeccaMahany:
### Description

Adds notification actions for notifications to open the new `action_uri` that can be included with notifications.

Adds documentation about using Multipass, Mockoon, and Microsoft Remote Desktop to test the UI on Linux.

### How it works

* MacOS: runs an app with a notification delegate to handle `didReceiveNotificationResponse` by opening the URL in the request content, if provided
* Windows: works out-of-the-box by specifying the activation argument "launch"
* Linux with dbus: saves sent notification IDs, then listens on dbus for `org.freedesktop.Notifications.ActionInvoked` signals with matching notification IDs, opening the URL if so
* Linux without dbus: appends `Open: <url>` to the end of the notify-send notification because URLs in notifications are clickable

### Testing notes

I tested on Ubuntu 22.04 with the following desktop environments:
* GNOME
* Cinnamon
* MATE
* Xfce4
* KDE

Per usual, testing via xrdp requires commenting out the seat check in `consoleuser_linux.go`.


ee/desktop/runner/runner_linux.go
```diff
@@ -54,5 +56,18 @@ func runAsUser(uid string, cmd *exec.Cmd) error {
 		},
 	}
 
+	// Get the user's session so we can get their display -- this is needed only for handling notifications
+	// at the moment, so we ignore any errors
+	ctx := context.Background()
+	sessionOutput, _ := exec.CommandContext(ctx, "loginctl", "show-user", uid, "--value", "--property=Sessions").Output()
+	session := strings.Trim(string(sessionOutput), "\n")
+	if session != "" {
+		displayOutput, _ := exec.CommandContext(ctx, "loginctl", "show-session", session, "--value", "--property=Display").Output()
```
RebeccaMahany, line 65: TODO -- discovered this is not always set -- may want to default to `:0`

ee/desktop/notify/notify_darwin.m
```diff
@@ -1,16 +1,46 @@
...
+    if ([actionUri length] != 0) {
+        [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:actionUri]];
+    }
+
+    completionHandler();
+}
+@end
+
+void runNotificationListenerApp(void) {
+    [NSAutoreleasePool new];
```
seejdev, line 22: I wonder if this whole function could just be wrapped like:

```
@autoreleasepool {
...
}
```

Then everything gets released automatically.

ee/desktop/notify/notify_linux.go
```diff
@@ -4,38 +4,149 @@
...
+	conn, err := dbus.ConnectSessionBus()
+	if err != nil {
+		level.Warn(logger).Log("msg", "couldn't connect to dbus to start notifier listener, proceeding without it", "err", err)
+	}
+
+	return &dbusNotifier{
+		iconFilepath:        iconFilepath,
+		logger:              logger,
+		conn:                conn,
+		signal:              make(chan *dbus.Signal, 10),
```
seejdev, line 45: Is 10 significant?

ee/control/control.go
```diff
@@ -79,7 +79,7 @@ func (cs *ControlService) Start(ctx context.Context) {
 	for {
 		// Fetch immediately on each iteration, avoiding the initial ticker delay
 		if err := cs.Fetch(); err != nil {
-			level.Debug(cs.logger).Log(
```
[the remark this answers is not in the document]
RebeccaMahany: I'm fine with anything above debug -- I just wanted them to show up in non-debug mode.

ee/desktop/notify/notify_linux.go
RebeccaMahany: It's not! I'll update

ee/desktop/notify/notify_darwin.m
RebeccaMahany: I'll give it a shot!

RebeccaMahany: Looks like it works 🙂

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -128,7 +130,7 @@ func (nc *NotificationConsumer) notify(notificationToSend notification) {
 		return
 	}
 
-	if err := nc.runner.SendNotification(notificationToSend.Title, notificationToSend.Body); err != nil {
+	if err := nc.runner.SendNotification(notificationToSend.Title, notificationToSend.Body, notificationToSend.ActionUri); err != nil {
```
James-Pickett, line 133: I _might_ have found a bit of a race here. 

When I run launcher, it will hit this line before the desktop process has been spawned. Maybe send notification should return an error if there is no one to send a notification to? Or it should return the count of notifications sent?

Or maybe this isn't a big deal? Just don't want it eating notifications.

RebeccaMahany: Looks like I can have the desktop processes runner return an error if `r.uidProcs` is empty, which I think should do it! Thanks for pointing this out, I'll get it sorted

RebeccaMahany: Updated, and added a test for same as well 🙂

ee/desktop/notify/notify_darwin.go
```diff
@@ -16,9 +18,51 @@ import (
...
+		return nil
+	} else {
+		for {
+			select {
+			case <-m.interrupt:
+				return nil
+			}
+		}
+	}
+}
```
James-Pickett, line 59: I think this can be simplified to:

```suggestion
func (m *macNotifier) Listen() error {
	if !isBundle() {
		<-m.interrupt
		return nil
	}

	// isBundle
	go func() {
		<-m.interrupt
		C.stopNotificationListenerApp()
		return
	}()

	C.runNotificationListenerApp()
	return nil
}
```

ee/desktop/notify/notify_linux.go
```diff
@@ -4,38 +4,149 @@
...
+			}
+			d.lock.RUnlock()
+
+			// Confirm that the URI is legitimate
+			actionUri := signal.Body[1].(string)
+			_, err := url.Parse(actionUri)
+			if err != nil {
+				level.Error(d.logger).Log("msg", "received invalid action URI from dbus", "action_uri", actionUri, "parse_err", err)
+				continue
+			}
```
James-Pickett, line 89: we do the same check in notify_consumer.go, can we drop it here?

ee/desktop/notify/notify_windows.go
```diff
@@ -4,19 +4,62 @@
...
+
+// Listen doesn't do anything on Windows -- the `launch` variable in the notification XML
+// automatically handles opening URLs for us.
+func (w *windowsNotifier) Listen() error {
+	for {
+		select {
+		case <-w.interrupt:
+			return nil
+		}
+	}
```
James-Pickett, line 33: I don't think we need the for loop since it's going to block either way?

```suggestion
    <-w.interrupt
    return nil
```

ee/desktop/notify/notify_linux.go
RebeccaMahany: Ah yup, forgot to pull it out here! Will update

ee/desktop/runner/runner_linux.go
```diff
@@ -54,5 +56,31 @@ func runAsUser(uid string, cmd *exec.Cmd) error {
 		},
 	}
 
+	// Get the user's session so we can get their display -- this is needed only for handling notifications
+	// at the moment, so we ignore any errors
+	ctx := context.Background()
+	sessionOutput, _ := exec.CommandContext(ctx, "loginctl", "show-user", uid, "--value", "--property=Sessions").Output()
```
directionless, line 62: Should this be a new context? 

Regardless, it should have a timeout.

directionless: error handling?

directionless: Hrm. This whole block is about trying to set some things into `cmd.Env` right?

I'd suggest we make it a helper func with a sig like `getDisplayInfo(uid int) map[string]string`

directionless (commented):
Not totally sure I absorbed it, so this is mostly nit level, and not deep structure comments. I suspect you can simplify a lot of signatures and future proof a bit by passing the notification object around.

ee/desktop/notify/notify_darwin.go
RebeccaMahany: Oh yes good catch, thanks! Will update here and in the windows listener

ee/control/control.go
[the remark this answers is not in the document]
RebeccaMahany: Oh okay, I don't think I realized the distinction -- I've been tailing debug.json while testing instead of the system logs. I can move these back to debug, then.

ee/desktop/server/server.go
```diff
@@ -27,17 +27,17 @@ type DesktopServer struct {
 	shutdownChan     chan<- struct{}
 	authToken        string
 	socketPath       string
-	notifier         *notify.DesktopNotifier
+	notifier         notify.DesktopNotifier
```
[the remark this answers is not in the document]
RebeccaMahany: `notify.DesktopNotifier` is an interface after the changes in this PR 🙂 -- it's here: https://github.com/kolide/launcher/blob/c9639060378b02c5f14c082083caf1b6712b4f41/ee/desktop/notify/notify.go#L7

ee/desktop/notify/notify.go
```diff
@@ -4,23 +4,18 @@ import (
 	"github.com/go-kit/kit/log"
 )
 
-type DesktopNotifier struct {
-	logger       log.Logger
-	iconFilepath string
+type DesktopNotifier interface {
```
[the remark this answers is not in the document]
RebeccaMahany: Yeah, so that it can be injected into the desktop server -- https://github.com/kolide/launcher/pull/1017#discussion_r1115993124. I could definitely make a separate interface in that package, but I'm not sure I see the benefit?

ee/desktop/runner/runner_linux.go
RebeccaMahany: This whole bit is still a bit TBD as I figure out the wayland issue -- it needs some tidying up! I can pull it out into a helper func once it's finalized.

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -28,14 +29,15 @@ type NotificationConsumer struct {
 
 // The desktop runner fullfils this interface -- it exists for testing purposes.
 type userProcessesRunner interface {
-	SendNotification(title, body string) error
+	SendNotification(title, body, actionUri string) error
```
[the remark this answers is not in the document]
RebeccaMahany: I think you're right that I should be passing around `notification` at this point, as we add more features to notifications. I hadn't before because it was a bit of a pain figuring out where to define it to avoid circular imports, but I may just need to do it once for the notification consumer and once for the notifier side.

ee/control/control.go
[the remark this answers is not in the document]
RebeccaMahany: These are back to debug and I updated the ones in notify_consumer as well 👍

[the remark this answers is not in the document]
directionless: I added a potential project of making this all simpler.

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
directionless: I don't think you can define it twice. But I think you could create a `types` somewhere if you needed to

ee/desktop/server/server.go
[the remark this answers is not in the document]
directionless: I think that's somewhat inverted. 

Consumers should declare the interface they need. The server can handle anything with `SendNotification` right?

ee/desktop/notify/notify.go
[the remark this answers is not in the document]
directionless: generally speaking, interfaces should be defined by the consumer and very slim.

ee/desktop/server/server.go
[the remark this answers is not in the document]
RebeccaMahany: Ah sorry my answer wasn't very complete. There's a second consumer, which is launcher/desktop, and it cares about Listen/Interrupt -- here: https://github.com/kolide/launcher/blob/36fc5aaf629f5652e3af95238d8599c637bd8b92/cmd/launcher/desktop.go#L106

I'll split it up so there's one interface in use by launcher/desktop for Listen/Interrupt, and a separate one here, both of which are fulfilled by the notifier.

ee/control/consumers/notificationconsumer/notify_consumer.go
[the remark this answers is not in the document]
RebeccaMahany: It's already defined twice -- once as `notificationconsumer.notification` (unexported, on the launcher root side) and once as `notify.Notification` (exported, on the desktop process side, now used in both `notify` and `server`). The desktop process side doesn't need the ID, ValidUntil, or SentAt fields, so I don't mind that it's split up -- but I think moving it to a `types` is the way to go, the dependency chain on the launcher root side is not working out well for doing it otherwise 🙂 .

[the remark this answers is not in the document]
RebeccaMahany: Never mind! I think the refactor I did in this PR and at the end of the last one made this much easier than I realized. There's now only 1 definition and we're passing `notification`s around everywhere.

ee/desktop/runner/runner_linux.go
RebeccaMahany: I've updated this to make a separate function and add a timeout, and additionally passed in the logger so that we can log errors. I don't think it makes sense to stop the desktop process from launching if anything here fails, so I think that should be all we need re: error handling too.

ee/desktop/notify/notify_linux.go
```diff
@@ -4,38 +4,141 @@
...
+	}
+
+	for {
+		select {
+		case signal := <-d.signal:
+			if signal == nil || signal.Name != signalActionInvoked {
+				continue
+			}
+
+			// Confirm that this is a Kolide-originated notification by checking for known notification IDs
```
directionless, line 72: Long term, is this the sort of thing we should have strong authentication for? Should we sign them, for example?

ee/desktop/runner/runner_linux.go
```diff
@@ -54,5 +59,73 @@ func runAsUser(uid string, cmd *exec.Cmd) error {
...
+		level.Debug(r.logger).Log(
+			"msg", "could not get user session",
+			"uid", uid,
+			"err", err,
+		)
+		return envVars
+	}
+
+	sessions := strings.Trim(string(sessionOutput), "\n")
+	if sessions != "" {
```
directionless, line 89: early return on `sessions =- ""` will be a little cleaner

directionless (dismissed):
I'm good with this. Maybe a couple nits, but we can get them in a future PR. 

@seejdev How does the c look? Does it fit with the other UI things happening for the menu?

ee/desktop/runner/runner_linux.go
RebeccaMahany: Oh yup for sure, will update!

ee/desktop/notify/notify_linux.go
RebeccaMahany: At the moment, it's possible but very unlikely that we'd act on a notification that we didn't originate, since we don't clear out these notification IDs from `d.sentNotificationIds` after acting on them. I don't think another application would be allowed to create an `ActionInvoked` signal with a previously-used ID until the ID resets after hitting the max of uint32. So I think in the future possibly worth adding strong authentication, and definitely worth removing a notification ID from `d.sentNotificationIds` after we act on it.

For how we could accomplish this: the only info we get from dbus that is usable (that I've found) is the notification ID assigned by dbus and the URL itself (i.e. the `action_key`) -- so we could possibly encrypt the URL, and then only act on URLs that we can decrypt.

directionless (approved):
I'm going to approve it. If we turn out to have issues, we can revisit

PR #1018, opened 2023-02-08
Menu icons and generator

directionless:
This brings in some initial icon work from Fritz for use in menus and notifications. It has source material, and then a generator to create the subsequent versions. 

Notes:
* It's meant to replace https://github.com/kolide/launcher/tree/main/ee/desktop/assets
* If we need these _on disk_ and not just as embeds, we could create something to instantiate them into launcher's root dir
* They will grow the launcher binary. Adding 10 ico files just does that 😦


ee/ui/assets/assets.go
```diff
@@ -0,0 +1,69 @@
...
+// THIS FILE IS AUTOGENERATED
+package assets
+
+import (
+	_ "embed"
+)
+
+
+//go:embed menubar-blocked-color-dark.ico
+var KolideMenuBlockedColorDarkIco []byte
```
James-Pickett, line 11: can't leave a comment on the pictures so, putting it here, I think some of the icons have dark / light switched

or maybe in the context of `menubar-will-block-color-light.png`, light means "light theme" as opposed tot he light color of the asset?

image

James-Pickett (commented):
This is really cool! Left one comment about possibly naming reversed in light vs dark.

ee/ui/assets/assets.go
directionless: The current names describe the intended usage. But I think it would be more re-usable for them to describe the _icon_ and leave the usage up to the application and control layers. I need to circle back with Fritz and discuss that. It's probably the biggest reason this is still in draft.

But I like writing go generators, so I did that anyhow.

ee/ui/assets/README.md
```diff
@@ -0,0 +1,29 @@
+# UI Assets
+
+These are various images from Fritz, that we bake into the Agent for use in parts
+of the Desktop UI.
+
+They originals from Fritz are in `source/`, and there's a generator to convert
+them into the desired formats and setup the embed. (In other versions, I've done
+this with make, or even shell snippets, but let's sometimes I try to stick with
```
seejdev, line 8:
```suggestion
this with make, or even shell snippets, but sometimes I try to stick with
```

ee/ui/assets/generator/generator.go
```diff
@@ -0,0 +1,238 @@
...
+	embeddedFiles = append(embeddedFiles, output)
+
+	if skip, err := skipFile(input, output); err != nil {
+		return err
+	} else if skip {
+		level.Debug(logger).Log("msg", "skipping", "input", input, "output", output)
+		return nil
+	}
+
+	// FIXME Does this need scaling?
```
seejdev, line 155: Confusingly, I can't find Apple official guidance on sizes. Some unofficial sources say macOS menu bar icons should be max 22x22, ideally 16x16? Seems kind of small... I believe macOS will scale larger icons down to proper scale regardless 🤷 

https://iconhandbook.co.uk/reference/chart/osx/
https://bjango.com/articles/designingmenubarextras/

directionless: I think it's better for us us to scale, less space in the binary

directionless: Scaling to 22x22 didn't change sizes. So I removed the FIXME, and left it as a `cp`

James-Pickett (approved):

PR #1020, opened 2023-02-08
Add auth to control server requests

RebeccaMahany:
Description:
* Implement control server auth as defined in this PR: https://github.com/kolide/k2/pull/7083
* Switch `valid_until` for notifications to a timestamp
* Set control server hostname + associated settings based on k2 hostname

### DIFF
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -180,7 +180,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 
 	// Set control server URL and control server TLS settings based on Kolide server URL, defaulting to local server
 	controlServerURL := ""
-	insecureTLS := *flInsecureTLS
+	insecureControlTLS := false
 	disableControlTLS := false
 	if *flKolideServerURL == "k2device.kolide.com" {
 		controlServerURL = "k2control.kolide.com"
@@ -191,7 +191,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 	} else if *flKolideServerURL == "localhost:3443" {
 		controlServerURL = *flKolideServerURL
 		// We don't plumb flRootPEM through to the control server, just disable TLS for now
-		insecureTLS = true
+		insecureControlTLS = true
 	} else if *flKolideServerURL == "localhost:3000" {
 		controlServerURL = *flKolideServerURL
 		disableControlTLS = true
@@ -208,11 +208,12 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		ControlRequestInterval:             *flControlRequestInterval,
 		Debug:                              *flDebug,
 		DisableControlTLS:                  disableControlTLS,
+		InsecureControlTLS:                 insecureControlTLS,
 		EnableInitialRunner:                *flInitialRunner,
 		EnrollSecret:                       *flEnrollSecret,
 		EnrollSecretPath:                   *flEnrollSecretPath,
 		AutoloadedExtensions:               flAutoloadedExtensions,
-		InsecureTLS:                        insecureTLS,
+		InsecureTLS:                        *flInsecureTLS,
 		InsecureTransport:                  *flInsecureTransport,
 		KolideHosted:                       *flKolideHosted,
 		KolideServerURL:                    *flKolideServerURL,
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -84,7 +84,6 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		flDebug             = flagset.Bool("debug", false, "Whether or not debug logging is enabled (default: false)")
 		flOsqueryVerbose    = flagset.Bool("osquery_verbose", false, "Enable verbose osqueryd (default: false)")
 		flDeveloperUsage    = flagset.Bool("dev_help", false, "Print full Launcher help, including developer options")
-		flDisableControlTLS = flagset.Bool("disable_control_tls", false, "Disable TLS encryption for the control features")
 		flInsecureTransport = flagset.Bool("insecure_transport", false, "Do not use TLS for transport layer (default: false)")
 		flInsecureTLS       = flagset.Bool("insecure", false, "Do not verify TLS certs for outgoing connections (default: false)")
 
@@ -92,6 +91,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		_ = flagset.String("debug_log_file", "", "DEPRECATED")
 		_ = flagset.Bool("control", false, "DEPRECATED")
 		_ = flagset.String("control_hostname", "", "DEPRECATED")
+		_ = flagset.Bool("disable_control_tls", false, "Disable TLS encryption for the control features")
 	)
 
 	flagset.Var(&flOsqueryFlags, "osquery_flag", "Flags to pass to osquery (possibly overriding Launcher defaults)")
@@ -181,7 +181,7 @@ func parseOptions(args []string) (*launcher.Options, error) {
 	// Set control server URL and control server TLS settings based on Kolide server URL, defaulting to local server
 	controlServerURL := "localhost:3000"
 	insecureTLS := *flInsecureTLS
-	disableControlTLS := *flDisableControlTLS
+	disableControlTLS := false
 	if *flKolideServerURL == "k2device.kolide.com" {
 		controlServerURL = "k2control.kolide.com"
 	} else if *flKolideServerURL == "k2device-preprod.kolide.com" {
@@ -323,7 +323,6 @@ func developerUsage(flagset *flag.FlagSet) {
 	printOpt("notary_prefix")
 	fmt.Fprintf(os.Stderr, "\n")
 	printOpt("control_get_shells_interval")
-	printOpt("disable_control_tls")
 	fmt.Fprintf(os.Stderr, "\n")
 	printOpt("osquery_flag")
 	fmt.Fprintf(os.Stderr, "\n")
--- a/cmd/launcher/options.go
+++ b/cmd/launcher/options.go
@@ -178,6 +178,24 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		return nil, err
 	}
 
+	// Set control server URL and control server TLS settings based on Kolide server URL, defaulting to local server
+	controlServerURL := "localhost:3000"
+	insecureTLS := *flInsecureTLS
+	disableControlTLS := *flDisableControlTLS
+	if *flKolideServerURL == "k2device.kolide.com" {
+		controlServerURL = "k2control.kolide.com"
+	} else if *flKolideServerURL == "k2device-preprod.kolide.com" {
+		controlServerURL = "k2control-preprod.kolide.com"
+	} else if strings.HasSuffix(*flKolideServerURL, "herokuapp.com") {
+		controlServerURL = *flKolideServerURL
+	} else if *flKolideServerURL == "localhost:3443" {
+		controlServerURL = *flKolideServerURL
+		insecureTLS = true
+	} else if *flKolideServerURL == "localhost:3000" {
+		controlServerURL = *flKolideServerURL
+		disableControlTLS = true
+	}
+
 	opts := &launcher.Options{
 		Autoupdate:                         *flAutoupdate,
 		AutoupdateInterval:                 *flAutoupdateInterval,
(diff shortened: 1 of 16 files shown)


ee/control/client_http.go
```diff
@@ -40,77 +61,140 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
+	if err != nil {
+		return nil, fmt.Errorf("could not create config request: %w", err)
+	}
+	configReq.Header.Set(HeaderChallenge, string(challenge))
+	configReq.Header.Set("Content-Type", "application/json")
+	configReq.Header.Set("Accept", "application/json")
+
+	// Calculate first signature
+	localDbKeys := agent.LocalDbKeys()
+	if localDbKeys.Type() != "noop" {
```
directionless, line 85: This would be better as an early return. And, honestly, as something throwing an error.

ee/control/client_http.go
```diff
@@ -40,77 +61,140 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
+		if err != nil {
+			return nil, fmt.Errorf("could not get signature header from local db keys: %w", err)
+		}
+		configReq.Header.Set(HeaderKey, key1)
+		configReq.Header.Set(HeaderSignature, sig1)
+	}
+
+	// Calculate second signature if available
+	hardwareKeys := agent.HardwareKeys()
+	if hardwareKeys.Type() != "noop" {
```
directionless, line 100: Might need a timeout on this. 

@James-Pickett I wonder if we should bake some kind of helper into krypto. Something like `echelper.Sign(signer1, signer2) (sig1, sig2, error)` or `echelper.Sign(signers ...crypto.Signers) ([]sigs, error)`

directionless (commented):
Seems okay to me. Seeing how those signatures are made, I think we could use a helper in `krypto` for it, but I don't know that needs to block this PR. (Though you do need a timeout in hardware signing. )

@seejdev you wrote most of the control server, does this look good to you?

ee/control/client_http.go
RebeccaMahany: Ah ok, my understanding was that maybe in the future it'd be possible for the request to be signed only by the hardware keys and not the db keys -- if that's the case, I could check both first to return an error if they're both noop? If not I'll just add an early return here.

ee/control/client_http.go
James-Pickett: I've exposed a func in krypto echelper that does signing with time out, accepting duration and interval as parameters. I played around with what @directionless suggested, but it felt weird because ...

what if it returns an error?
which one failed? 
I accept it if only one fails, but if I get back an array of signatures, how do I know which sig matches which key?

so I thought about having custom error messages that says which index of the given sigs failed ... but I feel like that added complexity is > the complexity of just calling sign twice.

I'm open to discussion =)

ee/control/client_http.go
```diff
@@ -40,77 +63,150 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
+			return nil, fmt.Errorf("no token set; could not make request to get one: %w", err)
+		}
+	}
+
+	dataReq, err := http.NewRequest(http.MethodGet, c.url(fmt.Sprintf("/api/agent/object/%s", hash)).String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not create subsystem data request: %w", err)
+	}
+
+	dataReq.Header.Set("Authorization", "Bearer "+c.token)
```
seejdev, line 145:
```suggestion
	dataReq.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.token))
```

ee/control/client_http.go
```diff
@@ -40,77 +63,150 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
 
-func (c *HTTPClient) do(verb, path string) (*http.Response, error) {
-	request, err := http.NewRequest(
-		verb,
-		c.url(path).String(),
-		nil,
-	)
+func (c *HTTPClient) GetSubsystemData(hash string) (io.Reader, error) {
+	if c.token == "" {
+		_, err := c.GetConfig()
```
seejdev, line 134: It may be better to just error out here instead of trying to handle an empty token. The caller needs to get hashes from the data returned by `GetConfig`, so if this function is being called without `GetConfig` called first, the hash is potentially stale or otherwise invalid.

seejdev:
So exciting!

seejdev:
> Seems okay to me. Seeing how those signatures are made, I think we could use a helper in `krypto` for it, but I don't know that needs to block this PR. (Though you do need a timeout in hardware signing. )
> 
> @seejdev you wrote most of the control server, does this look good to you?

LGTM - I have one concern on the `GetSubsystemData` implementation.

I'm glad this fit in relatively cleanly, and great to see @James-Pickett 's crypto work come to fruition here! Nice job!

ee/control/client_http.go
RebeccaMahany: Oh yup that absolutely makes more sense, I'll pull this out!

cmd/launcher/options.go
```diff
@@ -178,22 +178,40 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		return nil, err
 	}
 
+	// Set control server URL and control server TLS settings based on Kolide server URL, defaulting to local server
+	controlServerURL := "localhost:3000"
+	insecureTLS := *flInsecureTLS
+	disableControlTLS := *flDisableControlTLS
+	if *flKolideServerURL == "k2device.kolide.com" {
```
directionless, line 185: We should not allow a command line flag.

I'd probably recommend a switch statement, with the default case setting it to disabled.

RebeccaMahany: We're not allowing a command line flag for the control server -- https://github.com/kolide/launcher/blob/5f2907715481767a07f43dbbcdff45990eafcc32/cmd/launcher/options.go#L94

ee/control/client_http.go
```diff
@@ -40,77 +63,150 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
+		return nil, fmt.Errorf("could not create config request: %w", err)
+	}
+	configReq.Header.Set(HeaderChallenge, string(challenge))
+	configReq.Header.Set("Content-Type", "application/json")
+	configReq.Header.Set("Accept", "application/json")
+
+	// Calculate first signature
+	localDbKeys := agent.LocalDbKeys()
+	if localDbKeys.Type() != "noop" {
+		key1, err := keyHeaderValue(localDbKeys)
```
James-Pickett, line 89: the krypto PR I have open now exposes funcs like

`PublicEcdsaToB64Der`
`SignWithTimeout`

you don't have to use them, but I'll let you know when that PR merges if interested

cmd/launcher/options.go
directionless: `*flDisableControlTLS` I'm not sure we should expose this as a flag.

ee/control/client_http.go
directionless: I _suspect_ the common case is going to be consistent:
1. first sig is local, hard fail for that
2. second sig is hardware. Probably don't fail. maybe expose option.

I don't know what's more idiomatic. Make the callers do the work, or write a helper. I'm happy to roll with whatever.

cmd/launcher/options.go
RebeccaMahany: I'll move it to the deprecated section 👍

ee/control/client_http.go
RebeccaMahany: Yes definitely!

ee/control/client_http.go
James-Pickett: I think an uncommon success condition like "1st of n" should be the responsibility of the consumer .... If I'm using a library and I see a func `SignWithTimeout` that accepts multiple signers I expect an error if any of them fails. I know we are the only consumers of this library, but it just feels weird to me =). So if you good with whatever, lets let launcher handle it.

ee/control/client_http.go
RebeccaMahany: Updated to add an early return with an error here

cmd/launcher/options.go
```diff
@@ -178,22 +178,40 @@ func parseOptions(args []string) (*launcher.Options, error) {
...
+		controlServerURL = "k2control-preprod.kolide.com"
+	} else if strings.HasSuffix(*flKolideServerURL, "herokuapp.com") {
+		controlServerURL = *flKolideServerURL
+	} else if *flKolideServerURL == "localhost:3443" {
+		controlServerURL = *flKolideServerURL
+		insecureTLS = true
+	} else if *flKolideServerURL == "localhost:3000" {
+		controlServerURL = *flKolideServerURL
+		disableControlTLS = true
+	}
```
directionless, line 197: If we don't match any of these (a user running launcher outside kolide) we should disable the control server.

cmd/launcher/options.go
```diff
@@ -178,22 +178,40 @@ func parseOptions(args []string) (*launcher.Options, error) {
...
+	disableControlTLS := false
+	if *flKolideServerURL == "k2device.kolide.com" {
+		controlServerURL = "k2control.kolide.com"
+	} else if *flKolideServerURL == "k2device-preprod.kolide.com" {
+		controlServerURL = "k2control-preprod.kolide.com"
+	} else if strings.HasSuffix(*flKolideServerURL, "herokuapp.com") {
+		controlServerURL = *flKolideServerURL
+	} else if *flKolideServerURL == "localhost:3443" {
+		controlServerURL = *flKolideServerURL
+		insecureTLS = true
```
directionless, line 193:
```suggestion
		// We don't plumb flRootPEM through to the control server, just disable TLS for now
		insecureTLS = true
```

cmd/launcher/options.go
RebeccaMahany: `Control` is already hardcoded to false below (and the corresponding flag is marked as deprecated above) -- but I don't think that flag is actually in use right now. Is there another way you meant for disabling the control server?

directionless: Hrm. I think there's some out of date bits here. 

The intent, is to run the control server, with those hostnames. If launcher is not talking to one of these, no control server.

I think opts.Control isn't used any more. So in the default case, launcher would try to hit localhost. Which doesn't seem like what we want.

I think either we plumb opts.Control back. or we remove it, and do something with `controlServerURL` being `""`

RebeccaMahany: @seejdev do you have a suggestion? Otherwise I'm inclined to make `opts.Control` do something.

I think currently it will try to hit localhost by default, so I can also leave it as-is (no behavior change) for this PR and change it up in a subsequent PR once we know how we'd like this to work.

directionless: I think my bias would be to remove `opts.Control`, and just check for whether `controlServerURL` is empty. No real need for 2 variables.

directionless: Could probably kick that to the next PR. Once `krypto` merges, we'll want to update the calling functions

mouchar:
When we can expect this change to reach the stable channel? Now, the launcher is mindlessly trying to access https://localhost:3000 every minute.
It interferes with our apps running on this port.

ee/control/client_http.go
RebeccaMahany: Updated to use these!

cmd/launcher/options.go
RebeccaMahany: Discussed offline + updated accordingly -- now we don't run the control service if `controlServerURL` is empty, and also updated `krypto`

James-Pickett (approved):

RebeccaMahany:
> When we can expect this change to reach the stable channel? Now, the launcher is mindlessly trying to access https://localhost:3000 every minute. It interferes with our apps running on this port.

Hi @mouchar, thanks for bringing this issue to our attention. We will release a patch 0.13.6 to stable today to address it. 🙂

directionless:
> When we can expect this change to reach the stable channel? Now, the launcher is mindlessly trying to access https://localhost:3000 every minute. It interferes with our apps running on this port.

@mouchar I'm so sorry you ran into an issue. We didn't intend for that to go live. As @RebeccaMahany says, we'll get it backed out.

directionless:
@mouchar We have released 0.13.6, which should fix your issue.

ee/control/client_http.go
```diff
@@ -40,77 +63,129 @@ func NewControlHTTPClient(logger log.Logger, addr string, client *http.Client, o
...
+	sig1, err := signatureHeaderValue(localDbKeys, challenge)
+	if err != nil {
+		return nil, fmt.Errorf("could not get signature header from local db keys: %w", err)
+	}
+	configReq.Header.Set(HeaderKey, string(key1))
+	configReq.Header.Set(HeaderSignature, sig1)
+
+	// Calculate second signature if available
+	hardwareKeys := agent.HardwareKeys()
+	if hardwareKeys.Type() != "noop" {
```
directionless, line 103: `hardwareKeys.Public() != nil` is probably a little cleaner

directionless (commented):
TLS stuff seems a little weird, not sure I understand it right though

cmd/launcher/options.go
```diff
@@ -178,22 +178,41 @@ func parseOptions(args []string) (*launcher.Options, error) {
 		return nil, err
 	}
 
+	// Set control server URL and control server TLS settings based on Kolide server URL, defaulting to local server
+	controlServerURL := ""
+	insecureTLS := *flInsecureTLS
+	disableControlTLS := false
```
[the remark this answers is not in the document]
RebeccaMahany: Those options do [two different things](https://github.com/kolide/launcher/blob/main/ee/control/client_http_option.go) for the control service client, though -- `insecureTLS` sets `InsecureSkipVerify` on the client, and `disableTLS` sets the base URL's scheme to http.

I agree that it would be better to not overload `*flInsecureTLS` -- how about I add a separate `InsecureControlTLS` option (not controlled by a command line flag) for this?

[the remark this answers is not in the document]
RebeccaMahany: I went ahead and did it ⬆️ this way in the last commit, let me know what you think or if you'd like changes

directionless (approved):
Let's try!

cmd/launcher/options.go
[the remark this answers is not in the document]
directionless: Ah yes, good catch

PR #1021, opened 2023-02-09
Using Kolide's fork of systray

### LINKED ISSUE #1019: Improve menu bar to support colorized icons on macOS
The fyne systray project by default uses the [template](https://developer.apple.com/documentation/appkit/nsimage/1520017-template) property of `NSImage` when creating icons for the menu bar. This causes the icon to be rendered without color, and the alpha channel controls the opacity when in either light or dark mode. So even if we have colorized icons, they will never render with color.

Colorized icons will better help communicate launcher/k2 status to end users. The "right" solution here is probably moving to a better framework like fyne v2, but this is a potentially large ask. As a short-lived measure, we can simply fork and change the macOS code to create the menu bar icons as non-templated. We can use the modified fork until we decide on an improved approach to cross-platform GUI in Go, in general.

seejdev:
Fixed #1019 

Use our fork over the fyne one, so we can get colorized icons on macOS.

### DIFF
--- a/ee/desktop/menu/menu_systray.go
+++ b/ee/desktop/menu/menu_systray.go
@@ -3,7 +3,7 @@ package menu
 import (
 	"sync"
 
-	"fyne.io/systray"
+	"github.com/kolide/systray"
 
 	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/launcher/ee/desktop/assets"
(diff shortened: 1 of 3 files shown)


RebeccaMahany (approved):

directionless (approved):
Let's try!

PR #1026, opened 2023-02-11
Control Server Updates

directionless:
As I play with the control server, I've found several things I want to tweak. Unfortunately, this means this PR is a little sprawling. Nothing really huge, a little refactoring, a new idea... If anything here seems controversial, I'm happy to talk about it, drop it, etc.

### DIFF
--- a/pkg/agent/flags.go
+++ b/pkg/agent/flags.go
@@ -9,5 +9,5 @@ var Flags flagInt = &initialFlagTest{}
 type initialFlagTest struct{}
 
 func (initialFlagTest) DebugServerData() bool {
-	return true
+	return false
 }
--- a/pkg/agent/flags.go
+++ b/pkg/agent/flags.go
@@ -1,0 +1,13 @@
+package agent
+
+type flagInt interface {
+	DebugServerData() bool
+}
+
+var Flags flagInt = &initialFlagTest{}
+
+type initialFlagTest struct{}
+
+func (initialFlagTest) DebugServerData() bool {
+	return true
+}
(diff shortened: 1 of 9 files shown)


pkg/agent/flags.go
```diff
@@ -0,0 +1,13 @@
...
+type flagInt interface {
+	DebugServerData() bool
+}
+
+var Flags flagInt = &initialFlagTest{}
+
+type initialFlagTest struct{}
+
+func (initialFlagTest) DebugServerData() bool {
+	return true
```
directionless, line 12: This should be `false` or conditionalized before merge. But `true` is handy for debugging.

pkg/agent/flags.go
```diff
@@ -0,0 +1,13 @@
+package agent
+
+type flagInt interface {
+	DebugServerData() bool
+}
+
+var Flags flagInt = &initialFlagTest{}
```
James-Pickett, line 7: The goal with this is to make it so that select flags are globally available? If so, I like it, better than threading flags through all the time.

James-Pickett (commented):
All the updates LGTM, no approving due to agent/flags.go TODO

pkg/agent/flags.go
directionless, line 12:
```suggestion
	return false
```

pkg/agent/flags.go
directionless: Bigger discussion -- https://github.com/kolide/launcher/issues/1029

PR #1030, opened 2023-02-13
query over localserver

James-Pickett:
* adds local server endpoint for osq query queries
* updates local server to accept post or get
* updates internal ec middlewere routing to use posts


ee/localserver/mocks/querier.go
```diff
@@ -0,0 +1,51 @@
+// Code generated by mockery v2.20.0. DO NOT EDIT.
```
directionless, line 1: How was this generated? Should it be encoded in a `go:generate` line?

ee/localserver/krypto-ec-middleware.go
```diff
@@ -92,6 +93,9 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 		}
 
 		v := url.Values{}
+		for key, val := range cmdReq.UrlParameters {
+			v.Add(key, val)
+		}
```
directionless, line 98: This is basically supporting GET parameters. Do you think it makes sense to look at POST data instead? queries can get pretty big...

ee/localserver/request-query.go
```diff
@@ -0,0 +1,33 @@
...
+	"encoding/json"
+	"net/http"
+)
+
+func (ls *localServer) requestQueryHandler() http.Handler {
+	return http.HandlerFunc(ls.requestQueryHanlderFunc)
+}
+
+func (ls *localServer) requestQueryHanlderFunc(res http.ResponseWriter, req *http.Request) {
+	query := req.URL.Query().Get("query")
```
directionless, line 13: If you don't use POST, I think we should assume some encoding. base64 is probably fine. (Maybe need base64 for post too)

directionless (commented):
I love how this comes together with the muxes and wrapping.

ee/localserver/mocks/querier.go
James-Pickett: it was generated with [mockery](https://github.com/vektra/mockery) a cli tool for creating mocks, to use go:generate we would have to add mockery as a dependency I think?

ee/localserver/krypto-ec-middleware.go
James-Pickett: I don't know, adding url parameters just fit so neatly =) if we introduce posts, now the middleware has to make a decision about weather the next endpoint it a post or a get. I will add some tests with huge url params and see if it effects anything.

ee/localserver/request-query.go
James-Pickett: we call the standard url encoding when we make the request from krypto-ec-middleware.go do you think that is not sufficient? I tested with curl like.

```
curl -G localhost:40978/query --data-urlencode "query=select * from kolide_launcher_info where branch in ('unknown') AND branch like '%unknown%'"
```

I can also add some complex sql statements to the tests to see how it handles `string -> json -> bytes ->  json -> string`

James-Pickett: lol, the code snippet I posted is actually doing a post 😕

ee/localserver/mocks/querier.go
directionless: No. Only if you want `go generate ./` to work 😆 

It's not in the Makefile, because one of the generate commands require `thrift`, and I think one requires `grpc`

I think we should add this here as a way of documenting how it happened. I'd add some comments about how to install it or whatever.

ee/localserver/krypto-ec-middleware.go
directionless: Does it? Can we hint it from the cmd? 

I guess if the data fits in a GET, then maybe we can keep ignoring this. After all, at the point in time we need a POST, we probably need it to localserver as well

ee/localserver/request-query.go
directionless: I think we should base64 encode it.

James-Pickett: ecmiddleware now b64 encoding url parameters

ee/localserver/mocks/querier.go
James-Pickett: added a go generate line and some comments in the generated querier struct

ee/localserver/server.go
```diff
@@ -35,6 +35,7 @@ var portList = []int{
 	22322,
 }
 
+//go:generate mockery --name Querier
```
directionless, line 38: Could also put this into the `_test.go` file. You're call.

I'd add the comment about where mockery is from to this.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -23,7 +23,8 @@ const (
 )
 
 type v2CmdRequestType struct {
-	Path string
+	Path          string
+	UrlParameters map[string]string
```
directionless, line 27: I'm kinda willing to try, but I'm still really leery of max size stuff.

ee/localserver/krypto-ec-middleware.go
James-Pickett: After some testing, I did start getting errors when url params got too big. Since everything comming _into_ local server is a GET request, we will have some limitations there.

However, we can switch the internal routing to all POSTs, then we don't have to make decisions in the middle ware about weather it's a GET or POST and we just add a `body []byte` filed to the `cmdReq`

... what say you @directionless ?

ee/localserver/krypto-ec-middleware.go
James-Pickett: https://github.com/kolide/launcher/pull/1030#discussion_r1107406286 =)

directionless: Yeah, I lost track of the other comment. GitHub 🤷

ee/localserver/krypto-ec-middleware.go
directionless: I agree we'll need to handle the incoming data as POST. I think localserver can probably just handle both, I remember I did something obvious to limit it to GET.

I don't think I have an opinion about whether we push it all through post internally or not. 

We need to be consistent in the priority order of where `box` comes from, but aside from that, I think we can kinda just do whatever.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -137,3 +128,38 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+	if r.Body == nil {
+		return nil, fmt.Errorf("no box found in url params or request body: body nil")
+	}
+
+	var body map[string]string
+	if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+		return nil, fmt.Errorf("unmarshalling request body to json: %w", err)
+	}
+
+	val, ok := body["box"]
```
James-Pickett, line 154: not sure about this .... 

was thinking that if the url param was `box=VALUE`
the body ought to be `{"box":"VALUE"}`

but the body could just simply be `VALUE`

all tangled in my head .... please comment =)

ee/localserver/request-query.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"net/http"
+)
+
+func (ls *localServer) requestQueryHandler() http.Handler {
+	return http.HandlerFunc(ls.requestQueryHanlderFunc)
+}
+
+func (ls *localServer) requestQueryHanlderFunc(res http.ResponseWriter, r *http.Request) {
+	// now check body
```
James-Pickett, line 14:
```suggestion
```

ee/localserver/request-query.go
```diff
@@ -0,0 +1,50 @@
...
+		return
+	}
+
+	var body map[string]string
+	if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+		res.Write([]byte(fmt.Sprintf("error unmarshaling request body: %s", err)))
+		return
+	}
+
+	query, ok := body["query"]
```
James-Pickett, line 26: same here, should the body be in the form of json always?

ee/localserver/krypto-ec-middleware.go
James-Pickett: updated to handle both incoming get and post data, un sure of what post body should look like, see https://github.com/kolide/launcher/pull/1030#discussion_r1108773392

ee/localserver/krypto-ec-middleware.go
directionless: I had exactly the same thought... "I was thinking `{"box":"VALUE"}`, but I guess it could just be the whole body"

I suspect `{"box":"VALUE"}` gives us more rope in terms of additions/changes

ee/localserver/krypto-determiner-middleware.go
```diff
@@ -27,8 +27,9 @@ func (h *kryptoDeterminerMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Re
 	// Extract the box from the URL query parameters
 	boxRaw := r.URL.Query().Get("box")
 	if boxRaw == "" {
-		level.Debug(h.logger).Log("msg", "no data in box query parameter")
-		w.WriteHeader(http.StatusUnauthorized)
+		// if we dont box param, assume it's a post
```
James-Pickett, line 30:
```suggestion
		// if we dont have a box param, assume it's a post
```

ee/localserver/server.go
```diff
@@ -112,6 +115,11 @@ func New(logger log.Logger, db *bbolt.DB, kolideServer string) (*localServer, er
 	mux.HandleFunc("/", http.NotFound)
 	mux.Handle("/v0/cmd", kryptoDeterminerMiddleware)
 
+	// uncomment to test without going through middleware
+	// for example:
+	// curl -G localhost:40978/query?query=$(echo "select * from kolide_launcher_info" | base64)
```
James-Pickett, line 120:
```suggestion
	// curl localhost:40978/query --data '{"query":"select * from kolide_launcher_info"}'
```

ee/localserver/krypto-ec-middleware.go
```diff
@@ -137,3 +127,38 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+
+		return challenge.UnmarshalChallenge(decoded)
+	}
+
+	// now check body
+	if r.Body == nil {
+		return nil, fmt.Errorf("no box found in url params or request body: body nil")
+	}
+
+	var body map[string]string
```
directionless, line 148: Do you think `map[string]string` or `map[string]any` or even something structured? I'm kinda good with whatever, but have run into problems with booleans in json being unmarshalled to `map[string]string`

ee/localserver/krypto-ec-middleware.go
```diff
@@ -91,18 +80,19 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+				Scheme: r.URL.Scheme,
+				Host:   r.Host,
+				Path:   cmdReq.Path,
 			},
 		}
 
+		if cmdReq.Body != nil && len(cmdReq.Body) > 0 {
+			newReq.Body = io.NopCloser(bytes.NewBuffer(cmdReq.Body))
+		}
+
```
directionless, line 95: Hrm. This might be reasonable, but also seems like it merits a callout.

The post input to the middle way is this krypto format of `{box: <data>, future: stuff}`. But after we unpack it, we pass it with the body from the box. 

I think that's correct -- after all, what else would we do.

ee/localserver/request-query.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+func (ls *localServer) requestQueryHandler() http.Handler {
+	return http.HandlerFunc(ls.requestQueryHanlderFunc)
+}
+
+func (ls *localServer) requestQueryHanlderFunc(res http.ResponseWriter, r *http.Request) {
```
directionless, line 13:
```suggestion
	return http.HandlerFunc(ls.requestQueryHandlerFunc)
}

func (ls *localServer) requestQueryHandlerFunc(res http.ResponseWriter, r *http.Request) {
```

directionless: Usually signature is `w` and `r`. Or sometimes `resp` and `req` or `response` and `request`. But please don't mix.

ee/localserver/request-query.go
```diff
@@ -0,0 +1,49 @@
...
+		res.Write([]byte("no query key found in request body json"))
+		return
+	}
+
+	if query == "" {
+		res.Write([]byte("empty query"))
+		return
+	}
+
+	results, err := ls.querier.Query(query)
```
directionless, line 36: Do we need a timeout here? I forget if there's one implemented at a lower level. (I know the http server will time out, but will the query?)

ee/localserver/request-query.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func (ls *localServer) requestQueryHandler() http.Handler {
+	return http.HandlerFunc(ls.requestQueryHanlderFunc)
+}
+
+func (ls *localServer) requestQueryHanlderFunc(res http.ResponseWriter, r *http.Request) {
+	if r.Body == nil {
+		res.Write([]byte("request body is nil"))
+		return
```
directionless, line 16: For all of these, I think we should be returning a status. 

It might be worth making a helper along the lines of `sendClientError(w http.ResponseWriter, message string)`

ee/localserver/krypto-ec-middleware.go
James-Pickett: yea, I don't know what else we would do =)

I'll leave some comment explaining, but I think the v2CmdRequestType is pretty obvious

```go
type v2CmdRequestType struct {
	Path string
	Body []byte
}
```

ee/localserver/krypto-ec-middleware.go
James-Pickett: switch to any, just in case, add type check to string before decoding b64

ee/localserver/request-query.go
James-Pickett: fixed

ee/localserver/request-query.go
James-Pickett: added timeout ... unsure of reasonable duration interval, went with 1 sec && 250 milliseconds

ee/localserver/request-query.go
James-Pickett: done, returning bad request for errors

ee/localserver/krypto-ec-middleware.go
directionless: I think it's correct. The _outer_ body is our krypto protocol silliness, the inner body is whatever is being smuggled

directionless (approved):
Let's try. I think the timeout is probably going to need to change, but it's a find place to start.

(I also think we need to do timeout things down in the osquery SDK)

PR #1031, opened 2023-02-13
Integrating new icons into menu bar

### LINKED ISSUE #1027: Wire up new icons
Integrate the icons available here https://github.com/kolide/launcher/tree/main/ee/ui/assets so that systray can use them, and k2 can define menus with these icons.

seejdev:
Fixed #1027 

This replaces our menu bar icon implementation with one that chooses the appropriate icon based on what k2 is dictating.


ee/desktop/menu/menu_icon.go
```diff
@@ -0,0 +1,64 @@
...
+	default:
+		level.Debug(m.logger).Log(
+			"msg", "invalid icon",
+			"icon", icon)
+		return nil
+	}
+}
+
+func chooseIcon(darkIco, darkPng, lightIco, lightPng []byte) []byte {
+	if runtime.GOOS == "windows" {
```
directionless, line 49: This might be more clear with a switch statement.

ee/desktop/menu/menu_icon.go
```diff
@@ -0,0 +1,64 @@
...
+		)
+	default:
+		level.Debug(m.logger).Log(
+			"msg", "invalid icon",
+			"icon", icon)
+		return nil
+	}
+}
+
+func chooseIcon(darkIco, darkPng, lightIco, lightPng []byte) []byte {
```
directionless, line 48: Should we bring in monochrome as choice?

ee/desktop/menu/menu_icon.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"runtime"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/ee/ui/assets"
+)
+
+func (m *menu) getIcon(icon menuIcon) []byte {
+	switch icon {
+	case UnregisteredIcon:
```
directionless, line 12: I think we should name these akin to what the icons are named. `Translucent`, `Default`, `TriangleExclaimation`, and `CircleX`

directionless (commented):
Can you move the debug one to ee/ui/assets? We don't have a copy there

ee/desktop/menu/menu_icon.go
directionless: The monochrome ones are the alpha channel only ones, if we didn't have color. I'm not sure we'll ever chose them, but I feel like the hooks should be there

ee/desktop/menu/dark_mode_windows.go
```diff
@@ -0,0 +1,27 @@
...
+
+const (
+	regKey  = `Software\Microsoft\Windows\CurrentVersion\Themes\Personalize` // in HKCU
+	regName = `SystemUsesLightTheme`                                         // <- For taskbar & tray. Use AppsUseLightTheme for apps
+)
+
+func isDarkMode() bool {
+	k, err := registry.OpenKey(registry.CURRENT_USER, regKey, registry.QUERY_VALUE)
+	if err != nil {
+		log.Fatal(err)
```
James-Pickett, line 19: log.Fatal() will cause os.Exit(1). Are we sure we want this here? I guess if we couldn't read the registry we would end up in a crash loop.

If this was intentional, can you put some comments explaining why?

seejdev: oops no this was not intentional, I'll remove this

seejdev: fixed

James-Pickett (dismissed):

directionless (commented):
I'm real nervous about the light/dark detection. I think we should ship this without that, and then tinkering around.

ee/desktop/menu/dark_mode_darwin.go
```diff
@@ -0,0 +1,18 @@
...
+// +build darwin
+
+package menu
+
+import (
+	"os/exec"
+)
+
+func isDarkMode() bool {
+	cmd := exec.Command("defaults", "read", "-g", "AppleInterfaceStyle")
```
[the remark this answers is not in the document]
seejdev: In practice this would only called once as part of building the menu. The menu gets built only when either launcher desktop starts, or control server pushes a new menu/icon.

I borrowed this from some project I found, I'll look that up again. If there is a more formal API we can use, it's going to involve Objective-C.

ee/desktop/menu/menu.go
```diff
@@ -18,7 +18,10 @@ var InitialMenu []byte
 type menuIcon string
 
 const (
-	KolideDesktopIcon = "kolide-desktop"
+	Translucent         = "translucent"
```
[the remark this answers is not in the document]
seejdev: > I'm real nervous about the light/dark detection. I think we should ship this without that, and then tinkering around.

I have no problem with that - I can leave out the dark mode stuff for now, and default the icons to monochrome?

seejdev:
> I'm real nervous about the light/dark detection. I think we should ship this without that, and then tinkering around.

I have no problem with that - I can leave out the dark mode stuff for now, and default the icons to monochrome?

ee/desktop/menu/dark_mode_darwin.go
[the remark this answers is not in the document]
directionless: If I think through this, I think we have a bigger issue than it seems.

Users can swap light/dark, so we either need to poll for it, or we need a subscription style API. I don't see discussion about a subscription API, and polling comes with all sorts of downsides.

I do see some suggestions online about using a neutral menu, and then when the menu changes checking the light/dark state. But since I think folks will hang out in the yellow state, this doesn't wholly alleviate the need for polling.

Mostly this feels like a big can of worms.

directionless:
> > I'm real nervous about the light/dark detection. I think we should ship this without that, and then tinkering around.
> 
> I have no problem with that - I can leave out the dark mode stuff for now, and default the icons to monochrome?

I don't think it matters what we default to. Any of light/dark/monochrome seem fine. We should just pick something, and then dig into the light/dark, since I think it's a can of worms

ee/desktop/menu/dark_mode_darwin.go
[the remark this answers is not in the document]
seejdev: Honestly I think we'd be okay without monitoring light/dark changes, and just build the menu based on whatever the OS says at the time we ask.

Monitoring and reacting to changes is a bunch of complexity to handle what is effectively only a cosmetic bug, which will fix itself the next time launcher restarts. I have zero data to back up this claim, but I feel like users very rarely change their dark/light modes and, at least on Windows, the OS and first party applications themselves don't handle this edge case very well.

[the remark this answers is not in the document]
directionless: > Monitoring and reacting to changes is a bunch of complexity

Yes, that's exactly my concern.

On macOS it's common to let the OS swap between them based on time of day :|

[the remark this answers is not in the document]
seejdev: >On macOS it's common to let the OS swap between them based on time of day :|

oooh, great point. so more frequent than I made it out to be.

Either way, I've removed dark mode stuff from this PR, so we can tackle it separately.

ee/desktop/menu/menu.go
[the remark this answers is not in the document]
seejdev: Fixed!

ee/desktop/menu/menu.go
```diff
@@ -96,7 +99,7 @@ func (m *menu) getMenuData() *MenuData {
 // SetDefaults ensures we have the desired default values.
 func (md *MenuData) SetDefaults() {
 	if md.Icon == "" {
-		md.Icon = KolideDesktopIcon
+		md.Icon = TranslucentIcon
```
directionless, line 102: I'd have expected `DefaultIcon` here (and maybe translucent in the initial)

directionless (approved):

ee/desktop/menu/dark_mode_darwin.go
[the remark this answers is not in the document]
directionless: Thank you!

ee/desktop/menu/menu.go
seejdev: Sure, I wasn't exactly sure what was best choice here.

directionless (dismissed):

RebeccaMahany (dismissed):

RebeccaMahany (approved):

directionless (approved):

directionless:
Copying from slack, CJ changed the whitespace on the icons. Something like `for f in menubar-default-*.png; do convert "$f" -gravity center -crop 192x192+0+0 +repage "$f"; done`

PR #1032, opened 2023-02-14
Add local key info to initial enrollment request

directionless:


pkg/osquery/extension.go
```diff
@@ -1049,6 +1049,21 @@ func getEnrollDetails(client Querier) (service.EnrollmentDetails, error) {
 	details.GOOS = runtime.GOOS
 	details.GOARCH = runtime.GOARCH
 
+	// Pull in some launcher key info. These depend on the agent package, and we'll need to check for nils
+	if agent.LocalDbKeys().Public() != nil {
+		if key, err := x509.MarshalPKIXPublicKey(agent.LocalDbKeys().Public()); err == nil {
+			// der is a binary format, so convert to b64
+			details.LauncherLocalKey = base64.StdEncoding.EncodeToString(key)
```
RebeccaMahany, line 1056: I believe you should have `echelper.PublicEcdsaToB64Der(localDbKeys.Public().(*ecdsa.PublicKey))` available now to skip having to call both marshal + encode here and below 🙂

RebeccaMahany (approved):

pkg/osquery/extension.go
directionless: Yeah... It buffered the string out to make a `[]byte` and it felt weird casting it back again.

PR #1033, opened 2023-02-14
Enable non-prod menu items in pre-prod environments

seejdev:
This will enable menu items marked as non-prod-only to show up in launchers running against pre-prod.

### DIFF
--- a/ee/desktop/menu/menu_systray.go
+++ b/ee/desktop/menu/menu_systray.go
@@ -109,7 +109,7 @@ func (m *menu) cleanup() {
 
 // Returns true if launcher is running in production
 func (m *menu) isProd() bool {
-	return m.hostname == "k2device-preprod.kolide.com" || m.hostname == "k2device.kolide.com"
+	return m.hostname == "k2device.kolide.com"
 }
 
 // makeActionHandler creates a handler to execute the desired action when a menu item is clicked


ee/desktop/menu/menu_systray.go
```diff
@@ -109,7 +109,7 @@ func (m *menu) cleanup() {
 
 // Returns true if launcher is running in production
 func (m *menu) isProd() bool {
-	return m.hostname == "k2device-preprod.kolide.com" || m.hostname == "k2device.kolide.com"
+	return m.hostname == "k2device.kolide.com"
```
directionless, line 112: I'm not sure if we want this. 

`k2device-preprod.kolide.com` is prod (As compared to localhost development and review apps / staging)

directionless: Things that are kolide-production (k2device-preprod) vs customers (k2device) should probably be detected differently. Maybe depends what you're looking for

James-Pickett: agree, I think `isProd` and `shouldShouldSpawnDesktop` are 2 different things, at least until we release desktop into the wild.

seejdev: My intention was to enable things like debug menu items that we only want shown to internal Kolide launchers. This could be done via control server pushing unique menus for pre-prod instances, but this approach also allows us to have non-prod items in the default menu.

seejdev: >agree, I think isProd and shouldSpawnDesktop are 2 different things

I can get behind that, if we don't have enough information local to launcher for it to answer "Should I spawn desktop?" without having to talk to control server and ask, then I think I should remove the `nonProdOnly` concept because k2 will already know and doesn't need it conditionalized within launcher.

directionless: I think the control server should push them. This is pretty easy on the K2 side of things. 

I think the default menu is going to be something very basic like "Waiting to connect to Kolide", and then after that it's all pushed. I expect to get updates very quickly

PR #1035, opened 2023-02-14
Spawn launcher desktop for human console users only

### LINKED ISSUE #1034: Launcher can create two desktop processes after macOS software update
This has been observed at least twice, and the common thread was launcher running immediately after a macOS upgrade.

### Symptoms

Two desktop processes

```
    0   953   769   0 10:20AM ??         0:00.46 /usr/local/kolide-k2/bin/launcher-updates/1676352633/Kolide.app/Contents/MacOS/launcher desktop
    0  1127   769   0 10:20AM ??         0:00.50 /usr/local/kolide-k2/bin/launcher-updates/1676352633/Kolide.app/Contents/MacOS/launcher desktop
```

Two systray icons, potentially using different themes, light and dark:

image

Two launcher user dirs created:

```
drwx------    3 cj             wheel        96 Feb 14 10:20 desktop_501
drwx------    3 _windowserver  wheel        96 Feb 14 10:20 desktop_88
```

seejdev:
Fixed #1034 

I haven't been able to actually test this, and that will take a non-trivial amount of time and effort given that this bug is related to OS upgrades.

However, indications point towards this bug being caused by a console user called `_windowsserver` which normally isn't logged in, but will be logged in after the first boot post-upgrade of macOS. This was causing us to create more than one desktop process, since we interpreted the `_windowsuser` user and the actual human user as two different users.

The change here restricts desktop launcher spawning for console users with UID >= 501, which is the minimum UID for human console users.

### DIFF
--- a/ee/consoleuser/consoleuser_darwin.go
+++ b/ee/consoleuser/consoleuser_darwin.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


seejdev:
lmk if anyone has suggestions on how best to test this - this is hard to automate

directionless (approved):

PR #1036, opened 2023-02-14
Icons with moar pixels

seejdev:
Two changes:

- I found that when imagemagick converted the SVGs to .icos, we ended up with white squares particularly for dark mode icons. Changing the generator to convert from PNGs to .icos preserves the transparency
- Added some new icons, and now all icons are 256x256


James-Pickett (dismissed):

ee/ui/assets/assets.go
```diff
@@ -19,9 +19,21 @@ var MenubarCircleXLightmodeIco []byte
...
+var MenubarCircleXMonochromeDarkmodePng []byte
+
 //go:embed menubar-circle-x-monochrome.ico
 var MenubarCircleXMonochromeIco []byte
 
+//go:embed menubar-circle-x-monochrome-lightmode.ico
+var MenubarCircleXMonochromeLightmodeIco []byte
+
+//go:embed menubar-circle-x-monochrome-lightmode.png
+var MenubarCircleXMonochromeLightmodePng []byte
```
directionless, line 35: I'm a little confused by this. I thought we didn't need the dark/light monochrome versions. And if we do have them, we do we still have `menubar-circle-x-monochrome` ?

directionless (commented):
Can we ditch the svgs?

ee/ui/assets/assets.go
seejdev: Hmm I'm not entirely sure, I just dropped the new set of icons in and hit generate. Maybe Fritz renamed some?

seejdev: To better answer your question - yes, for macOS we would only need one monochrome version which can be used as a template image for both dark and light modes. Not sure if Fritz had anything else in mind.

I can remove the `-monochrome-dark/light` ones and leave just the `-monochrome` ones if we don't see any need for them right now.

directionless: I'd bet Fritz gave us more than we need

directionless: (mostly I'm trying minimize the size we use)

seejdev: Removed  the `-monochrome-dark/light` ones, regenerated.

RebeccaMahany (approved):

PR #1037, opened 2023-02-15
Fix some things in the menu relating to io.readers

directionless:


directionless:
`io.Readers` are good and all, but this PR is out of date.

PR #1038, opened 2023-02-15
Fix double read from an io.Reader inside the menu generation

directionless:
Inside the desktop runner's menu handling code, there was a double read of an io.Reader. This causes the second one to fail. The first one looks like it was from an earlier refactor, I removed it. 

Also adds another debugging flag and squelches a log

### DIFF
--- a/ee/desktop/runner/runner.go
+++ b/ee/desktop/runner/runner.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -272,6 +273,10 @@ func (r *DesktopUsersProcessesRunner) SendNotification(title, body string) error
 
 // Update handles control server updates for the desktop-menu subsystem
 func (r *DesktopUsersProcessesRunner) Update(data io.Reader) error {
+	if data == nil {
+		return errors.New("data is nil")
+	}
+
 	var dataCopy bytes.Buffer
 	dataTee := io.TeeReader(data, &dataCopy)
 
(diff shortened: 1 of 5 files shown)


ee/desktop/runner/runner.go
```diff
@@ -272,26 +273,24 @@ func (r *DesktopUsersProcessesRunner) SendNotification(title, body string) error
...
 
 	var dataCopy bytes.Buffer
 	dataTee := io.TeeReader(data, &dataCopy)
 
-	if err := json.NewDecoder(dataTee).Decode(&menu); err != nil {
-		if agent.Flags.DebugServerData() {
+	// Regardless, we will write the menu data out to a file that can be grabbed by
+	// any desktop user processes, either when they refresh, or when they are spawned.
+	if err := r.generateMenuFile(dataTee); err != nil {
+		if agent.Flags.DebugServerData() || true {
```
seejdev, line 286: This is always true?

directionless: Ha. Good catch
```suggestion
		if agent.Flags.DebugServerData() {
```

pkg/log/log.go
```diff
@@ -62,12 +62,13 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
...
+		return len(p), nil
 	}
 
 	msg := strings.TrimSpace(string(p))
 	caller := extractOsqueryCaller(msg)
 	if err := lf(l.logger).Log(append(l.extraKeyVals, "msg", msg, "caller", caller)...); err != nil {
-		return 0, err
+		return len(p), err
 	}
 	return len(p), nil
```
James-Pickett, line 73: both of the returns are returning the same, could be simplified to

```suggestion
return lf(l.logger).Log(append(l.extraKeyVals, "msg", msg, "caller", caller)...)
```

James-Pickett (dismissed):

seejdev (dismissed):

pkg/log/log.go
directionless: Good call. Though I opted to back this out.

James-Pickett (approved):

PR #1050, opened 2023-02-22
Monitor correct parent process

RebeccaMahany:
Pass in PPID as env var for monitoring.

### DIFF
--- a/cmd/launcher/desktop.go
+++ b/cmd/launcher/desktop.go
@@ -88,7 +88,7 @@ func runDesktop(args []string) error {
 		)
 	}
 
-	if *flPpid == 0 {
+	if *flPpid <= 1 {
 		ppid, err := parentProcessId()
 		if err != nil {
 			return fmt.Errorf("ppid not included in flags and could not be looked up: %w", err)
--- a/cmd/launcher/desktop.go
+++ b/cmd/launcher/desktop.go
@@ -17,7 +17,6 @@ import (
 	"github.com/kolide/launcher/ee/desktop/menu"
 	"github.com/kolide/launcher/ee/desktop/server"
 	"github.com/kolide/launcher/pkg/agent"
-	"github.com/mitchellh/go-ps"
 	"github.com/oklog/run"
 	"github.com/peterbourgon/ff/v3"
 	"github.com/shirou/gopsutil/process"
@@ -56,6 +55,11 @@ func runDesktop(args []string) error {
 			"",
 			"path to icon file",
 		)
+		flPpid = flagset.Int(
+			"ppid",
+			0,
+			"parent process ID to monitor",
+		)
 	)
 
 	if err := ff.Parse(flagset, args, ff.WithEnvVarNoPrefix()); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/launcher/desktop.go
```diff
@@ -186,6 +193,22 @@ func monitorParentProcess(logger log.Logger) {
...
 
+func parentProcessId() (int, error) {
+	if runtime.GOOS != "darwin" {
+		return os.Getppid(), nil
+	}
+
+	// On macOS, we run desktop as `launchctl asuser <UID> sudo --preserve-env -u <username> launcher desktop`, which
+	// means the parent process of the current process is the sudo process. We need to monitor one level above that
+	// to be monitoring the launcher root process.
+	sudoProcess, err := ps.FindProcess(os.Getppid())
```
directionless, line 204: I think this will be racy around updates.

Can we pass the parent process id in as a command line arg or env?

James-Pickett, line 204: could use `process.NewProcess()` (from `github.com/shirou/gopsutil/process` already imported) rather than importing an additional pkg

James-Pickett: I am in favor of passing the pid as en env var. However, I just want to call out that this really is a back up plan for when root launcher crashes or gets manually killed. During a normal shut down like an auto update, desktop is shut down via a desktop server http cmd.

RebeccaMahany: Yup, will update 👍

RebeccaMahany: Oh nice, thank you!

cmd/launcher/desktop.go
```diff
@@ -83,6 +88,18 @@ func runDesktop(args []string) error {
 		)
 	}
 
+	if *flPpid <= 1 {
+		ppid, err := parentProcessId()
```
James-Pickett, line 92: I think we should log here if the `flPpid` was not provided

RebeccaMahany: Agreed! Added

RebeccaMahany: Sample logs after update if PPID flag is omitted:

```
{"caller":"runner.go:647","component":"desktop_users_processes_runner","level":"info","msg":"{\"caller\":\"desktop.go:92\",\"fl_ppid\":0,\"msg\":\"expected PPID command-line flag for launcher desktop but did not receive it\",\"pid\":10985,\"severity\":\"warn\",\"subprocess\":\"desktop\",\"ts\":\"2023-02-22T20:43:27.218669Z\",\"uid\":\"501\"}","subprocess":"desktop","ts":"2023-02-22T20:43:27.218759Z","uid":"501"}
{"caller":"runner.go:647","component":"desktop_users_processes_runner","level":"info","msg":"{\"caller\":\"logutil.go:13\",\"err\":\"run with positional args: running subcommand desktop: valid ppid not included in flags and could not be looked up: could not find parent process for sudo launcher desktop: could not find parent line\",\"severity\":\"info\",\"ts\":\"2023-02-22T20:43:27.244036Z\"}","subprocess":"desktop","ts":"2023-02-22T20:43:27.244158Z","uid":"501"}
```

cmd/launcher/desktop.go
RebeccaMahany: Oh, interesting -- discovered when testing the new logging that gopsutil/process isn't able to find the parent process whereas go-ps could -- looks like the two packages use different methods to find the parent. In light of that, I'm removing this method to rely just on the PPID command-line flag.

cmd/launcher/desktop.go
RebeccaMahany: After latest commit:

```
{"caller":"runner.go:647","component":"desktop_users_processes_runner","level":"info","msg":"{\"caller\":\"logutil.go:13\",\"err\":\"run with positional args: running subcommand desktop: received invalid PPID command-line flag for launcher desktop: 0\",\"severity\":\"info\",\"ts\":\"2023-02-22T20:52:08.481721Z\"}","subprocess":"desktop","ts":"2023-02-22T20:52:08.481866Z","uid":"501"}
```

James-Pickett (approved):

PR #1051, opened 2023-02-22
proof-of-concept for mockoon-based integration test in pipeline

RebeccaMahany:
See `Integration test (ubuntu only)` here: https://github.com/kolide/launcher/actions/runs/4247187813/jobs/7384925543

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -100,6 +100,18 @@ jobs:
           throw "launcher.exe --version failed"
         }
 
+    - name: Integration test (ubuntu only)
+      if: ${{ contains(matrix.os, 'ubuntu') }}
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -100,6 +100,18 @@ jobs:
           throw "launcher.exe --version failed"
         }
 
+    - name: Integration test (ubuntu only)
+      if: ${{ contains(matrix.os, 'ubuntu') }}
+      run: |
+        sudo apt-get -y install nodejs npm
+        sudo npm install -g n
+        sudo n lts
```
directionless, line 108: Why not https://github.com/actions/setup-node

directionless:
I assume we've gone in a different direction. But feel free to open/revisit

Issue #1041, opened 2023-02-17
We should add our cert bundle to normal launcher and interactive usage


directionless:
osquery does not natively recognize the macOS root certificates. When I implemented osquery transport (#797) I added some logic to handle this in: https://github.com/kolide/launcher/blob/7b84aaaf112097ababf2a53bd238c42eff385742/cmd/launcher/extension.go#L231-L238

However, this logic is not applied to the normal launcher runtime, nor the `interactive` command. Which, ultimately, causes usage of the curl table to fail. 

We should apply similar logic on those code paths.

RebeccaMahany:
I think we fixed this in https://github.com/kolide/launcher/issues/2318 and https://github.com/kolide/launcher/issues/2350

PR #1054, opened 2023-02-23
adds default certs to interactive mode if no cert set as flag

James-Pickett:
Relates to #1041


directionless (approved):
Why limit this to interactive?

directionless:
> Why limit this to interactive?

Oh, of course, because interactive uses a totally different code path.

seejdev (approved):

Issue #1056, opened 2023-02-24
Table Request: software BOM via syft


directionless:
Some time ago I encountered https://github.com/anchore/syft and https://github.com/anchore/grype

The first of those is a general purpose software BOM. I think it would be really interesting to link that into launcher, so we could pull general SBOM info from files.

directionless:
We tested this, and decided it carried too many dependencies. Closing for now.

PR #1059, opened 2023-02-27
adds syft table to launcher

James-Pickett:
closes https://github.com/kolide/launcher/issues/1056


go.mod
```diff
@@ -32,85 +32,161 @@ require (
...
 
 require (
+	github.com/anchore/syft v0.73.0
 	github.com/kolide/systray v1.10.2
-	github.com/vmihailenco/msgpack/v5 v5.3.5
 )
 
 require (
 	github.com/BurntSushi/toml v1.1.0 // indirect
+	github.com/CycloneDX/cyclonedx-go v0.7.1-0.20221222100750-41a1ac565cce // indirect
```
James-Pickett, line 64: that's a lot of additionally dependency 🤔

directionless: Yeah.... Any idea what it does to the resultant size? I don't love it, but it might just be the cost of `syft`

pkg/osquery/tables/syft/table.go
```diff
@@ -0,0 +1,115 @@
...
+		return results, nil
+	}
+
+	for _, filePath := range filePaths {
+		for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
+			flattenOpts := []dataflatten.FlattenOpts{
+				dataflatten.WithQuery(strings.Split(dataQuery, "/")),
+			}
+
+			src, srcCleanup := source.NewFromFile(filePath)
```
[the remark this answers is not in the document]
seejdev: Does syft provide any tuning of resources or memory footprint? Maybe there's some knobs we can turn. Although this should happen infrequently, I can imagine a customer balking at a background service being that resource intensive.

pkg/osquery/tables/syft/table_test.go
```diff
@@ -0,0 +1,95 @@
...
+	if runtime.GOOS == "windows" {
+		executablePath = fmt.Sprintf("%s.exe", executablePath)
+	}
+
+	// due to flakey tests we are tracking the time it takes to build and attempting emit a meaningful error if we time out
+	timeout := time.Second * 60
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, "go", "build", "-o", executablePath, "../../../../cmd/launcher")
```
directionless, line 82: Is launcher going to take longer to build? We could also just build this test file. Or ship a trivial binary to build.

pkg/osquery/tables/syft/table.go
```diff
@@ -0,0 +1,115 @@
...
+			if srcCleanup != nil {
```
seejdev, line 61: Curious what cases does `source.NewFromFile` return nil?

pkg/osquery/tables/syft/table.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	filePaths := tablehelpers.GetConstraints(queryContext, "path")
+
+	if len(filePaths) == 0 {
+		level.Info(t.logger).Log(
+			"msg", fmt.Sprintf("no path provided to %s", tableName),
+			"table", tableName,
```
directionless, line 49:
```suggestion
			"msg", "no path provided",
			"table", tableName,
```

pkg/osquery/tables/syft/table.go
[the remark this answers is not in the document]
directionless: Not sure either. I think it's probably fine to leave it as-is.

pkg/osquery/tables/syft/table.go
```diff
@@ -0,0 +1,115 @@
...
+			flattenOpts := []dataflatten.FlattenOpts{
+				dataflatten.WithQuery(strings.Split(dataQuery, "/")),
+			}
+
+			src, srcCleanup := source.NewFromFile(filePath)
+			if srcCleanup != nil {
+				defer srcCleanup()
+			}
+
+			packageCatalog, relationships, linuxDistro, err := syft.CatalogPackages(&src, cataloger.DefaultConfig())
```
seejdev, line 65: Is `linuxDistro` typically an empty string on macOS/Windows?

go.mod
seejdev: I suppose that's due to the nature of syft, but yeah... hmmm

directionless (commented):
This is simpler than I expected. Nice

pkg/osquery/tables/syft/table.go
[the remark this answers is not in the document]
directionless: I don't think we're likely to write a check that uses this. I think it'll show up occasionally in "what is happening in this weird thing" or in LQ. And in those cases, it's what it is. But we should be cognizant of the cost.

RebeccaMahany (dismissed):
LGTM assuming we decide to not try to tune the search to consume fewer resources -- this looks super cool!

go.mod
James-Pickett: It nearly doubles the size of the build =( 

18.1 MB to 30.3 MB on MacOS

Additionally, there is some issues with building for windows in CI https://github.com/kolide/launcher/actions/runs/4308325630/jobs/7514473491

Not sure if the juice is worth the squeeze here.

seejdev: Wow, I would have not expected such a large increase

James-Pickett:
Closing this due to drastic increase in binary size, reach out if you would like to discuss.

directionless:
> Closing this due to drastic increase in binary size, reach out if you would like to discuss.

Sadly, I think this is the right call for now. It's a substantial increase, and I don't see a simple way to shrink it. 

`upx` is a thing we could look at. But I've never been sure.

Issue #1016, opened 2023-02-07
Replace bbolt DB usage in tests with an in-memory implementation


seejdev:
This has been discussed before.... our tests can be flakey and there is great concern that bbolt is a culprit involved.

Enabling our tests to avoid having to use a concrete bbolt implementation, and instead use an in-memory one would:

- Presumably speed up those tests, since the memory operations will be much faster than disk access
- Should improve performance within Github CI, since disk access is a bottleneck and can cause jobs to timeout unpredictably

The sequence here would be:

- [ ] Create an in-memory key-value store implementation
- [ ] Consolidate as much bbolt implementation code into a single key-value store implementation
- [ ] Replace non osquery/extension.go `*bbolt.DB` references with [key-value interfaces](https://github.com/kolide/launcher/blob/main/pkg/agent/keyvalue_store.go) 
- [ ] Replace osquery/extension.go (including launcher tables) `*bbolt.DB` references with key-value interfaces
- [ ] Make tests instantiate and use the in-memory implementation, replacing the `*bbolt.DB` usage

directionless:
Sadly, bolt db doesn't support this. https://github.com/etcd-io/bbolt/issues/227 is the usual thread about it

directionless:
Don't need this issue.

PR #1062, opened 2023-02-28
bbolt begone: Phase 1

seejdev:
Related to #1016 

This is the first phase of abstracting away our bbolt dependency. This is a large effort and although there is plenty of bbolt code still left to touch, I felt like this was a good point to submit a cohesive change without it becoming too overwhelming.

This PR has several changes:
- Introduces a set of interfaces which define generic key-value store operations
- Integrates these interfaces into places where I could fairly easily replace `*bbolt.DB` references and imports
- Refactors the BucketConsumer type and several other places where we repeated ourselves with similar bbolt implementation code, and consolidates this logic into one bbolt implementation file
- Introduces a simple in-memory key-value store which implements these same interfaces
- Enables some tests to use the in-memory key-value store implementation instead of bbolt when running in CI

These changes have some nice immediate benefits:
- I was able to remove the `import "go.etcd.io/bbolt"` dependency from nine files! These areas can now effectively be used with any concrete key-value store implementation which satisfies the key-value interfaces.
- Anything in the future that needs to use bbolt, doesn't need to write new bbolt code. The basic key-value operations are available for use to any clients and we don't need to reinvent it any further.

What this PR _does not_ do:

- I have purposefully left out osquery packages, the launcher tables which use bbolt, and the [extension.go](https://github.com/kolide/launcher/blob/main/pkg/osquery/extension.go) file. This area uses bbolt in more complex ways and will require some additional interfaces and code to truly abstract bbolt out. This area is next up for converting to a bbolt-less implementation.


pkg/agent/storage/keyvalue_store_ci.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/stretchr/testify/require"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	dbTestFileName = "test.db"
+)
+
+func NewCIKeyValueStore(t *testing.T, logger log.Logger, bucketName string) types.KVStore {
+	if os.Getenv("CI") == "true" {
```
seejdev, line 19: Not sure how folks feel about this - we would run against bbolt locally, but forgo bbolt in GitHub CI. There's probably an improvement here to be made, so we don't accidentally regress something in the bbolt implementation and not notice it in the CI runs.

pkg/agent/storage/keyvalue_store_bbolt_test.go
```diff
@@ -93,8 +92,8 @@ func Test_Updates(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			db := createDb(t)
-			bc := NewBucketConsumer(log.NewNopLogger(), db, tt.name)
```
seejdev, line 97: I'm going to refactor this test so it's agnostic to the key value store type. Then it can be run against each concrete key-value store implementation, since they all should behave the same from a client's perspective.

pkg/agent/keys.go
```diff
@@ -59,73 +59,44 @@ func SetupKeys(logger log.Logger, db *bbolt.DB) error {
...
-			return nil
-		}
-
-		_ = b.Delete([]byte(privateEccData))
-		_ = b.Delete([]byte(publicEccData))
-
-		return nil
-	})
+	_ = deleter.Delete([]byte(privateEccData))
+	_ = deleter.Delete([]byte(publicEccData))
```
directionless, line 101: This is now small enough you could probably drop `clearKeyData` and just inline the `Delete` calls. But 🤷 that's a bigger change

directionless (commented):
I like this.

Though I think the the bolt and memory things should be in their own packages. With no strong feelings, I'd suggest `pkg/agent/storage/bbolt` and `agentbbolt.NewStore` or something

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,175 @@
...
+func (s *bboltKeyValueStore) Get(key []byte) (value []byte, err error) {
+	if s == nil || s.db == nil {
+		return nil, errors.New("db is nil")
+	}
+
+	if err := s.db.View(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(s.bucketName))
+		if b == nil {
+			return fmt.Errorf("%s bucket not found", s.bucketName)
+		}
```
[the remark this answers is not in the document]
seejdev: I like the idea of putting it in New, and it avoids having these methods have the odd side-effect of creating a bucket.

pkg/agent/types/keyvalue_store.go
```diff
@@ -0,0 +1,56 @@
...
+	// If the key exist then its previous value will be overwritten.
+	// Returns an error if the key is blank, if the key is too large, or if the value is too large.
+	Set(key, value []byte) error
+}
+
+// Deleter is an interface for deleting data in a key/value store.
+type Deleter interface {
+	// Delete removes a key.
+	// If the key does not exist then nothing is done and a nil error is returned.
+	Delete(key []byte) error
```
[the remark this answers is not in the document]
seejdev: Makes sense!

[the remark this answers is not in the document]
seejdev: Fixed!

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,175 @@
...
+		db:         db,
+		bucketName: bucketName,
+	}
+
+	return m
+}
+
+func (s *bboltKeyValueStore) Get(key []byte) (value []byte, err error) {
+	if s == nil || s.db == nil {
+		return nil, errors.New("db is nil")
```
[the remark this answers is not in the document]
seejdev: Custom `storage` errors could be helpful. May be worth doing that now. I'll check it out

pkg/agent/storage/keyvalue_store_ci.go
seejdev: I've seen it be done, like in the fyne codebase they have some files with
```
//go:build ci
// +build ci
```
It doesn't look like we've done that anywhere, I assume that will require some changes to make to pipe that build var down.

directionless: That looks like it's adding in a `ci` build tag. Can we use the existing `test` build tag that's implicit for the test suite? Partly, if feels very weird pulling `testing` in to prod builds

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -253,16 +219,8 @@ func (nc *NotificationConsumer) cleanup() {
...
-				level.Debug(nc.logger).Log("msg", "could not delete old notification from bucket", "key", string(k), "err", err)
-			}
-		}
-
-		return nil
-	}); err != nil {
-		level.Debug(nc.logger).Log("msg", "could not delete old notifications from bucket", "err", err)
+	if err := nc.store.Delete(keysToDelete...); err != nil {
+		// Log, but don't error, since we might be able to delete some others
+		level.Error(nc.logger).Log("msg", "could not delete old notifications from bucket", "err", err)
```
RebeccaMahany, line 224: Since this isn't looping over the keys anymore, I think we could return the error now

pkg/agent/keys.go
```diff
@@ -59,73 +59,44 @@ func SetupKeys(logger log.Logger, db *bbolt.DB) error {
...
-			return nil
-		}
-
-		_ = b.Delete([]byte(privateEccData))
-		_ = b.Delete([]byte(publicEccData))
-
-		return nil
-	})
+	keys := [][]byte{[]byte(privateEccData), []byte(publicEccData)}
+	_ = deleter.Delete(keys...)
```
RebeccaMahany, line 101: I think you can simplify and avoid putting the keys into an array just to unpack it:

```
_ = deleter.Delete([]byte(privateEccData), []byte(publicEccData))
```

?

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,188 @@
...
+
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	m := &bboltKeyValueStore{
+		logger:     log.With(logger, "bucket", bucketName),
```
RebeccaMahany, line 35: Nice

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return s.db.Update(func(tx *bbolt.Tx) error {
+		b := tx.Bucket([]byte(s.bucketName))
+		if b == nil {
+			return fmt.Errorf("%s bucket does not exist", s.bucketName)
+		}
+
+		if value != nil {
+			if err := b.Put(key, value); err != nil {
+				return err
```
RebeccaMahany, line 76: Potentially worth wrapping this err with the key included? 🤷‍♀️  Same q in `Delete` below

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,188 @@
...
+			return fmt.Errorf("%s bucket does not exist", s.bucketName)
+		}
+
+		for key, value := range kvPairs {
+			if err := b.Put([]byte(key), []byte(value)); err != nil {
+				// Log errors but continue processing the remaining key-values
+				level.Error(s.logger).Log(
+					"msg", "failed to store key-value in bucket",
+					"key", key,
+					"value", value,
```
RebeccaMahany, line 147: Is the value always safe to log here?

pkg/agent/storage/keyvalue_store_bbolt.go
```diff
@@ -0,0 +1,175 @@
...
+
+		if err := b.ForEach(fn); err != nil {
+			return fmt.Errorf("error iterating over keys in bucket: %w", err)
+		}
+
+		return nil
+	})
+}
+
+func (s *bboltKeyValueStore) Update(data io.Reader) error {
```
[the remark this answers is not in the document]
RebeccaMahany: This tripped me up a tiny bit as well, but I think I can also go with slightly awkward 🙂

RebeccaMahany (commented):
This is nice! Feels very clean.

pkg/agent/storage/keyvalue_store_bbolt.go
directionless: Good question. Value might be sensitive or huge. Probably better to omit

pkg/agent/storage/keyvalue_store_ci.go
directionless: Eh, might not ultimately be worth it.

You know... If this was in it's own `pkg` it wouldn't get linked into production builds.

pkg/agent/storage/keyvalue_store_bbolt_test.go
```diff
@@ -1,10 +1,9 @@
-package control
+package storage
```
directionless, line 1: I think the bbolt, memory, and ci implementations should land in their own pkg directories.

pkg/agent/storage/keyvalue_store_test.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+
+				if !tt.expectedErr {
+					for k, v := range tt.gets {
+						val, err := s.Get([]byte(k))
+						require.NoError(t, err)
+						assert.Equal(t, v, string(val))
+					}
+				}
+			}
```
James-Pickett, line 73: might want to test the thread safety of the implementations like:

```suggestion
                  for _, s := range getStores(t) {
				wg := sync.WaitGroup{}
				for k, v := range tt.sets {
					wg.Add(1)
					go func() {
						defer wg.Done()
						err := s.Set([]byte(k), []byte(v))
						if tt.expectedErr {
							require.Error(t, err)
						} else {
							require.NoError(t, err)
						}
					}()
					wg.Wait()
				}

				if !tt.expectedErr {
					wg.Add(1)
					go func() {
						defer wg.Done()
						for k, v := range tt.gets {
							val, err := s.Get([]byte(k))
							require.NoError(t, err)
							assert.Equal(t, v, string(val))
						}
					}()
					wg.Wait()
				}
			}
```

pkg/agent/storage/keyvalue_store_test.go
```diff
@@ -0,0 +1,140 @@
...
+			t.Parallel()
+
+			for _, s := range getStores(t) {
+				for k, v := range tt.sets {
+					err := s.Set([]byte(k), []byte(v))
+					if tt.expectedErr {
+						require.Error(t, err)
+					} else {
+						require.NoError(t, err)
+					}
```
James-Pickett, line 63: for some reason I can't stand `else`s =)
```suggestion
					if tt.expectedErr {
						require.Error(t, err)
						continue
					}
					require.NoError(t, err)
```

pkg/agent/storage/keyvalue_store_test.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+
+				if !tt.expectedErr {
+					for k, v := range tt.gets {
+						val, err := s.Get([]byte(k))
+						require.NoError(t, err)
+						assert.Equal(t, v, string(val))
+					}
+				}
```
James-Pickett, line 72: probably worth testing for thread safety of the store implementations

something like:
```suggestion
			for _, s := range getStores(t) {
				wg := sync.WaitGroup{}
				for k, v := range tt.sets {
					k, v := k, v

					wg.Add(1)
					go func() {
						defer wg.Done()
						err := s.Set([]byte(k), []byte(v))
						if tt.expectedErr {
							require.Error(t, err)
						} else {
							require.NoError(t, err)
						}
					}()

				}

				wg.Wait()

				if !tt.expectedErr {
					wg.Add(1)

					go func() {
						defer wg.Done()
						for k, v := range tt.gets {
							val, err := s.Get([]byte(k))
							require.NoError(t, err)
							assert.Equal(t, v, string(val))
						}
					}()
				}

				wg.Wait()
			}
```

ee/control/consumers/notificationconsumer/notify_consumer.go
seejdev: I could have `cleanup()` return an error, but since it's run periodically I think the error will just get logged like this (either here or in `runCleanup`). I suppose the comment here is inaccurate now though...

pkg/agent/storage/keyvalue_store_bbolt.go
seejdev: Agree!

pkg/agent/storage/keyvalue_store_bbolt_test.go
seejdev: Done!

pkg/agent/storage/keyvalue_store_bbolt.go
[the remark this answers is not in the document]
seejdev: Added custom errors for nil databases and nonexistent buckets.

ee/control/consumers/notificationconsumer/notify_consumer.go
RebeccaMahany: I'm fine with just removing the comment, then!

directionless (approved):

ee/control/consumers/notificationconsumer/notify_consumer.go
seejdev: Cool, I'll adjust it in the next PR!

PR #1064, opened 2023-03-01
adds endpoint to run scheduled queries on demand over local server

James-Pickett:


ee/localserver/request-query.go
```diff
@@ -62,6 +64,79 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+		return err
+	}, 1*time.Second, 250*time.Millisecond)
+
+	if err != nil {
+		sendClientError(w, fmt.Sprintf("error executing query for scheduled query: %s, error: %s", scheduledQueryQuery, err))
+		return
+	}
+
+	if len(results) == 0 {
+		sendClientError(w, fmt.Sprintf("no query found with name '%s' using query %s", name, scheduledQueryQuery))
```
James-Pickett, line 115:
```suggestion
		sendClientError(w, fmt.Sprintf("no query found with name '%s' using \"%s\"", name, scheduledQueryQuery))
```

ee/localserver/request-query.go
```diff
@@ -62,6 +64,79 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+
+	var results []map[string]string
+	var err error
+	backoff.WaitFor(func() error {
+		results, err = ls.querier.Query(scheduledQueryQuery)
+		return err
+	}, 1*time.Second, 250*time.Millisecond)
+
+	if err != nil {
+		sendClientError(w, fmt.Sprintf("error executing query for scheduled query: %s, error: %s", scheduledQueryQuery, err))
```
James-Pickett, line 110:
```suggestion
		sendClientError(w, fmt.Sprintf("error executing query for scheduled query using \"%s\", error: %s", scheduledQueryQuery, err))
```

ee/localserver/request-query.go
```diff
@@ -62,6 +64,79 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+	if !ok {
+		sendClientError(w, fmt.Sprintf("no query found with name '%s' using query \"%s\"", name, scheduledQueryQuery))
+		return
+	}
+
+	newBody := map[string]string{
+		"query": query,
+	}
+
+	jsonBytes, err := json.Marshal(newBody)
```
James-Pickett, line 129: I think we can drop the check for "query" since thats the only column requested, the length check earlier should handle it, so we can simplify to
```suggestion
	jsonBytes, err := json.Marshal(results[0)
```

ee/localserver/request-query.go
```diff
@@ -19,48 +21,89 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+		return
+	}
+
+	w.Write(jsonBytes)
+}
+func (ls *localServer) requestRunScheduledQueryHandler() http.Handler {
+	return http.HandlerFunc(ls.requestRunScheduledQueryHanlderFunc)
+}
+
+func (ls *localServer) requestRunScheduledQueryHanlderFunc(w http.ResponseWriter, r *http.Request) {
```
RebeccaMahany, line 60: Nit -- there's a typo in `Handler` 🙂

ee/localserver/request-query.go
```diff
@@ -19,48 +21,89 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+	}, 1*time.Second, 250*time.Millisecond)
+
+	if err != nil {
+		sendClientError(w, fmt.Sprintf("error executing query: %s", err))
+		return
+	}
+
+	jsonBytes, err := json.Marshal(results)
+	if err != nil {
+		sendClientError(w, fmt.Sprintf("error marshalling results to json: %s", err))
```
RebeccaMahany, line 50: Optional -- maybe we'd want a `sendServerError` so we can return a 5xx instead of a 4xx in these cases where it's not a client-side error?

ee/localserver/request-query_test.go
```diff
@@ -79,3 +83,100 @@ func Test_localServer_requestQueryHandler(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.testName, func(t *testing.T) {
+			t.Parallel()
+
+			//go:generate mockery --name Querier
+			// https://github.com/vektra/mockery <-- cli tool to generate mocks for usage with testify
```
RebeccaMahany, line 140: THIS IS SO COOL!?

ee/localserver/server.go
```diff
@@ -90,6 +90,8 @@ func New(logger log.Logger, db *bbolt.DB, kolideServer string) (*localServer, er
 	ecAuthedMux.Handle("/id.png", ls.requestIdHandler())
 	ecAuthedMux.Handle("/query", ls.requestQueryHandler())
 	ecAuthedMux.Handle("/query.png", ls.requestQueryHandler())
+	ecAuthedMux.Handle("/runscheduledquery", ls.requestQueryHandler())
```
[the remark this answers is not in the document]
RebeccaMahany: I was thinking I preferred `/scheduledquery` (for more consistency with `/query`), but not enough to leave a comment about it while reviewing!

ee/localserver/request-query.go
```diff
@@ -19,48 +21,89 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
 
-	if query == "" {
-		sendClientError(w, "empty query")
+	name, ok := body["name"]
+	if !ok || name == "" {
+		sendClientError(w, "no name key found in request body json")
 		return
 	}
 
+	scheduledQueryQuery := fmt.Sprintf("select query from osquery_schedule where name = '%s'", name)
```
RebeccaMahany, line 78: Making sure I'm understanding right -- this is how it works?

1. K2 calls `/runscheduledquery` with the name of the scheduled query
1. Launcher returns the query
1. K2 calls `/query` with the query returned in the previous step to actually run it

James-Pickett: So the driver behind this is that the JS bridge has to use GET requests passing the query (in a nacl box) as a URL parameter. This means there is a limit on the size of the query. This endpoint is intended to be a work around for that. It ought to work like this:

1. K2 looks up the name of a osquery scheduled query it want to be run. The same query should be available to launcher in the osquery_schedule table.
2. K2 calls `/scheduledquery`, with body `{ "name": "name_of_scheduled_query" }`
3. Launcher looks up the query's sql using the provided name
4. Launcher executes the query
5. Launcher returns results

I'll add a comment ents to the func about this.

ee/localserver/server.go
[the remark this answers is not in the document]
James-Pickett: went with `/scheduledquery`

ee/localserver/request-query.go
James-Pickett: fixed

ee/localserver/request-query.go
```diff
@@ -19,48 +21,100 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+}
+
+// requestScheduledQueryHandlerFunc uses the name field in the request body to look up
+// an existing osquery scheduled query execute it, returning the results.
+func (ls *localServer) requestScheduledQueryHandlerFunc(w http.ResponseWriter, r *http.Request) {
+	// The driver behind this is that the JS bridge has to use GET requests passing the query (in a nacl box) as a URL parameter.
+	// This means there is a limit on the size of the query. This endpoint is intended to be a work around for that. It ought to work like this:
+	//
+	// 1. K2 looks up the name of a osquery scheduled query it want to be run. The same query should be available to launcher in the osquery_schedule table.
+	// 2. K2 calls `/scheduledquery`, with body `{ "name": "name_of_scheduled_query" }`
```
directionless, line 67: Do you think this should be a different endpoint, or a different parameter to the same endpoint? (Either seems fine)

ee/localserver/request-query.go
```diff
@@ -19,48 +21,100 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
+		return
+	}
+
+	name, ok := body["name"]
+	if !ok || name == "" {
+		sendClientError(w, "no name key found in request body json")
+		return
+	}
+
+	scheduledQueryQuery := fmt.Sprintf("select query from osquery_schedule where name = '%s'", name)
```
directionless, line 89: I think this would be interesting as a `LIKE` so we can pull more than one set of results.

ee/localserver/request-query.go
```diff
@@ -19,48 +21,100 @@ func (ls *localServer) requestQueryHanlderFunc(w http.ResponseWriter, r *http.Re
...
 		return
 	}
 
-	jsonBytes, err := json.Marshal(results)
+	if len(results) == 0 {
+		sendClientError(w, fmt.Sprintf("no scheduled query found using \"%s\"", scheduledQueryQuery))
+		return
+	}
+
+	jsonBytes, err := json.Marshal(results[0])
```
directionless, line 108: We should support multiple queries, and we should probably model the response off osquery's response format.

ee/localserver/request-query.go
RebeccaMahany: Ah ok, I just completely missed where it was calling `requestQueryHanlderFunc` at the end of this func and that's how the query was getting run! Thanks for adding the comments

ee/localserver/request-query.go
directionless: If you support `LIKE` and running multiple queries, we'll need the query in the response. (And I think we really should support `LIKE`)

ee/localserver/request-query.go
James-Pickett: @directionless, idea (possibly bad) 💡 

what if we just let accepted a query for the query? something like:

`/fetchedquery --data '{"query": "select query from osquery_scheduled where name like '%foo%"}'`

we would have to handle the case where the query in the body returns more than one column ... error or first alphabetically or something

directionless: I'm not sure it buys us much. 

The query names are, in theory, known. But there's some skew that comes from how everything is generated. But a `LIKE` should cover pretty much anything. 

Go really far down the dynamic query and it just get confusing. Would we support other tables? Full sql? Anything that returns rows with a `query` column? 

I guess I could see some nebulous future where sometimes you wanted `select query from osquery_scheduled where name like '%foo%"`, but sometimes you wanted a pivot on `kolide_server_data`.

But mostly I think it's complicated

ee/localserver/request-query.go
James-Pickett: where can I find an example of osquery's response format for multiple queries?

or is it just
`[][]map[string]string`

instead of
`[]map[string]string`

but then how does one know which result goes with which query? ordering maybe?

directionless: "Distributed Queries" section of https://osquery.readthedocs.io/en/latest/deployment/remote/#remote-server-api

Which is modeled https://github.com/osquery/osquery-go/blob/master/plugin/distributed/distributed.go#L41-L48

possibly we should just use that struct directly

ee/localserver/server.go
```diff
@@ -99,13 +101,16 @@ func New(logger log.Logger, db *bbolt.DB, kolideServer string) (*localServer, er
...
+	// mux.Handle("/scheduledquery", ls.requestScheduledQueryHandler())
 
 	srv := &http.Server{
 		Handler:           ls.requestLoggingHandler(ls.preflightCorsHandler(ls.rateLimitHandler(mux))),
 		ReadTimeout:       500 * time.Millisecond,
 		ReadHeaderTimeout: 50 * time.Millisecond,
-		WriteTimeout:      5 * time.Second,
-		MaxHeaderBytes:    1024,
+		// WriteTimeout very high due to retry logic in the scheduledquery endpoint
+		WriteTimeout:   30 * time.Second,
```
James-Pickett, line 112: too much?

ee/localserver/request-query.go
James-Pickett: thanks, using that struct now

ee/localserver/request-query.go
James-Pickett: I prefer a different endpoint .... seems easier that splitting logic path of single endpoint

ee/localserver/server.go
directionless: Probably fine. We can futz around as needed.

directionless (approved):
I love it. This just looks like fun

PR #1065, opened 2023-03-01
Support notification actions for Ubuntu 22.04/Wayland

RebeccaMahany:
When testing notification actions previously, @James-Pickett noticed that on his Linux laptop, notification actions would open the URL in a text editor instead of using the default browser. After more investigation, this issue appears to be present on Ubuntu 22.04, but not on the 22.04 VMs I have. My initial guess was that James's laptop uses Wayland and all of the VMs (accessed via VNC or RDP) are x11, but it could be some other difference between his environment and my VMs that causes the issue.

Fixes:

* First, default to x-www-browser over xdg-open, since x-www-browser does not appear to have the issue noted below figuring out what application to use to open the URL.
* Second, set `DISPLAY` for Wayland sessions as well. We cannot pull this from loginctl like we do for x11 sessions, so ultimately we look for the xwayland process, knowing that the display is the first command-line argument to this process. It's possibly a little brittle, but I couldn't find a better option.
* Finally, add the snapd directory where mimeinfo.cache lives to `XDG_DATA_DIRS`. If the user has installed their browsers only via snap (e.g. Firefox, even when installed via apt, actually installs via snap), then the mimeinfo defaults are stored in this separate directory that xdg-open cannot find, ultimately resulting in the behavior noted where the URL is opened in a text editor.


ee/desktop/notify/notify_linux.go
```diff
@@ -81,7 +81,9 @@ func (d *dbusNotifier) Listen() error {
 
 			// Attempt to open a browser to the given URL
 			actionUri := signal.Body[1].(string)
-			providers := []string{"xdg-open", "x-www-browser"}
+			// We default to x-www-browser first because, if available, it appears to be better at picking
+			// the correct default browser.
+			providers := []string{"x-www-browser", "xdg-open"}
```
directionless, line 86: Consider an unexported constant/var at the package level.

ee/desktop/runner/runner_linux.go
```diff
@@ -91,6 +94,8 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
 	}
 
 	sessionList := strings.Split(sessions, " ")
+
+CheckSessions:
```
directionless, line 98: Whoa. Labels. Do we need this, or is this indicative of being too complex and that we should have some other functions?

ee/desktop/runner/runner_linux.go
```diff
@@ -132,3 +156,63 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
 
 	return envVars
 }
+
+func (r *DesktopUsersProcessesRunner) displayFromXwayland(uid int32) string {
+	//For wayland, DISPLAY is not included in loginctl show-session output -- in GNOME,
+	// Mutter spawns Xwayland and sets $DISPLAY at the same time. Find $DISPLAY by finding
+	// the Xwayland process and examining its args.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
directionless, line 165: This should have a context passed in, and not create a new one

ee/desktop/runner/runner_linux.go
```diff
@@ -132,3 +156,63 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
...
+			if procUid != uid {
+				continue
+			}
+
+			// We have a match! Grab the display value. The xwayland process looks like:
+			// /usr/bin/Xwayland :0 -rootless -noreset -accessx -core -auth /run/user/1000/.mutter-Xwaylandauth.ROP401 -listen 4 -listen 5 -displayfd 6 -initfd 7
+			cmdlineArgs := strings.Split(cmdline, " ")
+			if len(cmdlineArgs) < 2 {
+				// Process is somehow malformed or not what we're looking for -- stop evaluating it
+				break
```
directionless, line 210: not continue?

ee/desktop/runner/runner_linux.go
```diff
@@ -132,3 +156,63 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
...
+			}
+
+			return cmdlineArgs[1]
```
directionless, line 213: This seems fine. I wonder if we'll end up needing a regex

directionless (approved):
Seems okay. Though the labels imply there's some refactoring that might help.

ee/desktop/runner/runner_linux.go
RebeccaMahany: I think you're right and this could use a refactor, I'll give it another look!

ee/desktop/runner/runner_linux.go
directionless: I'm mostly thinking about all the ugly corner cases.

What if the user has 2 displays, :0 and :1? Maybe we ignore that.

What if they have other processes running called `Xwayland`? (maybe `grep Xwayland`)? Is it better to continue and see if we find one that matches what we want? 

I think return-first-match makes sense. But then on not-match, continue seems okay.

RebeccaMahany: I've updated this so that it makes it clearer what we're doing -- we _are_ doing return-first-match and continue on no-match.

I don't have another process running called `Xwayland`, at least, but yeah I'm a little unsure about the edge cases too.

ee/desktop/runner/runner_linux.go
RebeccaMahany: This is updated and I think it's more readable now 🙂

seejdev (approved):

ee/desktop/runner/runner_linux.go
directionless: I think we only have imperfect choices. So this seems fine, and we'll iterate as we need

ee/desktop/runner/runner_linux.go
```diff
@@ -104,31 +112,98 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
...
-			xDisplayOutput, err := exec.CommandContext(ctx, "loginctl", "show-session", session, "--value", "--property=Display").Output()
-			if err != nil {
-				level.Debug(r.logger).Log(
-					"msg", "could not get Display from user session",
-					"uid", uid,
-					"err", err,
-				)
-				continue
-			}
+		if sessionType == "x11" {
```
directionless, line 115: I tend to prefer `switch`, but it's  bike shedding, and not worth the change.

ee/desktop/runner/runner_linux.go
```diff
@@ -104,31 +112,98 @@ func (r *DesktopUsersProcessesRunner) userEnvVars(uid string) map[string]string
...
+			"msg", "could not get Display from user session",
+			"err", err,
+		)
+		return defaultDisplay
+	}
+
+	return strings.Trim(string(xDisplayOutput), "\n")
+}
+
+func (r *DesktopUsersProcessesRunner) displayFromXwayland(uid int32, ctx context.Context) string {
```
directionless, line 151: `ctx` should be the first argument. (I think, this one is not a nit)

directionless (changes requested):
Looks solid. But please put `ctx` first

ee/desktop/runner/runner_linux.go
RebeccaMahany: I preferred it as well, but it was what necessitated the label (so that we could break out of the outer for loop once we found a good session)

ee/desktop/runner/runner_linux.go
RebeccaMahany: Updated all method signatures here so ctx is first ✅

directionless (approved):

ee/desktop/runner/runner_linux.go
directionless: Thank you

ee/desktop/runner/runner_linux.go
directionless: Seems weird. Feels like a hint at more refactoring. But 🤷, not clearly worth while.

Issue #1061, opened 2023-02-28
Potential improvements to the log checkpoint data


directionless:
Our agent generates useful data and includes it in ["log checkpoints"](https://github.com/kolide/launcher/tree/d3362e6b4ec66d8421654d5c9b9adba29495b090/pkg/log/checkpoint). Working with various customer things, I think there are some potential things we should add:
- Hardware info
  * serial number
  * the motherboard uuid
  * model information
  * OS versions
- something from the server provided control data
  * Not quite sure what yet, but I'm imagining that K2 can push down useful info
 - WAN ip address?  But how would we get this? Might be the server controlled data

PR #1070, opened 2023-03-04
Add additional log checkpoint information

directionless:
In thinking about how to add more information to the checkpoint logs it felt like this needed to become a struct -- easier to hold some state, add a querier, and whatnot.

1. Refactor to a checkpointer struct, moved the `Run` around a little
2. Add a `SetQuerier` function
3. When there is a querier, additional things are queried. 

Fixes: #1061

### DIFF
--- a/pkg/log/checkpoint/queried.go
+++ b/pkg/log/checkpoint/queried.go
@@ -49,12 +49,21 @@ func (c *checkPointer) logOsqueryInfo() {
 	c.logger.Log("osquery_info", info)
 }
 
+func (c *checkPointer) logQueriedInfo() {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	for k, v := range c.queriedInfo {
+		c.logger.Log(k, v)
+	}
+}
+
 // queryStaticInfo usually the querier to add additional static info.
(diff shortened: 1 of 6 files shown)


pkg/log/checkpoint/queried.go
```diff
@@ -0,0 +1,100 @@
...
+
+const osquerySqlQuery = `
+SELECT
+	osquery_info.version as osquery_version,
+	osquery_info.instance_id as osquery_instance_id
+FROM
+    osquery_info
+`
+
+func (c *checkPointer) logOsqueryInfo() {
```
James-Pickett, line 38: should we add some data from `kolide_launcher_osquery_instance_history`? could possibly help up identify if osquery is crashing often and why

James-Pickett (approved):
Nice!

pkg/log/checkpoint/queried.go
directionless: That's a good thought. I'm going to merge that, and flag that for the next pass

directionless: https://github.com/kolide/launcher/issues/1071

Issue #1072, opened 2023-03-06
Add menu support for relative dates


directionless:
Discussion with K2, is that we often want to display dates as "in 3 days" and other sorts of relative display.

Is this something we can support? Are there other things we might want to format client side? Does this fit inside go templates, or do we want something else?

seejdev:
This looks helpful https://github.com/xeonx/timeago

directionless:
One of the things I'm least sure about, is how we handle supporting relative dates, vs not. And tinkering a bit with templates, I think we have some really interesting rope. Templates aren't really typed -- if there's an if statement, the untaken branch can have unreferenced functions. eg, this works:
```golang
	t, err := template.New("").Parse("{{if false}}{{.doesnotexist}}{{else}}You have a task named \"{{ .Name}}\"{{end}}")
```

I think this gives us some really interesting rope, in that we can conditionally change behavior based on launcher versions. Though maybe we should make a custom function to expose capabilities.

PR #1078, opened 2023-03-07
Template functions to support relative times in menu items

seejdev:
Supports #1072 

This PR adds two template functions:

- `hasCapability` which enables K2 to generate data conditionalized on launcher functionality in the field
- `relativeTime` generates a relative fuzzy timestamp from a Unix timestamp

This enables menu items defined as:

`Your machine will be blocked {{if hasCapability "relativeTime"}}{{relativeTime 1678526503}}{{else}}{{end}}."`

to render as:

`Your machine will be blocked in 3 days.`


ee/desktop/menu/menu_template.go
```diff
@@ -4,6 +4,14 @@ import (
 	"fmt"
 	"strings"
 	"text/template"
+	"time"
+
+	"github.com/xeonx/timeago"
```
directionless, line 9: My inclination would be to write our own, with rules from the K2 side, over pulling in a library. Do you think this is worth pulling in?

ee/desktop/menu/menu_template.go
```diff
@@ -32,7 +40,26 @@ func (tp *templateParser) Parse(text string) (string, error) {
...
+		funcHasCapability: func(capability string) bool {
+			if capability == funcRelativeTime {
+				return true
+			}
+			return false
+		},
+		// relativeTime takes a RFC 339 formatted timestamp and returns a fuzzy timestamp
+		funcRelativeTime: func(timestamp string) string {
+			t, err := time.Parse(
+				time.RFC3339,
```
directionless, line 54: When we did the notifications, we ultimately concluded it was simpler to pass seconds-since-epoch, than try to get ruby and go talking via rfc3339. (It shouldn't have been, but 🤷 we took the easy path)

Recommend unix time here as well

directionless (commented):
A couple small bits, but I like it.

ee/desktop/menu/menu_template.go
seejdev: Nah, it does more than we need atm (like `i18n` support). I'll replace with something simple to get us going.

directionless (approved):

ee/desktop/menu/menu_template.go
```diff
@@ -32,7 +38,37 @@ func (tp *templateParser) Parse(text string) (string, error) {
 		return "", fmt.Errorf("templateData is nil")
 	}
 
-	t, err := template.New("menu_template").Parse(text)
+	t, err := template.New("menu_template").Funcs(template.FuncMap{
```
James-Pickett, line 41: template thing is so cool

James-Pickett (approved):

ee/desktop/menu/menu_template.go
seejdev: I know! I underestimated their power and originally thought this would be a much more complex implementation.

PR #1079, opened 2023-03-07
Periodically refresh the desktop menu

### LINKED ISSUE #1072: Add menu support for relative dates

seejdev:
Fixed #1072 (In conjunction with https://github.com/kolide/launcher/pull/1078)

This PR prevents desktop menu data from becoming too stale by making the runner periodically refresh menu data and notifying the desktop processes. I removed the refresh menu action since this functionality is no longer needed.


directionless (approved):
Seems simple enough. Woo

Issue #954, opened 2022-12-01
Begin work on Autoupdate: The Next Generation 


directionless:
I think it's finally time to revisit our autoupdate code! Excitement! Adventures! Bugs....

At a high level, I think our goals are:
* Remove Notary
* Use some modern libraries (Probably `go-tuf`, happy to hear other suggestions)
* Use a reasonable channel model for specifying which versions
* Expose more functionality inside launcher (right now it's all abstracted out into a goroutine that mostly does it's own thing)
* Improve testability / dev setups

I'd imagine some milestones like:
* [x] Play with go-tuf. What's it like running a server? A client? Can we _not_ have a server, and do it with a static set of files?
* [x] What would `porter` look like?
* [x] Launcher code? Proper channel and signaling usage
* [x] Test cases!
* [x] deploy a prod server
* [x] go!

launcher work remaining as of 29 June 2023:
* [x] Point to per-arch path https://github.com/kolide/launcher/pull/1195
* [x] Roll out new autoupdate code for internal testing on nightly -- https://github.com/kolide/launcher/pull/1391
* [x] Roll out new autoupdate code for internal testing on beta
* [x] Enable gradual rollout to stable (https://github.com/kolide/launcher/pull/1526)
* [x] Release autoupdate fully to stable

endpoint-releaser and monorepo work:
* [x] New TUF infrastructure for development
* [x] New TUF infrastructure for production
* [x] endpoint-releaser publishes and promotes releases to new TUF infrastructure
* [x] Monitor + alert when TUF keys are nearing expiration (https://github.com/kolide/endpoint-releaser/issues/71)
* [x] Allow for changing TUF passphrase (completed but no CI job)

RebeccaMahany:
Blocks https://github.com/kolide/launcher/issues/896

RebeccaMahany:
100% of devices now have the `use_tuf_autoupdater` feature flag set.

Planned improvements filed in https://github.com/kolide/launcher/issues/1577.

PR #1081, opened 2023-03-07
Run new TUF autoupdater side-by-side with notary autoupdater

RebeccaMahany:

This is part 1 of moving to use our new TUF infrastructure; here, we run our new TUF autoupdater side-by-side with our older notary autoupdater implementation. The new autoupdater only maintains a TUF repo but does not attempt to perform updates yet.

Subsequent PRs will tackle the following (order is not set in stone)
1. Improving TUF autoupdater metrics and exposing them in a launcher table
1. Point to production TUF instead of development
1. Download, verify, and untar downloads from new TUF, moving them to a new update directory located in the root directory
1. `findnew` looks for updates in new location, falling back to old location
1. Other improvements/refactors


pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+		interrupt:       make(chan struct{}),
+		checkInterval:   60 * time.Second,
+		errorCounter:    make([]int64, 0), // For now, the error counter is a simple list of timestamps when errors occurred
+		logger:          log.NewNopLogger(),
+	}
+
+	for _, opt := range opts {
+		opt(ta)
+	}
+
```
directionless, line 112: I recognize that this works, but this feels a little out of order to me.

In the `NewTufAutoupdater` function, it creates all the tuf things, then it creates the `TufAutoupdater` struct. But this means that if we want to use the functional argument pattern to change them, it's hard. I've generally reversed these.

Create `ta` first, parts opts, then create the TUF things based on the values in `ta`. (And add new things to `ta`)

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func WithUpdateCheckInterval(checkInterval time.Duration) TufAutoupdaterOption {
+	return func(ta *TufAutoupdater) {
+		ta.checkInterval = checkInterval
+	}
+}
+
+func NewTufAutoupdater(metadataUrl, mirrorUrl, binaryPath, rootDirectory string, opts ...TufAutoupdaterOption) (*TufAutoupdater, error) {
+	binaryName := filepath.Base(binaryPath)
```
directionless, line 68: Why pass `binaryPath` and not binaryName directly?

Is this going to need to strip off `.exe` on windows?

pkg/make/builder.go
```diff
@@ -474,6 +484,41 @@ func bootstrapFromNotary(notaryConfigDir, remoteServerURL, localRepo, gun string
 	return nil
 }
 
+func bootstrapFromTUF(remoteServerURL string, localRepo string) error {
+	req, err := http.NewRequest(http.MethodGet, remoteServerURL+"/repository/root.json", nil)
+	if err != nil {
+		return fmt.Errorf("could not create request to TUF: %w", err)
+	}
```
directionless, line 491: How do we vet this download?

cmd/launcher/internal/updater/updater.go
```diff
@@ -79,17 +97,22 @@ func NewUpdater(
 // updater allows us to mock *autoupdate.Updater during testing
 type updater interface {
 	Run(opts ...tuf.Option) (stop func(), err error)
+	RollingErrorCount() int
 }
 
 type updaterCmd struct {
```
directionless, line 103: For what it's worth, I don't think I have an opinion about whether this model is at all correct. It think it carries a lot of baggage.

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+
+	oneDayAgo := time.Now().Add(-24 * time.Hour).Unix()
+	errorCount := 0
+	for _, errorTimestamp := range ta.errorCounter {
+		if errorTimestamp < oneDayAgo {
+			continue
+		}
+		errorCount += 1
+	}
+
```
directionless, line 134: Does anything ever clean `errorCounter`?

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+	return nil
+}
+
+func (ta *TufAutoupdater) versionFromTarget(target string) string {
+	strippedBinary := strings.TrimSuffix(ta.binary, ".exe")
+
+	// The target is in the form `launcher/linux/launcher-0.13.6.tar.gz` -- trim the prefix and the file extension to return the version
+	prefixToTrim := fmt.Sprintf("%s/%s/%s-", strippedBinary, ta.operatingSystem, strippedBinary)
```
directionless, line 208: Urk. I've lost site of how we track archtecture

cmd/launcher/internal/updater/updater.go
RebeccaMahany: Yup, I think updating it falls more into a later phase though! I'd like to get new TUF just happily running alongside the current stuff before doing larger refactors.

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+	legacytuf "github.com/kolide/updater/tuf"
+	client "github.com/theupdateframework/go-tuf/client"
+	filejsonstore "github.com/theupdateframework/go-tuf/client/filejsonstore"
+)
+
+//go:embed assets/tuf-dev/root.json
+var rootJson []byte
+
+const (
+	DefaultTufServer = "https://tuf-devel.kolide.com"
```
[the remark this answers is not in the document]
RebeccaMahany: Yes! I realized that there's a fairly simple way to start out pointing to the dev infra and then move to the prod infra -- just changing the TUF repo location.

I do really think I'd like to roll it out like this first (at least for a beta release) before doing the production infrastructure, to minimize changes we have to make to production infrastructure.

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+)
+
+type TufAutoupdater struct {
+	metadataClient  *client.Client
+	mirrorClient    *http.Client
```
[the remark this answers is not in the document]
RebeccaMahany: We also pass a http client into the metadata client.Client, and I think it's possible we'd want to configure those two differently

pkg/autoupdate/tuf_autoupdate.go
RebeccaMahany: No, just a launcher or computer restart -- I wasn't super worried given the long default update interval and given that this will be updated in the next couple PRs (though that sort of comment will jinx it to hang around forever 😀). I can add some cleanup here.

cmd/launcher/internal/updater/updater.go
```diff
@@ -62,12 +63,29 @@ func NewUpdater(
 		return nil, err
 	}
 
+	// create the new updater
+	tufAutoupdater, err := autoupdate.NewTufAutoupdater(
```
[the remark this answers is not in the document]
RebeccaMahany: I'm fine with a new pkg as well! I started out here because I wasn't sure how much of the existing autoupdate functionality I wanted to lean on, but even with the full download/stage I only used `checkExecutable`.

Would definitely like suggestions for naming, `tufautoupdate` versus `autoupdate` is feeling a little murky

cmd/launcher/internal/updater/updater.go
```diff
@@ -23,6 +23,7 @@ type UpdaterConfig struct {
 	UpdateChannel      autoupdate.UpdateChannel
 	InitialDelay       time.Duration // start delay, to avoid whomping critical early data
 	NotaryURL          string
+	TufServerURL       string
```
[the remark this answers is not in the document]
RebeccaMahany: For this part cmd/launcher/internal/updater specifically I want to keep it here for now, to make it easier in future work to switch between the new autoupdater and the old one if the new one is erroring out.

pkg/autoupdate/tuf_autoupdate.go
RebeccaMahany: Sure, I'll update this and handle not using the default http clients at the same time as well 👍

pkg/autoupdate/tuf_autoupdate.go
RebeccaMahany: I don't believe it's present in the TUF/dl.kolide.co stuff

pkg/autoupdate/tuf_autoupdate.go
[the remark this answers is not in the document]
RebeccaMahany: Realized we aren't even using the mirrorClient/url right now anyway, I'll pull them out for simplicity's sake

pkg/autoupdate/tuf_autoupdate.go
```diff
@@ -0,0 +1,211 @@
...
+	return func(ta *TufAutoupdater) {
+		ta.logger = log.With(logger, "component", "tuf_autoupdater")
+	}
+}
+
+func WithChannel(channel UpdateChannel) TufAutoupdaterOption {
+	return func(ta *TufAutoupdater) {
+		ta.channel = channel
+	}
+}
```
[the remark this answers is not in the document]
RebeccaMahany: Right now we've got 1 and 2, and I imagine 3 would (eventually) be handled by a new consumer as a part of phase 5-ish. I can plan on coming back to runtime in part 5?

pkg/autoupdate/tuf_autoupdate.go
directionless: Maybe? I think we should change where we store updates. We'll need to for nix, and the current idiom is kinda wonky

pkg/make/builder.go
RebeccaMahany: Checking the tuf spec and the go-tuf library more, I think we should remove this download, and ship this file as a hardcoded asset instead -- ref [here](https://theupdateframework.github.io/specification/latest/#load-trusted-root). Let me know if that sounds right to you?

cmd/launcher/internal/updater/updater.go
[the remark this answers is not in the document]
RebeccaMahany: I went with package `tuf` and have started aliasing the other tuf to `legacytuf`, but I am not 100% sold on it and open to suggestions!

pkg/autoupdate/tuf_autoupdate.go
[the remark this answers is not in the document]
directionless: okay

cmd/launcher/internal/updater/updater.go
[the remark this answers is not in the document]
directionless: I would be unclever and name it `autoupdate2`. Or possible `autoupdate/tuf`. Or possible it doesn't matter, since we could, eventually, rename it.

pkg/autoupdate/tuf_autoupdate.go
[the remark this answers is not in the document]
directionless: Huh, I guess it could be a subscriber... Though I think it might be cleaner to just see where the runtime interfaces get us. Wither way, this is fine for now

pkg/autoupdate/tuf_autoupdate.go
directionless: I want to say it will be okay, but I'm sure they'll be _one_ person who has intermittent network, and it causes an endless stream of errors and never restarts.

RebeccaMahany: Makes sense, better to be on the safe side! I've added a cleanup job that runs 2x a day

cmd/launcher/internal/updater/updater.go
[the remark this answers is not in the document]
directionless: I suspect this is adding more complexity than benefit. But I'm willing to go with it

pkg/make/builder.go
directionless: I think, something like that. The process that our current code _should_ be doing is along the lines of:
1. Check in some initial trust information (this `root.json`, for example)
2. Update a build time copy 
3. Ship the updated version

The intent of (2) is to incur that initial download/sync cost at build time.

pkg/autoupdate/tuf_autoupdate.go
RebeccaMahany: I think per usual my worry is about changing too many preexisting things that work as-is 🙂 I'll update this to be the binaryName instead of path, and punt this down the road a bit.

pkg/make/builder.go
RebeccaMahany: This is updated now

cmd/launcher/internal/updater/updater.go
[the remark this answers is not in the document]
RebeccaMahany: Moved under autoupdate/tuf

cmd/launcher/internal/updater/updater.go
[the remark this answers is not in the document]
RebeccaMahany: I've pulled it out!

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,231 @@
...
+		type releaseFileCustomMetadata struct {
+			Target string `json:"target"`
+		}
+
+		var custom releaseFileCustomMetadata
+		if err := json.Unmarshal(*target.Custom, &custom); err != nil {
+			return fmt.Errorf("could not unmarshal release file custom metadata: %w", err)
+		}
+
+		level.Debug(ta.logger).Log(
```
James-Pickett, line 179: just to make sure I understand, this is all we are doing for now? logging when we find a newer version?

RebeccaMahany: Yes correct! Next steps (in next PRs) will be a) expose this info in a launcher table so we can query and see that devices are correctly pulling from new TUF and then b) start to actually download these newer versions.

pkg/autoupdate/tuf_autoupdate.go
James-Pickett: I also think this is something we'll need. There have been a few times that folks have asked for arm64 on linux / windows. I think number will continue to grow. @RebeccaMahany , do you think this would be difficult to bolt on later?

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,231 @@
...
+		select {
+		case <-checkTicker.C:
+			if err := ta.checkForUpdate(); err != nil {
+				ta.lock.Lock()
+				ta.errorCounter = append(ta.errorCounter, time.Now().Unix())
+				ta.lock.Unlock()
+
+				level.Debug(ta.logger).Log("msg", "error checking for update", "err", err)
+			}
+		case <-errorCheckTicker.C:
```
James-Pickett, line 134: Why are we doing the error count check and error clean up thing? Why don't we just log an error and move on? We are logging it in the checkTime.C block above.

RebeccaMahany: In the next PR, I want to expose the error count in a launcher table, kind of similar to `kolide_launcher_osquery_instance_history` -- I can add a comment about what it's for!

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not init metadata client: %w", err)
+	}
+
+	return ta, nil
+}
+
+func initMetadataClient(binary, rootDirectory, metadataUrl string, metadataHttpClient *http.Client) (*client.Client, error) {
+	// Set up the local TUF directory for our TUF client -- a dev repo, to be replaced once we move to production
+	localTufDirectory := filepath.Join(rootDirectory, fmt.Sprintf("%s-tuf-dev", binary))
```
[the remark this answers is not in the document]
RebeccaMahany: Yes, good idea!

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/kolide/kit/version"
+	client "github.com/theupdateframework/go-tuf/client"
+	filejsonstore "github.com/theupdateframework/go-tuf/client/filejsonstore"
+)
+
+//go:embed assets/tuf-dev/root.json
+var rootJson []byte
+
+const (
+	DefaultTufServer = "https://tuf-devel.kolide.com"
```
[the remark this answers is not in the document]
RebeccaMahany: I think @directionless named it 🙂 

I could make the local directory `tuf-devel` instead of `tuf-dev` if we think it'd be better for them to match -- I wasn't thinking too hard about it when I named the directory haha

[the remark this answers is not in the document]
James-Pickett: don't need to change anything, I was just curious

pkg/autoupdate/tuf/autoupdate.go
RebeccaMahany: Added a comment on the error counter!

James-Pickett (dismissed):

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,236 @@
...
+	client "github.com/theupdateframework/go-tuf/client"
+	filejsonstore "github.com/theupdateframework/go-tuf/client/filejsonstore"
+)
+
+//go:embed assets/tuf-dev/root.json
+var rootJson []byte
+
+const (
+	DefaultTufServer       = "https://tuf-devel.kolide.com"
+	DefaultChannel         = "stable"
```
directionless, line 30: Exporting these seems weird.

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+}
+
+func WithUpdateCheckInterval(checkInterval time.Duration) TufAutoupdaterOption {
+	return func(ta *TufAutoupdater) {
+		ta.checkInterval = checkInterval
+	}
+}
+
+func NewTufAutoupdater(metadataUrl, binary, rootDirectory string, metadataHttpClient *http.Client, opts ...TufAutoupdaterOption) (*TufAutoupdater, error) {
```
directionless, line 66: I suspect we can remove option. But let's go with it and iterate

directionless (approved):
I think there's going to be room for simplification, but we can use this for now!

pkg/autoupdate/tuf/autoupdate.go
RebeccaMahany: Looks like `DefaultChannel` was exported from a previous iteration -- unexported it. (`DefaultTufServer` is in use in `options`, matches similar consts for notary/mirror URLs that live in `autoupdate`.)

directionless (approved):

Issue #1089, opened 2023-03-09
I want table access to control data flags


directionless:
I keep trying to debug things around what the control data flags are set to. I think that means I want table access to them.

I don't know if it's worth doing this as a one off, or thinking about how it intersects with #1083 and a generalized flags environment.

PR #1092, opened 2023-03-09
converts kolide_server_data table to be generic launcher_db table, adds table for agent_flags

James-Pickett:
closes [1089](https://github.com/kolide/launcher/issues/1089)

### DIFF
--- a/pkg/osquery/tables/launcher_db/launcher_db.go
+++ b/pkg/osquery/tables/launcher_db/launcher_db.go
@@ -1,31 +1,28 @@
-package kolide_server_data
+package launcher_db
 
 import (
 	"context"
 	"fmt"
 
-	"github.com/kolide/launcher/pkg/osquery"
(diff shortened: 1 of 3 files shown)


pkg/osquery/tables/launcher_db/launcher_db.go
```diff
@@ -1,31 +1,28 @@
-package kolide_server_data
```
directionless, line 1: I do not think this is a good idea. 

Two reasons.

First, this table should be here, standing alone. The intent is not to debug, but to provide a hook for queries to join against.

Second, I think a launcher_db table would need an allowlist of buckets to iterate. There are buckets and keys that should not be exposed

directionless (changes requested):

directionless (approved):
@James-Pickett points out I didn't read this carefully. This is a generic interface, which is instantiated specifically. Seems fine!

PR #1094, opened 2023-03-10
Fix desktop not running when it should be

seejdev:
This PR addresses something we overlooked until desktop enablement became defaulted to false.

Control server will tell launcher to enable desktop by sending the `desktop_enabled_v1` flag to it. Launcher will store this in the database, however launcher was not actually querying this value in order to override the default value of `desktopProcessSpawningEnabled`. This PR simply adds a database query to check for this flag and set `desktopProcessSpawningEnabled` to `true` if the flag is present.

In practice, this meant killing the launcher proper process would prevent desktop processes from being spawned in the future. This change should enable desktop again seamlessly - launcher instances which should have desktop enabled should already have the `desktop_enabled_v1` in their databases. Upon upgrade, launcher should observe this now and launch desktop.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -195,11 +195,6 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		checkpointer.SetQuerier(extension)
 	}()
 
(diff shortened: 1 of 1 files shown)


directionless (approved):
Comment seems wrong, but I'll take it.

Issue #1086, opened 2023-03-08
It would be neat if localserver could trigger a control server fetch rate


directionless:
Talking about things on slack, we realized it would be valuable to be able to adjust the control server update timing. 

The specific use case is around a user in the device trust flow. they'll take some fix action, and the menu update will lag. So there are a couple of interesting I think we should do.

* [x] #1101
* [ ] Much as osquery has an "accelerated mode" we should create a way to adjust the control server poll interval. It would be really powerful for the localserver to be able to trigger something like "poll every 5s for the next 5 minutes"

directionless:
This was implemented in #1099

PR #1099, opened 2023-03-13
james/localserver fetch control data

James-Pickett:
- wip endpoint for updating control server refresh rate
- adds endpoint for fetching local server data on demand
- removes unneeded left over changes from previous branch
- clean up


ee/localserver/request-controlservice.go
```diff
@@ -0,0 +1,21 @@
+package localserver
+
+import (
+	"fmt"
+	"net/http"
+)
+
+func (ls *localServer) requestControlServiceFetchHanlder() http.Handler {
```
RebeccaMahany, line 8: Nit -- typo in Handler 🙂

ee/localserver/request-controlservice.go
```diff
@@ -0,0 +1,21 @@
...
+	return http.HandlerFunc(ls.requestControlServiceFetchFunc)
+}
+
+func (ls *localServer) requestControlServiceFetchFunc(w http.ResponseWriter, r *http.Request) {
+	if ls.controlService == nil {
+		sendClientError(w, fmt.Sprintf("control service not configured"))
+		return
+	}
+
+	if err := ls.controlService.Fetch(); err != nil {
```
RebeccaMahany, line 18: I wonder if we'd also want a flag to make the control service make requests for each subsystem, like temporarily flipping `agent.Flags.ForceControlSubsystems`? Maybe depends on what this PR is needed for.

ee/localserver/request-controlservice.go
James-Pickett: Eventually, I will learn to spell Handler ...

ee/localserver/request-controlservice.go
James-Pickett: There is a little more information about what this PR is for [here](https://github.com/kolide/launcher/issues/1086). I think that during the auth flow, if a user fixes an issue, we probably want to update the menu right meow. So the user doesn't still have a yellow / red icon on their task bar.

> I wonder if we'd also want a flag to make the control service make requests for each subsystem, like temporarily flipping agent.Flags.ForceControlSubsystems? Maybe depends on what this PR is needed for.

Do you mean like if a sub system was disabled, we want to for it to update?

RebeccaMahany: Ah ok, that makes sense and in that case we probably don't need the flag! I was thinking more along the lines of, "what if local data is bad/corrupt so we want to force launcher to refresh its data even if it doesn't think it's out-of-date", but I guess that'd be accomplished on the k2 side by regenerating hashes, maybe.

RebeccaMahany (approved):
🔥

ee/localserver/request-controlservice.go
directionless: I think the force rebuild is a good thing to consider. I'm not sure quote where it'll land. For the `flags`, it's easy for k2 to add a cache bust flag (it already has one)

But I'm less sure about the other systems.

ee/localserver/server.go
```diff
@@ -54,22 +58,42 @@ type localServer struct {
...
+func WithControlService(cs controlService) LocalServerOption {
+	return func(s *localServer) {
+		s.controlService = cs
+	}
+}
+
+func New(db *bbolt.DB, kolideServer string, opts ...LocalServerOption) (*localServer, error) {
 	ls := &localServer{
-		logger:                log.With(logger, "component", "localserver"),
+		logger:                log.NewNopLogger(),
```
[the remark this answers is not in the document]
James-Pickett: This is typically what we do for passing a logger using the options pattern. If the callers passes the `WithLogger()` option then that is used

ee/localserver/server.go
```diff
@@ -86,6 +110,8 @@ func New(logger log.Logger, db *bbolt.DB, kolideServer string) (*localServer, er
 	ecKryptoMiddleware := newKryptoEcMiddleware(ls.logger, ls.myLocalDbSigner, ls.myLocalHardwareSigner, *ls.serverEcKey)
 	ecAuthedMux := http.NewServeMux()
 	ecAuthedMux.HandleFunc("/", http.NotFound)
+	ecAuthedMux.Handle("/controlservicefetch", ls.requestControlServiceFetchHandler())
+	ecAuthedMux.Handle("/controlservicefetch.png", ls.requestControlServiceFetchHandler())
```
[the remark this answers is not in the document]
James-Pickett: This is a new endpoint. It's added to the server before the server starts up.

ee/localserver/server.go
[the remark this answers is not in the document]
seejdev: 🤦 I see that now!

seejdev (approved):

ee/localserver/request-controlservice.go
directionless: I agree -- it wouldn't belong in auth flow.

directionless (approved):
Nice!

A couple of comments, though I don't think they need to block anything here. Future thoughts...

1. Seeing the control service here makes sense, but also starts feeling tangly. TBD whether the idea of an agent runtime interface will help

2. I think we're going to want something to adjust the control timing. In my head, it's probably the same endpoint -- "Set the request interval to 10s, and start _now_"

PR #1102, opened 2023-03-13
bbolt begone: Phase 2

seejdev:
This PR is an extension of #1062 - the theme being replacing areas with a direct dependency on bbolt, with generic key-value store interfaces.

There was some low hanging fruit in some places like the launcher tables that rely on database access. These were easy enough to change over. The osquery extension class is a beast and somewhat tangled already, so I did my best at replacing bbolt where possible.

Extension also uses some functionality that does not fit very cleanly into key-value interfaces (at least without some significant refactoring of the extension code itself). This includes functionality like extracting database statistics information, and performing DB compaction, and log processing. These areas use bbolt stuff that is not necessarily specific to a key-value store, and it gets funky. Rather than make this PR larger and introduce more refactoring, I chose to leave those out. Perhaps in the future, when the extension can be broken out a bit better, there will be an opportunity to make this cleaner and get to a point where a direct dependency on bbolt can be removed.


cmd/launcher/launcher.go
```diff
@@ -170,13 +170,33 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+
 	// init osquery instance history
-	if err := osqueryInstanceHistory.InitHistory(db); err != nil {
+	if err := osqueryInstanceHistory.InitHistory(osqueryHistoryStore); err != nil {
 		return fmt.Errorf("error initializing osquery instance history: %w", err)
 	}
 
 	// create the osquery extension for launcher. This is where osquery itself is launched.
-	extension, runnerRestart, runnerShutdown, err := createExtensionRuntime(ctx, db, client, opts)
+	extension, runnerRestart, runnerShutdown, err := createExtensionRuntime(ctx, db, configStore, serverDataBucketConsumer, desktopFlagsBucketConsumer, client, opts)
```
RebeccaMahany, line 199: Any time we add a new bucket, we'll have to init it here and then thread it through `createExtensionRuntime` => `osqueryRunnerOptions` / `grpcRunnerOptions` => `commonRunnerOptions` => `LauncherTables`, which feels a decent amount of boilerplate/a lot of locations to touch in order to do an otherwise fairly small update. I get that the whole point is that we don't want to keep passing the bbolt db around everywhere -- would we want to consider another option to reduce boilerplate? E.g. maybe a very small interface for something like `GetOrCreateBucket(bucketName string) types.KVStore` that we can use for areas like `LauncherTables` that require access to a couple different buckets.

pkg/osquery/extension.go
```diff
@@ -170,39 +164,25 @@ func NewExtension(client service.KolideService, db *bbolt.DB, opts ExtensionOpts
...
-	err := db.Update(func(tx *bbolt.Tx) error {
-		for _, name := range bucketNames {
-			_, err := tx.CreateBucketIfNotExists([]byte(name))
-			if err != nil {
-				return err
-			}
-		}
-		return nil
-	})
+	stores, err := makeStores(opts.Logger, db)
```
RebeccaMahany, line 167: +1 for doing a refactor for the extension in a different/later PR

pkg/agent/storage/bbolt/keyvalue_store_bbolt.go
```diff
@@ -214,14 +215,32 @@ func (s *bboltKeyValueStore) Update(data io.Reader) error {
 // SetupDB is used for creating bbolt databases for testing
 func SetupDB(t *testing.T) *bbolt.DB {
```
James-Pickett, line 216: I believe this will cause the testing.T package to be included in our compiled binary. I suggested accepting a filepath as a parameter instead. Leave it up to the caller to use `SetupDb(t.TempDir())`

ee/control/consumers/notificationconsumer/notify_consumer.go
```diff
@@ -220,7 +220,6 @@ func (nc *NotificationConsumer) cleanup() {
 
 	// Delete all old keys
 	if err := nc.store.Delete(keysToDelete...); err != nil {
-		// Log, but don't error, since we might be able to delete some others
 		level.Error(nc.logger).Log("msg", "could not delete old notifications from bucket", "err", err)
```
directionless, line 223:
```suggestion
		level.Debug(nc.logger).Log("msg", "could not delete old notifications from bucket", "err", err)
```
(I know it's old code, but here we are)

cmd/launcher/launcher.go
directionless: I think we'll want something, yeah. This might be a good corner to test with. 

I've been assuming we'll evolve a layer above `store`. But I don't know what it'll look like

directionless (commented):
> functionality like extracting database statistics information, and performing DB compaction, and log processing

I think there are reasonable odds we'll want new functions on the interface. I could imagine `Stats`, `Compact`, `LogWrite`, and `LogRead`. I could also imagine a different interfaces for some of those. Regardless, pushing those forward feels good.

cmd/launcher/launcher.go
seejdev: I've introduced `Storage.GetStore()` to address this! This change has helped in two ways - reducing boilerplate and argument lists, and ensuring there is a single instance of each store.

pkg/agent/storage/bbolt/keyvalue_store_bbolt.go
seejdev: Ooh good catch! I see another spot in the codebase we've done a similar pattern. I go fix!

seejdev: I ended up moving this function to the `storageci` package which is only used by tests and keeps it from being compiled into our release binary.

pkg/agent/types/storage.go
```diff
@@ -0,0 +1,26 @@
+package types
+
+// Stores are named identifiers corresponding to key-value buckets
+type Store string
+
+const (
```
seejdev, line 6: Not sure if I like these defined in `types`, since they're not actually types. Open to suggestions of better arrangements.

pkg/agent/types/kontext.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import "go.etcd.io/bbolt"
+
+// Kontext ("Kolide context") holds references to data and structs which are used throughout
+// launcher code and are typically valid for the lifetime of the application instance.
+type Kontext struct {
```
directionless, line 7: I'll roll with if the team wants, but I am slightly hesitant to name it `kontext`. I think it servers a very different purpose than `context` and I think the name collision will be more frustrating than joyful.

(I would support `kontext` as a wrapper over context.Context`)

seejdev: I'm not attached to it -- `runtime` feels in the right area, but perhaps is too generic or could be confused with other things? I'm open to ideas... a name that helps indicate that this struct is a collection of common objects.

cmd/launcher/launcher.go
```diff
@@ -110,6 +107,12 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		return fmt.Errorf("write launcher pid to file: %w", err)
 	}
 
+	storage, err := agentbbolt.NewStorage(logger, db)
```
seejdev, line 110: I'm going to move this up to proper `main()` function that calls `runLauncher`

pkg/agent/types/kontext.go
James-Pickett: * `launchinator` ... `lchntr := launchinator.New(...)`
* `kolidergeist` ... `kgst := kolidergeist.New(...)`
* `kaunchpad` ... `kpd := kaunhcpad.New(...)`
* `kore` ... `kr := kore.New(...)`
* `knexus` ... `kxs := knexus.New(...)`

generated with some AI help =)

RebeccaMahany: I like `knapsack` -- like it's our inventory of useful services. I think in general bag/container metaphors make the most sense to me for this usage. (So maybe `kontainer`?)

I think we're at risk of overloading `agent`/`settings/`config` because those are already in use elsewhere in our codebase and mean different things, so I probably wouldn't go with something in that direction. I also really like `kontext` but agree with seph that the conflation with `context` will probably be a little confusing.

RebeccaMahany: Some sillier bag/container-based options:
* `capsule` / `kapsule` (bonus encapsulation reference)
* `locker` (looks a little similar to launcher, which is fun)
* `canteen` and `cauldron` are fun but I think they're too unreadable if you sub the c for a k, unfortunately

James-Pickett: +1 `knapsack`

@RebeccaMahany 's "inventory of useful services" description is pretty good.

seejdev: `knapsack` it is!

cmd/launcher/launcher.go
seejdev: Actually, I can't do this since bbolt db is created here.

pkg/agent/types/knapsack.go
```diff
@@ -0,0 +1,30 @@
...
+	// remove this field and prevent "leaking" bbolt into places it doesn't need to.
+	BboltDB *bbolt.DB
+
+	// This struct is a work in progress, and will be iteratively added to as needs arise.
+	// Some potential future additions include:
+	// Flags
+	// Querier
+}
+
+func NewKnapsack(s Storage, db *bbolt.DB) *Knapsack {
```
James-Pickett, line 23: Suggestion ... 

We might consider putting Knapsack in it's own pkg. Possibly at `pkg/agent/knapsack`. If we think this is a good idea, It could wait until next PR.

If this will become truly ubiquitous, I can imagine there could be several convenience constructors / options we would use in tests or other situations.

`knapsack.New(WithMemStore, WithFmtLogger, WithFlagFoo, ...)`

James-Pickett (approved):
Nice!

pkg/agent/types/knapsack.go
```diff
@@ -0,0 +1,30 @@
+package types
+
+import "go.etcd.io/bbolt"
+
+// Knapsack is an inventory of data and useful services which are used throughout
+// launcher code and are typically valid for the lifetime of the launcher application instance.
+type Knapsack struct {
```
directionless, line 7: I think there's a sprawling conversation around struct vs type here, but I'm happy to let that keep evolving.

pkg/agent/types/knapsack.go
```diff
@@ -0,0 +1,30 @@
...
+	Storage Storage
```
directionless, line 8: `Storage` feels wrong. Launcher has a bunch of different `storage` things under the hood, and I think they should each be a member here.

directionless (changes requested):

pkg/agent/types/knapsack.go
seejdev: I don't mind doing this now, I have a feeling it will become pervasive enough that doing this now is worth it. Good suggestion!

pkg/agent/types/knapsack.go
seejdev: Yeah originally I was thinking the `Storage` abstraction would be helpful when `Knapsack` isn't present, but it's probably unnecessary since we will inject `Knapsack` into plenty of places anyways.

I've moved all the functionality previously in `Storage` up to this top layer now, the storage objects get injected into `Knapsack`, and are immediately accessible from there. Feels better!

cmd/launcher/launcher.go
```diff
@@ -173,12 +176,12 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	}
 
 	// init osquery instance history
-	if err := osqueryInstanceHistory.InitHistory(db); err != nil {
+	if err := osqueryInstanceHistory.InitHistory(k.OsqueryHistoryInstanceStore()); err != nil {
```
RebeccaMahany, line 179: NICE

pkg/agent/storage/stores.go
```diff
@@ -0,0 +1,21 @@
...
+	AgentFlagsStore             Store = "agent_flags"              // The store used for agent control flags.
+	AutoupdateErrorsStore       Store = "tuf_autoupdate_errors"    // The store used for tracking new autoupdater errors.
+	ConfigStore                 Store = "config"                   // The store used for launcher configuration.
+	ControlStore                Store = "control_service_data"     // The store used for control service caching data.
+	InitialResultsStore         Store = "initial_results"          // The store used for initial runner queries.
+	ResultLogsStore             Store = "result_logs"              // The store used for buffered result logs.
+	OsqueryHistoryInstanceStore Store = "osquery_instance_history" // The store used for the history of osquery instances.
+	SentNotificationsStore      Store = "sent_notifications"       // The store used for sent notifications.
+	StatusLogsStore             Store = "status_logs"              // The store used for buffered status logs.
+	ServerProvidedDataStore     Store = "server_provided_data"     // The store used for pushing values from server-backed tables.
```
RebeccaMahany, line 16: 💯 thanks for including the extra documentation here

RebeccaMahany (approved):
I like it!

pkg/agent/storage/bbolt/stores_bbolt.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/go-kit/kit/log"
+
+	"github.com/kolide/launcher/pkg/agent/storage"
+	"github.com/kolide/launcher/pkg/agent/types"
+	"go.etcd.io/bbolt"
+)
+
+// MakeStores creates all the KVStores used by launcher
+func MakeStores(logger log.Logger, db *bbolt.DB) (map[storage.Store]types.KVStore, error) {
```
James-Pickett, line 14: .... just a thought, I wonder if we can live without this and just let knapsack lazily create the stores if they don't exist

pkg/agent/knapsack/knapsack.go
```diff
@@ -0,0 +1,84 @@
...
+	return k
+}
+
+func (k *Knapsack) getKVStore(storeType storage.Store) types.KVStore {
+	if k == nil {
+		return nil
+	}
+
+	// Ignoring ok value, this should only fail if an invalid storeType is provided
+	store, _ := k.stores[storeType]
```
James-Pickett, line 42: related to other comment, maybe create the store if it doesn't exist?

I guess that would mean rethinking the storage interface a bit.  Or perhaps a new interface that sits one level above store that allows for "bucket" creation. `StoreFactory` `StoreCreator` `StoreInit` etc

... just thinking out loud more than making a suggestion

pkg/agent/knapsack/knapsack.go
```diff
@@ -0,0 +1,84 @@
...
+	return store
+}
+
+func (k *Knapsack) AgentFlagsStore() types.KVStore {
```
James-Pickett, line 46: love this

```go
k := knapsack.New(...)
k.AgentFlagsStore.Get("foo")
```

feels very nice

James-Pickett (approved):

pkg/agent/knapsack/knapsack.go
seejdev: So, I think in the extension code or maybe osquery tables, there was some logic where the buckets were created upfront, and some corresponding error handling which expected buckets to exist when they were accessed later. That being said, we could change all the logic to be done either lazily or eagerly.


IMO one advantage of creating all the buckets upfront is that it makes error handling easier - any bucket creation errors can be handled independently, and later on you don't need to worry about if it exists or not. You can assume it does exists and If it doesn't exist, something is fubar.

James-Pickett: 👍 yea, that makes sense

PR #1103, opened 2023-03-14
Expose data and metrics about new TUF autoupdater

RebeccaMahany:

This is part 2 of moving to use our new TUF infrastructure.

Here, we expose data and metrics about the new TUF autoupdater:

1. New table `kolide_tuf_release_version` contains current release version for each binary/operating system/channel, according to the new TUF autoupdater, so that we can validate it pulls the correct version
1. New table `kolide_tuf_autoupdater_errors` contains TUF autoupdater errors that occurred within the last week 
1. Preexisting table `kolide_launcher_autoupdate_config` now contains the TUF server URL as well

I've also done a small refactor to run one autoupdater, rather than one for launcher and one for osqueryd. This is because they share the same TUF repo, so it didn't make sense to maintain two separate, identical TUF repos on disk.

<details>
<summary>Screenshots and results from example queries</summary>

`kolide_tuf_release_version`:

Screenshot 2023-03-15 at 12 16 04 PM

`kolide_tuf_autoupdater_errors`:

```
device_id,device_name,device_platform,device_os,device_owner,device_owner_type,status,error_message,warning_message,response_received_at,key,value

8,rebeccas-macbook-pro.local,darwin,macOS 13.2.1 Ventura,"","",successful,,,2023-03-15 16:10:35 UTC,1678896627,could not update metadata: tuf: missing remote metadata timestamp.json
```

`kolide_launcher_autoupdate_config`:

Screenshot 2023-03-14 at 5 44 20 PM

</details>
__________________________________

Subsequent PRs will tackle the following (order is not set in stone)
1. Point to production TUF instead of development
1. Download, verify, and untar downloads from new TUF, moving them to a new update directory located in the root directory
1. `findnew` looks for updates in new location, falling back to old location
1. Other improvements/refactors

Previous work:


pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -164,73 +163,90 @@ func (ta *TufAutoupdater) checkForUpdate() error {
...
 
 	return strings.TrimSuffix(strings.TrimPrefix(target, prefixToTrim), ".tar.gz")
 }
 
-func (ta *TufAutoupdater) rollingErrorCount() int {
-	ta.lock.RLock()
-	defer ta.lock.RUnlock()
+func (ta *TufAutoupdater) storeError(autoupdateErr error) {
+	ta.lock.Lock()
+	defer ta.lock.Unlock()
```
James-Pickett, line 212: do we still need a lock here at all? I think we can rely on the underlying storage to be thread safe.

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -164,73 +163,90 @@ func (ta *TufAutoupdater) checkForUpdate() error {
...
+		level.Debug(ta.logger).Log("msg", "could store autoupdater error", "err", err)
 	}
-
-	return errorCount
 }
 
-func (ta *TufAutoupdater) cleanUpOldErrorCounts() {
+func (ta *TufAutoupdater) cleanUpOldErrors() {
 	ta.lock.Lock()
 	defer ta.lock.Unlock()
```
James-Pickett, line 222: same question about lock here, I think storage is thread safe

James-Pickett (commented):
Nice!

pkg/autoupdate/tuf/autoupdate.go
RebeccaMahany: Ah yup you're right! I will triple-check and then remove 👍

James-Pickett (approved):

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -164,73 +161,84 @@ func (ta *TufAutoupdater) checkForUpdate() error {
...
+
+		return nil
+	}); err != nil {
+		level.Debug(ta.logger).Log("msg", "could not iterate over bucket items to determine which are expired", "err", err)
 	}
 
-	ta.errorCounter = errorsWithinLastDay
+	// Delete all old keys
+	if err := ta.store.Delete(keysToDelete...); err != nil {
+		level.Error(ta.logger).Log("msg", "could not delete old autoupdater errors from bucket", "err", err)
```
seejdev, line 242: Maybe should be at `Debug` level, as long as failing to delete the key doesn't cause bad behavior.

seejdev (dismissed):

pkg/autoupdate/tuf/autoupdate.go
RebeccaMahany: Yup good point, updated!

James-Pickett (approved):

PR #1105, opened 2023-03-16
adds logic and endpoint to accelerate control server fetch interval

James-Pickett:


ee/localserver/server.go
```diff
@@ -129,8 +129,8 @@ func New(db *bbolt.DB, kolideServer string, opts ...LocalServerOption) (*localSe
 	// mux.Handle("/query", ls.requestQueryHandler())
 	// curl localhost:40978/scheduledquery --data '{"name":"pack:kolide_device_updaters:agentprocesses-all:snapshot"}'
 	// mux.Handle("/scheduledquery", ls.requestScheduledQueryHandler())
-	// curl localhost:40978/controlservicefetch
-	// mux.Handle("/controlservicefetch", ls.requestControlServiceFetch())
+	// curl localhost:40978/acceleratecontrol  --data '{"interval":"250ms", "duration":"1s"}'
```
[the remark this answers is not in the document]
James-Pickett: cool, thanks! yes, the curl is just an example we use for local testing. The real thing would be in a krypto NaCL box from the JS bridge

ee/control/control.go
```diff
@@ -110,6 +112,45 @@ func (cs *ControlService) Stop() {
 	}
 }
 
+func (cs *ControlService) AccelerateRequestInterval(interval, duration time.Duration) error {
+	// preform a fetch now
```
seejdev, line 116:
```suggestion
	// perform a fetch now
```

cmd/launcher/control.go
```diff
@@ -42,7 +42,7 @@ func createControlService(ctx context.Context, logger log.Logger, store types.Ge
 		control.WithRequestInterval(opts.ControlRequestInterval),
 		control.WithStore(store),
 	}
-	service := control.New(logger, ctx, client, controlOpts...)
+	service := control.New(logger, client, controlOpts...)
```
directionless, line 45: Thank you for pulling out `ctx` it seems wrong here

ee/control/control.go
```diff
@@ -110,6 +112,45 @@ func (cs *ControlService) Stop() {
...
+	if err := cs.Fetch(); err != nil {
+		// if we got an error, log it and move on
+		level.Error(cs.logger).Log("msg", "failed to fetch data from control server. Not fatal, moving on", "err", err)
```
directionless, line 119: Usually s/error/debug/

ee/control/control.go
```diff
@@ -110,6 +112,45 @@ func (cs *ControlService) Stop() {
...
+	}
+
+	if interval <= 0 || duration <= 0 {
```
directionless, line 122: I think we should define a minimum interval, I'd say 5 or 10s, and use that if it's lower

directionless (commented):
I'm generally fine with this. I think parts will shuffle around as we bring in more interfaces and `kontext`

ee/control/control_test.go
```diff
@@ -273,3 +277,101 @@ func TestControlServicePersistLastFetched(t *testing.T) {
...
+			accelerationInterval: 250 * time.Millisecond,
+			accelerationDuration: 1 * time.Second,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// Overly verbose way to figure out how many fetches we expect to make it more human readable (hopefully)
```
RebeccaMahany, line 312: This test is readable and makes sense 👍

RebeccaMahany (dismissed):

ee/control/control.go
James-Pickett: done

ee/control/control.go
James-Pickett: done, nice use of sed =)

cmd/launcher/control.go
James-Pickett: it wasn't being used

RebeccaMahany (approved):

seejdev (approved):

PR #1109, opened 2023-03-22
adds server for desktop process to monitor parent

James-Pickett:


ee/desktop/runner/monitor_server.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type monitorServer struct {
+	server   *http.Server
+	listener net.Listener
+	mux      *http.ServeMux
+}
+
+func newMonitorServer() (*monitorServer, error) {
+	listener, err := net.Listen("tcp", "localhost:0")
```
James-Pickett, line 19: Maybe we should use a socket? I have a feeling eventually someone will want to know why we are always listening on a random ports

directionless: As you point out, we've seen weird socket issues. I say we try this, it's not that uncommon

cmd/launcher/desktop.go
```diff
@@ -174,20 +170,22 @@ func listenSignals(logger log.Logger) {
...
-		// https://cs.opensource.google/go/go/+/master:src/context/context.go;l=456?q=func%20WithDeadline&ss=go%2Fgo
-		// but the linter and the context.WithTimeout docs say to do it
-		cancel()
-		if err != nil || !exists {
+		response, err := client.Get(monitorUrl)
+		if response != nil && response.Body != nil {
+			response.Body.Close()
+		}
+
+		if err != nil {
```
RebeccaMahany, line 186: Would we potentially want to wait for maybe ~3 consecutive errors, maybe with a small amount of backoff? Just so that launcher being slow to respond once doesn't immediately kill the desktop process.

cmd/launcher/desktop.go
```diff
@@ -174,20 +170,22 @@ func listenSignals(logger log.Logger) {
...
 			level.Error(logger).Log(
```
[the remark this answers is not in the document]
James-Pickett: now debug

cmd/launcher/desktop.go
James-Pickett: I added some retries, I started to add a backoff ... but it's fussy to test so I gave up. You think retries is enough?

seejdev: 👍 I was worried about something like this too - we could end up with the opposite problem, where desktop/menubar can vanish.

cmd/launcher/desktop.go
```diff
@@ -174,24 +171,62 @@ func listenSignals(logger log.Logger) {
...
-		if err != nil || !exists {
-			level.Error(logger).Log(
-				"msg", "parent process gone",
+		response, err := client.Get(monitorUrl)
+		if err != nil {
+			errCount++
+
+			if errCount < maxErrCount {
+
+				backOff := interval + (interval * time.Duration(errCount))
```
directionless, line 191: This seems fine, but is also a bit overkill for `maxErrCount = 3`

ee/desktop/runner/monitor_server.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+
+	"github.com/kolide/kit/ulid"
+)
+
+type monitorServer struct {
```
directionless, line 12: I don't love the name `monitorServer` but okay

directionless (dismissed):
I didn't test this, but it seems fine

seejdev (approved):

RebeccaMahany (approved):

seejdev (approved):

seejdev:
LGTM! The connection re-use is nifty.

seejdev:
LGTM! The connection re-use is nifty.

PR #1108, opened 2023-03-22
Point to production TUF infra

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/954.

This is part 3 of moving to use our new TUF infrastructure: pointing to new production.
__________________________________

Subsequent PRs will tackle the following (order is not set in stone)
1. Download, verify, and untar downloads from new TUF, moving them to a new update directory located in the root directory
1. `findnew` looks for updates in new location, falling back to old location
1. Other improvements/refactors
1. An "update now" functionality tied to control server
1. Eventually removing the old notary autoupdater

Previous work:
1. https://github.com/kolide/launcher/pull/1081
1. https://github.com/kolide/launcher/pull/1103

PR #1111, opened 2023-03-24
Add library manager to handle TUF downloads

RebeccaMahany:

This is the next step toward using our new TUF infrastructure: actually downloading new versions of osqueryd and launcher from the mirror. We are using a new directory for staged updates and for our update library -- this should help improve our Nix support. We are also using the version of the download, rather than the timestamp, as the directory name for each update.

<details>

<summary>Sample directory structure</summary>

Run from root directory:

```
├── launcher
│   └── 1.0.3-12-g3e481cf
│       ├── Kolide.app
│       │   └── Contents
│       │       ├── CodeResources
│       │       ├── Info.plist
│       │       ├── MacOS
│       │       │   └── launcher
│       │       ├── Resources
│       │       │   └── Kolide.icns
│       │       ├── _CodeSignature
│       │       │   └── CodeResources
│       │       └── embedded.provisionprofile
│       └── launcher
└── osqueryd

11 directories, 7 files
```

</details>

<details>

<summary>Test notes</summary>

After decreasing the autoupdate interval to a short interval, run launcher.

Observe that after the interval, launcher has downloaded a new version of launcher to `<root directory>/updates/launcher`. It should not download a version of osquery unless its version of osquery is outdated.

Rename the version in `<root directory>/updates/launcher` (e.g. `mv <root directory>/updates/launcher/1.0.3-12-g3e481cf <root directory>/updates/launcher/1.0.1`) to make launcher perform another download.

After the interval, observe that launcher has downloaded the new version again. Repeat the rename with a new version name.

After the next interval, launcher should download a new version, and additionally remove the oldest versions in the update directory, preserving only two.

Repeat the rename, but this time choose a name that is not a valid semantic version. Observe that after the interval, the update that does not have a valid version tied to it is removed.
</details>
__________________________________

1. Add wait groups so launcher and osqueryd downloads can run in parallel
1. Other improvements/refactors (use knapsack/flags for configuration)

Previous work:


pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+			return nil, fmt.Errorf("could not make updates directory for %s: %w", binary, err)
+		}
+	}
+
+	return &ulm, nil
+}
+
+// updatesDirectory returns the update library location for the given binary.
+func (ulm *updateLibraryManager) updatesDirectory(binary string) string {
+	return filepath.Join(ulm.rootDirectory, fmt.Sprintf("%s-updates", binary))
```
directionless, line 72: seems okay. Maybe a slight preference for `<basedir>/<binary>` which would commonly expand to `<launcher root>/updates/<binary>`

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+
+// updateLibraryManager manages the update libraries for launcher and osquery.
+// It downloads and verifies new updates, and moves them to the appropriate
+// location in the library specified by the version associated with that update.
+// It also ensures that old updates are removed when they are no longer needed.
+type updateLibraryManager struct {
+	metadataClient  *client.Client // used to validate downloads
+	mirrorUrl       string         // dl.kolide.co
+	mirrorClient    *http.Client
+	rootDirectory   string
```
directionless, line 33: Is this meant to be the launcher root, or the ulm root? If it's the ulm root, I'd rather call it `baseDir` (or something). And I'm not sure it should get the launcher root

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+// stagedUpdatesDirectory returns the location for staged updates -- i.e. updates
+// that have been downloaded, but not yet verified.
+func (ulm *updateLibraryManager) stagedUpdatesDirectory(binary string) string {
+	return filepath.Join(ulm.rootDirectory, fmt.Sprintf("%s-staged-updates", binary))
+}
+
+// addToLibrary adds the given target file to the library for the given binary,
+// downloading and verifying it if it's not already there. After any addition
+// to the library, it cleans up older versions that are no longer needed.
+func (ulm *updateLibraryManager) AddToLibrary(binary string, targetFilename string) error {
```
directionless, line 84: Should binary be it's own type?

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+	// Check to see if the current running version is the version we were requested to add;
+	// return early if it is, but don't error out if we can't determine the current version.
+	currentVersion, err := ulm.currentRunningVersion(binary)
+	if err != nil {
+		level.Debug(ulm.logger).Log("msg", "could not get current running version", "binary", binary, "err", err)
+	} else {
+		if currentVersion.Original() == ulm.versionFromTarget(binary, targetFilename) {
```
directionless, line 91: This makes sense, but `currentVersion.Original() == ulm.versionFromTarget(binary, targetFilename)` feels awkward. Maybe the simplest though

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// updatesDirectory returns the update library location for the given binary.
+func (ulm *updateLibraryManager) updatesDirectory(binary string) string {
+	return filepath.Join(ulm.rootDirectory, fmt.Sprintf("%s-updates", binary))
+}
+
+// stagedUpdatesDirectory returns the location for staged updates -- i.e. updates
+// that have been downloaded, but not yet verified.
+func (ulm *updateLibraryManager) stagedUpdatesDirectory(binary string) string {
```
directionless, line 77: What happens if, somehow, an update triggers while one is running? They'll both end up in the same staging directory.

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+	_, err = io.Copy(out, resp.Body)
+	if err != nil {
+		return stagedUpdatePath, fmt.Errorf("could not write downloaded target %s to file %s: %w", targetFilename, stagedUpdatePath, err)
+	}
+
+	return stagedUpdatePath, nil
+}
+
+// verifyStagedUpdate checks the downloaded update against the metadata in the TUF repo.
+func (ulm *updateLibraryManager) verifyStagedUpdate(binary string, stagedUpdate string) error {
```
directionless, line 169: This is re-implementing TUF verification? Why not use TUF's Download directly?

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+	metadataClient  *client.Client // used to validate downloads
+	mirrorUrl       string         // dl.kolide.co
+	mirrorClient    *http.Client
+	rootDirectory   string
+	operatingSystem string
+	osquerier       localserver.Querier // used to query for current running osquery version
+	logger          log.Logger
+}
+
+func newUpdateLibraryManager(metadataClient *client.Client, mirrorUrl string, mirrorClient *http.Client, rootDirectory string, operatingSystem string, osquerier localserver.Querier, logger log.Logger) (*updateLibraryManager, error) {
```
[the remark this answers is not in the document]
RebeccaMahany: We can! I don't feel strongly about it. It's convenient to have them together because I'm reusing some test setup functions for both sets of tests, but that's not difficult to break apart either.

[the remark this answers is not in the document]
directionless: You're call! I don't feel strongly

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Sure, I like that 👍

cmd/launcher/launcher.go
```diff
@@ -346,11 +346,16 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		// Create a new TUF autoupdater
 		metadataClient := http.DefaultClient
 		metadataClient.Timeout = 1 * time.Minute
+		mirrorClient := http.DefaultClient
+		mirrorClient.Timeout = 1 * time.Minute
```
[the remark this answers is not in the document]
directionless: I don't know :)

Maybe 5minutes?

I wonder if we can track it and see.

pkg/autoupdate/tuf/library_manager.go
directionless: Implied here, is that we need to be able to configure what `basedir`, but it would default to root/updates

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: I agree it's a little awkward -- I'd looked into having `ulm.versionFromTarget` parse and return a `*semver.Version` but most of the places consuming that function just needed the string, so it didn't feel worth it.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: I didn't put it in this PR because it felt like it already contained enough changes, but my plan was to add two workers and a wait group to the TUF autoupdater, so that the `addToLibrary` calls for both launcher + osqueryd could run at the same time without one cutting off the other and triggering a launcher restart too soon. I think that'd also prevent an update happening when one is already running. Let me know what you think or if you have another suggestion?

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: We can't because we host our metadata on a different server than the binaries, and the go-tuf client implementation doesn't support that (see [here](https://github.com/theupdateframework/go-tuf/blob/master/client/remote_store.go)).

This function uses go-tuf/client's `VerifyDigest` for verification, which does the majority of the work and was the most I was able to pull out.

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,341 @@
...
+		operatingSystem: operatingSystem,
+		osquerier:       osquerier,
+		logger:          log.With(logger, "component", "tuf_autoupdater_library_manager"),
+	}
+
+	// Ensure our staged updates and updates directories exist
+	for _, binary := range binaries {
+		// Attempt out the staged updates directory, in case there are any archives there that we
+		// were unable to clean up previously.
+		_ = os.RemoveAll(ulm.stagedUpdatesDirectory(binary))
```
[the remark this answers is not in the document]
RebeccaMahany: I like the suggestion to move it to the delete/tidy routine -- that's a more reasonable location for it.

Right now, the delete/tidy routine just runs at the end of download, so it's not super regular. I think I probably need to adjust to have it run on an interval instead, but haven't thought that through all the way yet.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Sure!

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: I've updated this!

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: ☝️ this is what I went with

pkg/autoupdate/tuf/library_manager.go
directionless: So either we import the SplitStore, or do this?

RebeccaMahany: SplitStore only implements it on the server side, so we'd still have to add an equivalent for the client there anyway

cmd/launcher/launcher.go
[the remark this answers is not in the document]
RebeccaMahany: From https://pkg.go.dev/net/http#Client:

> 	// Timeout specifies a time limit for requests made by this
	// Client. The timeout includes connection time, any
	// redirects, and reading the response body. The timer remains
	// running after Get, Head, Post, or Do return and will
	// interrupt reading of the Response.Body.

I've bumped it up to 5min accordingly.

We could start timing download requests and saving that data to get a sense for a reasonable timeout, but I think that's probably enough of a change that it'd be outside the scope of this PR.

pkg/autoupdate/tuf/library_manager.go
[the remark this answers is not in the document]
RebeccaMahany: I've moved it to the tidy routine for now

pkg/autoupdate/tuf/library_manager.go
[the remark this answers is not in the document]
RebeccaMahany: I think I'll leave them together for now and we can always split them in the future if something changes

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,359 @@
...
+			// either in this updates library or in the original install location.
+			return nil
+		}
+	}
+
+	if ulm.alreadyAdded(binary, targetFilename) {
+		return nil
+	}
+
+	defer ulm.tidyLibrary(binary, currentVersion)
```
RebeccaMahany, line 103: TODO tidying should happen on startup and not post-update to avoid accidentally removing an in-use intermediate binary

seejdev: Would it make sense to split the "tidying" responsibility out? This is a large-ish file, maybe the "tidier" could be injected in and run immediately, or guaranteed to run before the manager does it's thing.

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -27,18 +26,34 @@ import (
...
 	Target string `json:"target"`
 }
 
+type librarian interface {
+	AddToLibrary(binary autoupdatableBinary, targetFilename string) error
+}
+
 type TufAutoupdater struct {
 	metadataClient  *client.Client
+	libraryManager  librarian
```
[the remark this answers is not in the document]
RebeccaMahany: I was trying very hard not to devolve to constant library-related puns in this PR 😅

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Maybe a `LibraryCustodian` 🤔 

Yeah that's reasonable, this is getting a little bit unwieldy. Let me see what makes sense for splitting it up!

RebeccaMahany: Both adding to the library and tidying the library require a) being able to query the current running version and b) knowing where to find the updates directories and staged update directories per binary. I couldn't find a way to split this up (given the shared knowledge and responsibilities) where the abstraction felt meaningful.

I think in the future we'll have the tidying routine run very early on, right after version selection for launcher and osqueryd, and then it might feel a little easier to refactor?

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -67,7 +83,8 @@ func WithUpdateCheckInterval(checkInterval time.Duration) TufAutoupdaterOption {
 	}
 }
 
-func NewTufAutoupdater(metadataUrl, rootDirectory string, metadataHttpClient *http.Client, store types.KVStore, opts ...TufAutoupdaterOption) (*TufAutoupdater, error) {
+func NewTufAutoupdater(metadataUrl, rootDirectory string, updateDirectory string, metadataHttpClient *http.Client,
+	mirrorUrl string, mirrorHttpClient *http.Client, store types.KVStore, osquerier localserver.Querier, opts ...TufAutoupdaterOption) (*TufAutoupdater, error) {
```
James-Pickett, line 87: I don't think we should pull in the `localserver` pkg just for this interface. I'm not sure why `localserver.Querier` was exported in the first placed. We should declare a new interface here. Perhaps if querier is becoming ubiquitious enough, we could create one some where like the `osquery` or `agent` pkg.

pkg/autoupdate/tuf/autoupdate_test.go
```diff
@@ -14,6 +17,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	localservermocks "github.com/kolide/launcher/ee/localserver/mocks"
```
James-Pickett, line 20: same as other comment, feels weird to have dependency on local server here

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+// downloading and verifying it if it's not already there. After any addition
+// to the library, it cleans up older versions that are no longer needed.
+func (ulm *updateLibraryManager) AddToLibrary(binary autoupdatableBinary, targetFilename string) error {
+	// Check to see if the current running version is the version we were requested to add;
+	// return early if it is, but don't error out if we can't determine the current version.
+	currentVersion, err := ulm.currentRunningVersion(binary)
+	if err != nil {
+		level.Debug(ulm.logger).Log("msg", "could not get current running version", "binary", binary, "err", err)
+	} else {
+		if currentVersion.Original() == ulm.versionFromTarget(binary, targetFilename) {
```
James-Pickett, line 93:
```suggestion
	} else if currentVersion.Original() == ulm.versionFromTarget(binary, targetFilename) {
```

James-Pickett (commented):
Nice work!

pkg/autoupdate/tuf/autoupdate.go
directionless: Yeah, don't pull in `localserver`. Just declare it here.

Counterpoint -- I think it'll move into knapsack, so maybe we just ignore it

pkg/autoupdate/tuf/util_darwin.go
```diff
@@ -0,0 +1,21 @@
+//go:build darwin
```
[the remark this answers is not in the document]
directionless: I don't think I've really reasoned about it, but I got the other way -- I'd rather split this into platform specific helpers, and not have a switch at runtime.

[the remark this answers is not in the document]
James-Pickett: My reasoning behind it is that it's simply less files to maintain. I get annoyed when I update a MacOS specific file and then CI dings me because I did not make the corresponding update on Windows. However, I would not hold up the PR for this, so I'm down with whatever =)

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,359 @@
...
+// updateLibraryManager manages the update libraries for launcher and osquery.
+// It downloads and verifies new updates, and moves them to the appropriate
+// location in the library specified by the version associated with that update.
+// It also ensures that old updates are removed when they are no longer needed.
+type updateLibraryManager struct {
+	metadataClient  *client.Client // used to validate downloads
+	mirrorUrl       string         // dl.kolide.co
+	mirrorClient    *http.Client
+	baseDir         string
+	operatingSystem string
```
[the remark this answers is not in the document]
RebeccaMahany: Nah, I think I carried this over from the release side without thinking about it much -- I'll remove.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Sure, that's a good point. I'll update.

pkg/autoupdate/tuf/mock_querier.go
James-Pickett, line 1: I think it's better to put mocks in their own package like `./tuf/mocks` so that the mock and any dependencies are not included in the binary. I bet using `mock_querier_test.go` would work as well.

RebeccaMahany: I had them in `./tuf/mocks` originally since that's the mockery default, but had to move them in-package to avoid an import cycle: when I created a `type AutoupdatableBinary` used by the librarian interface that we're mocking (`AddToLibrary(binary AutoupdatableBinary, targetFilename string) error`), the mocks package would have to import the tuf package for the function signature, and the tuf package would import the mocks package for the test. So I moved them in-package to fix this (and stopped exporting `autoupdatableBinary`). Let me know if there's another way around this that I missed?

RebeccaMahany: Sorry, missed the last suggestion you made, which solves the problem -- updated both!

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Added to the other PR here 🙂 https://github.com/kolide/launcher/pull/1119/commits/c75e4385550bab643665b788e38c94433a54608c

directionless (commented):
A couple of comments.

I'm not any really need to be addressed. Semver seems a little weird, but I see it's there for the sorting.

I think merging the verify into the staging would be cleaner -- we wouldn't need to write then read back. And we'd ensure we never write bad data to disk. But I could live with it.

seejdev (dismissed):

seejdev:
I wasn't too familiar with this area of the code base, and this change makes it a lot easier to understand what's going on!

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+
+		// Get the current running version to preserve it when tidying the available updates
+		var currentVersion *semver.Version
+		var err error
+		switch binary {
+		case binaryOsqueryd:
+			// The osqueryd client may not have initialized yet, so retry the version
+			// check a couple times before giving up
+			osquerydVersionCheckRetries := 5
+			for i := 0; i < osquerydVersionCheckRetries; i += 1 {
```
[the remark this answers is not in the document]
RebeccaMahany: Currently, all of this is ultimately run by a `rungroup.Run` (Run => TufAutoupdater.Execute => updateLibraryManager.TidyLibrary), so yeah, within a goroutine.

We did discuss doing the tidying on launcher startup (after version selection), which sounds like it would be blocking in that case. Should we talk through that idea more? I am also fine with execing it with --version. 🤷‍♀️

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	defer out.Close()
+
+	resp, err := ulm.mirrorClient.Get(ulm.mirrorUrl + fmt.Sprintf("/kolide/%s/%s/%s", binary, runtime.GOOS, targetFilename))
+	if err != nil {
+		return stagedUpdatePath, fmt.Errorf("could not make request to download target %s: %w", targetFilename, err)
+	}
+	defer resp.Body.Close()
+
+	_, err = io.Copy(out, resp.Body)
```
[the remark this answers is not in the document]
RebeccaMahany: I think I was modeling it off the old code, yeah! Looking at how go-tuf does it -- they write to a file (os.File fulfilling the [Destination interface](https://github.com/theupdateframework/go-tuf/blob/master/client/client.go#L868-L871)) that is also deleted if verification fails: https://github.com/theupdateframework/go-tuf/blob/master/client/client.go#L873. It is a little more efficient because they're writing while generating the metadata at the same time -- I'll take a look at switching to something similar!

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+
+	const numberOfVersionsToKeep = 3
+
+	rawVersionsInLibrary, err := filepath.Glob(filepath.Join(ulm.updatesDirectory(binary), "*"))
+	if err != nil {
+		level.Debug(ulm.logger).Log("msg", "could not glob for updates to tidy updates library", "err", err)
+		return
+	}
+
+	removeUpdate := func(v string) {
```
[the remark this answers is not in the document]
RebeccaMahany: I don't think it's increasing readability -- I think I could also combine it with the inline `removeBrokenUpdateDir` above in `moveVerifiedUpdate`, since they're basically doing the same thing. I'll do that 💪

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	return nil
+}
+
+// currentRunningVersion returns the current running version of the given binary.
+func (ulm *updateLibraryManager) currentRunningVersion(binary autoupdatableBinary) (*semver.Version, error) {
+	switch binary {
+	case binaryLauncher:
+		currentVersion, err := semver.NewVersion(version.Version().Version)
```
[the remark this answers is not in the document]
RebeccaMahany: Looks like we don't need it! I'll pull it out.

pkg/autoupdate/tuf/library_manager.go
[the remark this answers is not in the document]
directionless: I don't think we know the simple answer, so I might be inclined to solve it later.

With the initial delay on autoupdater, I think we don't have an immediate problem. But it feels like it's not right.

pkg/autoupdate/tuf/library_manager.go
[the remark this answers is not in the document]
directionless: I'm coming at it thinking that writing and deleting feels like extra steps, potential for errors, and leaving known-bad data around. I'd probably dump the body into an io.Tee, checksum and buffer. If the checksum passes, then we can write it to disk and move on.

Though that Desintaition interface is cute too. But I don't think we need the flexibility.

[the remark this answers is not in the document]
RebeccaMahany: Just seeing this after going with a slightly different implementation that mostly leans on go-tuf -- 

```
	// Request download from mirror
	resp, err := ulm.mirrorClient.Get(ulm.mirrorUrl + fmt.Sprintf("/kolide/%s/%s/%s", binary, runtime.GOOS, targetFilename))
	if err != nil {
		return stagedUpdatePath, fmt.Errorf("could not make request to download target %s: %w", targetFilename, err)
	}
	defer resp.Body.Close()

	// Wrap the download in a LimitReader so we download at most localMeta.Length bytes
	stream := io.LimitReader(resp.Body, localTargetMetadata.Length)

	// Read the target file, simultaneously writing it to the staging directory and generating its metadata
	actualTargetMeta, err := util.GenerateTargetFileMeta(io.TeeReader(stream, out), localTargetMetadata.HashAlgorithms()...)
	if err != nil {
		return stagedUpdatePath, fmt.Errorf("could not write downloaded target %s to file %s and compute its metadata: %w", targetFilename, stagedUpdatePath, err)
	}

	// Verify the actual download against the confirmed local metadata
	if err := util.TargetFileMetaEqual(actualTargetMeta, localTargetMetadata); err != nil {
		return stagedUpdatePath, fmt.Errorf("verification failed for target %s staged at %s: %w", targetFilename, stagedUpdatePath, err)
	}
```

I'll go ahead and add a buffer instead of directly teeing it to `out`, but otherwise I think this addresses what you're suggesting!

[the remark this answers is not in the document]
RebeccaMahany: Updated ✅

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,389 @@
...
+	if err != nil {
+		return fmt.Errorf("could not compute digest for target %s to verify it: %w", stagedUpdate, err)
+	}
+
+	fileInfo, err := os.Stat(stagedUpdate)
+	if err != nil {
+		return fmt.Errorf("could not get info for staged update at %s: %w", stagedUpdate, err)
+	}
+
+	// Where the file lives in the binary bucket -- we can't use filepath.Join here because on Windows,
```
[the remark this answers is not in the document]
RebeccaMahany: Ended up getting to remove this entirely

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	return nil
+}
+
+// alreadyAdded checks if the given target already exists in the update library.
+func (ulm *updateLibraryManager) alreadyAdded(binary autoupdatableBinary, targetFilename string) bool {
+	updateDirectory := filepath.Join(ulm.updatesDirectory(binary), ulm.versionFromTarget(binary, targetFilename))
+
+	return autoupdate.CheckExecutable(context.TODO(), executableLocation(updateDirectory, binary), "--version") == nil
```
directionless, line 121: Does this actually work on windows given the weird GUI output issues we had?

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,349 @@
...
+
+// versionFromTarget extracts the semantic version for an update from its filename.
+func (ulm *updateLibraryManager) versionFromTarget(binary autoupdatableBinary, targetFilename string) string {
+	// The target is in the form `launcher-0.13.6.tar.gz` -- trim the prefix and the file extension to return the version
+	prefixToTrim := fmt.Sprintf("%s-", binary)
+
+	return strings.TrimSuffix(strings.TrimPrefix(targetFilename, prefixToTrim), ".tar.gz")
+}
+
+// stageUpdate downloads the update indicated by `targetFilename` and verifies it against
```
directionless, line 132:
```suggestion
// stageAndVerifyUpdate downloads the update indicated by `targetFilename` and verifies it against
```

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,349 @@
...
+// the given, validated local metadata.
+func (ulm *updateLibraryManager) stageAndVerifyUpdate(binary autoupdatableBinary, targetFilename string, localTargetMetadata data.TargetFileMeta) (string, error) {
+	// Create the staging file
+	stagedUpdatePath := filepath.Join(ulm.stagedUpdatesDirectory(binary), targetFilename)
+	out, err := os.Create(stagedUpdatePath)
+	if err != nil {
+		return "", fmt.Errorf("could not create file at %s: %w", stagedUpdatePath, err)
+	}
+	defer out.Close()
```
directionless, line 141: I think this is fine, but it might make sense to explicitly close before the return and check the error.

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	// Make sure that the binary is executable
+	if err := os.Chmod(executableLocation(newUpdateDirectory, binary), 0755); err != nil {
+		ulm.removeUpdate(binary, targetVersion)
+		return fmt.Errorf("could not set +x permissions on executable: %w", err)
+	}
+
+	// Validate the executable
+	if err := autoupdate.CheckExecutable(context.TODO(), executableLocation(newUpdateDirectory, binary), "--version"); err != nil {
```
directionless, line 194: I don't love that these checks are happening in the library path and not a staged or random path. But I can live with it. (we check the executability before invoking anyhow)

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -0,0 +1,349 @@
...
+	"time"
+
+	"github.com/Masterminds/semver"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/fsutil"
+	"github.com/kolide/kit/version"
+	"github.com/kolide/launcher/pkg/autoupdate"
+	"github.com/theupdateframework/go-tuf/data"
+	"github.com/theupdateframework/go-tuf/util"
```
directionless, line 24: mild preference for something like 
```suggestion
	tufutil "github.com/theupdateframework/go-tuf/util"
```

I spent a minute trying to figure out what util was when I read it in the code. 😆

directionless (approved):

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Will address in https://github.com/kolide/launcher/pull/1119

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Updated to a defer func() that logs the error if present -- I think I prefer that over having to call + check out.Close() in the 5 places we could return, and if something's very wrong we should catch it when verifying the executable anyway

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: It's hard to tell because when testing, the running version will always be a local dirty build that doesn't match what TUF has. (I do know the check worked correctly for osquery on Windows but I don't think that was affected by the same issue?)

But this is reusing the same check from the autoupdate package, so it should at least be consistent with what we have currently?

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -138,7 +138,11 @@ func (ulm *updateLibraryManager) stageAndVerifyUpdate(binary autoupdatableBinary
 	if err != nil {
 		return "", fmt.Errorf("could not create file at %s: %w", stagedUpdatePath, err)
 	}
-	defer out.Close()
+	defer func() {
+		if err := out.Close(); err != nil {
```
directionless, line 142: I think returning the error is more correct. If the file can't close, it means it's corrupt on disk.

RebeccaMahany: Will update in next PR

RebeccaMahany: Addressed in https://github.com/kolide/launcher/pull/1119

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Addressed in https://github.com/kolide/launcher/pull/1119/commits/51286ad8923a7cf0562f2d9eeb291580af702e6f

PR #1112, opened 2023-03-27
Made relativeTime text output title case to match k2 style

seejdev:
Made some letters go 🆙


RebeccaMahany (approved):
I'm wondering for the future if we'd want to think about handling casing entirely on the launcher side -- we could get into the reverse of this situation easily if the text formatting changes on the k2 side. Maybe something to chat about in the future with @blaedj?

I think usually "a" and "in" wouldn't be capitalized in these strings, but it doesn't look like we have a formal style guide at the moment and this is more visually consistent than it was before, so 🚀 🔥

James-Pickett (approved):

blaedj:
> Maybe something to chat about in the future with @blaedj?

Issue #1083, opened 2023-03-08
Our agent should make better use of go interfaces


directionless:
This is a bit of internal happiness, developer UX, and the realization that we didn't get our code organization quite right. In short, we should make better use of interfaces. Some examples
* We pass the bbolt db client around. This needlessly ties us to bbolt, and makes our testing slower than it should be WRT to disk 
* I kinda feel like there should be some `agent.Runtime` interface. Something that ties together flags, keys, database things. I'm not really sure quite what this looks like, but I can imagine some things
  - A good answer here, will make it much simpler to use the control server to set and update flags.
  - Some discussion in #1029

directionless:
We made good progress here, as we moved to things like knapsack. And it's really an idea, not a project or bug,

PR #1114, opened 2023-03-28
Introducing Flags interface, with implementation for control server related flags

seejdev:
Related to #1086, and #1083

### Motivation for these changes:

We have a number of launcher flags controlling various behaviors, specified through different means, and generally accessed and modified in non-standard and inconsistent ways. This PR lays the foundation for streamlining:

- Hooking into our pervasive `Knapsack` to enable disparate areas of the code base to utilize common flag functionality easily and with low developer overhead.
- How flags are defined, their defaults, their available command line options, and how they are stored in key-value stores.
- What flag values take precedence over others.
- Sanitizing flag values to prevent unreasonable values being used.
- Enabling temporary overrides for flags in a generic fashion, avoiding custom one-off code written to accomplish short-lived modifications to flags.

### What is in this PR:

- The `Flags` interface: provides an easy-to-use and readable API for setting/getting launcher flags. It's been added to the `Knapsack` struct, so anywhere `Knapsack` is passed to, you can access `Flags` just as easily.
- Implementations of default flag values, and command line options which can take precedence over defaults.
- Implementation of stored flag values, backed by a key-value store. This is principally designed for control server to push updates of launcher flags, and these can take precedence over defaults, and command line options.
- Implementation of flag value overrides, where a flag can be temporarily overridden for a period of time with a special-case value.
- Sanitization of integer flag values to constrain values to sensible limits.
- Re-working of the `types.Updater` interface to better separate JSON parsing from the logic of bulk-replacing the contents of the `agent_flags` key-value store. This was necessary to enable clients of the interface to have better control and knowledge of what keys are being changed.
- Implementation of a flag change notification framework, where clients can register as observers of various `FlagKey` and receive notifications when they change.
- Re-working of the control server request interval acceleration feature to tie into the `Flags` implementation, notification framework, and common temporary override functionality.
- Using all of the above to re-implement control server flags: `DesktopEnabled`, `DebugServerData`, `ForceControlSubsystems`, `ControlServerURL`, `ControlRequestInterval`, `DisableControlTLS`, `InsecureControlTLS`

### DIFF
--- a/cmd/launcher/desktop.go
+++ b/cmd/launcher/desktop.go
@@ -68,18 +68,26 @@ func runDesktop(args []string) error {
 		return fmt.Errorf("parsing flags: %w", err)
 	}
 
-	user, err := user.Current()
-	if err != nil {
-		return fmt.Errorf("getting current user: %w", err)
-	}
-
 	// set up logging
 	logger := logutil.NewServerLogger(*fldebug)
 	logger = log.With(logger,
 		"subprocess", "desktop",
 		"pid", os.Getpid(),
-		"uid", user.Uid,
 	)
+
+	// Try to get the current user, so we can use the UID for logging. Not a fatal error if we can't, though
+	user, err := user.Current()
+	if err != nil {
+		level.Debug(logger).Log(
+			"msg", "error getting current user",
+			"err", err,
+		)
+	} else {
+		logger = log.With(logger,
+			"uid", user.Uid,
+		)
+	}
+
 	level.Info(logger).Log("msg", "starting")
 
 	if *flsocketpath == "" {
@@ -173,6 +181,7 @@ func listenSignals(logger log.Logger) {
 // monitorParentProcess continuously checks to see if parent is a live and sends on provided channel if it is not
 func monitorParentProcess(logger log.Logger, monitorUrl string, interval time.Duration) {
 	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
 
 	client := http.Client{
 		Timeout: interval,
--- a/ee/consoleuser/consoleuser_linux.go
+++ b/ee/consoleuser/consoleuser_linux.go
@@ -39,7 +39,7 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 		// get the active property of the session, this command does not respect the --output=json flag
 		output, err := exec.CommandContext(ctx, "loginctl", "show-session", s.Session, "--value", "--property=Active").Output()
 		if err != nil {
-			return nil, fmt.Errorf("loginctl show-session: %w", err)
+			return nil, fmt.Errorf("loginctl show-session (for uid %d): %w", s.UID, err)
 		}
 
 		if strings.Trim(string(output), "\n") != "yes" {
--- a/ee/consoleuser/consoleuser_windows.go
+++ b/ee/consoleuser/consoleuser_windows.go
@@ -6,7 +6,6 @@ package consoleuser
 import (
 	"context"
 	"fmt"
-	"os/user"
 	"path/filepath"
 
 	"github.com/shirou/gopsutil/process"
@@ -28,7 +27,7 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 	for _, explorerProc := range explorerProcs {
 		uid, err := processOwnerUid(ctx, explorerProc)
 		if err != nil {
-			return nil, fmt.Errorf("getting process owner uid: %w", err)
+			return nil, fmt.Errorf("getting process owner uid (for pid %d): %w", explorerProc.Pid, err)
 		}
 		uidsMap[uid] = struct{}{}
 	}
@@ -53,7 +52,7 @@ func ExplorerProcess(ctx context.Context, uid string) (*process.Process, error) 
 	for _, proc := range explorerProcs {
 		procOwnerUid, err := processOwnerUid(ctx, proc)
 		if err != nil {
-			return nil, fmt.Errorf("getting process owner uid: %w", err)
+			return nil, fmt.Errorf("getting explorer process owner uid (for pid %d): %w", proc.Pid, err)
 		}
 
 		if uid == procOwnerUid {
@@ -91,13 +90,11 @@ func explorerProcesses(ctx context.Context) ([]*process.Process, error) {
 func processOwnerUid(ctx context.Context, proc *process.Process) (string, error) {
 	username, err := proc.UsernameWithContext(ctx)
 	if err != nil {
-		return "", fmt.Errorf("getting process username: %w", err)
+		return "", fmt.Errorf("getting process username (for pid %d): %w", proc.Pid, err)
 	}
 
-	user, err := user.Lookup(username)
-	if err != nil {
-		return "", fmt.Errorf("looking up user: %w", err)
-	}
-
-	return user.Uid, nil
+	// Looking up the proper UID (which on Windows, is a SID) seems to be problematic and
+	// can fail for reasons we don't quite understand. We just need something to uniquely
+	// identify the user, so on Windows we use the username instead of numeric UID.
+	return username, nil
 }
--- a/ee/control/consumers/notificationconsumer/notify_consumer.go
+++ b/ee/control/consumers/notificationconsumer/notify_consumer.go
@@ -190,6 +190,8 @@ func (nc *NotificationConsumer) Interrupt(err error) {
 func (nc *NotificationConsumer) runCleanup(ctx context.Context) {
 	ctx, nc.cancel = context.WithCancel(ctx)
 	t := time.NewTicker(nc.cleanupInterval)
+	defer t.Stop()
+
 	for {
 		select {
 		case <-ctx.Done():
--- a/ee/control/control.go
+++ b/ee/control/control.go
@@ -112,6 +112,7 @@ func (cs *ControlService) Interrupt(err error) {
 
 func (cs *ControlService) Stop() {
 	level.Info(cs.logger).Log("msg", "control service stopping")
+	cs.requestTicker.Stop()
 	if cs.cancel != nil {
 		cs.cancel()
 	}
--- a/ee/desktop/menu/action_open_url.go
+++ b/ee/desktop/menu/action_open_url.go
@@ -1,9 +1,6 @@
 package menu
 
 import (
-	"os/exec"
-	"runtime"
-
 	"github.com/go-kit/kit/log/level"
 )
 
@@ -19,23 +16,4 @@ func (a actionOpenURL) Perform(m *menu) {
 			"URL", a.URL,
 			"err", err)
 	}
-}
-
-// open opens the specified URL in the default browser of the user
-// See https://stackoverflow.com/a/39324149/1705598
-func open(url string) error {
-	var cmd string
-	var args []string
-
-	switch runtime.GOOS {
-	case "windows":
-		cmd = "cmd"
-		args = []string{"/c", "start"}
-	case "darwin":
-		cmd = "/usr/bin/open"
-	default: // "linux", "freebsd", "openbsd", "netbsd"
-		cmd = "xdg-open"
-	}
-	args = append(args, url)
-	return exec.Command(cmd, args...).Start()
 }
--- a/ee/desktop/menu/action_open_url_darwin.go
+++ b/ee/desktop/menu/action_open_url_darwin.go
@@ -1,0 +1,14 @@
+//go:build darwin
+// +build darwin
+
+package menu
+
+import (
+	"os/exec"
+)
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	return exec.Command("/usr/bin/open", url).Start()
+}
--- a/ee/desktop/menu/action_open_url_linux.go
+++ b/ee/desktop/menu/action_open_url_linux.go
@@ -1,0 +1,14 @@
+//go:build linux
+// +build linux
+
+package menu
+
+import (
+	"os/exec"
+)
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	return exec.Command("xdg-open", url).Start()
+}
--- a/ee/desktop/menu/action_open_url_windows.go
+++ b/ee/desktop/menu/action_open_url_windows.go
@@ -1,0 +1,19 @@
+//go:build windows
+// +build windows
+
+package menu
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	cmd := exec.Command("cmd", "/C", "start", url)
+	// https://stackoverflow.com/questions/42500570/how-to-hide-command-prompt-window-when-using-exec-in-golang
+	// Otherwise the cmd window will appear briefly
+	cmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true}
+	return cmd.Start()
+}
--- a/ee/desktop/runner/runner_test.go
+++ b/ee/desktop/runner/runner_test.go
@@ -151,7 +151,11 @@ func TestDesktopUserProcessRunner_Execute(t *testing.T) {
 			if tt.cleanShutdown || (os.Getenv("CI") == "true" && runtime.GOOS == "linux") {
 				assert.Len(t, r.uidProcs, 0, "unexpected process: logs: %s", logBytes.String())
 			} else {
-				assert.Contains(t, r.uidProcs, user.Uid)
+				if runtime.GOOS == "windows" {
+					assert.Contains(t, r.uidProcs, user.Username)
+				} else {
+					assert.Contains(t, r.uidProcs, user.Uid)
+				}
 				assert.Len(t, r.uidProcs, 1)
 
 				if len(tt.logContains) > 0 {
--- a/ee/localserver/request-id.go
+++ b/ee/localserver/request-id.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os/user"
+	"runtime"
 	"time"
 
 	"github.com/go-kit/kit/log/level"
@@ -107,12 +108,18 @@ func consoleUsers() ([]*user.User, error) {
 		}
 
 		for _, uid := range uids {
-			user, err := user.LookupId(uid)
+			var err error
+			var u *user.User
+			if runtime.GOOS == "windows" {
+				u, err = user.Lookup(uid)
+			} else {
+				u, err = user.LookupId(uid)
+			}
 			if err != nil {
 				return err
 			}
 
-			users = append(users, user)
+			users = append(users, u)
 		}
 		return nil
 	}, maxDuration, 250*time.Millisecond)
--- a/pkg/autoupdate/tuf/autoupdate.go
+++ b/pkg/autoupdate/tuf/autoupdate.go
@@ -153,7 +153,9 @@ func (ta *TufAutoupdater) Execute() (err error) {
 	ta.libraryManager.TidyLibrary()
 
 	checkTicker := time.NewTicker(ta.checkInterval)
+	defer checkTicker.Stop()
 	cleanupTicker := time.NewTicker(12 * time.Hour)
+	defer cleanupTicker.Stop()
 
 	for {
 		select {
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -2,11 +2,16 @@ package log
 
 import (
 	"bytes"
+	"fmt"
+	"os"
 	"regexp"
+	"strconv"
 	"strings"
 
 	kitlog "github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/shirou/gopsutil/host"
+	"github.com/shirou/gopsutil/process"
 )
 
 // OsqueryLogAdapater creates an io.Writer implementation useful for attaching
@@ -32,6 +37,8 @@ func WithLevelFunc(lf func(kitlog.Logger) kitlog.Logger) Option {
 }
 
 var callerRegexp = regexp.MustCompile(`[\w.]+:\d+]`)
+
+var pidRegex = regexp.MustCompile(`Refusing to kill non-osqueryd process (\d+)`)
 
 func extractOsqueryCaller(msg string) string {
 	return strings.TrimSuffix(callerRegexp.FindString(msg), "]")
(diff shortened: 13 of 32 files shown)


James-Pickett (commented):
I like the general shape of this. Cool to see generics in action. Look forward to more tests.

ee/desktop/runner/runner.go
```diff
@@ -101,10 +103,10 @@ func WithProcessSpawningEnabled(enabled bool) desktopUsersProcessesRunnerOption
 	}
 }
 
-// WithGetter sets the key/value getter for agent flags
-func WithGetter(getter types.Getter) desktopUsersProcessesRunnerOption {
+// WithKnapsack sets the knapsack
+func WithKnapsack(knapsack *knapsack.Knapsack) desktopUsersProcessesRunnerOption {
```
directionless, line 107: I'm fine with this, but I think `knapsack` is basically required, and should be a primary argument to `New`

ee/desktop/runner/runner.go
```diff
@@ -179,6 +181,9 @@ func New(opts ...desktopUsersProcessesRunnerOption) (*DesktopUsersProcessesRunne
 	runner.writeDefaultMenuTemplateFile()
 	runner.refreshMenu()
 
+	// Observe DesktopEnabled changes to know when to enable/disable process spawning
+	runner.knapsack.Flags.RegisterChangeObserver(runner, flags.DesktopEnabled)
```
directionless, line 185: Hrm. This feels clean, but I kinda wish we didn't have to pull in `flags` for it.

pkg/agent/flags/flag_values_constraints.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// FlagValueConstraints returns a map of FlagKey->flagValueConstraints
+func FlagValueConstraints() map[FlagKey]flagValueConstraint {
+	constraints := make(map[FlagKey]flagValueConstraint)
+
+	// Below is a list of integer FlagKeys and their constraints
+	constraints[ControlRequestInterval] = flagValueConstraint{
+		min: int64(5 * time.Second),
+		max: int64(1 * time.Hour),
```
directionless, line 14:
```suggestion
		max: int64(10 * time.Minute),
```

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+}
+
+func (fc *FlagController) SetDesktopEnabled(enabled bool) error {
+	return set(fc, DesktopEnabled, enabled)
+}
+func (fc *FlagController) DesktopEnabled() bool {
+	return get[bool](fc, DesktopEnabled)
+}
```
directionless, line 304: I think I like the explicit interfaces for these. I'm inclined to add explicit overrides here, instead of a general `SetOverride`.

Thinking a bit about how to simply further, I think of what's hard to track, is that `DesktopEnabled` calls `get` which spiders out to several other layers. Which totally makes sense, and might be the most correct thing. But is it cleaner to put more logic here? And code gen for the boilerplate?

Or is stashing it all in `get` the correct method?

directionless (commented):
There's a lot of meat here. But it feels very hard to understand.

pkg/agent/flags/flag_values_stored.go
```diff
@@ -0,0 +1,47 @@
+package flags
+
+import (
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/agent/types"
+)
+
+type storedFlagValues struct {
```
RebeccaMahany, line 9: I had to go back and check the docs a couple times to figure out/understand where the stored flag values came from/what they were for. I think the naming was a little confusing for me because the fact that they live in a KVStore isn't the distinguishing thing about them -- it's that they're set by k2/the control server, i.e. a cloud override to the flags shipped with the installer. Could we consider a name that explains their purpose/source? Maybe `cloudFlagValues` or `cloudControlledFlagValues`? (Not suggesting we use `cloud override` because `override` already has another meaning here 😅 )

ee/desktop/runner/runner.go
seejdev: Agree, fixed.

docs/architecture/knapsack.md
```diff
@@ -0,0 +1,27 @@
+# Knapsack
+
+`Knapsack` is an object instantiated very early on in the launcher startup process, and contains an inventory of useful services which are used throughout launcher. Its purpose is to provide a convenient API to access commonly used resources and functionality, while minimizing function argument list bloat.
```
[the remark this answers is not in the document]
seejdev: I can write up some thoughts on guidelines. My general thought is use the most minimally scoped dependencies when possible - this makes testing/mocking easier. I've done this in many places that only require a `KVStore` and nothing else. Once you need more than one dependency, like a store and access to flags, I think it makes sense to inject `Knapsack`, as that's really part of the purpose it's serving - minimizing ever-growing argument lists of dependencies.

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,346 @@
...
+				anyvalue = int64Value
+			}
+		}
+	case reflect.String:
+		stringValue, ok := override.Value().(string)
+		if ok {
+			anyvalue = stringValue
+		}
+	default:
+		level.Debug(fc.logger).Log("msg", "unsupported type of override flag", "type", reflect.TypeOf(t).Kind())
```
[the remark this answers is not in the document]
seejdev: Agreed!

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,346 @@
...
+	if !exists {
+		return *new(T), false
+	}
+
+	typedValue, ok := value.(T)
+	if !ok {
+		var t T
+		level.Debug(fc.logger).Log("msg", "cmd line flag type assertion failed", "type", reflect.TypeOf(t).Kind())
+		return *new(T), false
+	}
```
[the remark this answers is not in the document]
seejdev: I'll add this in - I thought I had done this, but may have lost it in some reorganization.

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,346 @@
...
+		boolValue, ok := override.Value().(bool)
+		if ok {
+			anyvalue = boolValue
+		}
+	case reflect.Int64:
+		int64Value, ok := override.Value().(int64)
+		if ok {
+			// Integers are sanitized to avoid unreasonable values
+			if fc.sanitizer != nil {
+				anyvalue = fc.sanitizer.Sanitize(key, int64Value)
```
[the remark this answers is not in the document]
seejdev: I opted for doing only on the `get` side because the values could still be modified after storing - i.e. modify the database and replace values that were previously sanitized. Although it could lead to `Sanitize` being called more frequently, it's a very cheap function so I'm not concerned about a performance hit, and overall this seemed like a simpler solution.

pkg/agent/flags/flag_values_defaults.go
```diff
@@ -0,0 +1,21 @@
...
+// DefaultFlagValues returns a FlagValues struct of default FlagKeys and values
+func DefaultFlagValues() *FlagValues {
+	f := NewFlagValues()
+
+	// Below is a list of every FlagKey and it's default value.
+	// When adding a new FlagKey, make sure to add it's default value below.
+	f.Set(DesktopEnabled, false)
+	f.Set(DebugServerData, false)
+	f.Set(ForceControlSubsystems, false)
+	f.Set(ControlServerURL, "")
```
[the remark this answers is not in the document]
seejdev: Possibly - worth a discussion!

cmd/launcher/launcher.go
```diff
@@ -265,16 +259,15 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 
 	// runEECode feels like it should move up to the opts level.
 	// We have some stuff there that sets `controlServerURL`
-	runEECode := opts.ControlServerURL != "" || opts.IAmBreakingEELicense
+	runEECode := opts.ControlServerURL != "" || opts.IAmBreakingEELicense // TODO
```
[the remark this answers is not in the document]
seejdev: Nope I'll remove this.

pkg/agent/flags/flag_value_sanitizer.go
```diff
@@ -0,0 +1,41 @@
...
+	s := &flagValueSanitizer{
+		constraints: constraints,
+	}
+
+	return s
+}
+
+// Sanitize returns a sanitized (clamped) value for a flag, if constraints for
+// the flag value have been provided. If undefined, the original value is returned.
+func (s *flagValueSanitizer) Sanitize(key FlagKey, value int64) int64 {
```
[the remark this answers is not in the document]
seejdev: Yeah I thought about string value sanitization. The idea I had was a client provides a sanitizer func like `func (string value) string` which is invoked by the sanitizer code. This would enable each disparate component to specify their own logic for what is, and what is not, a valid string value.

pkg/agent/flags/flag_controller.go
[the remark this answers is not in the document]
seejdev: Done!

pkg/agent/flags/flag_controller.go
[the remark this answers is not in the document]
RebeccaMahany: That makes sense, thanks for explaining!

docs/architecture/knapsack.md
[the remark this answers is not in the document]
RebeccaMahany: That sounds reasonable to me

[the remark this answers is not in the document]
directionless: I think we should always pass knapsack, but that a receiver may use an interface to only get a subset. 

I'm imagining a model that commonly looks like:
```
foo := foo.New(k)
foo.run(ctx)
```

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,346 @@
...
+	observers        map[FlagsChangeObserver][]FlagKey
+}
+
+func NewFlagController(logger log.Logger, defaultValues *FlagValues, cmdLineValues *FlagValues, storedFlagValues *storedFlagValues, sanitizer *flagValueSanitizer) *FlagController {
+	fc := &FlagController{
+		logger:           logger,
+		defaultValues:    defaultValues,
+		cmdLineValues:    cmdLineValues,
+		storedFlagValues: storedFlagValues,
+		sanitizer:        sanitizer,
```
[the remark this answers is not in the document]
seejdev: Refactored this, I think it's simpler and easier to use now!

pkg/agent/flags/flag_value_sanitizer.go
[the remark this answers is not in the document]
seejdev: I've added a way to easily add sanitization for string flags in the future now.

pkg/agent/flags/flag_values_stored.go
seejdev: I got rid of stored flag value stuff, and renamed this area to `getControlServerValue`.

ee/desktop/runner/runner.go
seejdev: I moved the `FlagKeys` to a `keys` package for a couple reasons, but it seems more readable too.

pkg/agent/knapsack/knapsack.go
```diff
@@ -43,42 +51,98 @@ func (k *Knapsack) getKVStore(storeType storage.Store) types.KVStore {
...
+}
+
+func (k *knapsack) SetDebugServerData(debug bool) error {
+	return k.flags.SetDebugServerData(debug)
+}
+func (k *knapsack) DebugServerData() bool {
+	return k.flags.DebugServerData()
+}
+
+func (k *knapsack) SetForceControlSubsystems(force bool) error {
```
James-Pickett, line 112: The above funcs to retrieve KVStores are suffixed with `Store` should we consider suffixing the flags with `Flag`?

pkg/agent/types/keyvalue_store.go
```diff
@@ -35,10 +33,10 @@ type Iterator interface {
 
 // Updater is an interface for bulk replacing data in a key/value store.
 type Updater interface {
-	// Update decodes data as a JSON-encoded map of key-value pairs, and inserts
+	// Update takes a sequence of alternating key-value pairs, and inserts
 	// these key-values into the store. Any preexisting keys in the store which
 	// do not exist in data will be deleted.
-	Update(data io.Reader) error
```
James-Pickett, line 41: why use `Update(pairs ...string)` instead of `Update(map[string]string)`? Browsing through a few of the implementations seems many convert from or to `map[string]string`

James-Pickett (commented):
... still digesting

ee/desktop/runner/runner.go
```diff
@@ -140,8 +135,8 @@ type DesktopUsersProcessesRunner struct {
 	// processSpawningEnabled controls whether or not desktop user processes are automatically spawned
 	// This effectively represents whether or not the launcher desktop GUI is enabled or not
 	processSpawningEnabled bool
-	// flagsGetter gets agent flags
-	flagsGetter types.Getter
+	// knapsack is the almighty sack of knaps
```
RebeccaMahany, line 138: 🔥

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		level.Debug(fc.logger).Log("msg", "failed to get control server key", "key", key, "err", err)
+		return nil
+	}
+
+	return value
+}
+
+// set stores the value for a FlagKey in the agent flags store. Typically, this is used by the control
+// server to provide a control-server-provided value
```
RebeccaMahany, line 53: Docs are really helpful, thanks for adding them

pkg/agent/flags/flag_value_duration.go
```diff
@@ -0,0 +1,103 @@
...
+	"math"
+	"strconv"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/agent/flags/keys"
+)
+
+type durationOption func(*durationFlagValue)
```
RebeccaMahany, line 13: I like this!

RebeccaMahany (approved):

pkg/agent/types/keyvalue_store.go
seejdev: Good catch - this was a change originally as a side effect of using generics. Now the JSON decoding is removed, and the map is more simple to use!

pkg/agent/knapsack/knapsack.go
seejdev: I could do that if folks think it's more readable (or, conversely remove the `Store` suffix) from the ones above. Always a balance between readability and conciseness.

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	changedKeys := append(updatedKeys, deletedKeys...)
+
+	// Now observers can be notified these keys have possibly changed
+	fc.notifyObservers(keys.ToFlagKeys(changedKeys)...)
+
+	return changedKeys, err
+}
+
+func (fc *FlagController) RegisterChangeObserver(observer types.FlagsChangeObserver, flagKeys ...keys.FlagKey) {
+	fc.observers[observer] = append(fc.observers[observer], flagKeys...)
```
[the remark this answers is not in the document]
seejdev: I could probably put some safeguards in here to prevent something like that. I can put it in a follow up PR.

RebeccaMahany (approved):

James-Pickett (approved):

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,189 @@
...
+			observer.FlagsChanged(changedKeys...)
+		}
+	}
+}
+
+func (fc *FlagController) SetDesktopEnabled(enabled bool) error {
+	return fc.set(keys.DesktopEnabled, boolToBytes(enabled))
+}
+func (fc *FlagController) DesktopEnabled() bool {
+	return NewBoolFlagValue(WithDefaultBool(false)).get(fc.getControlServerValue(keys.DesktopEnabled))
```
directionless, line 108: I think this is fine, but I also wonder about the underlying performance. 

`NewBoolFlagValue(WithDefaultBool(false)).get` looks like it does a bunch of allocation and option parsing. Vs `BoolFlagValue(fc.getControlServerValue(keys.DesktopEnabled, false)`. But I also think I'm probably over thinking this, since we always have to do a bunch of conversations and storage reads anyhow. 

So probably optimizing for developer UX is good. Let's try!

pkg/agent/flags/flag_value_duration.go
```diff
@@ -0,0 +1,103 @@
...
+	return func(d *durationFlagValue) {
+		d.override = override
+	}
+}
+
+func WithDefault(defaultVal time.Duration) durationOption {
+	return func(d *durationFlagValue) {
+		d.defaultVal = int64(defaultVal)
+	}
+}
```
directionless, line 25: You can probably roll these together into something that just appends to an array. Then iterate the array and take the first non-nil

pkg/agent/flags/flag_value_string.go
```diff
@@ -0,0 +1,43 @@
...
+
+type stringFlagValueOption func(*stringFlagValue)
+
+func WithDefaultString(defaultVal string) stringFlagValueOption {
+	return func(s *stringFlagValue) {
+		s.defaultVal = defaultVal
+	}
+}
+
+func WithSanitizer(sanitizer func(value string) string) stringFlagValueOption {
```
directionless, line 11: Makes sense. Though I wonder if we should write a `WithEnum` as a precanned sanitizer. Eh, wait till we need it...

pkg/agent/flags/keys/keys.go
```diff
@@ -0,0 +1,49 @@
+package keys
+
+// FlagKeys are named identifiers corresponding to flags
+type FlagKey string
+
+// When adding a new FlagKey:
+// 1. Define the FlagKey identifier, and the string key value it corresponds to, in the block below
+// 2. Add a getter and setter to the Flags interface (flags.go)
```
directionless, line 8: is this still correct?

cmd/launcher/launcher.go
```diff
@@ -200,51 +205,40 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+			k,
 			desktopRunner.WithLogger(logger),
 			desktopRunner.WithUpdateInterval(time.Second*5),
 			desktopRunner.WithMenuRefreshInterval(time.Minute*15),
 			desktopRunner.WithHostname(opts.KolideServerURL),
 			desktopRunner.WithAuthToken(ulid.New()),
 			desktopRunner.WithUsersFilesRoot(rootDirectory),
-			desktopRunner.WithProcessSpawningEnabled(desktopProcessSpawningEnabled),
-			desktopRunner.WithGetter(k.AgentFlagsStore()),
+			desktopRunner.WithProcessSpawningEnabled(k.DesktopEnabled()),
```
directionless, line 234: I think, long term, we should replace things like `desktopRunner.WithProcessSpawningEnabled(k.DesktopEnabled())` with having the underlying thing just call `k.DesktopEnabled()` directly.

We don't expect there to be other consumers of `desktopRunner` and the functional arguments exist mostly as a style to manage the developer experience around them. I think knapsack replaces them.

docs/architecture/flags.md
```diff
@@ -0,0 +1,84 @@
...
+- Default values, which are used when no other value has been provided.
+- Command line values, which can be fed into launcher via command line options, config files, or environmental variables.
+- Control server updates, which are ingested by the control server client and stored in a key-value store.
+- Temporary overrides, where a client can request the `Flags` interface to override the current value with a different value for a duration of time.
+
+Launcher flag values can also have constraints defined, which provide safeguards to prevent unreasonable values being used.
+
+## Retrieving Flags
+
+```mermaid
```
directionless, line 16: This diagram is different than the implementation. It's semantically the same, but the implementation calculates each value, and then adds the next overlay. This diagram implies early returns.

pkg/agent/flags/flag_controller_option.go
```diff
@@ -0,0 +1,12 @@
+package flags
+
+import "github.com/kolide/launcher/pkg/launcher"
+
+type Option func(*FlagController)
+
+// WithStore sets the key/value store for control data
+func WithCmdLineOpts(cmdLineOpts *launcher.Options) Option {
```
directionless, line 8: Is this basically required? Do you think we'll have more options? (eg: should we just add it to the method signature)

directionless (approved):
There's a lot here, so I'm not confident I absorbed it all. But this looks like it's a huge improvement.  I say we merge, and iterate based on what we learn.

Caveat -- don't merge until we cut 1.0.7, since I'm hoping we can come up with that tomorrow

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+// FlagController is responsible for retrieving flag values from the appropriate sources,
+// determining precedence, sanitizing flag values, and notifying observers of changes.
+type FlagController struct {
+	logger                 log.Logger
+	cmdLineOpts            *launcher.Options
+	agentFlagsStore        types.KVStore
+	overrideMutex          sync.RWMutex
+	controlRequestOverride FlagValueOverride
+	observers              map[types.FlagsChangeObserver][]keys.FlagKey
```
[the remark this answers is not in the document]
seejdev: Doesn't hurt to add one - in practice I think most observers will register early and wait for notifications, so modifications to the map would be rare later on. But better to be safe.

pkg/agent/flags/flag_controller.go
```diff
@@ -0,0 +1,189 @@
...
+		opt(fc)
+	}
+
+	return fc
+}
+
+// getControlServerValue looks for a control-server-provided value for the key and returns it.
+// If a control server value is not found, nil is returned.
+func (fc *FlagController) getControlServerValue(key keys.FlagKey) []byte {
+	value, err := fc.agentFlagsStore.Get([]byte(key))
```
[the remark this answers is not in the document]
seejdev: Yeah I can clean up the language inconsistency here.

cmd/launcher/launcher.go
seejdev: Yeah that's a good point, I'll add this to a list of cleanup items. I'll check and see if there are other option style functions that can be eliminated as well.

docs/architecture/flags.md
seejdev: You're right, I'll make this more accurate.

cmd/launcher/launcher.go
directionless: Yeah, it's all the long arc of using knapsack.

PR #1115, opened 2023-03-29
bufferedHttpResponse makes http.Header if nil when accessed

James-Pickett:
this updates buffered response to initialize its  http.Header if it's nil at the time of access

the built in 404 func calls `ResponseWriter.Header().Set(...)` but this is not initialized by the bufferedHttpResponse, so it was causing launcher panics

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -207,7 +207,7 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	if opts.ControlServerURL == "" {
 		level.Debug(logger).Log("msg", "control server URL not set, will not create control service")
 	} else {
-		controlService, err := createControlService(ctx, logger, k.ControlStore(), opts)
+		controlService, err = createControlService(ctx, logger, k.ControlStore(), opts)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


seejdev (approved):

directionless (dismissed):
I'd take this. 

I might also suggest we add some tests around this, since I'm kinda embarrassed I missed this panic.

James-Pickett:
@directionless 

> I might also suggest we add some tests around this, since I'm kinda embarrassed I missed this panic.

It looks like there is really not a way to test for this unless we use reflection to to check if the interface is nil.

https://go.dev/play/p/uGQHfOWvcor

```go
package main

import "fmt"

type Interface interface{}

type Concrete struct{}

func main() {
	var concretePointer *Concrete

	// it's always not nil
	fmt.Println(isNil(concretePointer))

	var concreteValue Concrete

	// it's always not nil
	fmt.Println(isNil(concreteValue))
}

func isNil(i Interface) string {
	if i == nil {
		return "nil"
	}

	return "not nil"
}
```

directionless:
> It looks like there is really not a way to test for this unless we use reflection to to check if the interface is nil.

The error was because nothing had exercised the 404 path. Is a test that hits the 404 path enough?

directionless (approved):

PR #1123, opened 2023-04-10
Add kolide_softwareupdate table

RebeccaMahany:
Adds a new execparser table to run `softwareupdate --list` to return recommended updates for macOS.

Sample query and output:

```
osquery> select * from kolide_softwareupdate;
+---------------+--------+-------------+-----------------------------+-------+
| fullkey       | parent | key         | value                       | query |
+---------------+--------+-------------+-----------------------------+-------+
| 0/Label       | 0      | Label       | macOS Ventura 13.3.1-22E261 | *     |
| 0/Recommended | 0      | Recommended | YES                         | *     |
| 0/Size        | 0      | Size        | 868648KiB                   | *     |
| 0/Title       | 0      | Title       | macOS Ventura 13.3.1        | *     |
| 0/Version     | 0      | Version     | 13.3.1                      | *     |
| 0/Action      | 0      | Action      | restart                     | *     |
+---------------+--------+-------------+-----------------------------+-------+
```

### DIFF
--- a/pkg/osquery/tables/execparsers/softwareupdate/parse_test.go
+++ b/pkg/osquery/tables/execparsers/softwareupdate/parse_test.go
@@ -1,0 +1,152 @@
+//go:build darwin
+// +build darwin
+
+package softwareupdate
+
+import (
+	"bytes"
+	_ "embed"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test-data/beta-update-available-noscan.txt
+var beta_update_available_noscan []byte
+
+//go:embed test-data/beta-update-available-scan.txt
+var beta_update_available_scan []byte
+
+//go:embed test-data/error-scan.txt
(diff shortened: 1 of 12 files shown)


pkg/osquery/tables/execparsers/softwareupdate/parse_test.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"bytes"
+	_ "embed"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test-data/beta-update-available-noscan.txt
+var beta_update_available_noscan []byte
```
directionless, line 15: Probably not worth changing, but given the number of these I'd probably _not_ embed, and instead make something like `readTestFile(t, filename) []byte` and stick that into the table definition

seejdev (approved):

PR #1126, opened 2023-04-12
Add kolide_softwareupdate_scan table

RebeccaMahany:
We are occasionally seeing the new `kolide_softwareupdate` table return 0 rows from devices who should have a software update available. The explanation appears to be that those devices have not scanned for updates recently -- so this PR adds a second table `kolide_softwareupdate_scan` that will perform the scan. With this data, hopefully we should be able to determine whether the lack of scan is why `kolide_softwareupdate` sometimes returns 0 rows when it should return some.


James-Pickett (approved):

directionless (approved):
I think the _right_ way is to plumb some options through the exec helper. But this seems like a fine hack

PR #1128, opened 2023-04-13
Allowing Windows to find console users without requiring User.UiD

seejdev:
This change is an attempt to fix errors like:

image

We don't actually need the proper UID to do anything critical - it's just used for bookkeeping. In the case that we can't lookup the UID, we fallback to using the username.


ee/consoleuser/consoleuser_windows.go
```diff
@@ -88,16 +90,21 @@ func explorerProcesses(ctx context.Context) ([]*process.Process, error) {
...
 	}
 
+	// Attempt to lookup the user by userid. This can fail on Windows, but it's not strictly
+	// necessary - we just need something to uniquely identify the user. Fallback to the username
+	// when this happens.
 	user, err := user.Lookup(username)
 	if err != nil {
-		return "", fmt.Errorf("looking up username %s: %w", username, err)
+		level.Debug(logger).Log("msg", "could not lookup windows username", "err", err)
+		return username, nil
```
directionless, line 105: So this changes the semantics here. We were returning a `uid` and now we're returning the `username`. Is it going to be a problem at the point we subsequently exec as the user? Basically, everything over in https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner.go#L484

directionless: Especially https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner_windows.go#L15

ee/consoleuser/consoleuser_windows.go
```diff
@@ -88,16 +90,21 @@ func explorerProcesses(ctx context.Context) ([]*process.Process, error) {
...
 		return "", fmt.Errorf("getting process username (for pid %d): %w", proc.Pid, err)
 	}
 
+	// Attempt to lookup the user by userid. This can fail on Windows, but it's not strictly
+	// necessary - we just need something to uniquely identify the user. Fallback to the username
+	// when this happens.
 	user, err := user.Lookup(username)
 	if err != nil {
-		return "", fmt.Errorf("looking up username %s: %w", username, err)
+		level.Debug(logger).Log("msg", "could not lookup windows username", "err", err)
```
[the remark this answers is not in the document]
directionless: Small preference for `err` to be the last key pair

ee/consoleuser/consoleuser_windows.go
James-Pickett: I think the only time it would be an issue is if we fell back to username ... and then did a check again and got a uid instead. Might be best to just switch to using username across the board on windows.

seejdev: Removing the user lookup entirely will make this change simpler!

James-Pickett (approved):

James-Pickett (approved):

James-Pickett (approved):

directionless:
I'm good with this if @James-Pickett and @seejdev are. I say we merge and get it into beta asap

directionless:
I'm good with this if @James-Pickett and @seejdev are. I say we merge and get it into beta asap

PR #1129, opened 2023-04-13
Add option to include stderr with stdout for execparser

RebeccaMahany:
We discovered that `softwareupdate` writes `No new software available` to stderr instead of stdout, meaning we were unable to actually determine whether the device is up-to-date in the kolide_softwareupdate and kolide_softwareupdate_scan tables. This PR adds an option to the execparser to include stderr in stdout; it accomplishes this the same way that exec's `CombinedOutput` function works.

### DIFF
--- a/pkg/osquery/tables/tablehelpers/exec.go
+++ b/pkg/osquery/tables/tablehelpers/exec.go
@@ -20,7 +20,7 @@ import (
 //  3. It moves the stderr into the return error, if needed.
 //
 // This is not suitable for high performance work -- it allocates new buffers each time.
-func Exec(ctx context.Context, logger log.Logger, timeoutSeconds int, possibleBins []string, args []string) ([]byte, error) {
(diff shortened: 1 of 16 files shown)


pkg/osquery/tables/tablehelpers/exec.go
```diff
@@ -20,7 +20,7 @@ import (
 //  3. It moves the stderr into the return error, if needed.
 //
 // This is not suitable for high performance work -- it allocates new buffers each time.
-func Exec(ctx context.Context, logger log.Logger, timeoutSeconds int, possibleBins []string, args []string) ([]byte, error) {
+func Exec(ctx context.Context, logger log.Logger, timeoutSeconds int, possibleBins []string, args []string, includeStderr bool) ([]byte, error) {
```
directionless, line 23: Arg list is unweildly, but this seems better than starting the work of moving to functional arguments

directionless (approved):
good enough

PR #1130, opened 2023-04-13
Log process info on osqueryd init error

RebeccaMahany:
This PR logs additional info whenever we detect [this osquery issue](https://github.com/osquery/osquery/issues/7796), in an attempt to understand why we occasionally see another unrecognized process holding the osquery lockfile.

Example log:

```
{
  "caller": "log.go:147",
  "cmdline": "./a.out",
  "create_time": "1681494252000",
  "launcher_pid": 38413,
  "msg": "detected non-osqueryd process using pidfile",
  "name": "a.out",
  "parent_cmdline": "sudo ./a.out",
  "parent_pid": 38409,
  "parent_status": "S",
  "pid": 38410,
  "severity": "debug",
  "status": "S",
  "system_uptime": 84572,
  "ts": "2023-04-14T17:44:18.867178Z",
  "uids": "[0]",
  "username": "root"
}
```

<details>
<summary>Test notes</summary>

In order to reproduce this error, you must stop launcher, then create another process that will use and lock the osqueryd pidfile. The only way I found that didn't result in an invalid PID error from osquery was to create, write, and lock the pidfile more or less exactly the way osquery does it [here](https://github.com/osquery/osquery/blob/master/osquery/utils/pidfile/pidfile_posix.cpp).

I tested on macOS since that's where we've been seeing these init errors, so the test steps are MacOS-specific.

Test steps:

1. Build launcher with changes and move it to a new update directory with a current timestamp
1. Unload launcher: `sudo launchctl unload /Library/LaunchDaemons/com.kolide-k2.launcher.plist`
1. Run the test script (below), which will overwrite the osquery pidfile for launcher and hold a lock on it for 60 seconds
1. Begin tailing launcher logs
1. Load launcher: `sudo launchctl load /Library/LaunchDaemons/com.kolide-k2.launcher.plist`
1. Observe a log `Refusing to kill non-osqueryd process`, which now includes additional context about the process that's holding the lock on the pidfile

```
#include <iostream>
#include <sstream>
#include <sys/file.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>
#include <chrono>
#include <thread>

using namespace std;

int main() {
  stringstream stream;
  stream << to_string(getpid());
  auto buffer = stream.str();

  int file_handle;
  string path ("/var/kolide-k2/k2device-preprod.kolide.com/osquery.pid");
  file_handle = open(path.c_str(), O_CREAT | O_RDWR | O_APPEND, 0600);
  if (file_handle == -1) {
    printf("open err\n");
    return 1;
  }

  if (ftruncate(file_handle, 0) != 0) {
    printf("truncate error\n");
    return 1;
  }

  auto buffer_size = static_cast<ssize_t>(buffer.size());
  auto remaining_bytes = buffer_size;

  buffer_size = remaining_bytes = {static_cast<ssize_t>(buffer.size())};

  for (int retry = 0; retry < 5 && remaining_bytes > 0; ++retry) {
    auto buffer_ptr = buffer.data() + buffer_size - remaining_bytes;

    auto count = write(file_handle, buffer_ptr, remaining_bytes);
    if (count == -1) {
      if (errno == EINTR) {
        continue;
      }

      break;
    }

    remaining_bytes -= count;
  }

  if (remaining_bytes != 0) {
    printf("io error\n");
    return 1;
  }

  if (fchmod(file_handle, 0600) != 0) {
    printf("chmod\n");
    return 1;
  }

  if (fchown(file_handle, getuid(), getgid()) != 0) {
    printf("chown\n");
    return 1;
  }

  int lock_status{};
  lock_status = flock(file_handle, LOCK_NB | LOCK_EX);
  if (lock_status == -1) {
    printf("flock\n");
    return 1;
  }

  printf("locked, sleeping\n");
  this_thread::sleep_for(chrono::milliseconds(60000));

  printf("done sleeping\n");

  flock(file_handle, LOCK_UN);
  close(file_handle);

  return 0;
}
```

</details>

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -10,6 +10,7 @@ import (
 
 	kitlog "github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/shirou/gopsutil/host"
 	"github.com/shirou/gopsutil/process"
 )
 
@@ -121,7 +122,6 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
 
 	// Gather as much info as we can about the process
 	processInfo := []interface{}{"pid", pid}
-	processInfo = append(processInfo, "launcher_pid", os.Getpid())
 	processInfo = append(processInfo, "name", getStringStat(unknownProcess.Name))
 	processInfo = append(processInfo, "cmdline", getStringStat(unknownProcess.Cmdline))
 	processInfo = append(processInfo, "status", getStringStat(unknownProcess.Status))
@@ -129,11 +129,19 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
 	processInfo = append(processInfo, "username", getStringStat(unknownProcess.Username))
 	processInfo = append(processInfo, "uids", getSliceStat(unknownProcess.Uids))
 
+	// Add info about the parent, if available
 	unknownProcessParent, _ := unknownProcess.Parent()
 	if unknownProcessParent != nil {
 		processInfo = append(processInfo, "parent_pid", unknownProcessParent.Pid)
 		processInfo = append(processInfo, "parent_cmdline", getStringStat(unknownProcessParent.Cmdline))
 		processInfo = append(processInfo, "parent_status", getStringStat(unknownProcessParent.Status))
+	}
+
+	// Add system-level info
+	processInfo = append(processInfo, "launcher_pid", os.Getpid())
+	uptime, err := host.Uptime()
+	if err == nil {
+		processInfo = append(processInfo, "system_uptime", uptime)
 	}
 
 	level.Debug(l.logger).Log(append(processInfo, "msg", "detected non-osqueryd process using pidfile")...)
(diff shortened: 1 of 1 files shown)


pkg/log/log.go
```diff
@@ -65,10 +71,103 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
...
+
 	msg := strings.TrimSpace(string(p))
 	caller := extractOsqueryCaller(msg)
 	if err := lf(l.logger).Log(append(l.extraKeyVals, "msg", msg, "caller", caller)...); err != nil {
 		return 0, err
 	}
 	return len(p), nil
 }
+
+// Occasionally, launcher will fail to start osquery -- in this case, osquery fails
```
directionless, line 87: The comment is good, but it's in the place of a godoc function comment. So it should either be formatted as such, or moved.

directionless (commented):
Generally seems fine. One nit about comment formats

pkg/log/log.go
RebeccaMahany: Addressed, thank you!

directionless (approved):

PR #1131, opened 2023-04-14
Hide command prompt window when launching URLs on Windows

seejdev:
This gets rid of the ugly cmd window appearing when users click on menu items with `action: open-url`

### DIFF
--- a/ee/desktop/menu/action_open_url_linux.go
+++ b/ee/desktop/menu/action_open_url_linux.go
@@ -1,0 +1,14 @@
+//go:build linux
+// +build linux
+
+package menu
+
+import (
+	"os/exec"
+)
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	return exec.Command("xdg-open", url).Start()
+}
--- a/ee/desktop/menu/action_open_url.go
+++ b/ee/desktop/menu/action_open_url.go
@@ -1,10 +1,6 @@
 package menu
 
 import (
-	"os/exec"
-	"runtime"
-	"syscall"
-
 	"github.com/go-kit/kit/log/level"
 )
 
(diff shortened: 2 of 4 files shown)


James-Pickett (approved):

ee/desktop/menu/action_open_url_linux.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"os/exec"
+)
+
+// open opens the specified URL in the default browser of the user
+// See https://stackoverflow.com/a/39324149/1705598
+func open(url string) error {
+	return exec.Command("xdg-open", url).Start()
+}
```
RebeccaMahany, line 14: Not necessary for this PR, but maybe worth taking a look at standardizing this with the [notify](https://github.com/kolide/launcher/blob/main/ee/desktop/notify/notify_linux.go#L34-L36) implementation in the future. I ran into a lot of weirdness there, and it's totally possible that x-www-browser wasn't actually more reliable at picking the correct default browser than xdg-open -- but either way we should probably do the same in both places?

RebeccaMahany (approved):

ee/desktop/menu/action_open_url_linux.go
seejdev: Agree, this should be standardized since they're doing the same thing. Maybe I'll make an issue, or add it to the list of things to do with future GUI launcher land.

RebeccaMahany: (We don't have anything to standardize here on the Windows or macOS side -- the URL opening is handled by the toast on Windows, and on macOS the URL opening happens in the obj-c code.)

RebeccaMahany: Sure, either works! If you end up making an issue I can track down my notes from when I was working through different options for URL opening on Linux too.

PR #1133, opened 2023-04-17
Upgrading gopsutil module to v3

seejdev:
This upgrades the [gopsutil dependency](https://github.com/shirou/gopsutil) to it's latest release, which moves us to the [v3 implementation](https://pkg.go.dev/github.com/shirou/gopsutil@v3.21.11+incompatible#readme-v3-migration).

The initial impetus for upgrading this was to resolve this annoying build warning:

```
go run cmd/make/make.go -targets=launcher -linkstamp   
# github.com/shirou/gopsutil/host
smc_darwin.c:75:41: warning: 'kIOMasterPortDefault' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:133:19: note: 'kIOMasterPortDefault' has been explicitly marked deprecated here
```

The only non-trivial change I had to make was accounting for a breaking API change where the `Process.Status()` method now returns a [slice of strings](https://github.com/shirou/gopsutil/blob/master/process/process.go#L419-L421) instead of string. This was very simple, as [gopsutil is putting the equivalent value in the first element](https://github.com/shirou/gopsutil/blob/master/process/process_test.go#L184).


directionless (dismissed):

pkg/log/log.go
```diff
@@ -160,6 +160,17 @@ func getStringStat(getFunc func() (string, error)) string {
...
 
+// getStringSliceStat is a small wrapper around gopsutil/process functions
+// to return the stat if available, or an error message if not, so
+// that either way the info will be captured in the log.
+func getStringSliceStat(getFunc func() ([]string, error)) string {
+	stat, err := getFunc()
+	if err != nil {
+		return fmt.Sprintf("could not get stat: %v", err)
+	}
+	return stat[0]
```
RebeccaMahany, line 171:
```suggestion
	// We only use this function for `Status` at the moment, which is guaranteed to have one element when successful.
	return stat[0]
```

Just adding what you have in the MR description so that it doesn't get lost/we don't misuse this in the future!

RebeccaMahany (approved):

RebeccaMahany (approved):

James-Pickett (approved):

PR #1134, opened 2023-04-17
Fixing a nondeterministic flag controller test

seejdev:
I noticed this test failed in CI: https://github.com/kolide/launcher/actions/runs/4720672135/jobs/8372995737?pr=933

```
mock: Unexpected Method Call
2023-04-17T11:44:36.4119310Z             FlagsChanged([]keys.FlagKey)
2023-04-17T11:44:36.4120510Z             		0: []keys.FlagKey{"control_server_url", "control_request_interval"}
2023-04-17T11:44:36.4136890Z             
2023-04-17T11:44:36.4221620Z             The closest call I have is: 
2023-04-17T11:44:36.4226430Z             
2023-04-17T11:44:36.4239170Z             FlagsChanged([]keys.FlagKey)
2023-04-17T11:44:36.4326010Z             		0: []keys.FlagKey{"control_request_interval", "control_server_url"}
```

This occurs because the FlagKey slice is generated from ranging over a map, which does not guarantee order.
Comparing the test results with `ElementsMatch` effectively makes the comparison ignore order.

### DIFF
--- a/pkg/agent/flags/flag_controller.go
+++ b/pkg/agent/flags/flag_controller.go
@@ -1,6 +1,7 @@
 package flags
 
 import (
+	"sort"
 	"sync"
 	"time"
 
@@ -79,6 +80,7 @@ func (fc *FlagController) Update(kvPairs map[string]string) ([]string, error) {
 
 	// Changed keys is the union of updated keys and deleted keys
 	changedKeys := append(updatedKeys, deletedKeys...)
+	sort.Strings(changedKeys)
 
 	// Now observers can be notified these keys have possibly changed
 	fc.notifyObservers(keys.ToFlagKeys(changedKeys)...)
--- a/pkg/agent/flags/flag_controller.go
+++ b/pkg/agent/flags/flag_controller.go
@@ -1,7 +1,6 @@
 package flags
 
 import (
-	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


pkg/agent/flags/flag_controller.go
```diff
@@ -79,6 +80,7 @@ func (fc *FlagController) Update(kvPairs map[string]string) ([]string, error) {
 
 	// Changed keys is the union of updated keys and deleted keys
 	changedKeys := append(updatedKeys, deletedKeys...)
+	sort.Strings(changedKeys)
```
directionless, line 83: Do we need this to be deterministic here, or just in tests? I'm not sure how I feel about determinism.

directionless (commented):
Instead of `assert.Equal` `assert.ElementsMatch` might do what you want.

James-Pickett (commented):
I agree with seph, I think updating the test to use `elementsMatch` is better than sorting each time.

seejdev:
> Instead of `assert.Equal` `assert.ElementsMatch` might do what you want.

Good point! However I'm not sure if the equivalent exists for mocking a function call.

RebeccaMahany (approved):

James-Pickett (approved):

PR #1135, opened 2023-04-17
Scheduling background activities on macOS instead of using timers to improve launcher energy efficiency

seejdev:
To Be Filled By O.E.M.


directionless (commented):
This is pretty neat. No idea if we should do this, vs pulling all the timers together into one meta timer. Or both.

directionless:
You're thinking not in this form?

PR #1138, opened 2023-04-18
Moving remaining launcher flags to Knapsack

seejdev:
This PR continues where #1114 left off.

The theme is coalescing agent flags into the `Flags` interface and consequently, `Knapsack`. The remaining flags largely belong to some categories: launcher application configurations, osquery configurations, and autoupdate configurations.

There's a lot of lines of code touched in this PR, but a large chunk is mocks and test support.

Other worthwhile notes:

- Tests which interact with flags are largely in better shape now, because of the mock ability of the `Knapsack`/`Flags` interface.
- Some flags do not have setters. These can always be added in the future, but the nature of these flags is that they are not typically meant to be changed by the control server, for example.
- Where possible, I removed `launcher.Options` in favor of `Knapsack`, or `Flags`.
- Where possible, I removed superfluous `With...` option style functions in favor of getting the value from `Knapsack`, since it is becoming a pervasive required argument for `New...` functions.
- Not all `launcher.Options` are included in `Flags`, there are some that are used in only one spot not under `runLauncher`, or deprecated all together.


James-Pickett (approved):
Nice!

pkg/agent/flags/flag_controller.go
```diff
@@ -101,29 +104,132 @@ func (fc *FlagController) notifyObservers(flagKeys ...keys.FlagKey) {
...
+}
+
+func (fc *FlagController) Transport() string {
+	return NewStringFlagValue(
+		WithDefaultString(fc.cmdLineOpts.Transport),
+	).get(nil)
+}
+
+func (fc *FlagController) LogMaxBytesPerBatch() int {
+	return fc.cmdLineOpts.LogMaxBytesPerBatch
```
RebeccaMahany, line 185: Would we want to restrict this with a min/max?

RebeccaMahany (approved):
LGTM

pkg/agent/flags/flag_controller.go
seejdev: @directionless Do you have any strong feelings on the limits here? I don't have a good enough sense of what's reasonable here.

directionless: Anything I know I recorded in https://github.com/kolide/launcher/blob/fb81fd8c680e6db53f6ebf045ef947a574e3fc80/cmd/launcher/extension.go#L69-L90

Does that help?

pkg/agent/flags/flag_controller.go
```diff
@@ -101,29 +104,132 @@ func (fc *FlagController) notifyObservers(flagKeys ...keys.FlagKey) {
...
+		WithDefaultString(fc.cmdLineOpts.KolideServerURL),
+	).get(fc.getControlServerValue(keys.KolideServerURL))
+}
+
+func (fc *FlagController) SetKolideHosted(hosted bool) error {
+	return fc.setControlServerValue(keys.KolideHosted, boolToBytes(hosted))
+}
+func (fc *FlagController) KolideHosted() bool {
+	return NewBoolFlagValue(WithDefaultBool(false)).get(fc.getControlServerValue(keys.KolideHosted))
+}
```
directionless, line 125: I don't think we're going to override this one. It's really just a thing to set other defaults with. (And we might phase it out)

pkg/agent/flags/flag_controller.go
```diff
@@ -175,15 +281,163 @@ func (fc *FlagController) ControlRequestInterval() time.Duration {
...
+func (fc *FlagController) InsecureTransportTLS() bool {
+	return NewBoolFlagValue(WithDefaultBool(fc.cmdLineOpts.InsecureTransport)).get(fc.getControlServerValue(keys.InsecureTransportTLS))
+}
+
+func (fc *FlagController) SetIAmBreakingEELicense(disabled bool) error {
+	return fc.setControlServerValue(keys.IAmBreakingEELicense, boolToBytes(disabled))
+}
+func (fc *FlagController) IAmBreakingEELicense() bool {
+	return NewBoolFlagValue(WithDefaultBool(fc.cmdLineOpts.IAmBreakingEELicense)).get(fc.getControlServerValue(keys.IAmBreakingEELicense))
+}
```
directionless, line 316: I don't see the control server setting this one. 😆

pkg/agent/flags/flag_controller.go
```diff
@@ -175,15 +281,163 @@ func (fc *FlagController) ControlRequestInterval() time.Duration {
...
+
+func (fc *FlagController) SetDebugLogFile(file string) error {
+	return fc.setControlServerValue(keys.DebugLogFile, []byte(file))
+}
+func (fc *FlagController) DebugLogFile() string {
+	return NewStringFlagValue(
+		WithDefaultString(fc.cmdLineOpts.DebugLogFile),
+	).get(fc.getControlServerValue(keys.DebugLogFile))
+}
```
directionless, line 332: Or this one

directionless (commented):
Shape looks right. I didn't follow all the bits, but I'm game. 

some questions about places where I'm not sure we'll want to let the control server set things

Issue #1140, opened 2023-04-18
Agent is slow to start on windows


directionless:
@RebeccaMahany was commenting that the Agent felt slow to start on windows, and we saw a similar complaint from a customer. 

This is probably because we set [`DelayedAutoStart`(https://github.com/kolide/launcher/blob/25c0c352ed9963f67b554f0c685682f736b80cdd/pkg/packagekit/package_wix.go#L123)] in our MSI.  This was introduced in https://github.com/kolide/launcher/pull/568 as an attempt to ensure launcher starts _after_ the network services. 

There are likely better solutions. Perhaps something with `LoadOrderGroup` or just better retry logic inside launcher.

RebeccaMahany:
The motivation for fixing this is that when a user reboots and then tries to access an app behind Kolide, if they make it to Kolide verification before launcher starts up, they will be prompted to download the agent. This is a confusing user experience, and it's not that hard to end up in this state -- I've run into this a couple times without trying.

Issue #563, opened 2020-01-23
Delayed Start for windows services?


directionless:
I've seen a couple of windows clients run into issues with DNS resolution. This usually manifests as an error about the server being unresolvable.

This trouble seems to be correlated with hosts that _also_ have a VPN setup. My theory is that there's some kind of resolution issue during the early boot. I have seen launcher stay up after a post boot manual service restart.

One possible fix would be to look at the delayed start option for windows.

Possible another is to implement some better retry logic.

PR #1142, opened 2023-04-18
Replace delayed auto-start with service dependency in Wix configuration

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1140

Removes delayed auto-start from the Wix configuration, on the assumption that this is what causes launcher to take 2+ minutes to start up after reboot on Windows. In its place, adds service dependency for `Dnscache` to ensure we don't regress on https://github.com/kolide/launcher/issues/232 and https://github.com/kolide/launcher/issues/563.

Reference: https://www.oreilly.com/library/view/wix-36-a/9781782160427/ch11s06.html

This PR also includes a small amount of hardening to hopefully make us more robust to temporary DNS failures:
1. Allow for retries on creating the osquery extension manager client
1. Attempts to lookup the K2 server host before starting launcher, to confirm that DNS works first
1. Adds a flag `delay_start` to allow for delaying launcher -- in case this fix isn't sufficient for some folks, they can configure `delay_start` for long enough that they see the previous behavior again

<details>
<summary>Test notes</summary>

#### Prerequisites 

You will want to have the [Wix Toolset](https://github.com/wixtoolset/wix3/releases/tag/wix3112rtm) installed.

If you get an error `WiX Toolset requires the .NET Framework 3.5.1 Windows feature to be enabled`, you can fix that by running the following:

```
dism.exe /online /enable-feature /featurename:NetFx3
```

#### Building the MSI

Build an MSI using local changes, using the correct enroll secret and adjusting the Wix path as necessary:

```
make -j2 github-build

./build/package-builder.exe make --debug=true --transport=jsonrpc --hostname=k2device-preprod.kolide.com --enroll_secret=<secret goes here> --output_dir=./build/ --wix_path=/c/Program\ Files\ \(x86\)/Wix\ Toolset\ v3.11/bin --identifier=kolide-k2 --launcher_version=./build/launcher.exe
```

#### Test steps

Install the MSI. Validate that launcher is running:

1. Running `get-process | where ProcessName -like "*launcher*"` returns a result that looks reasonable
1. Kolide icon appears in the task bar, with a working menu
1. You can authenticate to k2
1. You can LQ your device from k2
1. Check debug logs (Program Files => Kolide => Launcher-kolide-k2 => data => debug.json) and confirm no very strange errors -- a couple initial DNS failures from the checkpointer seem harmless

Reboot your computer. Validate that launcher begins running very shortly after reboot (within a matter of seconds). Run through the steps above to confirm the same.

<details>

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -56,6 +56,16 @@ const (
 // enabled, the finalizers will trigger various restarts.
 func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) error {
 	logger := log.With(ctxlog.FromContext(ctx), "caller", log.DefaultCaller)
+
+	// If delay_start is configured, wait before running launcher.
+	if opts.DelayStart > 0*time.Second {
+		level.Debug(logger).Log(
+			"msg", "delay_start configured, waiting before starting launcher",
+			"delay_start", opts.DelayStart.String(),
+		)
+		time.Sleep(opts.DelayStart)
+	}
+
 	level.Debug(logger).Log("msg", "runLauncher starting")
 
 	// We've seen launcher intermittently be unable to recover from
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -30,6 +31,7 @@ import (
 	"github.com/kolide/launcher/pkg/agent/storage"
 	agentbbolt "github.com/kolide/launcher/pkg/agent/storage/bbolt"
 	"github.com/kolide/launcher/pkg/autoupdate/tuf"
+	"github.com/kolide/launcher/pkg/backoff"
 	"github.com/kolide/launcher/pkg/contexts/ctxlog"
 	"github.com/kolide/launcher/pkg/debug"
 	"github.com/kolide/launcher/pkg/launcher"
@@ -55,6 +57,21 @@ const (
 func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) error {
 	logger := log.With(ctxlog.FromContext(ctx), "caller", log.DefaultCaller)
 	level.Debug(logger).Log("msg", "runLauncher starting")
+
+	// We've seen launcher intermittently be unable to recover from
+	// DNS failures in the past, so this check gives us a little bit
+	// of room to ensure that we are able to resolve DNS requests
+	// before proceeding with starting launcher.
+	if err := backoff.WaitFor(func() error {
+		_, lookupErr := net.LookupIP(opts.KolideServerURL)
+		return lookupErr
+	}, 10*time.Second, 1*time.Second); err != nil {
+		level.Info(logger).Log(
+			"msg", "could not successfully perform IP lookup before starting launcher, proceeding anyway",
+			"kolide_server_url", opts.KolideServerURL,
+			"err", err,
+		)
+	}
 
 	// determine the root directory, create one if it's not provided
 	rootDirectory := opts.RootDirectory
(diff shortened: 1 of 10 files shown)


pkg/packagekit/package_wix.go
```diff
@@ -120,7 +120,8 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
 
 	if includeService {
 		launcherService := wix.NewService("launcher.exe",
-			wix.WithDelayedStart(),
+			// Ensure that the service does not start until DNS is available, to avoid unrecoverable DNS failures in launcher.
+			wix.WithServiceDependency("Dnscache"),
```
directionless, line 124: How confident are we that this is the correct thing across ALL THE WINDOWS VERSIONS?

RebeccaMahany: This service exists for Windows 11, 10, 8, 7, Vista, and XP. Whether it's the best solution for us is still up in the air.

directionless: That does cover all the versions. 😆

RebeccaMahany: It does also appear to work -- launcher + launcher desktop both start up super fast! But I'm hesitant to go for it till I'm more sure that we aren't going to hit a regression w the unrecoverable DNS issue, so that's what I've been working through.

cmd/launcher/launcher.go
```diff
@@ -54,8 +56,33 @@ const (
...
+	// We've seen launcher intermittently be unable to recover from
+	// DNS failures in the past, so this check gives us a little bit
+	// of room to ensure that we are able to resolve DNS requests
+	// before proceeding with starting launcher.
+	if err := backoff.WaitFor(func() error {
+		_, lookupErr := net.LookupIP(opts.KolideServerURL)
+		return lookupErr
+	}, 10*time.Second, 1*time.Second); err != nil {
+		level.Info(logger).Log(
+			"msg", "could not successfully perform IP lookup before starting launcher, proceeding anyway",
```
directionless, line 80: Hrm. I was going to suggest a longer timeout before exiting, but if you're going to proceed anyhow, I guess this is reasonable

James-Pickett (approved):

James-Pickett:
Nice!

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -537,6 +537,22 @@ func (opts *osqueryOptions) createOsquerydCommand(osquerydBinary string, paths *
 	return cmd, nil
 }
 
+// StartOsqueryClient will create and return a new osquery client with a connection
+// over the socket at the provided path. It will retry for up to 10 seconds to create
+// the connection in the event of a failure.
+func (o *OsqueryInstance) StartOsqueryClient(paths *osqueryFilePaths) (*osquery.ExtensionManagerClient, error) {
```
directionless, line 543: This feels a little awkward. Why wrap NewClient, vs adding some retry logic to newclient directly?

I can roll with it if pressed, but the two different timeouts in play makes this feel weird.

cmd/launcher/launcher.go
RebeccaMahany: In test, mine was resolving within 3-5 seconds fairly consistently, so I think this is safe 🤞

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: I did it this way a) for consistency with [how we start the server](https://github.com/kolide/launcher/blob/main/pkg/osquery/runtime/osqueryinstance.go#L542) and b) because `NewClient` is defined in a [separate package outside this repo](https://github.com/osquery/osquery-go/blob/master/client.go#L35) and it didn't feel worth it to update that for a relatively small change.

cmd/launcher/launcher.go
RebeccaMahany: I am also happy to bump it up, though, I don't feel very strongly about it being 10 sec if you have a timeout you prefer!

directionless: Nah, it seems fine to try.

pkg/osquery/runtime/osqueryinstance.go
directionless: Oh your right. That does make it messier. Okay

directionless (approved):

PR #1154, opened 2023-04-24
Change menu template data to a map type

seejdev:
This PR adds a few things to make menu template data and variables easier to define.

Changing the `TemplateData` struct to a map allows template expansion to avoid parsing errors when looking up nonexistent template variables. Avoiding the errors means that one undefined template variable no longer causes the entire input to fail and results in no menu update.

Previous versions of launcher can handle this without failing:

```
{{if hasCapability `errorlessTemplateVars`}}{{.MenuVersion}}{{else}}{{end}}
```

While new versions of launcher can just as easily handle template variables whether or not they are defined:

```
{{if .MenuVersion}}Menu Version is: {{.MenuVersion}}{{else}}{{end}} // evaluates to "Menu Version is: 0.1.0"
{{if .UndefinedVar}}This is never used: {{.UndefinedVar}}{{else}}Fallback{{end}} // evaluates to "Fallback"
Undefined: {{.UndefinedVar}} // evaluates to "Undefined: <no value>"
```

### DIFF
--- a/ee/desktop/menu/menu_template.go
+++ b/ee/desktop/menu/menu_template.go
@@ -8,17 +8,23 @@ import (
 )
 
 const (
-	funcHasCapability = "hasCapability"
-	funcRelativeTime  = "relativeTime"
+	CurrentMenuVersion string = "0.1.0" // Bump menu version when major changes occur to the TemplateData format
+
+	// Capabilities queriable via hasCapability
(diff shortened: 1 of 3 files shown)


James-Pickett (approved):

directionless (approved):
good idea

PR #1156, opened 2023-04-24
Modify registry data to adjust Windows service configuration

RebeccaMahany:
Applies the changes from https://github.com/kolide/launcher/pull/1142 to existing installations.


cmd/launcher/svc_windows.go
```diff
@@ -190,3 +194,43 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
...
+
+	// DependOnService is type REG_MULTI_SZ, i.e. a list of strings
+	dependOnServiceName = `DependOnService`
+	dependOnService     = `Dnscache`
+)
+
+func editRegistry(logger log.Logger) {
+	// Get launcher service key
+	// TODO: ideally we'd pull the identifier from somewhere legitimate
+	launcherServiceKeyName := fmt.Sprintf(launcherServiceRegistryKeyNameFormat, "KolideK2")
```
directionless, line 213: Can we pull it from the config file?

directionless (commented):
This looks pretty simple. What, if any, do you think the risks are?

cmd/launcher/svc_windows.go
RebeccaMahany: My thought was that we'd extract it from the server hostname in the options, but hadn't spent the time chasing that down yet to make sure that was correct. If yes I can definitely do that, the options are available here.

RebeccaMahany:
> This looks pretty simple. What, if any, do you think the risks are?

In test, I didn't run into any issues -- it seems to be fairly accepted to use this API to edit your own application's data. To avoid unnecessary writes, I would update to check the values first and only re-set them if they aren't already correct, instead of writing every time.

directionless:
> > This looks pretty simple. What, if any, do you think the risks are?
> 
> In test, I didn't run into any issues -- it seems to be fairly accepted to use this API to edit your own application's data. To avoid unnecessary writes, I would update to check the values first and only re-set them if they aren't already correct, instead of writing every time.

If this is accepted, then it's probably consistent across windows versions? We should probably do something to check that there's any kind of expected value, before overwriting it. (eg: if it's not set, then the system is different, and 🤷 )

Any chance we'd have a partial write that might break service restart?

RebeccaMahany:
> If this is accepted, then it's probably consistent across windows versions? We should probably do something to check that there's any kind of expected value, before overwriting it. (eg: if it's not set, then the system is different, and 🤷 )

I added a check yesterday to see if the values are already set to what we expect, so that we don't perform unnecessary writes in that case. If we fail to pull the values, then we don't perform the write. I think that should cover this? Let me know if I'm missing something.

> Any chance we'd have a partial write that might break service restart?

I don't think we're in too much danger with flipping the delayed autostart bool, but it might be possible that we could write bad/incomplete data to the list of services that launcher depends on? I'll try out writing a nonexistent service to the array and see if it affects whether launcher can start itself after reboot.

For reference, it's ultimately using [this API](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regsetvalueexw) to set the values.

RebeccaMahany:
> I'll try out writing a nonexistent service to the array and see if it affects whether launcher can start itself after reboot.

So yep that will absolutely stop launcher from being able to start itself after reboot. I feel like the chances of us getting it into this state are low, but maybe to be safest, we should not set `DependsOn` here -- that seems fine given that we still have the DNS checks on launcher first starting anyway. Thoughts?

James-Pickett (approved):

cmd/launcher/svc_windows.go
```diff
@@ -190,3 +196,74 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 		}
 	}
 }
+
+const (
```
directionless, line 200: I'd toss all this into `cmd/launcher/svc_checker_windows.go` or something

directionless (approved):
I'd take it, but I think the flow is very hard to follow, especially around the service dependancy.

I suspect splitting some things into other functions and using early returns would help. At least, that's my usual go-to.

cmd/launcher/svc_windows.go
RebeccaMahany: Will do!

seejdev (approved):
LGTM

cmd/launcher/svc_config_windows.go
```diff
@@ -0,0 +1,114 @@
...
+		// we don't want to wipe any current data from the list.
+		return
+	}
+
+	// Check whether the service configuration already includes Dnscache in the list of services it depends on
+	foundDnscacheInList := false
+	for _, service := range serviceList {
+		if service == dnscacheService {
+			foundDnscacheInList = true
+			break
```
directionless, line 100: You can early return the whole function, right?

directionless (dismissed):

cmd/launcher/svc_config_windows.go
RebeccaMahany: Yes, thank you! updated

directionless (approved):

cmd/launcher/svc_config_windows.go
directionless: I <3 early returns

Issue #1155, opened 2023-04-24
Multiple desktops appearing occasionally 


James-Pickett:
We have received reports of multiple desktops (systrays) showing up.

To help prevent and diagnose this here are some possible improvements:

* Add the `pid` field to all logs associated with a desktop process. This should make it easier to trace through large log files.
* Ensure that when desktop runner declares a desktop process as unresponsive, it is also rejected by the "monitor_server". If for some reason desktop runner is unable to kill the process, the process should terminate it self if it doesn't get a response from the monitor server.

directionless:
I would also suggest that if desktop sees it's parent pid is 1, that it considers itself an orphan and exists.

Issue #1141, opened 2023-04-18
Can we accelerate the agent on menu interaction?


directionless:
Looking at how users work with the the menu, it feels a little bit like _most_ times the user is interacting with the menu. we're going to want to accelerate the control interval. 

I'm not sure if that's all interactions, or maybe just interactions that trigger an action. Or maybe we need it as a boolean on menu items passed in from K2.

James-Pickett:
Acceleration of the fetching of the control server data is handled in launcher root. All the interaction between the user and the desktop app is handled by the desktop subprocess. We'll need away for the subprocess to communicate the user interaction back to launcher root. We already have some plumbing in place for this with the [desktop runner monitor server](https://github.com/kolide/launcher/blob/main/ee/desktop/runner/monitor_server.go) which provides the desktop subsystem an endpoint to make sure launcher root is healthy. We could repurpose / expand this to let launcher root know when a user is interacting with the menu.

Regarding what "interaction with the menu" means, out of the box the only means we have of capturing the user interaction is from OnClicked listeners than can be applied to all items except the very top level stats bar item. We could:
* anytime anything (other than the top level status bar items) is clicked, we start acceleration
* created a dedicated "refresh" button
* allow k2 to add a specific property to the menu data to specify acceleration on click

PR #1159, opened 2023-04-25
adds control server acceleartion on menu open event, adds more failsafes for shutting down desktop when parent gone

James-Pickett:
This PR repurposes the "monitor_server" that was used only for the desktop user process to check on the root processes' health. The new runner server has an additional endpoint for "menuopened" that the desktop user process can call when it gets a notification from systray that the menu has been opened.

It introduces RunnerServer and renames DesktopServer to UserServer. Both servers are related to the desktop but one is run by the DesktopRunner and one is run by the Desktop User Process.

After this PR merges, I will do some reshuffling of the desktop package structure and put packages into either a `user` or `runner` folder based on their role.

It also:

* Updates DesktopRunner to DeRegister desktop user clients form the RunnerServer when they are assumed to be dead. This way if the assumption was false and a new desktop user process was spun up, the old one should terminate it self since the RunnerServer will stop responding to it.
* Updates the Desktop User Process to self terminate if it's ppid is less than 2. We've seen some funny cases where a user has 2 desktop processes running and one has a parent of 1 or 0.

closes: https://github.com/kolide/launcher/issues/1155
closes: https://github.com/kolide/launcher/issues/1141


cmd/launcher/desktop.go
```diff
@@ -18,8 +17,11 @@ import (
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/desktop/menu"
 	"github.com/kolide/launcher/ee/desktop/notify"
-	"github.com/kolide/launcher/ee/desktop/server"
+	rootserver "github.com/kolide/launcher/ee/desktop/runner/server"
+	desktopserver "github.com/kolide/launcher/ee/desktop/server"
```
James-Pickett, line 21: wonder if we should just go ahead an name these packages `rootserver` and `desktopserver`

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,141 @@
+package server
```
James-Pickett, line 1: should we call this `runnerserver` or `rootserver`?

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	})
+
+	// menu opened endpoint
+	mux.HandleFunc(MenuOpenedEndpoint, func(w http.ResponseWriter, r *http.Request) {
+		if r.Body != nil {
+			r.Body.Close()
+		}
+
+		ms.knapsack.SetControlRequestIntervalOverride(5*time.Second, 1*time.Minute)
```
James-Pickett, line 64: hooray for `Knaspsack`, it made this part so easy

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/kolide/launcher/pkg/agent/types"
+)
+
+type RootServer struct {
+	server                *http.Server
+	listener              net.Listener
+	logger                log.Logger
+	desktopProcAuthTokens map[string]string
+	mutex                 sync.Mutex
+	knapsack              types.Knapsack
```
directionless, line 24: Optional, but you can declare what you need in an interface. And not the whole knapsack

cmd/launcher/desktop.go
James-Pickett: rootserver is the server run by the root process

desktopserver is the server run by the desktop process

maybe we should call them `rootprocserver` and `desktopprocserver`

any suggestions?

cmd/launcher/desktop.go
```diff
@@ -75,6 +82,10 @@ func runDesktop(args []string) error {
 		"pid", os.Getpid(),
 	)
 
+	if *fldebug {
+		level.AllowDebug()
+	}
```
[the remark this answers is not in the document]
James-Pickett: you correct and this is actually set earlier when we create the logger, removing

cmd/launcher/desktop.go
```diff
@@ -32,20 +34,25 @@ func runDesktop(args []string) error {
 			"",
 			"hostname launcher is connected to",
 		)
-		flauthtoken = flagset.String(
-			"authtoken",
+		flUserServerAuthToken = flagset.String(
+			"user_server_auth_token",
 			"",
-			"auth token for desktop server",
+			"auth token for user user server",
```
RebeccaMahany, line 40: I think typo 🙂 

```suggestion
			"auth token for user server",
```

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/kit/ulid"
+)
+
+type RunnerServer struct {
```
RebeccaMahany, line 17: Mind adding a short comment about what the runner server's responsibilities are? `Allows user desktop processes to monitor + notify parent process` or similar

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	})
+
+	// menu opened endpoint
+	mux.HandleFunc(MenuOpenedEndpoint, func(w http.ResponseWriter, r *http.Request) {
+		if r.Body != nil {
+			r.Body.Close()
+		}
+
+		controlRequestIntervalOverrider.SetControlRequestIntervalOverride(5*time.Second, 1*time.Minute)
```
RebeccaMahany, line 67: Could we make these durations consts so it's easier to identify/update them in the future?

RebeccaMahany (approved):
Just a couple small comments, LGTM overall!

cmd/launcher/desktop.go
James-Pickett: done

ee/desktop/runner/server/server.go
James-Pickett: done

ee/desktop/runner/server/server.go
James-Pickett: done

RebeccaMahany (approved):
🔥

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+const (
+	HealthCheckEndpoint                = "/health"
+	MenuOpenedEndpoint                 = "/menuopened"
+	controlRequestAccelerationInterval = 5 * time.Second
+	controlRequestAcclerationDuration  = 1 * time.Minute
+)
+
+type controlRequestIntervalOverrider interface {
```
seejdev, line 35: Yay, cool to see the knapsack sub-interface in action. This feels clean too

ee/desktop/runner/server/server.go
```diff
@@ -0,0 +1,148 @@
...
+	SetControlRequestIntervalOverride(time.Duration, time.Duration)
+}
+
+func New(logger log.Logger, controlRequestIntervalOverrider controlRequestIntervalOverrider) (*RunnerServer, error) {
+	listener, err := net.Listen("tcp", "localhost:0")
```
seejdev, line 40: Does using port zero mean it will choose a dynamically generated unused port?

seejdev (approved):

ee/desktop/runner/server/server.go
James-Pickett: yea, it just picks a random port, it's what the current "monitor server" that is doing now on stable

PR #1162, opened 2023-04-27
Make test failures more specific to track down flaky TUF test

RebeccaMahany:
Attempt to make it easier to track down test failure: https://github.com/kolide/launcher/actions/runs/4821553131/jobs/8587578725


RebeccaMahany:
Haven't been able to reproduce this again

RebeccaMahany:
Seen again here https://github.com/kolide/launcher/actions/runs/4876196913/jobs/8699373158

For some reason in the test `TestAddToLibrary_alreadyAdded`, it's not correctly identifying that the desired update has already been downloaded, so it attempts to download it and correctly hits a panic because there's no http client set on that test library manager. It's unclear to me why the check isn't working.

RebeccaMahany:
Still not sure why this test failed twice and haven't been able to reproduce it, so I've removed the debugging from the tests and added a lot more checks so that hopefully the next time we see this, we get a more useful, specific error.

James-Pickett (approved):

PR #1173, opened 2023-05-02
Increasing thrift's ServerConnectivityCheckInterval from 5ms to 100ms

seejdev:
This changes the Thrift server-side connectivity checks to be 20 times less frequent.

From Thrift documentation:

> ServerConnectivityCheckInterval defines the ticker interval used by
> connectivity check in thrift compiled TProcessorFunc implementations.

So we have two-way connectivity checks occurring where:

- Every 5 seconds, launcher osquery client is pinging the osquery Thrift server to check that the osquery process is still running.
- Every 5 milliseconds, the osquery Thirft server is checking that the socket connected to the client is still open.

Increasing the latency to 100ms seems reasonable.

### DIFF
--- a/cmd/grpc.ext/grpc.go
+++ b/cmd/grpc.ext/grpc.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/apache/thrift/lib/go/thrift"
 	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/kit/env"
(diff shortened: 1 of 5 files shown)


James-Pickett (dismissed):

James-Pickett (approved):

directionless (approved):
Let's try! I don't love the globals, but here we are.

RebeccaMahany (approved):

PR #1174, opened 2023-05-02
Handle `Features` array within remotectl properties

RebeccaMahany:
Fixes issue noted by @FritzX6 where the table will fail to parse properties when they have a nested `Features` array within, resulting in the following error in launcher logs: `error parsing data: top-level key/value pair in remotectl output is in an unexpected format`.

Adds test for same.

### DIFF
--- a/pkg/osquery/tables/execparsers/remotectl/parse_test.go
+++ b/pkg/osquery/tables/execparsers/remotectl/parse_test.go
@@ -27,6 +27,9 @@ var malformed_dumpstate_at_top_level string
 //go:embed test-data/malformed_dumpstate_in_properties.txt
 var malformed_dumpstate_in_properties string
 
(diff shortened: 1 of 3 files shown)


pkg/osquery/tables/execparsers/remotectl/parse_test.go
```diff
@@ -78,6 +88,9 @@ func TestParse(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
+			if tt.name != "dumpstate_with_features" {
```
seejdev, line 91: Left behind from testing?

RebeccaMahany: 🤦‍♀️ yup, thanks

directionless (approved):

PR #1168, opened 2023-04-28
Pass knapsack in to tuf autoupdater to simplify configuration

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/954.

This is an incremental improvement of TUF work -- passes the knapsack in to the TUF autoupdater. The next piece of work requires some refactoring/configuration changes, and using the knapsack is more flexible.

Subsequent PRs will tackle the following (order is not set in stone):
1. New implementation of findNew (in progress: https://github.com/kolide/launcher/pull/1165)
1. Trigger a reload/restart after a successful update; launcher init performs version selection using the new update library (potentially falling back to `findNew`); launcher performs library tidying after version selection
1. Other improvements/refactors (ship a more up-to-date TUF repo)
1. An "update now" functionality tied to control server
1. Eventually removing the old notary autoupdater

Previous work:
1. https://github.com/kolide/launcher/pull/1081
1. https://github.com/kolide/launcher/pull/1103
1. https://github.com/kolide/launcher/pull/1108
1. https://github.com/kolide/launcher/pull/1110
1. https://github.com/kolide/launcher/pull/1111
1. https://github.com/kolide/launcher/pull/1119

PR #1119, opened 2023-04-06
Small autoupdate improvements

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/954.

This is the next step toward using our new TUF infrastructure: a couple small improvements, including addressing code review comments left on https://github.com/kolide/launcher/pull/1111 and others:
* Ensure that repeated errors when trying to autoupdate launcher will not prevent us from trying to update osquery
* Use a temp directory for staging downloads
* Add a lock to the library to ensure updates aren't running in multiple places for the same library

Subsequent PRs will tackle the following (order is not set in stone):
1. Trigger a reload/restart after a successful update; launcher init performs version selection using the new update library (potentially falling back to `findNew`); launcher performs library tidying after version selection
1. Other improvements/refactors (use knapsack/flags for configuration)
1. An "update now" functionality tied to control server
1. Eventually removing the old notary autoupdater

Previous work:
1. https://github.com/kolide/launcher/pull/1081
1. https://github.com/kolide/launcher/pull/1103
1. https://github.com/kolide/launcher/pull/1108
1. https://github.com/kolide/launcher/pull/1110
1. https://github.com/kolide/launcher/pull/1111

PR #1178, opened 2023-05-04
[TUF autoupdater] Remove osquery client dependency

RebeccaMahany:

This PR extracts the osquerier dependency from the library, so that we can init the library before we have an osquery client.

It also pulls out a shared function for sorting versions in the library for reuse later.

Adds a small amount of documentation about how to generate mocks with mockery, since it's done in a slightly non-standard way in this package.

<details>
<summary>Future work + linked PRs</summary>

1. A new implementation of findNewest -- already in progress here: https://github.com/kolide/launcher/pull/1185
1. Point to release file `<binary>/<os>/<arch>/<channel>/release.json`, once available
1. Other improvements/refactors (ship a more up-to-date TUF repo, maybe split library parts of `autoupdate/tuf` out into its own package)

Previous work:
1. https://github.com/kolide/launcher/pull/1111
1. https://github.com/kolide/launcher/pull/1119
1. https://github.com/kolide/launcher/pull/1168

</details>


pkg/autoupdate/tuf/read_only_library.go
```diff
@@ -0,0 +1,269 @@
...
+
+	"github.com/Masterminds/semver"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/autoupdate"
+)
+
+// readOnlyLibrary provides a read-only view into the updates libraries for all
+// binaries.
+type readOnlyLibrary struct {
```
directionless, line 23: I don't wholly understand the motivation for splitting this out. It feels like it's adding a lot of complexity, since the RO and RW versions are inherently intertwinned.

RebeccaMahany: I went to make a diagram to explain why I've split this out, but in the process I've realized I've split out the wrong bit.

The original motivation was to remove the osquerier dependency, because `findNewest` needs to run before we have an osquery client. However, we only need the osquerier to get the current running version, which we only need to run the routine to tidy the library (so that we don't delete the currently-running version).

I'll think on it a little more, but I think the direction I'm leaning is to move the osquerier dependency out of the library altogether and re-combine.

directionless: Happy to rubber duck or brainstorm if it would help.

RebeccaMahany: I've just pushed this update, if you'd like to take a quick look and have thoughts! Wondering if I should maybe move some of the install version stuff into a new file, or maybe it's fine to keep in the library.

pkg/autoupdate/tuf/autoupdate.go
```diff
@@ -46,14 +48,20 @@ type ReleaseFileCustomMetadata struct {
 }
 
 type librarian interface {
+	IsInstallVersion(binary autoupdatableBinary, targetFilename string) bool
 	Available(binary autoupdatableBinary, targetFilename string) bool
 	AddToLibrary(binary autoupdatableBinary, targetFilename string, targetMetadata data.TargetFileMeta) error
-	TidyLibrary()
+	TidyLibrary(binary autoupdatableBinary, currentVersion string)
```
directionless, line 54: Why change this? I imagine most callers would just want to tidy everything.

directionless: Ah, reading more code, I think it's so that we shift the responsibility of finding the running version to the caller.

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -320,53 +246,246 @@ func (ulm *updateLibraryManager) tidyUpdateLibrary(binary autoupdatableBinary, c
...
+	rawVersionsInLibrary, err := filepath.Glob(filepath.Join(ulm.updatesDirectory(binary), "*"))
+	if err != nil {
+		return nil, nil, fmt.Errorf("could not glob for updates in library: %w", err)
+	}
+
+	versionsInLibrary := make([]*semver.Version, 0)
+	invalidVersions := make([]string, 0)
+	for _, rawVersionWithPath := range rawVersionsInLibrary {
+		rawVersion := filepath.Base(rawVersionWithPath)
+		v, err := semver.NewVersion(rawVersion)
```
directionless, line 296: I haven't tested this, but how does it handle the git versions we use? `v1.0.7-16-g6e6704e1dc33`

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -320,53 +246,246 @@ func (ulm *updateLibraryManager) tidyUpdateLibrary(binary autoupdatableBinary, c
...
+	for i, v := range versionsInLibrary {
+		versionsInLibraryStr[i] = v.Original()
 	}
+
+	return versionsInLibraryStr, invalidVersions, nil
+}
+
+// IsInstallVersion checks whether the version in the given target is the same as the one
+// contained in the installation.
+func (ulm *updateLibraryManager) IsInstallVersion(binary autoupdatableBinary, targetFilename string) bool {
```
directionless, line 325: Hrm. So how will this get used?

1. installed as 1.0.8
2. upgraded to 1.0.9
3. downgraded to 1.0.8

So what is executed?

My gut sense is that the complexity isn't worth it, and just downloading 1.0.8 is fine

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -320,53 +246,246 @@ func (ulm *updateLibraryManager) tidyUpdateLibrary(binary autoupdatableBinary, c
...
+	}
+
+	ulm.cacheInstalledVersion(binary, v)
+
+	return v, pathToBinary, nil
+}
+
+// findInstallLocation attempts to find the install location for the given binary, looking
+// in well-known locations.
+func (ulm *updateLibraryManager) findInstallLocation(binary autoupdatableBinary) string {
```
directionless, line 374: This, especially, feels like it's going to be fragile and hard to manage.

If we really want to track the installed version, I think we should set an ENV on launch and carry it through.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: It parses them and sorts them correctly, I think. It will say that `1.0.7-16-abcdabcd` is more recent than `1.0.7-15-defgdefg`. It also says that `1.0.7-16-g6e6704e1dc33` is _not_ more recent than `1.0.7`, which looks like it's expected and adheres to the semver spec, but is maybe not what we expect?

RebeccaMahany: (There's a test `Test_sortedVersionsInLibrary` that covers at least some of these test cases -- I'm happy to expand it too.)

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: I think you're right -- it's already felt pretty fragile in development. Maybe it's fine to just not care about it -- I'm fine with setting it on launch and carrying it through, too.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Pulled it out, more trouble than it was worth!

directionless: Having slept on it... I think we need _something_ here. 

There's a case that's like:
1. installed as 1.0.8
2. stable is 1.0.8
3. download 1.0.8

And that feels kinda -_-

Bt my inclination would be to push the initial versioning to `main` or potentially even the packaging layer.

directionless: Main might be clean enough. I think the hardest part here was the find-binary code. And main / extension setup should know that

directionless: I feel like I'm talking myself in circles here. I'm sorry the requirements aren't more clear.

RebeccaMahany: I've just wrapped up this commit https://github.com/kolide/launcher/pull/1178/commits/72db9509dd977bf2ed2e6c68b250805e9afa07d1, which I think handles what we're talking about:

1. install version is 1.0.8
1. current running version is therefore 1.0.8
1. there is no update downloaded for 1.0.8
1. we see that current running version is 1.0.8 so we don't perform the download

seejdev (approved):

PR #1183, opened 2023-05-08
Setup Loggers

directionless:
This PR is an attempt at setting up logging. The goal is to create something like:
```mermaid
flowchart LR
    subgraph inputs
        direction TB
        systemLogger>systemLogger]
        logger>logger]

        %% Force formatting
        systemLogger ~~~ logger
    end

    subgraph destinations
        direction TB
        destSys[(System Log Store)]
        destDebugLog[(debug.json)]
        destMem[(inmemory ring)]
        destStore[(persisted ring)]
    end

    systemLogger --> slTee{tee} --> logger
    slTee --> slFilter[[filter out debug]] --> destSys

    %% force formatting
    logger ~~~ slTee ~~~ systemLogger

    logger --> lTee{tee}
    lTee{tee} --> destDebugLog
    lTee --> destMem
    lTee --> f2[[filter: info higher]] --> destStore

    osquery -. kolide_launcher_logs .-> destMem
    osquery -. kolide_launcher_crit_logs .-> destStore

```

To achieve that, this PR adds:
1. a simple ring buffer, backed by a `store`
2. building on that, it creates a `ringlogger`
3. a log thicket, which uses these to create that routing.

TODO:
* replace the logger in main
* setup osquery table access


pkg/agent/logrouter/logrouter.go
```diff
@@ -0,0 +1,199 @@
...
+// TODO: This should probably move to the underlying ringlogger
+func (lr *logRouter) Replay(logger log.Logger) error {
+	logs, err := lr.logRing.GetAll()
+	if err != nil {
+		return fmt.Errorf("gettings logs to replay: %w", err)
+	}
+	for _, l := range logs {
+		var levelFn func(logger log.Logger) log.Logger
+
+		pairs := make([]any, 2*len(l))
```
James-Pickett, line 121: why is the length of the pairs `2*len(l)`?

I think the intention here is to add 2 for the "replayed" & true values? If this is the case `2+len(l)` would make more sense to me.

pkg/agent/logrouter/logrouter.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (lr *logRouter) GetRecentDebugLogs() ([]map[string]any, error) {
+	return lr.logRing.GetAll()
+}
+
+// Replay will replay the logs from the memory ring to a new logger. Because we're serliazing everything down to a
+// string, handling the leveling is cumbersome.
+//
+// TODO: This should probably move to the underlying ringlogger
+func (lr *logRouter) Replay(logger log.Logger) error {
```
James-Pickett, line 113: this is really cool and will work nicely with log shipping

James-Pickett (commented):
This looks pretty good, only one tiny nit.

pkg/agent/logrouter/logrouter.go
directionless: I built it for that! Well, and the idea of an action triggering a log replay

directionless: I think there are some ways we can figure out how to clean it up. What methods live where. But the bones are shaping up

directionless:
Replaced by #1322

Issue #1149, opened 2023-04-19
Make launcher available per-architecture


RebeccaMahany:
We will likely want to start supporting launcher on ARM for Windows and Linux, and potentially may want to ship separate builds per-architecture for macOS instead of shipping a universal build.

This encompasses:

- [x] Making launcher builds available per-architecture in Github artifacts on merges to main: https://github.com/kolide/launcher/pull/1152
- [x] Publishing launcher builds per-architecture to our mirror and to TUF: https://github.com/kolide/endpoint-releaser/pull/89
- [x] Publishing osquery builds per-architecture to our mirror and to TUF: https://github.com/kolide/endpoint-releaser/pull/94
- [x] Promoting launcher and osquery builds per-architecture to TUF: https://github.com/kolide/endpoint-releaser/pull/94
- [x] New autoupdate code should look for the release for its architecture at `<binary>/<os>/<arch>/<channel>/release.json`: https://github.com/kolide/launcher/pull/1195 (waiting on next osquery release)
- [ ] Create installer packages per-architecture
- [x] Publishing and promoting osquery arm64 windows builds once available (available in next release after https://github.com/osquery/osquery/pull/7918): https://github.com/kolide/endpoint-releaser/pull/108
- [x] New autoupdate work (https://github.com/kolide/launcher/issues/954) must be released

I think it's likely not worthwhile to update the old TUF server and old autoupdate code to account for this.

directionless:
Looks right. Order seems about right too. I agree we don't care about the old notary stuff.

tstromberg:
I learned today that we have employees using Raspberry Pi (arm64) devices for software development, but are unable to use Kolide for compliance monitoring on them so I am very much interested in this issue. Thanks!

directionless:
Hi @tstromberg! We've started digging into windows arm64 support and have a beta for that. We haven't done nearly as much for linux arm64 support. 

Which distros are people using?

How do you imagine the installation working? I'm somewhat concerned that the number of possible downloads becomes unmanageable as it grows.

zenador:
It'd be great to be able to use this on Ubuntu on ARM!

yumcimil:
XAM customer here - we're just starting to prototype Ubuntu 24.10's very early support for Snapdragon X Elite processors. Maybe Fedora would be the other option for us, but for supportability, we wouldn't be looking at anything more esoteric.

krab-skunk:
[...]

PR #1185, opened 2023-05-08
[TUF autoupdater] Check out latest

RebeccaMahany:

This is the next step toward using our new TUF infrastructure: a new implementation of `findNew` that looks at our new update library.

Includes some drive-by changes:
* I've increased the download timeout from 5 to 8 minutes. I've seen just a couple download timeouts in the `kolide_tuf_autoupdater_errors` table that I'm hoping this will resolve.
* Adds a `GetChannelVersionFromTufServer` to be used by `launcher doctor` and by packaging changes for https://github.com/kolide/launcher/issues/1149

<details>
<summary>How we pick which version of launcher/osqueryd to run</summary>

This flowchart is also committed to the docs in this PR.

```mermaid
flowchart TB
    A[Library lookup] --> B{Do we have a local TUF repo?}
    B ---->|No| C[Get most recent version from update library]
    C --> D[Return path to most recent version of executable]
    D --> H[End]
    B -->|Yes| E{release.json target metadata exists?}
    E -->|No| C
    E --> |Yes| F{Target indicated by release.json\nis downloaded to update library?}
    F --> |Yes| G[Return path to selected executable in update library]
    F --> |No| C
    G --> H
```

</details>

<details>
<summary>Future work + linked PRs</summary>

1. Point to release file `<binary>/<os>/<arch>/<channel>/release.json`, once available (in progress but on hold here: https://github.com/kolide/launcher/pull/1195)
1. Other improvements/refactors (ship a more up-to-date TUF repo)

Previous work:
1. https://github.com/kolide/launcher/pull/1111
1. https://github.com/kolide/launcher/pull/1119
1. https://github.com/kolide/launcher/pull/1168
1. https://github.com/kolide/launcher/pull/1178
1. https://github.com/kolide/launcher/pull/1188

</details>


seejdev (approved):

directionless:
I was slacking about this. It feels confusing and complex.

RebeccaMahany:
> I was slacking about this. It feels confusing and complex.

@directionless I think this is simpler and more lightweight now -- let me know what you think.

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+	"errors"
+	"fmt"
+	"path/filepath"
+
+	"github.com/kolide/launcher/pkg/autoupdate"
+)
+
+// CheckOutLatest returns the path to the latest downloaded executable for our binary, as well
+// as its version.
+func CheckOutLatest(binary autoupdatableBinary, rootDirectory string, updateDirectory string, channel string) (string, string, error) {
```
James-Pickett, line 14: I always find >2 return values confusing, consider naming the return values
```suggestion
func CheckOutLatest(binary autoupdatableBinary, rootDirectory string, updateDirectory string, channel string) (path string, version string, err error) {
```

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	// If we can't find the specific release version that we should be on, then just return the executable
+	// with the most recent version in the library
+	return mostRecentVersion(binary, updateDirectory)
+}
+
+// findExecutableFromRelease looks at our local TUF repository to find the release for our
+// given channel. If it's already downloaded, then we return its path and version.
+func findExecutableFromRelease(binary autoupdatableBinary, tufRepositoryLocation string, channel string, baseUpdateDirectory string) (string, string, error) {
```
James-Pickett, line 31: same consider naming return values when > 2

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+	if autoupdate.CheckExecutable(context.TODO(), targetPath, "--version") != nil {
+		return "", "", fmt.Errorf("version %s from target %s either not yet downloaded or corrupted: %w", targetVersion, targetName, err)
+	}
+
+	return targetPath, targetVersion, nil
+}
+
+// mostRecentVersion returns the path to the most recent, valid version available in the library for the
+// given binary, along with its version.
+func mostRecentVersion(binary autoupdatableBinary, baseUpdateDirectory string) (string, string, error) {
```
James-Pickett, line 59: consider naming return values

James-Pickett (commented):
I lack the understanding to review the code on a macro level. I'll try to look again later. Left some "micro" comments =)

pkg/autoupdate/tuf/library_lookup.go
directionless: I agree that more than 2 return values is confusing, but I think naming them is an antipattern. (Though there's disagreement in go)

It looks pretty, but it creates them as implicit vars, which always feels like it breaks scoping to me

directionless: eg: I just accept it as imperfect. Since it feels least bad. We could talk about changing that?

pkg/autoupdate/tuf/library_lookup.go
directionless: I agree that more than 2 return values is confusing, but I think naming them is an antipattern. (Though there's disagreement in go)

It looks pretty, but it creates them as implicit vars, which always feels like it breaks scoping to me

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// findExecutableFromRelease looks at our local TUF repository to find the release for our
+// given channel. If it's already downloaded, then we return its path and version.
+func findExecutableFromRelease(binary autoupdatableBinary, tufRepositoryLocation string, channel string, baseUpdateDirectory string) (string, string, error) {
+	// Initialize a read-only TUF metadata client to parse the data we already have downloaded about releases.
+	metadataClient, err := readOnlyTufMetadataClient(tufRepositoryLocation)
+	if err != nil {
+		return "", "", errors.New("could not initialize TUF client, cannot find release")
+	}
```
directionless, line 36: What's the behavior here if there's no local repo? (eg: first boot)

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+	// From already-downloaded metadata, look for the release version
+	targets, err := metadataClient.Targets()
+	if err != nil {
+		return "", "", fmt.Errorf("could not get target: %w", err)
+	}
+
+	targetName, _, err := findRelease(binary, targets, channel)
+	if err != nil {
+		return "", "", fmt.Errorf("could not find release: %w", err)
+	}
```
directionless, line 47: I think this is okay, but it feels weird. These are functions that are hard to follow, and work a lot with `github.com/theupdateframework/go-tuf/data`. It feels like we should bake them into something like `metadataClient`. Not sure it's worth it though

pkg/autoupdate/tuf/library_lookup.go
```diff
@@ -0,0 +1,75 @@
...
+// as its version.
+func CheckOutLatest(binary autoupdatableBinary, rootDirectory string, updateDirectory string, channel string) (string, string, error) {
+	if updateDirectory == "" {
+		updateDirectory = defaultLibraryDirectory(rootDirectory)
+	}
+
+	releasePath, releaseVersion, err := findExecutableFromRelease(binary, LocalTufDirectory(rootDirectory), channel, updateDirectory)
+	if err == nil {
+		return releasePath, releaseVersion, nil
+	}
```
directionless, line 22: Can we log the error?

directionless (commented):
I think this is mostly okay. (See questions)

this feels much more readable than the prior. I suspect some of this should land in the library manager, and we should call it. Having this iterate files on disk feels wrong. (What happens if we move the library into ram?) But we can iterate there if we want to.

pkg/autoupdate/tuf/channel_version.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/kolide/launcher/pkg/agent"
+)
+
+// VersionForChannel returns the tagged version of the given binary for the given release channel.
+func VersionForChannel(binary, channel, tufServerUrl string) (string, error) {
```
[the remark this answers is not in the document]
RebeccaMahany: What calls this:

> VersionForChannel to be used by launcher doctor and by packaging changes for https://github.com/kolide/launcher/issues/1149

Discussion on the launcher doctor PR: https://github.com/kolide/launcher/pull/1197#discussion_r1195584997

Since package build doesn't necessarily run on a machine with launcher installed, it can't use the populated TUF repo -- and either way, this feels a little safer than having separate routines updating the same TUF directory potentially at the same time.

I have a TODO to ship a more complete/up-to-date TUF repo -- hopefully that addresses the performance concern some?

pkg/autoupdate/tuf/library_lookup.go
RebeccaMahany: If we can't read from the TUF repo for whatever reason, then we'd fall back to picking the most recent version in the update library. If there are no updates in the update library either, then we'd assume that the current running executable (likely the installed version, at this point) is the correct one to be running.

pkg/autoupdate/tuf/read_only_tuf_client.go
```diff
@@ -0,0 +1,78 @@
...
+	return r.localFilestore.Close()
+}
+
+func (r *readOnlyLocalStore) GetMeta() (map[string]json.RawMessage, error) {
+	return r.localFilestore.GetMeta()
+}
+
+func (r *readOnlyLocalStore) SetMeta(name string, meta json.RawMessage) error {
+	return nil
+}
```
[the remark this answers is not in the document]
RebeccaMahany: I was treating these like a no-op -- I'm not sure we'd want to error out here? Erroring out would only be preventing us from executing the lookup. Same goes for https://github.com/kolide/launcher/pull/1185#discussion_r1198465017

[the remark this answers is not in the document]
directionless: Happy to defer to you!

I agree it's a no-op, but I don't know if the caller should get an error since it's an invalid action on a RO

pkg/autoupdate/tuf/library_lookup.go
RebeccaMahany: I think there are a couple places in `autoupdate.go` that feel similar, where it's a lot of work with go-tuf/data. Maybe could be wrapped in a `tufReleaseClient` (or something) that performs the repeated tasks (update from remote TUF repo to download updated targets, find release target among those targets, find what target it's pointing at, return that target). I could do that here, or add it to the TODOs for a subsequent PR if we want to think about it more.

pkg/autoupdate/tuf/channel_version.go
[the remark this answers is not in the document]
directionless: _ah_

So it's not used by this PR, and will be used by things that operate occasionally (package build, doctor). Given their nature, they need to talk to the server, and the additional overhead is okay. That makes sense.

I think we should name this something more clear, and godoc it appropriately. Maybe `GetVersionFromTufServer` or something?

pkg/autoupdate/tuf/library_lookup.go
directionless: Perfect. I'm getting turned around because there are many functions that all call each other so tracing the flow isn't super clear to me.

pkg/autoupdate/tuf/channel_version.go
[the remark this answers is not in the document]
RebeccaMahany: Yup I like that better, I will rename + add some docs!

pkg/autoupdate/tuf/library_lookup.go
RebeccaMahany: Discussed offline, updated to return a struct

pkg/autoupdate/tuf/library_lookup.go
RebeccaMahany: I'll leave as-is for now, I guess, and think on it

James-Pickett (approved):

PR #1186, opened 2023-05-08
Allow future new menu item actions to be easier to introduce

seejdev:
Similar in vein to https://github.com/kolide/launcher/pull/1154

While testing out an idea I realized that as it stands, launcher will fail to parse a menu entirely if/when a new menu item action is introduced. It's preferable to gracefully tolerate actions we don't recognize instead of failing. This change eliminates the error, and adds a `errorlessAction` capability which can be used in conjunction with `hasCapability` to introduce new actions while maintaining backwards compatibility.

### DIFF
--- a/ee/desktop/user/menu/action.go
+++ b/ee/desktop/user/menu/action.go
@@ -49,7 +49,9 @@ func (a *Action) UnmarshalJSON(data []byte) error {
 		}
 		a.Performer = openURL
 	default:
-		return fmt.Errorf("unknown action type: %s", action.Type)
+		// Silently ignore unrecognized actions because:
+		// 1. We don't have a logger reference here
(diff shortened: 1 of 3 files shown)


ee/desktop/user/menu/action.go
```diff
@@ -49,7 +49,9 @@ func (a *Action) UnmarshalJSON(data []byte) error {
 		}
 		a.Performer = openURL
 	default:
-		return fmt.Errorf("unknown action type: %s", action.Type)
+		// Silently ignore unrecognized actions because:
```
directionless, line 52: Should we log anything here?

directionless (approved):

ee/desktop/user/menu/action.go
directionless: Eh maybe not important

Issue #1104, opened 2023-03-15
`kolide_json` cannot read the `debug.json` file


directionless:
While poking around, I was surprised top see that `kolide_json` was unable to read the debug logs. 

```
pemberton:~ seph$ jq . /var/kolide-k2/k2device-preprod.kolide.com/debug.json | wc -l
    8569

pemberton:~ seph$ /usr/local/kolide-k2/bin/launcher interactive --osquery_flag verbose
osquery> select * from kolide_json where path = '/var/kolide-k2/k2device-preprod.kolide.com/debug.json';
osquery> 
```

Digging into it, it's because the debug logs, are a series of json objects. Eg:
```
{ "a": 1 }
{ "b": 2}
```

Which is probably a json extension. Or maybe an array. 

I'm not sure there's a simple compatible change for this.

PR #1189, opened 2023-05-08
james/json-line-format

James-Pickett:
adds new table `kolide_jsonl` for parsing json line formatted files

```sql
osquery> select fullkey, parent, key, value from kolide_json where path = '/var/kolide-k2/k2device-preprod.kolide.com/debug.json' limit 5;
+----------+--------+--------+-----------------------------+
| fullkey  | parent | key    | value                       |
+----------+--------+--------+-----------------------------+
| 0/ts     | 0      | ts     | 2023-04-28T16:33:58.362167Z |
| 0/caller | 0      | caller | launcher.go:70              |
| 0/level  | 0      | level  | debug                       |
| 0/msg    | 0      | msg    | runLauncher starting        |
| 1/caller | 1      | caller | checkpoint.go:93            |
+----------+--------+--------+-----------------------------+
```

closes https://github.com/kolide/launcher/issues/1104

### DIFF
--- a/pkg/dataflatten/flatten_test.go
+++ b/pkg/dataflatten/flatten_test.go
@@ -472,7 +472,7 @@ func TestFlatten(t *testing.T) {
 	}
 }
 
-func TestFlattenJsonl(t *testing.T) {
+func TestFlattenJsonlErrors(t *testing.T) {
 	t.Parallel()
 
 	var tests = []flattenTestCase{
@@ -481,11 +481,17 @@ func TestFlattenJsonl(t *testing.T) {
 			err: true,
 		},
 		{
+			// this test case was left over from attempting to parse json that
+			// is contained within a file that is not stricly jsonl
+			// it should error, maybe look at this again in the future?
 			comment: "valid json inline text",
 			in:      `valid json is hidden["a"]in me`,
 			err:     true,
 		},
 		{
+			// this test case was left over from attempting to parse json that
+			// is contained within a file that is not stricly jsonl
+			// it should error, maybe look at this again in the future?
 			comment: "valid json sandwich",
 			in: `
 			there is some json under me
--- a/pkg/dataflatten/flatten_test.go
+++ b/pkg/dataflatten/flatten_test.go
@@ -1,10 +1,12 @@
 package dataflatten
 
 import (
+	"bytes"
 	"encoding/json"
 	"os"
 	"path/filepath"
 	"sort"
+	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


pkg/dataflatten/json.go
```diff
@@ -1,25 +1,49 @@
...
+		var object interface{}
+		err := decoder.Decode(&object)
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+
+			return nil, fmt.Errorf("unmarshalling json: %w", err)
+		}
+		objects = append(objects, object)
```
directionless, line 36: This presents an interesting question. As written here, this will parse well formed jsonl data. But if we replaced the return with a continue, we'd also be able to parse:
```
Some test that might, happen 
{ a: 1, b: 2 }
To have a json blob in the middle
```

But I'm not sure that's a good idea.

pkg/dataflatten/json.go
```diff
@@ -1,25 +1,49 @@
...
 
-	if err := json.Unmarshal(rawdata, &data); err != nil {
-		return nil, fmt.Errorf("unmarshalling json: %w", err)
+	// if we have a single json object, don't flatten as an array
+	if len(objects) == 1 {
+		return Flatten(objects[0], opts...)
 	}
 
-	return Flatten(data, opts...)
+	return Flatten(objects, opts...)
```
directionless, line 44: I'm a little leery of this switch. My gut sense is that it's going to be hard to write SQL that handles it.

directionless (commented):
Code looks good, but I'm a little leery of changing the table semantics on a table we use so heavily, I'd suggest we keep `kolide_json` as is, and make a `kolide_jsonl`. Thoughts?

pkg/dataflatten/json.go
```diff
@@ -1,25 +1,49 @@
...
+	return Json(bufio.NewReader(f), opts...)
 }
 
-func Json(rawdata []byte, opts ...FlattenOpts) ([]Row, error) {
-	var data interface{}
+func Json(r io.Reader, opts ...FlattenOpts) ([]Row, error) {
+	decoder := json.NewDecoder(r)
+	var objects []interface{}
+
+	for {
```
directionless, line 26: I like how simple this is.

pkg/dataflatten/flatten_test.go
```diff
@@ -115,6 +115,126 @@ func TestFlatten_NestingBug(t *testing.T) {
...
+	// We do a bunch of tests to select this user. So we'll pull
+	// this out here and make the testcases more DRY
+	testdataUser0 := []Row{
+		{Path: []string{"2", "users", "0", "favorites", "0"}, Value: "ants"},
+		{Path: []string{"2", "users", "0", "name"}, Value: "Alex Aardvark"},
+		{Path: []string{"2", "users", "0", "uuid"}, Value: "abc123"},
+		{Path: []string{"2", "users", "0", "id"}, Value: "1"},
+	}
+
+	var tests = []flattenTestCase{
```
RebeccaMahany, line 134: Nice 🔥

pkg/dataflatten/json.go
James-Pickett: split out to `kolide_jsonl` table

pkg/dataflatten/json.go
James-Pickett: I tried messing with that ... it was very easy to hit infinite loop ... it might be possible, but not sure it's worth the effort

James-Pickett:
> Code looks good, but I'm a little leery of changing the table semantics on a table we use so heavily, I'd suggest we keep `kolide_json` as is, and make a `kolide_jsonl`. Thoughts?

ended up doing this

RebeccaMahany (approved):

pkg/dataflatten/jsonl.go
```diff
@@ -0,0 +1,36 @@
...
+	"io"
+	"os"
+)
+
+func JsonlFile(file string, opts ...FlattenOpts) ([]Row, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
+	return Jsonl(bufio.NewReader(f), opts...)
```
directionless, line 16: Why `bufio.NewReader(f)` here?

pkg/dataflatten/jsonl.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+func JsonlFile(file string, opts ...FlattenOpts) ([]Row, error) {
+	f, err := os.Open(file)
+	if err != nil {
+		return nil, err
+	}
```
directionless, line 15:
```suggestion
	f, err := os.Open(file)
	if err != nil {
		return nil, err
	}
	defer f.Close()
```
yah?

James-Pickett: ooops, thank you

pkg/dataflatten/jsonl.go
James-Pickett: I thought that bufio would do some magic under the the hood to make reading the file more efficient. I ran some bench marks using a file with millions of lines and there was no noticeable difference. Dropped bufio, now just passing file.

directionless (approved):
Wooooo!

PR #1190, opened 2023-05-09
Get `make containers` running with recent version of Golang and launcher

RebeccaMahany:
* Upgrades to golang 1.20, since that's what we build with now
* Updates to use most recent launcher tag instead of hardcoded tag
* Specify platform/arch everywhere for M1 compatibility

Testing steps:

Ran `make containers`, then ran with a few of the resulting images and confirmed it stayed up: `docker run --platform linux/amd64 -i -t gcr.io/kolide-public-containers/launcher-fakedata-ubuntu18:latest -debug -hostname test.ngrok.io -i-am-breaking-ee-license -enroll_secret "secret" -transport jsonrpc` (also tested with ubuntu20, centos7).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,8 +259,9 @@ build-docker: sha = $(shell git describe --always --abbrev=12)
 build-docker:
 	docker build -t launcher-build --build-arg gitver=$(sha) .
 
+build-dockerfake: version = $(shell git describe --always --tags --abbrev=12)
 build-dockerfake:
-	docker build -t launcher-fakedata-build --build-arg gitver=v0.11.21 --build-arg FAKE=-fakedata .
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -259,8 +259,9 @@ build-docker: sha = $(shell git describe --always --abbrev=12)
 build-docker:
 	docker build -t launcher-build --build-arg gitver=$(sha) .
 
+build-dockerfake: version = $(shell git describe --always --tags --abbrev=12)
```
directionless, line 262: If this works with docker, 👍

directionless (approved):

Makefile
RebeccaMahany: It's working all right (as far as I can tell) with `make containers` -- it's pretty much the same as the `sha = ` above

James-Pickett (approved):

PR #1195, opened 2023-05-12
[TUF autoupdater] Include arch in release file path and download file path

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1149, https://github.com/kolide/launcher/issues/954.


seejdev (approved):

RebeccaMahany:
Waiting until we've published at least one osquery release (now that changes in https://github.com/kolide/launcher/issues/1149 have been merged) to merge this

jaredallard:
Sorry to ping on this, but is there any news of when this would be? We're looking to run this on linux/arm64 and it seems like it's so close! @RebeccaMahany :pray:

RebeccaMahany:
@jaredallard Apologies for the delay on the response, I missed this! I'll be getting this PR through today since we've had a new osquery release -- however, the larger issue https://github.com/kolide/launcher/issues/1149 is also tied to an in-progress autoupdate rewrite https://github.com/kolide/launcher/issues/954 that I anticipate taking a little bit more time. I'll get https://github.com/kolide/launcher/issues/954 updated so that it'll be easier to follow it/track the status.

pkg/autoupdate/tuf/library_manager.go
```diff
@@ -121,7 +122,8 @@ func (ulm *updateLibraryManager) stageAndVerifyUpdate(binary autoupdatableBinary
 	stagedUpdatePath := filepath.Join(ulm.stagingDir, targetFilename)
 
 	// Request download from mirror
-	resp, err := ulm.mirrorClient.Get(ulm.mirrorUrl + fmt.Sprintf("/kolide/%s/%s/%s", binary, runtime.GOOS, targetFilename))
+	downloadPath := path.Join("/", "kolide", string(binary), runtime.GOOS, PlatformArch(), targetFilename)
+	resp, err := ulm.mirrorClient.Get(ulm.mirrorUrl + downloadPath)
```
directionless, line 126: `path.Join(ulm.mirrorUrl, "kolide", ...)` and you don't need the leading `/`, right?

directionless (approved):
Seems reasonable.

I think all the path generation is hard to follow. I don't have an easy suggestion for a fix, though. I think over in the MirrorTool library, that's probably why I ended up with the abstraction I did -- so the objects can generate their own paths. But I'm not sure to easily apply that here.

pkg/autoupdate/tuf/library_manager.go
RebeccaMahany: Unfortunately looks like this doesn't work because path.Join will strip the second / out of the scheme: https://play.golang.com/p/qD-wPK4x6Zw 😩

directionless: Oh.. This is URL constructions. I bet there's a URL helper. But this seems fine

Issue #1145, opened 2023-04-18
Add a `launcher doctor` command


directionless:
placeholder

PR #1197, opened 2023-05-16
launcher doctor subcommand

seejdev:
Fixed https://github.com/kolide/launcher/issues/1145

This PR adds a `launcher doctor` subcommand which runs several checkups to verify launcher installation health.

Notable things in this change:

- Doctor can be used with standard launcher options, since it's using the same underlying `parseOption` function.
- It can also fallback to reasonable defaults, even if a launcher installation, or config file, doesn't exist.
- Checkups are run sequentially and results outputted in a consistent manner via writers; this enables the output to easily be directed to stdout or a file or whatever.
- When run via CLI, the output will be colorized with a black background.
- Adding new checkups and modifying existing ones is straightforward and the underlying reporting/outputting can interact with arbitrary diagnostic functionality.
- Flare will now invoke doctor and pipe the output to the flare log file. Flare needs more work, but this is a start, and makes it more valuable.

### DIFF
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -30,12 +30,12 @@ import (
 )
 
 var (
+	// Command line colors and printing functions
 	cRunning = color.New(color.FgCyan)
 	cHeader  = color.New(color.FgHiWhite).Add(color.Bold)
 	cWarn    = color.New(color.FgHiYellow)
-
-	pass = color.New(color.FgGreen).PrintlnFunc()
-	fail = color.New(color.FgRed).Add(color.Bold).PrintlnFunc()
+	pass     = color.New(color.FgGreen).PrintlnFunc()
+	fail     = color.New(color.FgRed).Add(color.Bold).PrintlnFunc()
 
 	// Println functions for checkup details
 	cInfo = color.New(color.FgWhite)
@@ -55,7 +55,7 @@ var (
 	configFile string
 )
 
-// checkup is
+// checkup encapsulates a launcher health checkup
 type checkup struct {
 	name   string
 	check  func() (string, error)
@@ -138,6 +138,7 @@ func runDoctor(args []string) error {
 	return nil
 }
 
+// parseDoctorOptions parses command line options and provides defaults
 func parseDoctorOptions(args []string) (*launcher.Options, error) {
 	flagset := flag.NewFlagSet("launcher doctor", flag.ExitOnError)
 	flagset.Usage = func() { usage(flagset) }
@@ -350,6 +351,7 @@ func parseDoctorOptions(args []string) (*launcher.Options, error) {
 	return opts, nil
 }
 
+// run logs the results of a checkup being run
 func (c *checkup) run() {
 	if c.check == nil {
 	}
@@ -368,6 +370,7 @@ func (c *checkup) run() {
 	}
 }
 
+// checkupPlatform verifies that the current platform is supported by launcher
 func checkupPlatform(os string) (string, error) {
 	if slices.Contains([]string{"windows", "darwin", "linux"}, os) {
 		return fmt.Sprintf("Platform: %s", os), nil
@@ -380,6 +383,7 @@ type launcherFile struct {
 	found bool
 }
 
+// checkupRootDir tests for the presence of important files in the launcher root directory
 func checkupRootDir(filepaths []string) (string, error) {
 	importantFiles := []*launcherFile{
 		{
@@ -420,6 +424,7 @@ func checkupRootDir(filepaths []string) (string, error) {
 	return "", fmt.Errorf("%d root directory files not found", failures)
 }
 
+// checkupConnectivity tests connections to Kolide cloud services
 func checkupConnectivity(logger log.Logger, k types.Knapsack) (string, error) {
 	var failures int
 	checkpointer := checkpoint.New(logger, k)
@@ -451,6 +456,7 @@ func checkupConnectivity(logger log.Logger, k types.Knapsack) (string, error) {
 	}
 
 	for k, v := range notaryVersions {
+		// Check for failure if the notary version isn't a parsable integer
 		if _, err := strconv.ParseInt(v, 10, 32); err == nil {
 			good(fmt.Sprintf("%s - %s", k, v))
 		} else {
@@ -466,6 +472,7 @@ func checkupConnectivity(logger log.Logger, k types.Knapsack) (string, error) {
 	return "", fmt.Errorf("%d failures encountered while attempting communication with Kolide", failures)
 }
 
+// checkupVersion tests to see if the current launcher version is up to date
 func checkupVersion(v version.Info) (string, error) {
 	info(fmt.Sprintf("Current version: %s", v.Version))
 	// TODO: Query TUF for latest available version for this launcher instance
@@ -479,6 +486,7 @@ func checkupVersion(v version.Info) (string, error) {
 	return "", fmt.Errorf("launcher is out of date")
 }
 
+// checkupConfigFile tests that the config file is valid and logs it's contents
 func checkupConfigFile(filepath string) (string, error) {
 	file, err := os.Open(filepath)
 	if err != nil {
@@ -494,6 +502,7 @@ func checkupConfigFile(filepath string) (string, error) {
 	return "Config file found", nil
 }
 
+// checkupLogFiles checks to see if expected log files are present
 func checkupLogFiles(filepaths []string) (string, error) {
 	var foundCurrentLogFile bool
 	for _, f := range filepaths {
@@ -519,6 +528,7 @@ func checkupLogFiles(filepaths []string) (string, error) {
 	return "", fmt.Errorf("No log file found")
 }
 
+// getFilepaths returns a list of file paths matching the pattern
 func getFilepaths(elem ...string) []string {
 	fileGlob := filepath.Join(elem...)
 	filepaths, err := filepath.Glob(fileGlob)
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -325,11 +325,7 @@ func checkupOsquery(updateChannel, tufServerURL, osquerydPath string) (string, e
 	}
 
 	info(fmt.Sprintf("Target version:\t%s", targetVersion))
-
-	if targetVersion != currentVersion {
-		return "", fmt.Errorf("osqueryd is out of date")
-	}
-	return "Up to date!", nil
+	return "Osquery version checks complete", nil
 }
 
 func checkupFilesPresent(filepaths []string, importantFiles []*launcherFile) (string, error) {
@@ -421,11 +417,7 @@ func checkupVersion(updateChannel, tufServerURL string, v version.Info) (string,
 	}
 
 	info(fmt.Sprintf("Target version:\t%s", targetVersion))
-
-	if targetVersion != v.Version {
-		return "", fmt.Errorf("launcher is out of date")
-	}
-	return "Up to date!", nil
+	return "Launcher version checks complete", nil
 }
 
 // checkupConfigFile tests that the config file is valid and logs it's contents
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -82,7 +82,7 @@ func runDoctor(args []string) error {
 
 	checkups := []*checkup{
 		{
-			name: "Check platform",
+			name: "Platform",
 			check: func() (string, error) {
 				return checkupPlatform(runtime.GOOS)
 			},
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -30,12 +30,12 @@ var (
 	doctorWriter io.Writer
 
 	// Command line colors
-	cyanText   = color.New(color.FgCyan)
-	headerText = color.New(color.Bold, color.FgHiWhite)
-	yellowText = color.New(color.FgHiYellow)
-	whiteText  = color.New(color.FgWhite)
-	greenText  = color.New(color.FgGreen)
-	redText    = color.New(color.Bold, color.FgRed)
+	cyanText   = color.New(color.FgCyan, color.BgBlack)
+	headerText = color.New(color.Bold, color.FgHiWhite, color.BgBlack)
+	yellowText = color.New(color.FgHiYellow, color.BgBlack)
+	whiteText  = color.New(color.FgWhite, color.BgBlack)
+	greenText  = color.New(color.FgGreen, color.BgBlack)
+	redText    = color.New(color.Bold, color.FgRed, color.BgBlack)
 
 	// Printf functions
 	cyan = func(format string, a ...interface{}) {
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -379,9 +379,10 @@ func runCheckups(checkups []*checkup) {
 		for _, c := range failedCheckups {
 			fail(fmt.Sprintf("    %s\n", c.name))
 		}
-	} else {
-		greenText("\nAll checkups passed! Your Kolide launcher is healthy.")
+		return
 	}
+	
+	greenText("\nAll checkups passed! Your Kolide launcher is healthy.")
 }
 
 // run logs the results of a checkup being run
@@ -399,11 +400,10 @@ func (c *checkup) run() error {
 		fail(err)
 		redText("𐄂 Checkup failed!")
 		return err
-	} else {
-		pass(result)
-		greenText("✔ Checkup passed!")
 	}
-
+	
+	pass(result)
+	greenText("✔ Checkup passed!")
 	return nil
 }
 
@@ -496,23 +496,23 @@ func checkupConnectivity(logger log.Logger, k types.Knapsack) (string, error) {
 	checkpointer := checkpoint.New(logger, k)
 	connections := checkpointer.Connections()
 	for k, v := range connections {
-		if v == "successful tcp connection" {
-			pass(fmt.Sprintf("%s - %s", k, v))
-		} else {
+		if v != "successful tcp connection" {
 			fail(fmt.Sprintf("%s - %s", k, v))
 			failures = failures + 1
+			continue
 		}
+		pass(fmt.Sprintf("%s - %s", k, v))
 	}
 
 	ipLookups := checkpointer.IpLookups()
 	for k, v := range ipLookups {
 		valStrSlice, ok := v.([]string)
-		if ok && len(valStrSlice) > 0 {
-			pass(fmt.Sprintf("%s - %s", k, valStrSlice))
-		} else {
+		if !ok || len(valStrSlice) == 0 {
 			fail(fmt.Sprintf("%s - %s", k, valStrSlice))
 			failures = failures + 1
+			continue
 		}
+		pass(fmt.Sprintf("%s - %s", k, valStrSlice))
 	}
 
 	notaryVersions, err := checkpointer.NotaryVersions()
@@ -523,12 +523,12 @@ func checkupConnectivity(logger log.Logger, k types.Knapsack) (string, error) {
 
 	for k, v := range notaryVersions {
 		// Check for failure if the notary version isn't a parsable integer
-		if _, err := strconv.ParseInt(v, 10, 32); err == nil {
-			pass(fmt.Sprintf("%s - %s", k, v))
-		} else {
+		if _, err := strconv.ParseInt(v, 10, 32); err != nil {
 			fail(fmt.Sprintf("%s - %s", k, v))
 			failures = failures + 1
+			continue
 		}
+		pass(fmt.Sprintf("%s - %s", k, v)
 	}
 
 	if failures == 0 {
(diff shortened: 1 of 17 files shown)


cmd/launcher/doctor.go
```diff
@@ -0,0 +1,531 @@
...
+	if failures == 0 {
+		return "Successfully communicated with Kolide", nil
+	}
+
+	return "", fmt.Errorf("%d failures encountered while attempting communication with Kolide", failures)
+}
+
+func checkupVersion(v version.Info) (string, error) {
+	info(fmt.Sprintf("Current version: %s", v.Version))
+	// TODO: Query TUF for latest available version for this launcher instance
```
seejdev, line 471: Not yet sure if this is easy to do? Have to factor in current version, update channel, etc.

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,531 @@
...
+			fail("%s\n", c.name)
+		}
+	} else {
+		pass("\nAll checkups passed! Your Kolide launcher is healthy.")
+	}
+
+	return nil
+}
+
+func parseDoctorOptions(args []string) (*launcher.Options, error) {
```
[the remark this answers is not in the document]
RebeccaMahany: I think some consolidation makes sense -- I would be worried about getting out of sync between parseOptions and this one when adding a new command-line flag

cmd/launcher/doctor.go
RebeccaMahany: I think you could do this fairly simply once https://github.com/kolide/launcher/pull/1185 is merged in:

1. Export this function so you can init a metadata client for yourself here: https://github.com/kolide/launcher/blob/main/pkg/autoupdate/tuf/autoupdate.go#L117
1. Export `findRelease`: https://github.com/kolide/launcher/blob/0ae45a14810c2d3be77d786840def2bfe997cad0/pkg/autoupdate/tuf/autoupdate.go#L328
1. Export `versionFromTarget` or implement it yourself: https://github.com/kolide/launcher/blob/main/pkg/autoupdate/tuf/library_manager.go

Then something like:

```
// Make a temp TUF directory so that we don't change anything about our currently in-use local TUF repo
tempTufRepoDir, _ := agent.MkdirTemp("temp-tuf")

// Get a metadata client -- this will be initialized with the root metadata we distribute, so we can trust the downloads it performs
tufClient, _ := tuf.InitMetadataClient(tempTufRepoDir, k.TufServerURL(), http.DefaultClient)

// Talks to TUF server to fetch all targets
tufClient.Update()

// Finds the release target (release.json), finds the target it points to, gets the version from that target
targets, _ := tufClient.Targets()
releaseVersion, _, _ := tuf.FindRelease("launcher", targets, k.UpdateChannel())
fmt.Println(tuf.VersionFromTarget("launcher", releaseVersion))
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,541 @@
...
+
+	// TODO: Choose failure based on current >= target
+	if true {
+		return "Up to date!", nil
+	}
+
+	return "", fmt.Errorf("launcher is out of date")
+}
+
+// checkupConfigFile tests that the config file is valid and logs it's contents
```
RebeccaMahany, line 489: How does this validate the contents?

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,541 @@
...
+
+	cRunning.Println("Kolide launcher doctor version:")
+	version.PrintFull()
+	cRunning.Println("\nRunning Kolide launcher checkups...")
+
+	checkups := []*checkup{
+		{
+			name: "Check platform",
+			check: func() (string, error) {
+				return checkupPlatform(runtime.GOOS)
```
RebeccaMahany, line 85: With https://github.com/kolide/launcher/issues/1149 upcoming, maybe helpful to do a similar check for runtime.GOARCH?

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,541 @@
...
+
+	fcOpts := []flags.Option{flags.WithCmdLineOpts(opts)}
+	flagController := flags.NewFlagController(logger, nil, fcOpts...)
+	k := knapsack.New(nil, flagController, nil)
+
+	cRunning.Println("Kolide launcher doctor version:")
+	version.PrintFull()
+	cRunning.Println("\nRunning Kolide launcher checkups...")
+
+	checkups := []*checkup{
```
RebeccaMahany, line 81: Would we want to also check for --

* Launcher process exists?
* Launcher desktop process exists?
* Is there anything we could check for osquery?

cmd/launcher/doctor.go
seejdev: Added a `PlainParser` to actually validate now, thanks!

cmd/launcher/doctor.go
seejdev: Done!

cmd/launcher/doctor.go
seejdev: I'll add osquery binary and logs checks next.

A launcher process report could be useful info

pkg/log/checkpoint/checkpoint.go
```diff
@@ -128,29 +138,28 @@ func (c *checkPointer) logKolideServerVersion() {
...
-		c.logger.Log("notary versions", err)
+		return nil, err
 	} else {
-		c.logger.Log("notary versions", fetchNotaryVersions(httpClient, notaryUrl))
+		return fetchNotaryVersions(httpClient, notaryUrl), nil
 	}
 }
 
-func (c *checkPointer) logConnections() {
+func (c *checkPointer) Connections() map[string]string {
```
directionless, line 155: Seems fine. Another route would be to take a logger in.

cmd/launcher/paths.go
```diff
@@ -0,0 +1,57 @@
...
+func getAppBinaryPaths() []string {
+	var paths []string
+	switch runtime.GOOS {
+	case "darwin":
+		paths = []string{
+			filepath.Join(binDir, "Kolide.app", "Contents", "MacOS", "launcher"),
+		}
+	case "linux":
+		paths = []string{
+			filepath.Join(binDir, "launcher.exe"),
```
directionless, line 37:
```suggestion
			filepath.Join(binDir, "launcher"),
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	}
+
+	configFile string
+	binDir     string
+)
+
+// checkup encapsulates a launcher health checkup
+type checkup struct {
+	name  string
+	check func() (string, error)
```
directionless, line 64: This signature feels very simple, maybe a good enough start. But I think there are a couple things I don't know how to do. (but maybe they don't matter)

First, I don't think we can differentiate between an unexpected failure in code, vs code works but a broken condition. Like, if we failed to open a boltdb connection that feels different than if some piece of data isn't what we need. Both are the same failure to the user, but one feels like a protocol error, and the other feels like the correct reporting of data.

Second, I'm not sure how to represent warnings here. 

If I combine those, I wonder about `func() (info string, status enum, error)`

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	flagController := flags.NewFlagController(logger, nil, fcOpts...)
+	k := knapsack.New(nil, flagController, nil)
+
+	cyanText.Println("Kolide launcher doctor version:")
+	version.PrintFull()
+	cyanText.Println("\nRunning Kolide launcher checkups...")
+
+	checkups := []*checkup{
+		{
+			name: "Check platform",
```
directionless, line 85:
```suggestion
			name: "Platform",
```

cmd/launcher/doctor.go
directionless: But that suggestion makes it hard to differentiate the output for flare vs doctor.

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+			name: "Check architecture",
+			check: func() (string, error) {
+				return checkupArch(runtime.GOARCH)
+			},
+		},
+		{
+			name: "Root directory contents",
+			check: func() (string, error) {
+				return checkupRootDir(getFilepaths(k.RootDirectory(), "*"))
+			},
```
directionless, line 100: I haven't run this, but thinking aloud I see a difference in `doctor` and `flare`.

For `doctor`, I think this would just want to check if the directory exists. Maybe permissions. But `flare` would want to display the contents. I suspect with reasonable clever callers and signatures, this is supportable by the same underlying checkup function

cmd/launcher/doctor.go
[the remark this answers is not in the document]
directionless: Sadly, I'm not sure we have great options.

I think we're carrying a lot of weird baggage because:
* launcher have a lot of options
* we use go's flags library
* flags does not support ignoring unknown options

So I guess we maybe only have a couple choices. We could push flag parsing into something shared.

Or we could use a different flag/config file parser for things like `doctor` which have pretty narrow needs.

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+func parseDoctorOptions(args []string) (*launcher.Options, error) {
+	flagset := flag.NewFlagSet("launcher doctor", flag.ExitOnError)
+	flagset.Usage = commandUsage(flagset, "launcher doctor")
+
+	var defaultRootDir, defaultEtcDir, defaultConfigFile string
+	getDefaults(&defaultRootDir, &defaultEtcDir, &binDir, &defaultConfigFile)
+
+	var (
+		// Primary options
+		flAutoloadedExtensions   arrayFlags
```
directionless, line 161: If we don't need these, I think we can use a lot of `_`

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+		os.Exit(0)
+	}
+
+	configFile = *flConfigFile
+
+	// If launcher is using a kolide host, we may override many of
+	// the settings. When we're ready, we can _additionally_
+	// conditionalize this on the ServerURL to get all the
+	// existing deployments
+	if *flKolideHosted {
```
directionless, line 246: definitely worried about divergence with launcher

cmd/launcher/doctor.go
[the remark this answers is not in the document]
directionless: Though some of this could, maybe, be absorbed by knapsack's flag stuff. Not sure?

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+}
+
+func runCheckups(checkups []*checkup) {
+	failedCheckups := []*checkup{}
+
+	// Sequentially run all of the checkups
+	for _, c := range checkups {
+		err := c.run()
+		if err != nil {
+			failedCheckups = append(failedCheckups, c)
```
directionless, line 372: Why build the array, and not just print here?

cmd/launcher/doctor.go
directionless: Oh, I see now. It's in the check's run call. hmmmmm

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	"github.com/shirou/gopsutil/v3/process"
+
+	"github.com/fatih/color"
+	"golang.org/x/exp/slices"
+)
+
+var (
+	// Command line colors
+	cyanText   = color.New(color.FgCyan)
+	headerText = color.New(color.FgHiWhite).Add(color.Bold)
```
directionless, line 35: This is unreadable on a white background. I don't think it needs color

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	"github.com/peterbourgon/ff/v3"
+	"github.com/shirou/gopsutil/v3/process"
+
+	"github.com/fatih/color"
+	"golang.org/x/exp/slices"
+)
+
+var (
+	// Command line colors
+	cyanText   = color.New(color.FgCyan)
```
directionless, line 34: This one is also hard to read. I wonder if bold would fit better.

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	headerText = color.New(color.FgHiWhite).Add(color.Bold)
+	yellowText = color.New(color.FgHiYellow)
```
directionless, line 36: Also cannot read

directionless (commented):
Looks pretty good

cmd/launcher/doctor.go
RebeccaMahany: I'm also going to have to do something very similar for the packaging/installer part of https://github.com/kolide/launcher/issues/1149 (make sure we download the correct version for the given channel, once we can't just download `launcher-stable.tar.gz` any more). So maybe I could just add an exported function to the `tuf` package that does this.

RebeccaMahany: Added 🙂 https://github.com/kolide/launcher/pull/1185/commits/b612e8158037119191261db6e769200219b05bdf

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+
+	if len(failedCheckups) > 0 {
+		redText("\nSome checkups failed:")
+
+		for _, c := range failedCheckups {
+			fail(fmt.Sprintf("    %s\n", c.name))
+		}
+	} else {
+		greenText("\nAll checkups passed! Your Kolide launcher is healthy.")
+	}
```
James-Pickett, line 384:
```suggestion
	if len(failedCheckups) > 0 {
		redText("\nSome checkups failed:")

		for _, c := range failedCheckups {
			fail(fmt.Sprintf("    %s\n", c.name))
		}
		return
	}
	
	greenText("\nAll checkups passed! Your Kolide launcher is healthy.")
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+		info(result)
+		fail(err)
+		redText("𐄂 Checkup failed!")
+		return err
+	} else {
+		pass(result)
+		greenText("✔ Checkup passed!")
+	}
+
+	return nil
```
James-Pickett, line 407:
```suggestion
	if err != nil {
		info(result)
		fail(err)
		redText("𐄂 Checkup failed!")
		return err
	}
	
	pass(result)
	greenText("✔ Checkup passed!")
	return nil
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	checkpointer := checkpoint.New(logger, k)
+	connections := checkpointer.Connections()
+	for k, v := range connections {
+		if v == "successful tcp connection" {
+			pass(fmt.Sprintf("%s - %s", k, v))
+		} else {
+			fail(fmt.Sprintf("%s - %s", k, v))
+			failures = failures + 1
+		}
+	}
```
James-Pickett, line 505:
```suggestion
	for k, v := range connections {
		if v != "successful tcp connection" {
			fail(fmt.Sprintf("%s - %s", k, v))
			failures = failures + 1
			continue
		}
		pass(fmt.Sprintf("%s - %s", k, v))
	}
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	ipLookups := checkpointer.IpLookups()
+	for k, v := range ipLookups {
+		valStrSlice, ok := v.([]string)
+		if ok && len(valStrSlice) > 0 {
+			pass(fmt.Sprintf("%s - %s", k, valStrSlice))
+		} else {
+			fail(fmt.Sprintf("%s - %s", k, valStrSlice))
+			failures = failures + 1
+		}
+	}
```
James-Pickett, line 516:
```suggestion
	for k, v := range ipLookups {
		valStrSlice, ok := v.([]string)
		if !ok || len(valStrSlice) == 0 {
			fail(fmt.Sprintf("%s - %s", k, valStrSlice))
			failures = failures + 1
			continue
		}
		pass(fmt.Sprintf("%s - %s", k, valStrSlice))
	}
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+
+	for k, v := range notaryVersions {
+		// Check for failure if the notary version isn't a parsable integer
+		if _, err := strconv.ParseInt(v, 10, 32); err == nil {
+			pass(fmt.Sprintf("%s - %s", k, v))
+		} else {
+			fail(fmt.Sprintf("%s - %s", k, v))
+			failures = failures + 1
+		}
+	}
```
James-Pickett, line 532:
```suggestion
	for k, v := range notaryVersions {
		// Check for failure if the notary version isn't a parsable integer
		if _, err := strconv.ParseInt(v, 10, 32); err != nil {
			fail(fmt.Sprintf("%s - %s", k, v))
			failures = failures + 1
			continue
		}
		pass(fmt.Sprintf("%s - %s", k, v)
	}
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+	// Parse the config file how launcher would
+	err = ff.PlainParser(file, func(name, value string) error {
+		info(fmt.Sprintf("%s %s", name, value))
+		return nil
+	})
+
+	if err == nil {
+		return "Config file found", nil
+	}
+	return "", fmt.Errorf("Invalid config file")
```
James-Pickett, line 572: I think we should keep the consistency of checking for err != nil when possible and keep happy path to the left.

```suggestion
	if err != nil {
		return "", fmt.Errorf("Invalid config file")
	}
	return "Config file found", nil
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+
+			fi, err := os.Stat(f)
+			if err == nil {
+				info("")
+				info(fmt.Sprintf("Most recent log file: %s", filename))
+				info(fmt.Sprintf("Latest modification: %s", fi.ModTime().String()))
+				info(fmt.Sprintf("File size (B): %d", fi.Size()))
+			}
+		}
+	}
```
James-Pickett, line 593: lets try to reduce some nesting:
```suggestion
	for _, f := range filepaths {
		filename := filepath.Base(f)
		info(filename)
		
		if filename != "debug.json" {
		    continue
		}

	        foundCurrentLogFile = true
        
	        fi, err := os.Stat(f)
	        if err != nil {
	            continue
	        }

		info("")
		info(fmt.Sprintf("Most recent log file: %s", filename))
		info(fmt.Sprintf("Latest modification: %s", fi.ModTime().String()))
		info(fmt.Sprintf("File size (B): %d", fi.Size()))
	}
```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+
+	if foundCurrentLogFile {
+		return "Log file found", nil
+	}
+	return "", fmt.Errorf("No log file found")
```
James-Pickett, line 598: happy path to the left

```suggestion
	if !foundCurrentLogFile {
	    return "", fmt.Errorf("No log file found")
	}
	
	return "Log file found", nil

```

cmd/launcher/doctor.go
```diff
@@ -0,0 +1,625 @@
...
+			args, _ := p.Cmdline()
+			user, _ := p.Username()
+			info(fmt.Sprintf("%s %d  %s  %s", user, p.Pid, name, args))
+		}
+	}
+
+	if foundKolide {
+		return "Launcher processes found", nil
+	}
+	return "", fmt.Errorf("No launcher processes found")
```
James-Pickett, line 624:
```suggestion
	if !foundKolide {
	        return "", fmt.Errorf("No launcher processes found")
	}
	return "Launcher processes found", nil
```

James-Pickett (commented):
nice work! I like the overall approach and the UX. Looks like it will be straight forward to add new checkups in the future. like the other comments, I think we should get smart about consolidating flags and preventing drift between what launcher needs and what doctor thinks it needs

cmd/launcher/doctor.go
seejdev: Merged these modifications back into the original `parseOptions` function so we're not diverging.

cmd/launcher/doctor.go
seejdev: I felt it was easier to quickly understand what checkups succeeded/failed if the summary is the last thing printed, rather than whoever is looking through this needing to scan through the whole file to find failed checkups.

directionless: Okay!

cmd/launcher/doctor.go
seejdev: I've forced this always be printed to a black background now so we get consistent results regardless of end user shell.

cmd/launcher/options.go
```diff
@@ -24,6 +24,17 @@ const (
 	skipEnvParse         = runtime.GOOS == "windows" // skip environmental variable parsing on windows
 )
 
+var (
+	// When launcher proper runs, it's expected that these defaults are their zero values
+	// However, special launcher subcommands such as launcher doctor can override these
```
directionless, line 29: Huh, this is an interesting way to do this.

directionless (approved):
This has value as it is, and I think it's isolated enough to be safe to merge.

I kinda want to quibble over:
* how the output looks -- I really like the grouping, but not the forced colors
* flare's internals
* checkup interfaces

But I think we can iterate.

directionless (dismissed):

James-Pickett (approved):
nice!

PR #1198, opened 2023-05-16
Start to include arch in packaging targets + packaging commands

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1149

This is a small part of the packaging updates: specifying the architecture we're building for when building packages, to allow us to later specify arm64 when we'd like to build arm64 packages. For macOS, this amounted to moving a hardcoded line out of the dist file, to be set by an option instead. For our FPM packaging, this amounted to passing in the `-a` flag, again set by an option. For our Wix build, this is already set: https://github.com/kolide/launcher/blob/main/pkg/packagekit/wix/wix.go#L274

### DIFF
--- a/pkg/packaging/target.go
+++ b/pkg/packaging/target.go
@@ -86,11 +86,11 @@ func (t *Target) Parse(s string) error {
 	}
 
 	// For now, set the default arch according to the given platform
-	if defaultArch, ok := defaultArchMap[t.Platform]; ok {
-		t.Arch = defaultArch
-	} else {
+	defaultArch, ok := defaultArchMap[t.Platform]
+	if !ok {
 		return fmt.Errorf("cannot select default arch for unknown platform %s", t.Platform)
 	}
+	t.Arch = defaultArch
 
 	return nil
 }
--- a/pkg/packaging/target.go
+++ b/pkg/packaging/target.go
@@ -12,6 +12,7 @@ type Target struct {
 	Init     InitFlavor
 	Package  PackageFlavor
 	Platform PlatformFlavor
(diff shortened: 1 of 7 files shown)


pkg/packaging/target.go
```diff
@@ -70,6 +85,13 @@ func (t *Target) Parse(s string) error {
 		return err
 	}
 
+	// For now, set the default arch according to the given platform
+	if defaultArch, ok := defaultArchMap[t.Platform]; ok {
+		t.Arch = defaultArch
+	} else {
+		return fmt.Errorf("cannot select default arch for unknown platform %s", t.Platform)
+	}
+
```
James-Pickett, line 94: prefer happy path on left, reduce nesting

```suggestion
	// For now, set the default arch according to the given platform
	if defaultArch, ok := defaultArchMap[t.Platform]; !ok {
		return fmt.Errorf("cannot select default arch for unknown platform %s", t.Platform)
	}
	t.Arch = defaultArch
```

directionless (commented):
Shape looks okay. Do you also need to patch kworker, or will this fallback to universal? I agree with james about early returns.

RebeccaMahany:
> Shape looks okay. Do you also need to patch kworker, or will this fallback to universal? I agree with james about early returns.

@directionless it'll fall back to universal for darwin and amd64 for the other two -- no new packaging options or command-line flags at the moment!

pkg/packaging/target.go
RebeccaMahany: Updated!

seejdev (approved):

PR #1210, opened 2023-05-25
osquery runner waits for socket to appear before moving on

James-Pickett:
by throwing the osquery start cmd in a go routine and having it wait for half a second before starting, while allowing the rest of the runner to keep going, I was able to produce the error. Kind a feels like monteray bug, but not exactly. Still playing with it. 

```
{"caller":"queried.go:71","component":"log checkpoint","err":"error querying for static: could not query the extension manager client: write unix -\u003e/var/kolide-k2/k2device-preprod.kolide.com/osquery.sock: write: broken pipe","msg":"failed to query os info","severity":"info","ts":"2023-05-25T18:52:16.878345Z"}
```

you can wrap the [osquery start function](https://github.com/James-Pickett/launcher/blob/james/osquery-wait-for-socket/pkg/osquery/runtime/runner.go#L302) in something like this:
```go
	go func() {
		sleeptime := time.Millisecond * 1000
		time.Sleep(sleeptime)
		err = o.cmd.Start()
		if err != nil {
			// Failure here is indicative of a failure to exec. A missing
			// binary? Bad permissions? TODO: Consider catching errors in the
			// update system and falling back to an earlier version.
			msgPairs := append(
				getOsqueryInfoForLog(o.cmd.Path),
				"msg", "Fatal error starting osquery. Could not exec.",
				"err", err,
			)

			level.Info(o.logger).Log(msgPairs...)
			// return fmt.Errorf("fatal error starting osqueryd process: %w", err)
		}
	}()
```

to try and simulate a slow start

with out the remove socket / wait for socket before an after start it launcher will fail

adding the socket checks hardens the osquery runtime


pkg/osquery/runtime/runner.go
```diff
@@ -290,6 +291,13 @@ func (r *Runner) launchOsqueryInstance() error {
 		"args", strings.Join(o.cmd.Args, " "),
 	)
 
+	// remove osquery socket if exists
+	if err := backoff.WaitFor(func() error {
```
directionless, line 295: Do you think this needs the retry?

pkg/osquery/runtime/runner.go
```diff
@@ -306,6 +314,14 @@ func (r *Runner) launchOsqueryInstance() error {
 		return fmt.Errorf("fatal error starting osqueryd process: %w", err)
 	}
 
+	// wait for osquery to create the socket before moving on
+	if err := backoff.WaitFor(func() error {
+		_, err := os.Stat(paths.extensionSocketPath)
+		return err
+	}, 1*time.Minute, 1*time.Second); err != nil {
+		return fmt.Errorf("timeout waiting for osqueryd to create socket at %s: %w", paths.extensionSocketPath, err)
+	}
```
directionless, line 323: Hrm.... You are extending the dragon region.

I think what this is trying to do is detect if osquery us running and healthy. (The comment should reflect that)

I naively think we should be using some `Healthy()` routine, and not just checking that the path exists. But if I go reading `Health()` I see that that we don't really separate the underlying health of osquery from the health of the extension manager. _And_ even if we did, we'd need to carefully work through socket contention and locks

So many checking for the socket is a simple thing we can do.

Otherwise, I think were looking at pulling in https://github.com/osquery/osquery-go/pull/107 and generally creating much more careful sequencing here.

directionless (commented):
Hrm. I think you're onto something here, and oddly, I thought this worked differently. 

I had thought that launcher created the socket, and that osquery then utilized it. But I don't see that happening. 

So I think what you're trying to do is add some kind of health check around the osquery startup. This makes sense, and even thought we had one, but I don't see that either.

pkg/osquery/runtime/runner.go
James-Pickett: Actually I added the retry because I was thinking of windows and that you can't delete an opened file ... but on windows we create a random socket path so it doesn't matter. I'll remove it

pkg/osquery/runtime/runner.go
James-Pickett: I think the dragon region is below a bit =), but yea it makes this func even longer.

directionless (approved):
I think this fits in https://github.com/kolide/launcher/issues/1211 somehow. My guess is this is a step towards some of the refactoring we'll need to do in here. And I really like the slow start test functionality. 

I think it's fine to ship

pkg/osquery/runtime/runner.go
```diff
@@ -290,8 +291,16 @@ func (r *Runner) launchOsqueryInstance() error {
 		"args", strings.Join(o.cmd.Args, " "),
 	)
 
+	if err := os.RemoveAll(paths.extensionSocketPath); err != nil {
```
RebeccaMahany, line 294: Maybe add a quick comment about why we do this here?

RebeccaMahany (dismissed):
LGTM, just a small suggestion re documentation

pkg/osquery/runtime/runner.go
James-Pickett: good idea

RebeccaMahany (approved):

directionless:
This relates to #1211

Issue #1211, opened 2023-05-30
Inconsistent issues when starting up


directionless:
We have recently started seeing reports around startup issues. These manifest as needing to wait a long time, or install a couple times (which triggers process restarts). Log files tend to be a bit muddled (partly due to #1205), but often will show some issues around the socket.

I believe this is a return of _The Monterey Bug_ (#813 and [internal issue](https://github.com/kolide/k2/issues/4737)).

I spent awhile staring at how launcher starts up, and I'm sure it's still kinda racy. There is huge amount of complexity in our runtime. https://github.com/kolide/launcher/blob/7a926d039a1c1ed5515f16759653d68343be7c0d/pkg/osquery/runtime/runner.go#L354-L365

I believe the first steps are to:
1. **Move enrollment info _off_ the socket:** I think a big part of the startup contention is that osquery enrollment goes through launcher, which needs to talk to osquery to get the extra details. While this isn't _inherently_ a race, it's very very finicky. I've come to think this is way too clever, and we should simply invoke osquery in single mode and run what we need. https://github.com/kolide/launcher/pull/1213
2. **`osquery-go` should enforce a single user:** One of the deep issues here, is that osquery/thrift only supports a single user of the thrift socket at a time. And, if you have goroutines, it's very easy to get that confused. See work in https://github.com/osquery/osquery-go/pull/107 and https://github.com/osquery/osquery-go/pull/108 (both extracted from https://github.com/osquery/osquery-go/pull/99)

I suspect that will completely remove underlying startup issue.

I think we should _also_ rewrite the entire runtime to be much simpler. At the very core, its:
1. Starting osquery
2. Registering `kolide_grpc`
3. Registering `kolide_tables`
4. Providing a `Querier` to the rest of the code

But it has mountains of spaghetti that have accrued over the years.

directionless:
Possible one precipitating event, is that macOS needs to verify the validity of signed things. This means that it'll hit Apple's servers to verify notarization, which might cause unexpected and inconsistent startup delays.

If we shipped osquery as a bundle, we could staple it.

directionless:
I think we've fixed this

PR #1212, opened 2023-05-30
Use shared osquery extension client, with timeout

RebeccaMahany:
Pulls in changes from https://github.com/osquery/osquery-go/pull/108.

Removes the client lock since it's now handled in osquery-go.

Relates To: https://github.com/kolide/launcher/issues/1211


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -219,10 +219,9 @@ type OsqueryInstance struct {
 	doneCtx                 context.Context
 	cancel                  context.CancelFunc
 	cmd                     *exec.Cmd
-	emsLock                 sync.RWMutex
+	emsLock                 sync.RWMutex // Lock for extensionManagerServers
```
directionless, line 222: I'm not sure still need this. Maybe? I can't remember if the intent was to gate access to the thrift socket, which osquery-go should be doing, or gate behavior inside our extensions

RebeccaMahany: This is just a lock for the array so that we don't add/remove servers from the array while we're reading from it, so I think we do still want it. I added a comment about this below.

directionless: Oh, I see. I guess that makes sense. I'm not sure our normal usage hits it, but sure.

directionless (approved):

directionless:
This relates to #1211

PR #1213, opened 2023-05-30
Get enrollment details via exec, not the thrift socket

directionless:
Part of my theory around #1211 is that some of the initial startup raciness comes from the initial enrollment details. Launcher is using the socket to fetch them, at the time it's registering, and, and and and...

This PR shifts the enrollment details from an thrift API fetch, to an exec of osquery. It's an extra exec, which isn't ideal, but seems like the lessor of the evils here

TODO:
* [x] Fix tests
* [x] Update ADR
* [x] manually test enrollment macos
* [ ] manually test enrollment windows
* [ ] manually test enrollment linux

### DIFF
--- a/docs/architecture/2023-09-29_request_enrollment_details_updates.md
+++ b/docs/architecture/2023-09-29_request_enrollment_details_updates.md
@@ -1,0 +1,46 @@
+# Revisiting enrollment details
+
+## Authors
+
+- seph ([@directionless](https://github.com/directionless))
+
+## Status
+
+Accepted (2023-09)
+
+## Context
+
+Over the last several years, roughly since late 2021, we've seen
+occasional problems starting up. Internally, we've called this _The
+Monterey Bug_, and there is some information and links in [GitHub
+Issue #1211(https://github.com/kolide/launcher/issues/1211).
+
+We have never managed to diagnose this. We have reproduced small parts
+of it, but nothing that holds up over all.
+
+Our leading theories are that it is somehow related to runtime
+complexity, osquery startup time, and thrift socket contention. There
+may be multiple related and unrelated issues at play.
+
+There is additional complexity that stems from the original
+implementation of
+[`getEnrollmentDetails`](https://github.com/kolide/launcher/blob/ab411f07d1d147b963809df2e1fdb04cb574d1a3/pkg/osquery/extension.go#L934). Because
+it use the osquery socket, it cannot run until osquery is started. But
+simultaneously launcher is trying to register extensions, and osquery
+is trying to enroll. 
+
+## Decision
+
+To both simplify startup ordering _and_ reduce socket contention, we
+can gather enrollment details via execing osquery. Semantically, this
+is a fairly simple change -- we can use the same query, and the same
+osquery.
+
+## Consequences
+
+We incur an exec call during startup. But in return, we can gain
+several benefits: 
+- Decouple enrollment details from the main osquery startup.
+- Reduce contention on the socket during early startup
+- Enrollment no longer has a circular dependency
+- Enables future work to completely pull enrollment into launcher 
--- a/docs/architecture/2023-09-29_request_enrollment_details_updates.md
+++ b/docs/architecture/2023-09-29_request_enrollment_details_updates.md
@@ -1,5 +1,7 @@
 # Revisiting enrollment details
 
+This continues the work started in [Initial Host Details](2018-06-15_request_enrollment_details.md)
+
 ## Authors
 
 - seph ([@directionless](https://github.com/directionless))
--- a/pkg/osquery/enrollment_details.go
+++ b/pkg/osquery/enrollment_details.go
@@ -65,13 +65,13 @@ func getEnrollDetails(osquerydPath string) (service.EnrollmentDetails, error) {
 		return details, fmt.Errorf("create osquery for enrollment details: %w", err)
 	}
 
-	osqCtx, _ := context.WithTimeout(context.TODO(), 5*time.Second)
+	osqCtx, osqCancel := context.WithTimeout(context.TODO(), 5*time.Second)
+	defer osqCancel()
 
 	if err := osq.Execute(osqCtx); osqCtx.Err() != nil {
 		return details, fmt.Errorf("query enrollment details context error: %w", osqCtx.Err())
 	} else if err != nil {
 		return details, fmt.Errorf("query enrollment details: %w", err)
-
 	}
 
 	var resp []map[string]string
--- a/pkg/osquery/enrollment_details.go
+++ b/pkg/osquery/enrollment_details.go
@@ -69,10 +69,10 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
 	osqCtx, osqCancel := context.WithTimeout(ctx, 5*time.Second)
 	defer osqCancel()
 
-	if err := osq.RunSql(osqCtx, []byte(query)); osqCtx.Err() != nil {
+	if sqlErr := osq.RunSql(osqCtx, []byte(query)); osqCtx.Err() != nil {
 		return details, fmt.Errorf("query enrollment details context error: %w", osqCtx.Err())
-	} else if err != nil {
-		return details, fmt.Errorf("query enrollment details: %w", err)
+	} else if sqlErr != nil {
+		return details, fmt.Errorf("query enrollment details: %w", sqlErr)
 	}
 
 	var resp []map[string]string
--- a/pkg/osquery/enrollment_details.go
+++ b/pkg/osquery/enrollment_details.go
@@ -60,7 +60,6 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
 
 	osq, err := runsimple.NewOsqueryProcess(
 		osquerydPath,
-		runsimple.RunSql([]byte(query)),
 		runsimple.WithStdout(&respBytes),
 	)
 	if err != nil {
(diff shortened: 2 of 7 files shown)


pkg/osquery/enrollment_details.go
```diff
@@ -0,0 +1,140 @@
...
+	// To facilitate manual testing around missing enrollment details,
+	// there is a environmental variable to trigger the failure condition
+	if os.Getenv("LAUNCHER_DEBUG_ENROLL_DETAILS_ERROR") == "true" {
+		return details, errors.New("Skipping enrollment details")
+	}
+
+	// If the binary doesn't exist, bail out early.
+	if info, err := os.Stat(osquerydPath); os.IsNotExist(err) {
+		return details, fmt.Errorf("no binary at %s", osquerydPath)
+	} else if info.IsDir() {
```
RebeccaMahany, line 33: Do we want to also return early in case of other types of errors?

pkg/osquery/enrollment_details.go
```diff
@@ -0,0 +1,140 @@
...
+	osq, err := runsimple.NewOsqueryProcess(
+		osquerydPath,
+		runsimple.RunSql([]byte(query)),
+		runsimple.WithStdout(&respBytes),
+	)
+	if err != nil {
+		return details, fmt.Errorf("create osquery for enrollment details: %w", err)
+	}
+
+	osqCtx, osqCancel := context.WithTimeout(context.TODO(), 5*time.Second)
```
RebeccaMahany, line 68: This function is called by `Enroll` which has access to a preexisting `ctx` -- do we want to pass that in to this function and use it here instead of `context.TODO()`?

I saw later in the simple runner you left a comment that osquery should have its own context -- why?

pkg/osquery/extension.go
```diff
@@ -403,19 +400,26 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+	if e.knapsack.OsquerydPath() == "" {
+		level.Info(logger).Log("msg", "Cannot get additional enrollment details without an osqueryd path. This is probably CI")
+	} else {
+		if err := backoff.WaitFor(func() error {
+			enrollDetails, err = getEnrollDetails(e.knapsack.OsquerydPath())
+			if err != nil {
+				level.Debug(logger).Log("msg", "getEnrollDetails failed in backoff", "err", err)
+			}
+			return err
+		}, 5*time.Second, 5*time.Second); err != nil {
```
RebeccaMahany, line 415: This will only run once -- we should make the timeout larger than the interval. Maybe --

```suggestion
		}, 30*time.Second, 5*time.Second); err != nil {
```

pkg/osquery/runsimple/osqueryrunner.go
```diff
@@ -0,0 +1,137 @@
...
+	return p, nil
+}
+
+func (p *osqueryProcess) Execute(ctx context.Context) error {
+	// TODO: Not totally sure on ctx here. I think that osquery should probably have it's own context, but also
+	// that it's a better method signature.
+
+	// cmd.Start does not block
+	// Need to call cmd.Wait after it
+	// So how do we manage the start, health, wait in the rest of the control flow?
```
RebeccaMahany, line 91: Calling cmd.Run instead of cmd.Start -- do you still need this comment?

pkg/osquery/runsimple/osqueryrunner.go
```diff
@@ -0,0 +1,137 @@
...
+	if p.sql != nil {
+		args = append(args, []string{
+			"-S",
+			"--config_path", "/dev/null",
+			"--disable_events",
+			"--disable_database",
+			"--disable_audit",
+			"--ephemeral",
+			"--json",
+		}...)
```
RebeccaMahany, line 104: Curious -- why do we only add these args in this case?

RebeccaMahany:
- [ ] Able to enroll on Ubuntu 22.04 successfully with this PR
- [ ] Able to enroll on Windows 11 successfully with this PR

pkg/osquery/enrollment_details.go
directionless: Two answers. 

Good point -- yes, this should use the context. I've updated it.

The later comment is about running osquery as a subprocess. If we wanted it to run longer, it might want it's own context so we can separate it and manage the shutdown more cleanly. Not totally sure. <handwave about rungroup semantics>

pkg/osquery/runsimple/osqueryrunner.go
directionless: Well, maybe. Depends on whether we're trying to replace the larger osquery runtime with this.

pkg/osquery/runsimple/osqueryrunner.go
directionless: The idea, which might be imperfect, is that if SQL is provided, we're running in a one-shot mode. And these are flags that make sense for that.

pkg/osquery/runsimple/osqueryrunner.go
```diff
@@ -0,0 +1,114 @@
...
+		opt(p)
+	}
+
+	return p, nil
+}
+
+// RunSql will run a given blob by passing it in as stdin. Note that osquery is perticular. It needs the
+// trailing semicolon, but it's real weird about line breaks, an may return as multiline json output. It
+// is the responsibility of the caller to get the details right.
+func (p osqueryProcess) RunSql(ctx context.Context, sql []byte) error {
```
directionless, line 75: This should probably _not_ be `(p osqueryProcess)`

pkg/osquery/enrollment_details.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	osqCtx, osqCancel := context.WithTimeout(ctx, 5*time.Second)
+	defer osqCancel()
+
+	if err := osq.RunSql(osqCtx, []byte(query)); osqCtx.Err() != nil {
+		return details, fmt.Errorf("query enrollment details context error: %w", osqCtx.Err())
+	} else if err != nil {
+		return details, fmt.Errorf("query enrollment details: %w", err)
+	}
```
James-Pickett, line 76: I think the err being compared here will be the same err as on line 61 since this will be scoped to the if statement

```suggestion
	err = osq.RunSql(osqCtx, []byte(query))
	
	if osqCtx.Err() != nil {
		return details, fmt.Errorf("query enrollment details context error: %w", osqCtx.Err())
	}
	if err != nil {
		return details, fmt.Errorf("query enrollment details: %w", err)
	}
```

directionless: Testing, it looks like I had worked. (I didn't actually notice this, this is an excellent catch). I renamed the error to make the scoping more clear.

docs/architecture/2023-09-29_request_enrollment_details_updates.md
```diff
@@ -0,0 +1,48 @@
...
+osquery.
+
+## Consequences
+
+We incur an exec call during startup. But in return, we can gain
+several benefits: 
+- Decouple enrollment details from the main osquery startup.
+- Reduce contention on the socket during early startup
+- Enrollment no longer has a circular dependency
+- Enables future work to completely pull enrollment into launcher
```
RebeccaMahany, line 48: Thanks for adding the docs, they're great

pkg/osquery/runsimple/osqueryrunner.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// It's okay for these to be nil, so we can just set them without checking.
+	cmd.Stdin = p.stdin
+	cmd.Stdout = p.stdout
+	cmd.Stderr = p.stderr
+
+	return cmd.Run()
+}
+
+func (p osqueryProcess) RunVersion(ctx context.Context) error {
```
RebeccaMahany, line 101: Do we want to update launcher main to use this? (https://github.com/kolide/launcher/blob/main/cmd/launcher/launcher.go#L480)

RebeccaMahany (approved):
LGTM

pkg/osquery/runsimple/osqueryrunner.go
RebeccaMahany: Fine to do this in a followup PR at another time, feel free to resolve

James-Pickett (commented):
Nice! Looks good except for some questions about tests.

pkg/osquery/runsimple/osqueryrunner_test.go
```diff
@@ -0,0 +1,108 @@
...
+		t.Skip("No osquery. Not implemented")
+	}
+
+	osq, err := NewOsqueryProcess(osquerydPath)
+	require.NoError(t, err)
+
+	require.NoError(t, osq.RunSql(context.TODO(), []byte("select 1")))
+}
+
+func Test_OsquerySingleSqlErr(t *testing.T) {
```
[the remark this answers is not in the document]
directionless: Good catch. I hadn't updated these since several refactors, and because they're skipped in ci 🤷 

I updated them. and added a download-osquery

pkg/osquery/runsimple/osqueryrunner_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if tt.jsonMulti {
+				// This is discouraged, and hard to test. So let's bail for now
+				return
+			}
+
+			var result []map[string]string
+			require.NoError(t, json.Unmarshal(stdout.Bytes(), &result))
+
+			fmt.Println(stderr.String())
+			spew.Dump(result)
```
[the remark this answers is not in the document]
directionless: Eh. This was an artifact from the printf period. It's gone now

RebeccaMahany (approved):

PR #1215, opened 2023-05-31
OpenTelemetry for launcher

RebeccaMahany:
Adds traces to a couple locations in launcher, as well as an optional exporter using OTLP over HTTP. These changes work out-of-the-box with the Jaeger all-in-one container documented [here](https://www.jaegertracing.io/docs/1.45/getting-started/). Adds a command-line flag to disable exporting traces.


pkg/osquery/tables/tablehelpers/exec.go
```diff
@@ -21,6 +26,11 @@ import (
 //
 // This is not suitable for high performance work -- it allocates new buffers each time.
 func Exec(ctx context.Context, logger log.Logger, timeoutSeconds int, possibleBins []string, args []string, includeStderr bool) ([]byte, error) {
+	ctx, span := traces.New(ctx, trace.WithAttributes(
+		attribute.StringSlice(traces.AttributeName("tablehelpers", "possible_binaries"), possibleBins),
+		attribute.StringSlice(traces.AttributeName("tablehelpers", "args"), args)))
+	defer span.End()
```
directionless, line 32: This `traces.New(ctx, trace.WithAttributes(...)` feels awkward. 

We have this `traces` wrapper, which seems good. But then we also need to bring in otel for a bunch more, so that feels weird. I kinda think we should find ways to push more through our wrapper. 

I think we should consider the common usage, and then make that easy with the wrapper. Thinking aloud...
* do spans have names? if so `trace.New(ctx, name, ...)`.
* Are most of our attributes going to be `traces.AttributeName`? If so, we can take `keyvals` and process them out. 
* We probably sometimes need an escape value, so maybe a `New(ctx context.Context, keyvals ...interface{})` and then `NewRawOtel(ctx context.Context, opts ...trace.SpanStartOption)`

pkg/traces/traces.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	ApplicationName = "launcher"
+	defaultSpanName = "launcher/unknown"
+)
+
+// New returns a new context and span, including information about the calling function
+// as appropriate. Standardizes the tracer name. The caller is always responsible for
+// ending the span.
+func New(ctx context.Context, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
```
directionless, line 22: Consider `Start` or `StartSpan`. It's not a new trace.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -47,18 +50,27 @@ func newKryptoEcMiddleware(logger log.Logger, localDbSigner, hardwareSigner cryp
 
 func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, span := traces.New(r.Context())
+		defer span.End()
```
directionless, line 54: This is an interesting question... I don't know if we should add spans to `kryptoEcMiddleware.Wrap` vs wrapping with a span middleware. Thoughts?

ee/localserver/krypto-ec-middleware.go
```diff
@@ -68,13 +80,19 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 		// reusing it a bunch. However, it will fail if the clocks are too far out of sync.
 		timestampDelta := time.Now().Unix() - challengeBox.Timestamp()
 		if timestampDelta > timestampValidityRange || timestampDelta < -timestampValidityRange {
+			span.SetAttributes(attribute.Int64("timestamp_delta", timestampDelta))
+			span.SetStatus(codes.Error, "timestamp is out of range")
+
 			level.Debug(e.logger).Log("msg", "timestamp is out of range", "delta", timestampDelta)
```
directionless, line 86: One thing I want to call out here, is that I think the _eventual_ convergence here is that we merge logging and spans. And I have no idea what that looks like

ee/localserver/request-controlservice.go
```diff
@@ -5,37 +5,54 @@ import (
...
 	return http.HandlerFunc(ls.requestAccelerateControlFunc)
 }
 
 func (ls *localServer) requestAccelerateControlFunc(w http.ResponseWriter, r *http.Request) {
+	_, span := traces.New(r.Context(), trace.WithAttributes(attribute.String(traces.AttributeName("localserver", "path"), r.URL.Path)))
+	defer span.End()
+
 	if r.Body == nil {
+		span.SetStatus(codes.Error, "request body is nil")
 		sendClientError(w, "request body is nil")
```
directionless, line 25: Why not pass the span to `sendClientError`?

(Or make a wrapper, that pulls errors out of request context)

ee/localserver/request-id.go
```diff
@@ -68,6 +72,9 @@ func (ls *localServer) requestIdHandler() http.Handler {
 }
 
 func (ls *localServer) requestIdHandlerFunc(res http.ResponseWriter, req *http.Request) {
+	_, span := traces.New(req.Context(), trace.WithAttributes(attribute.String(traces.AttributeName("localserver", "path"), req.URL.Path)))
```
directionless, line 75: i wonder if the ctx should replace the req one

ee/localserver/server.go
```diff
@@ -129,7 +130,9 @@ func New(k types.Knapsack, opts ...LocalServerOption) (*localServer, error) {
 	// mux.Handle("/acceleratecontrol", ls.requestAccelerateControlHandler())
 
 	srv := &http.Server{
-		Handler:           ls.requestLoggingHandler(ls.preflightCorsHandler(ls.rateLimitHandler(mux))),
+		Handler: otelhttp.NewHandler(ls.requestLoggingHandler(ls.preflightCorsHandler(ls.rateLimitHandler(mux))), "localserver", otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
```
directionless, line 133: I like this.

pkg/osquery/tables/dataflattentable/exec_and_parse.go
```diff
@@ -77,6 +82,9 @@ func NewExecAndParseTable(logger log.Logger, tableName string, p parser, execCmd
 }
 
 func (t *execTableV2) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	ctx, span := traces.New(ctx, trace.WithAttributes(attribute.String(traces.AttributeName("dataflattentable", "table_name"), t.tableName)))
```
directionless, line 85: I _want_ to add spans to the generate, but I think we should be doing it at a different level.

pkg/traces/exporter/exporter.go
```diff
@@ -0,0 +1,138 @@
...
+	go traceExporter.configureProvider(exp, serverProvidedDataStore, client)
+
+	return traceExporter, nil
+}
+
+// configureProvider sets up the trace provider attributes describing this application and installation,
+// uniquely identifying the source of these traces.
+func (t *TraceExporter) configureProvider(exp *otlptrace.Exporter, serverProvidedDataStore types.Getter, client osquery.Querier) {
+	// Set all the attributes that we know we can get first
+	attrs := []attribute.KeyValue{
```
directionless, line 55: Is `attrs` something that we can change in runtime? Because I think it's going to be hard to populate some of this early in startup. `client osquery.Querier`, for example, won't be online as early as we want this.

But if we can adjust them after we're running, we get nice behavior. (And CJ has a PR with a callback mechanism somewhere)

directionless (commented):
I love it, but I want to noodle on the function signatures

pkg/osquery/tables/tablehelpers/exec.go
RebeccaMahany: I think more of a wrapper makes sense! I was waffling and not sure if it was optimizing too early.

Re: span name, I want to set the span name inside `traces.New`, rather than taking it as an arg, so that we can make sure we're setting span names consistently/avoid accidentally copy-pasting a different span's name.

Right now the only span options I'm using are `WithAttributes`, yeah -- it's possible we may want some of the others in the future.

ee/localserver/krypto-ec-middleware.go
RebeccaMahany: I thought it was helpful this way because it gives us more specific timing info for the `kryptoEcMiddleware`, which would've been helpful in the past. We do already effectively wrap the entire localhost handler with the span middleware -- we use `otelhttp.NewHandler` to wrap all our other handlers. So we have an outer span for the initial request, a child span for the kryptoEcMiddleware, and then a final child span for the request handling (e.g. control server acceleration or whatever we're doing).

ee/localserver/request-id.go
RebeccaMahany: I don't think it matters

ee/localserver/krypto-ec-middleware.go
directionless: That's a good point -- maybe we don't _usually_ want to instrument a span, but this one might be special.

pkg/traces/exporter/exporter.go
RebeccaMahany: I can take another look, but I didn't see support for updating the provider after it's already been set. The way I have this working now is that the exporter just waits till the osquery client is ready, so the first ~30sec-1min of traces wouldn't be reported -- which would be a bummer for e.g. trying to figure out osquery extension startup behavior via traces.

I think we could replace the provider, so we could do something like initially writing traces to a file until the osquery client is ready, and then starting to export traces to an external collector once we can get the identifiable information we need.

I'll research more.

pkg/osquery/tables/dataflattentable/exec_and_parse.go
RebeccaMahany: Where do you have in mind?

pkg/osquery/tables/tablehelpers/exec.go
directionless: pushing name to the wrapper makes sense. 

Not sure which way is over/under optimizing. I'm looking at the diff, and seeing a lot of `otel` scattered about, and it just feels wrong.

ee/localserver/request-id.go
RebeccaMahany: This is the innermost handler -- there shouldn't be any more child spans

pkg/traces/exporter/exporter.go
RebeccaMahany: So, unfortunately, it looks like the provider's resource (i.e. all these identifying attributes about the device) is expected to be immutable once created. I did get it working by creating an initial provider with the info we had and replacing it with a new provider with more attributes once we could access the osquery client -- but then it would refuse to ship traces if I called `Shutdown` on the old provider, which we are required to do to free up resources. Trying something different next.

RebeccaMahany: Never mind! This approach actually worked great once I made sure to use a new exporter for the new provider, since the old exporter was getting shut down.

pkg/osquery/tables/tablehelpers/exec.go
RebeccaMahany: I've updated -- let me know how this looks to you.

ee/localserver/krypto-ec-middleware.go
directionless: There were a couple aspects to the original comment, but I'm not sure they're germane. 

Foremost, my point is about how if we want to instrument the entire span, we'd be better off doing it in middleware. Though you're correct that `otelhttp.NewHandler` is already that middleware.

The secondary point is that if we're instrumenting the span  with `otelhttp.NewHandler`, then immediately starting a subspan isn't going to be meaningful. It should be the same timing period. If there are smaller parts we want to trace it makes sense, but this looks like:
```
trace1{
  trace2{
    stuff
  }
}
```

So I don't see it adding much

ee/localserver/request-id.go
directionless: For now. :)

(probably fine. but I want to kinda note it as almost an antipattern)

pkg/osquery/tables/dataflattentable/exec_and_parse.go
directionless: Probably the osquery-go SDK, or even thrift.

ee/localserver/krypto-ec-middleware.go
directionless: Though I see that you need the `span` to add in the error. But I wonder if `SpanFromContext` is a better fit. 

Basically, why add the span? Does it help any?

ee/localserver/request-id.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/consoleuser"
 	"github.com/kolide/launcher/pkg/backoff"
+	"github.com/kolide/launcher/pkg/traces"
+	"go.opentelemetry.io/otel/codes"
```
directionless, line 18: I don't love needing codes here. But maybe it's a good enough

go.mod
```diff
@@ -59,6 +59,15 @@ require (
 	github.com/kolide/systray v1.10.4
 	github.com/kolide/toast v1.0.0
 	github.com/shirou/gopsutil/v3 v3.23.3
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.16.0
+)
+
+require (
+	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/felixge/httpsnoop v1.0.3 // indirect
```
directionless, line 67: 😦  🤷

pkg/agent/knapsack/knapsack.go
```diff
@@ -343,3 +343,10 @@ func (k *knapsack) SetUpdateDirectory(directory string) error {
 func (k *knapsack) UpdateDirectory() string {
 	return k.flags.UpdateDirectory()
 }
+
+func (k *knapsack) SetExportTraces(enabled bool) error {
+	return k.flags.SetExportTraces(enabled)
+}
+func (k *knapsack) ExportTraces() bool {
+	return k.flags.ExportTraces()
```
directionless, line 351: I suspect we're not going to keep this, but I appreciate the plumbing

pkg/traces/traces.go
```diff
@@ -0,0 +1,99 @@
...
+	"runtime"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const (
+	ApplicationName           = "launcher"
```
directionless, line 17: I wonder if we'll want to split launcher and desktop apart here. But maybe TBD?

pkg/traces/traces.go
```diff
@@ -0,0 +1,99 @@
...
+	return otel.Tracer(ApplicationName).Start(ctx, spanName, opts...)
+}
+
+// buildAttributes takes the given keyVals, expected to be pairs representing the key
+// and value of each attribute, and parses them appropriately, ensuring that the keys
+// have consistent and specific names. Pairs that cannot be parsed will be discarded.
+func buildAttributes(callerFile string, keyVals ...interface{}) []attribute.KeyValue {
+	callerDir := defaultAttributeNamespace
+	if callerFile != "" {
+		callerDir = filepath.Base(filepath.Dir(callerFile))
```
directionless, line 60: I don't know what this value is -- is the path separator going to vary by platform?

pkg/traces/traces.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	attrs := make([]attribute.KeyValue, 0)
+
+	for i := 0; i < len(keyVals); i += 2 {
+		// Keys must always be strings
+		if reflect.TypeOf(keyVals[i]).String() != "string" {
+			continue
+		}
```
directionless, line 69: Feels weird to mix reflect and the type switch below. Either seems okay, but both feels weird. Why not
```
if _, ok := keyVals[i].(string); !ok {
     attrs = append(attrs, <bad key type at ...>)
     continue
}

File: pkg/traces/traces.go [hunk 16]
```diff
@@ -0,0 +1,99 @@
...
+			attrs = append(attrs, attribute.Int64Slice(key, keyVals[i+1].([]int64)))
+		case float64:
+			attrs = append(attrs, attribute.Float64(key, keyVals[i+1].(float64)))
+		case []float64:
+			attrs = append(attrs, attribute.Float64Slice(key, keyVals[i+1].([]float64)))
+		case string:
+			attrs = append(attrs, attribute.String(key, keyVals[i+1].(string)))
+		case []string:
+			attrs = append(attrs, attribute.StringSlice(key, keyVals[i+1].([]string)))
+		}
```
directionless, line 95: Consider a default that sets an error

directionless (commented):
I'm not sure I can follow the exporter, but this seems about right. 

I think we can add some sugar around adding errors. `trace.AddErrorToCtx(ctx, err)` and trace.AddErrorToSpan(span, err)`maybe? But we could kick that forward.

pkg/traces/traces.go
RebeccaMahany: It's the closest I got to getting the package name -- e.g. when the caller is ee/localserver/request-controlservice.go, this will be `localserver`. I can document this better.

pkg/traces/traces.go
RebeccaMahany: Updated!

pkg/traces/traces.go
RebeccaMahany: Added, thank you

pkg/traces/traces.go
RebeccaMahany: Added a comment to make this clearer

ee/localserver/request-id.go
RebeccaMahany: Updated traces pkg to add a `SetError` function that handles this, so we don't need codes here anymore

ee/localserver/krypto-ec-middleware.go
RebeccaMahany: In general I like more child spans because they help us narrow down causes of performance issues. E.g. here, if we have a localserver request that we see take a long time, we can use the multiple child spans to identify if it's the krypto middleware that's taking a while, or the inner request to e.g. accelerate the control server interval. Without the child spans, all we have is one span that tells us it takes a certain amount of time for the request to be processed and a response to be returned, which isn't really more info than what we've got in the browser dev console.

pkg/traces/traces.go
RebeccaMahany: I'm actually not sure this is right altogether -- we use this for the name of the tracer, and the TracerProvider interface has this documentation:

```
// The passed name needs to uniquely identify instrumentation code.
// Therefore, it is recommended that name is the Go package name of the
// library providing instrumentation (note: not the code being
// instrumented).
```

So I think I should update to either `go.opentelemetry.io/otel/trace` or `github.com/kolide/launcher/pkg/traces`, and rename this const to something more appropriate.

To your point, yes we'll definitely want a way to distinguish between launcher proper and launcher desktop traces -- I haven't thought that through much yet.

RebeccaMahany: Yeah, I am going to update this -- this is used to set the `otel.library.name` tag.

ee/localserver/krypto-ec-middleware.go
directionless: Right. But only if there are those inner spans. So why wrap the whole function, vs just the blocks you want? There's an extra span here

go.mod
directionless: Like I said 🤷 

Gotta just live with the extra dependancies. I guess `backoff` is coming from some notary thing

pkg/traces/traces.go
directionless: On windows, are the paths `/` or `\`? (I'm never sure, since it's oddly inconsistent. windows paths are, of course `\`, but sometimes the internal go paths are `/`)

pkg/traces/traces.go
```diff
@@ -0,0 +1,115 @@
...
+				fmt.Sprintf("bad key type %T: %v", keyVals[i], keyVals[i]),
+				fmt.Sprintf("%v", keyVals[i+1]),
+			))
+			continue
+		}
+
+		key := fmt.Sprintf("launcher.%s.%s", callerDir, keyVals[i])
+
+		// Create an attribute of the appropriate type
+		switch keyVals[i+1].(type) {
```
directionless, line 88:
```suggestion
		switch v := keyVals[i+1].(type) {
```

You can set `v` here, and not need all the additional typecasts below

pkg/traces/traces.go
```diff
@@ -0,0 +1,115 @@
...
+	spanName := defaultSpanName
+
+	opts := make([]trace.SpanStartOption, 0)
+
+	// Extract information about the caller to set some standard attributes (code.filepath,
+	// code.lineno, code.function) and to set more specific span and attribute names.
+	// runtime.Caller(0) would return information about `StartSpan` -- calling
+	// runtime.Caller(1) will return information about the function calling `StartSpan`.
+	pc, callerFile, callerLine, ok := runtime.Caller(1)
+	if ok {
```
directionless, line 36:
```suggestion
	if pc, callerFile, callerLine, ok := runtime.Caller(1); ok {
```

directionless (approved):
I'd say fix the typecasts, and it's a solid base to iterate with

pkg/traces/traces.go
RebeccaMahany: Oh yup that's much tidier, thank you! Updated

pkg/traces/traces.go
RebeccaMahany: Oh sorry, I didn't realize I'd missed that part of the question -- let me confirm.

RebeccaMahany: Forward slash also -- so no, path separator will not vary by platform.

Screenshot 2023-06-09 122315

pkg/osquery/tables/dataflattentable/exec_and_parse.go
RebeccaMahany: I imagine we'll come back to that idea a little later?

pkg/agent/flags/keys/keys.go
```diff
@@ -9,6 +9,7 @@ type FlagKey string
 // 3. Implement the getter and setter in the Knapsack, which delegates the call to the FlagController
 // 4. Implement the getter and setter in the FlagController, providing defaults, limits, and overrides
 // 4. Implement tests for any new APIs, sanitizers, limits, overrides.
+// 5. Update mocks -- in pkg/agent/types, run `mockery --name Knapsack` and `mockery --name Flags`.
```
seejdev, line 12: Good call!

seejdev (dismissed):
Seems like a solid starting point! Great job!

pkg/traces/exporter/exporter.go
```diff
@@ -0,0 +1,191 @@
...
+	system_info.hostname
+FROM
+	os_version,
+	system_info,
+	osquery_info;
+`
+
+		resp, err := t.osqueryClient.Query(query)
+		if err != nil || len(resp) == 0 {
+			time.Sleep(30 * time.Second)
```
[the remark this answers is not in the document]
RebeccaMahany: Not especially -- I think usually the client is up within 30 sec to 2 mins, so it seemed like a good re-check interval. I can make this a `osqueryClientRecheckInterval` variable if that'd improve clarity?

pkg/traces/exporter/exporter.go
```diff
@@ -0,0 +1,191 @@
...
+// to our resource attributes. Since this is called on startup when the osquery client
+// may not be ready yet, we perform a few retries.
+func (t *TraceExporter) addAttributesFromOsquery() {
+	t.attrLock.Lock()
+	defer t.attrLock.Unlock()
+
+	// The osqueryd client may not have initialized yet, so retry a few times on error.
+	osquerydRetries := 5
+	for i := 0; i < osquerydRetries; i += 1 {
+		query := `
```
James-Pickett, line 124: readability nit, might be easier to understand this code if we put the query as a const outside the for loop. Consider using backoff.watifor. Not that there is anything wrong with the loop, it's just kind of a common practice in our code base.

pkg/traces/traces.go
```diff
@@ -0,0 +1,115 @@
...
+func StartSpan(ctx context.Context, keyVals ...interface{}) (context.Context, trace.Span) {
+	spanName := defaultSpanName
+
+	opts := make([]trace.SpanStartOption, 0)
+
+	// Extract information about the caller to set some standard attributes (code.filepath,
+	// code.lineno, code.function) and to set more specific span and attribute names.
+	// runtime.Caller(0) would return information about `StartSpan` -- calling
+	// runtime.Caller(1) will return information about the function calling `StartSpan`.
+	pc, callerFile, callerLine, ok := runtime.Caller(1)
```
James-Pickett, line 35: what is pc here? could we use a more descriptive name?

James-Pickett (dismissed):
awesome work!

pkg/traces/exporter/exporter.go
RebeccaMahany: Updated, let me know how it looks to you!

James-Pickett (approved):

pkg/traces/traces.go
James-Pickett: thank you!

pkg/traces/traces.go
directionless: Then it'll need `path` not `filepath` and a comment.

pkg/traces/exporter/exporter.go
James-Pickett: look good!

ee/localserver/krypto-ec-middleware.go
directionless: I think we're talking past each other. But we can point at a screen with real data and iterate later

PR #1217, opened 2023-06-02
Replace deadcode linter with unused linter

RebeccaMahany:
Replaces deprecated `deadcode` linter with maintained `unused` linter and cleans up or ignores resulting errors. This will prevent us from having to see the following on every single lint job:

```
level=warning msg="[runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused."
```

### DIFF
--- a/cmd/launcher/doctor.go
+++ b/cmd/launcher/doctor.go
@@ -35,7 +35,6 @@ var (
 	// Command line colors
 	cyanText   *color.Color
 	headerText *color.Color
-	yellowText *color.Color
 	whiteText  *color.Color
 	greenText  *color.Color
 	redText    *color.Color
@@ -50,7 +49,6 @@ var (
 
 	// Indented output for checkup results
 	info func(a ...interface{})
-	warn func(a ...interface{})
 	fail func(a ...interface{})
 	pass func(a ...interface{})
 )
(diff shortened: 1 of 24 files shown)


pkg/agent/keys_darwin.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/kolide/launcher/pkg/agent/types"
 )
 
-// nolint: deadcode
+// nolint:unused
```
James-Pickett, line 15: this is used... does the linter not respect build flags?

James-Pickett (approved):

cmd/launcher/doctor.go
```diff
@@ -50,7 +49,6 @@ var (
 
 	// Indented output for checkup results
 	info func(a ...interface{})
-	warn func(a ...interface{})
```
directionless, line 53: Feels weird removing this stuff. It's probably not a problem, per se, but it makes it harder to make a base of things ready to use. :|

Would the linter be appeased if we put it in a test?

ee/localserver/server.go
```diff
@@ -336,12 +336,3 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
...
 }
-
-func (ls *localServer) kryptoBoxOutboundHandler(http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-	})
-}
-
-func (ls *localServer) verify(message []byte, sig []byte) error {
-	return krypto.RsaVerify(ls.serverKey, message, sig)
-}
```
directionless, line 347: I think is probably fine. This is probably the old v1 code. (It'll be obvious if DT breaks)

cmd/launcher/doctor.go
directionless: I could live with it though. It _is_ unused after all

RebeccaMahany: I'm happy to flag these as unused and keep them

seejdev (approved):

PR #1218, opened 2023-06-03
Remove old tables

directionless:
This removes several old, and likely broken, tables:
* `kolide_best_practices`
* `kolide_email_addresses`
* `kolide_vulnerabilities`

These are all unmaintained, and I know several are missing important functionality. Our current best practice is to build data oriented tables and to do this work inside SQL, I am opting to remove these instead of trying to fix them.

@terracatta I'm tagging you here. I don't think we've used these in the entirely life K2, but I wanted to give you a chance to object


terracatta:
> I don't think we've used these in the entirely life K2,

Correct. Approved, but obv some issues with the CI need to fixed first it seems.

terracatta (dismissed):

PR #1226, opened 2023-06-12
Auth for exporting traces

RebeccaMahany:
Adds a gRPC wrapper that adds a bearer auth header to all outgoing trace exports. Also allows the ingest server URL and the export traces flag to be set by the control server.

```mermaid
sequenceDiagram
    Participant Control server
    Box Launcher
        Participant Control service
        Participant Agent flags subsystem
        Participant Auth token subsystem
        Participant Traces exporter
    end
    Participant Agent collector

    loop Config generation
        Control server ->> Control server: Generate JWT with TTL of 24 hours
        Control service ->> Control server: Perform regular check to see if subsystem data has changed
        Control server ->> Control service: Return config, including JWT
    end

    opt Ingest URL update
        Control service ->> Agent flags subsystem: Perform update
        Agent flags subsystem ->> Traces exporter: Notify that URL has changed
        Traces exporter ->> Traces exporter: Init new exporter using new ingest URL
    end

    opt Token update
        Control service ->> Auth token subsystem: Perform update
        Auth token subsystem ->> Auth token subsystem: Store new token
        Control service ->> Traces exporter: Ping
        Traces exporter ->> Auth token subsystem: Fetch new token
        Traces exporter ->> Traces exporter: Replace token used for export
    end

    Traces exporter ->> Agent collector: Send traces with bearer auth header to stored ingest URL
    
    Agent collector ->> Agent collector: Validate JWT
    Agent collector ->> Agent collector: Process and store traces
```


cmd/launcher/options.go
```diff
@@ -83,6 +83,7 @@ func parseOptions(subcommandName string, args []string) (*launcher.Options, erro
 		flCompactDbMaxTx         = flagset.Int64("compactdb-max-tx", 65536, "Maximum transaction size used when compacting the internal DB")
 		flConfigFilePath         = flagset.String("config", defaultConfigFilePath, "config file to parse options from (optional)")
 		flExportTraces           = flagset.Bool("export_traces", false, "Whether to export traces")
+		flIngestServerURL        = flagset.String("ingest_url", "localhost:4317", "Where to export traces and logs")
```
directionless, line 86: Naive question -- with this default, does it default to sending logs there? It feels like we should default to `""`

pkg/traces/exporter/exporter.go
```diff
@@ -160,7 +200,13 @@ func (t *TraceExporter) addAttributesFromOsquery() {
 
 // setNewGlobalProvider creates and sets a new global provider with the currently-available
 // attributes. If a provider was previously set, it will be shut down.
-func (t *TraceExporter) setNewGlobalProvider(exp sdktrace.SpanExporter) {
+func (t *TraceExporter) setNewGlobalProvider() {
+	exp, err := newExporter(context.Background(), t.ingestAuthToken, t.ingestUrl)
```
directionless, line 204: Something feels off here. See comments on the authenticator

pkg/traces/exporter/client_authenticator.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"fmt"
+)
+
+// Implements google.golang.org/grpc/credentials.PerRPCCredentials interface
+type clientAuthenticator struct {
+	token string
+}
```
directionless, line 11: This makes sense to me, but I think it would benefit from the addition of `SetNewToken(token string)`

That would allow the token to be updated, without create a new exporter.

directionless: I suspect it would be super clean to create one, and register it with `knapsack.RegisterChangeObserver`

Maybe pull that into `New(k knapsack)`? Not sure...

cmd/launcher/options.go
RebeccaMahany: It won't, just because the default for `export_traces` is false. But I can make this an empty string!

pkg/traces/exporter/exporter.go
```diff
@@ -53,37 +72,50 @@ func NewTraceExporter(ctx context.Context, serverProvidedDataStore types.Getter,
...
 	return t, nil
 }
 
-// newExporter returns an exporter that will send traces with OTLP over HTTP.
-func newExporter(ctx context.Context) (sdktrace.SpanExporter, error) {
-	traceClient := otlptracegrpc.NewClient(otlptracegrpc.WithInsecure())
+// newExporter returns an exporter that will send traces with OTLP over gRPC.
+func newExporter(ctx context.Context, token string, url string) (sdktrace.SpanExporter, error) {
+	traceClient := otlptracegrpc.NewClient(
+		otlptracegrpc.WithInsecure(),
```
James-Pickett, line 115: do we want to default to `WithInsecure`? or have a flag?

James-Pickett (commented):
nice!

pkg/agent/flags/flag_controller.go
```diff
@@ -453,3 +453,12 @@ func (fc *FlagController) ExportTraces() bool {
 		WithDefaultBool(fc.cmdLineOpts.ExportTraces),
 	).get(fc.getControlServerValue(keys.ExportTraces))
 }
+
+func (fc *FlagController) SetIngestServerURL(url string) error {
```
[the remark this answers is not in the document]
RebeccaMahany: I like ObservabilityIngest, I can update!

pkg/traces/exporter/exporter.go
RebeccaMahany: We don't want to default to `WithInsecure` ultimately -- I will add a flag, just hadn't gotten to it yet 😅

pkg/traces/exporter/client_authenticator.go
RebeccaMahany: Yes, definitely cleaner! I'll update

cmd/launcher/launcher.go
```diff
@@ -288,6 +288,19 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+
+		// Set up our tracing instrumentation
+		if exp, err := exporter.NewTraceExporter(ctx, k, extension, logger); err != nil {
+			level.Debug(logger).Log(
+				"msg", "could not set up trace exporter",
+				"err", err,
+			)
+		} else {
+			runGroup.Add(exp.Execute, exp.Interrupt)
+			if err := controlService.RegisterConsumer(exporter.IngestSubsystem, exp); err != nil {
```
RebeccaMahany, line 300: TODO:
* New consumer for ingest subsystem
* Make exporter a subscriber of the ingest subsystem (lets @James-Pickett 's work in https://github.com/kolide/launcher/pull/1228 also subscribe to the ingest subsystem)
* Update diagram

pkg/traces/exporter/exporter.go
RebeccaMahany: This is added now!

pkg/traces/exporter/client_authenticator.go
RebeccaMahany: This is updated -- I ended up making `TraceExporter` subscribe to the `ingest` updates, and swap in the token when pinged.

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
```diff
@@ -0,0 +1,41 @@
...
+
+type IngestTokenConsumer struct {
+	store types.KVStore
+}
+
+type traceSubsystemConfig struct {
+	IngestToken string `json:"ingest_token"`
+}
+
+func NewIngestTokenConsumer(store types.KVStore) *IngestTokenConsumer {
```
James-Pickett, line 23: if the idea of this consumer is that it is just for bearer auth tokens, I think it should be a little more generally names. Like `tokenconsumer` or `authtokenconsumer`

```suggestion
func NewTokenConsumer(store types.KVStore) *tokenConsumer {
```

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/kolide/launcher/pkg/agent/types"
+)
+
+const IngestSubsystem = "ingest"
```
James-Pickett, line 11: The functionality of the code all looks pretty good ... now we quibble about naming =). I think this subsystem should just be called `tokens` or `auth_tokens`. Since it's not necessarily related to any type of ingest. We may use it these tokens later to fetch data from somewhere.

If this is only intended to be used for `ObservabiltyIngest` then I think it should be named so

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
```diff
@@ -0,0 +1,41 @@
...
+
+func NewIngestTokenConsumer(store types.KVStore) *IngestTokenConsumer {
+	return &IngestTokenConsumer{
+		store: store,
+	}
+}
+
+// Update satisfies control.consumer interface
+func (i *IngestTokenConsumer) Update(data io.Reader) error {
+	var updatedCfg traceSubsystemConfig
```
James-Pickett, line 31: I think we ought to just unmarshal to something like

`map[string]string`

and iterate over the kvps setting the new keys

cmd/launcher/launcher.go
```diff
@@ -288,6 +289,22 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 		if err := controlService.RegisterConsumer(notificationconsumer.NotificationSubsystem, notificationConsumer); err != nil {
 			return fmt.Errorf("failed to register notify consumer: %w", err)
 		}
+
+		// Set up our tracing instrumentation
+		ingestConsumer := tokenconsumer.NewIngestTokenConsumer(k.TokenStore())
```
James-Pickett, line 294: if all this does is store tokens, maybe call it something like `tokenConsumer`, see comments on tokenconsumer

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
RebeccaMahany: Also covering https://github.com/kolide/launcher/pull/1226#discussion_r1230159791 and https://github.com/kolide/launcher/pull/1226#discussion_r1230153466 --

I was waffling about naming for this whole package mostly because I'm unsure about how it would expand. Will we want it to handle other tokens used for other things (so, subsystem `auth_tokens`, consumer `TokenConsumer`), or will we want it to handle other ingest-related items (so, subsystem `ingest`, consumer `IngestConfig(?)Consumer`)?

All the other ingest-related control server items ended fitting up better into the agent flags subsystem, so I don't really have any plans to put other ingest-related items in this subsystem's update data. But I also don't know if we plan to ship other tokens, either. Do you have thoughts/plans one way or the other? If not, maybe the clearest option would be to go super-specific (subsystem `ingest_auth_tokens`, consumer `IngestAuthTokensConsumer`)?

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
James-Pickett: discussed here https://github.com/kolide/launcher/pull/1226#discussion_r1230219621

cmd/launcher/launcher.go
James-Pickett: discussed here https://github.com/kolide/launcher/pull/1226#discussion_r1230219621

ee/control/consumers/tokenconsumer/ingest_token_consumer.go
RebeccaMahany: Moved to KeyValue consumer, no need for this anymore!

James-Pickett (approved):
awesome sauce

James-Pickett (approved):

PR #1228, opened 2023-06-13
ship logs to http endpoint

James-Pickett:
* adds a new logger "LogShipper" that gets added via TeeLogger at launcher startup
* LogShipper ship logs on an interval using endpoint and token provided by control server or flags
* updates tracing code to use "TraceIngestURL" flag and LogShipper to use "LogIngestUrl"


pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+func WithLogger(logger log.Logger) httpBufferOption {
+	return func(hb *httpBuffer) {
+		hb.logger = logger
+	}
+}
+
+func New(endpoint string, opts ...httpBufferOption) *httpBuffer {
```
James-Pickett, line 42: instead of accepting a string endpoint, probably accept an http client or perhaps a custom interface that will handle auth and all that

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+	if hb.buf.Len() > hb.maxSize {
+		// just drop our logs on the floor and start over
+		hb.logger.Log("msg", "buffer is full, dropping logs", "size", hb.buf.Len())
+		hb.buf.Reset()
+	}
+
+	n, err = hb.buf.Write(p)
+
+	// only try to send if we're at send size and not currently sending
+	if hb.buf.Len() >= hb.sendSize && hb.sendMutex.TryLock() {
```
James-Pickett, line 69: This limits us to sending one http request at a time, I think this is probably what we want to make sure we don't give users sudden http traffic spikes?

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+		go func() {
+			defer hb.sendMutex.Unlock()
+			hb.send(sendBuf)
+		}()
+	}
+
+	return n, err
+}
+
+func (hb *httpBuffer) Flush() error {
```
James-Pickett, line 82: should we leave flush blocking? I would think so, if I call flush I want it to be flushed when it's done

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+	hb.bufMutex.Lock()
+	defer hb.bufMutex.Unlock()
+	return hb.send(&hb.buf)
+}
+
+func (hb *httpBuffer) send(r io.Reader) error {
+	resp, err := http.Post(hb.endpoint, "application/octet-stream", r)
```
James-Pickett, line 89: when switched to accepting a client instead of just a string endpoint, ensure client is configured with time outs

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+	hb.bufMutex.Lock()
+	defer hb.bufMutex.Unlock()
+
+	if hb.buf.Len() > hb.maxSize {
+		// just drop our logs on the floor and start over
+		hb.logger.Log("msg", "buffer is full, dropping logs", "size", hb.buf.Len())
+		hb.buf.Reset()
+	}
+
+	n, err = hb.buf.Write(p)
```
James-Pickett, line 66: err check?

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func WithLogger(logger log.Logger) httpBufferOption {
+	return func(hb *httpBuffer) {
+		hb.logger = logger
+	}
+}
+
+func New(endpoint string, opts ...httpBufferOption) *httpBuffer {
+	buffer := &httpBuffer{
```
James-Pickett, line 43: set default noop logger

James-Pickett (commented):
meh

pkg/log/httpbuffer/httpbuffer_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHttpBuffer(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name                                          string
+		maxSize, sendSize                             int
+		startingData, writeData, expectedDataReceived []byte
```
James-Pickett, line 20: making write data a `[][]bbyte` with multiple writes would make test more robust

directionless (commented):
This looks pretty clean, I think it'll be easy to play with.

I've been stalling on https://github.com/kolide/launcher/pull/1183 but I think it would be easy to integrate this with it

pkg/log/httpbuffer/httpbuffer.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(buffer)
+	}
+
+	return buffer
+}
+
+func (hb *httpBuffer) Write(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
James-Pickett: looks like the `log.NewJSONLogger` [does write immediately](https://github.com/go-kit/kit/blob/v0.9.0/log/json_logger.go#L34). In that case I would just create a new logger that wraps a JSONLogger using a buffer like this. It feels a little weird, make a convenience method that does it for us?

I'll play with it some and see how it feels.

pkg/log/httpbuffer/httpbuffer.go
James-Pickett: In that code it looks like we write to the db, then periodically flush and purge. Are you thinking we should also back the http sending logger with the db or is in memory okay? I guess the advantage of being db backed is that it would persist through crashes ... starting to talk myself into it.

pkg/log/httpbuffer/httpbuffer.go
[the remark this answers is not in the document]
directionless: log does, but when you get deep enough I don't think you have any guarantee from the json encoder. This stuff is all based around stream interfaces. https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/encoding/json/stream.go;l=201-238

This is what [`pkg/log/ringlogger/ringlogger.go`](https://github.com/kolide/launcher/pull/1183/files#diff-87710cac52bf05de1f0c1fc3592b8d14e9074ef4192ad095a28d0b714307f0da) does. It creates a buffer and a json logger, and it does some log and resets based on those.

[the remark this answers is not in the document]
James-Pickett: I don't really see an alternative

Looking at your example:
* log is called
* buffer is reset
* call the underlying jsonlogger.log <- here were still expecting the json logger to write immediately to the buffer?
* write contents of buffer to persistent ring

in this example, were still relying on json logger which calls json encoder to write to the buffer.

Either way, please take a look at the most recent push. I broke the "sendbuffer" to it's own package with some robust testing.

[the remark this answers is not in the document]
directionless: Hrm. Maybe. I'm pretty sure I saw issue when I tried doing what you had originally. Like, try passing several megabyte hunks through it

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+		// add to buffer to be sent
+		if _, err := sendBuffer.Write(data); err != nil {
+			return fmt.Errorf("writing data to buffer: %w", err)
+		}
+
+		bytesToDelete += len(data) + len(dataKey)
+		removeDataKeysToIndex++
+	}
+
+	if !sb.sendMutex.TryLock() {
```
directionless, line 224: Shouldn't this happen prior to the iteration and buffer building?

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+		sb.logger.Log("msg", "could not get lock on send mutex, will retry")
+		return nil
+	}
+
```
directionless, line 228: defer unlock?

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+	// have lock on write mutex and send mutex at this point
+	if err := sb.purge(removeDataKeysToIndex); err != nil {
+		sb.sendMutex.Unlock()
+		return fmt.Errorf("purging data from kv store: %w", err)
+	}
+
+	sb.size -= bytesToDelete
+
+	// send off async
+	go func() {
```
directionless, line 238: `sendAndPurge` is already in a goroutine. Why spawn another?

directionless (commented):
(early comments)

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+
+	if len(sb.dataKeys) == 0 {
+		return nil
+	}
+
+	removeDataKeysToIndex := 0
+	sendBuffer := &bytes.Buffer{}
+	bytesToDelete := 0
+
+	for _, dataKey := range sb.dataKeys {
```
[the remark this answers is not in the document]
James-Pickett: The foreach doesn't guarantee any kind of order, this was the driving factor behind tracking the data keys. If were not concerned about order we could drop it.

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+		isSending:       false,
+	}
+
+	for _, opt := range opts {
+		opt(sb)
+	}
+
+	sb.logger = log.With(sb.logger, "component", "sendbuffer")
+
+	// get all data keys and calculate starting size
```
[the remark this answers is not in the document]
James-Pickett: it would get purged on the first write attempt

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+	return func(sb *SendBuffer) {
+		sb.logger = logger
+	}
+}
+
+func WithKVStore(kvStore types.KVStore) option {
+	return func(sb *SendBuffer) {
+		sb.kvStore = kvStore
+	}
+}
```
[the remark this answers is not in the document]
James-Pickett: I suppose it's easy to add it in later if we decided that we want to back with on disk KVStore, I can drop it for simplicity sake

pkg/sendbuffer/sendbuffer.go
James-Pickett: I suppose it's easy to add it in later if we decided that we want to back with on disk KVStore, I can drop it for simplicity sake

James-Pickett: I did this so that the write mutex does not have to wait on the send

pkg/sendbuffer/sendbuffer.go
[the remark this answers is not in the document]
directionless: _ah_. That does feel harder. I think that means we'd either end up tracking the data keys (as you do here) or doing an iterate and sort each time. 

I think tracking the data keys brings in another mutex, and some amount of synchronization between them. But using foreach has extra communication. 

I wonder if trying to use kvstore ends up working against this. the ring stuff I was playing with in #1183 is probably a better fit, but I'm probably being overly eager in optimizing. Okay.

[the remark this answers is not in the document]
James-Pickett: If we don't care about care about persistance between restarts ... the yea I suppose there is no reason to use the KV store. I chose it would make it easy to add persistance later if we wanted, but for simplicity may we just keep a list? And if we ever want persistance we'll add it later, would not be too hard.

pkg/sendbuffer/sendbuffer.go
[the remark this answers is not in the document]
James-Pickett: dropped

pkg/sendbuffer/sendbuffer.go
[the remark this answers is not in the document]
James-Pickett: no longer relevant, not using kv store anymore

pkg/sendbuffer/sendbuffer.go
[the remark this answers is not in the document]
James-Pickett: dropped KV store

pkg/sendbuffer/sendbuffer.go
James-Pickett: done

pkg/sendbuffer/sendbuffer.go
James-Pickett: yep, rewrote this part

pkg/sendbuffer/sendbuffer.go
James-Pickett: busted out the flushing to another func with it's own write lock, so no more go routine, looks cleaner

pkg/sendbuffer/sendbuffer.go
```diff
@@ -0,0 +1,264 @@
...
+			"size_of_data", len(data),
+			"max_send_size", sb.maxSendSize,
+			"head", string(data)[0:minInt(len(data), 100)],
+		)
+		return len(data), nil
+	}
+
+	// if we are full, something has backed up
+	// purge everything
+	if len(data)+sb.size > sb.maxSize {
```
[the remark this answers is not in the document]
James-Pickett: so this is actually is when we've reached the maximum storage size, I've renamed the var to reflect this. So if we hit this limit it's because were not even attempting to send for some reason.

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,70 @@
...
+
+// StartShipping is a no-op -- the exporter is already running in the background. The TraceExporter
+// otherwise only responds to control server events.
+func (ls *LogShipper) StartShipping() error {
+	ls.sendBuffer.StartSending()
+
+	// nothing else to do, wait for launcher to exit
+	<-context.Background().Done()
+	return nil
+}
```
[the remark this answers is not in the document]
James-Pickett: done ... not sure if I got context pattern correct, wish rungroup.Execute accepted a context

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,77 @@
...
+	ls.sender.authtoken = string(token)
+	ls.sender.endpoint = logEndpoint(ls.knapsack)
+}
+
+// StartShipping is a no-op -- the exporter is already running in the background. The TraceExporter
+// otherwise only responds to control server events.
+func (ls *LogShipper) StartShipping() error {
+	ls.sendBuffer.Run(ls.ctx)
+	return nil
+}
```
directionless, line 64:
```suggestion
func (ls *LogShipper) Run(ctx) error {
	return ls.sendBuffer.Run(ctx)
}
```

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,77 @@
...
+type LogShipper struct {
+	sender     *authedHttpSender
+	sendBuffer *sendbuffer.SendBuffer
+	logger     *logger
+	knapsack   types.Knapsack
+	ctx        context.Context
+	cancel     context.CancelFunc
+}
+
+func New(ctx context.Context, k types.Knapsack) *LogShipper {
```
directionless, line 25: Why does new take a ctx?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,77 @@
...
+// StartShipping is a no-op -- the exporter is already running in the background. The TraceExporter
+// otherwise only responds to control server events.
+func (ls *LogShipper) StartShipping() error {
+	ls.sendBuffer.Run(ls.ctx)
+	return nil
+}
+
+func (ls *LogShipper) StopShipping(_ error) {
+	ls.cancel()
+}
```
directionless, line 68:
```suggestion
func (ls *LogShipper) Stop(_ error) {
	ls.cancel()
}
```

pkg/log/logshipper/logger.go
```diff
@@ -0,0 +1,43 @@
...
+	"io"
+
+	"github.com/go-kit/kit/log"
+)
+
+const (
+	truncatedFormatString = "%s[TRUNCATED]"
+)
+
+type logger struct {
```
directionless, line 14: I'm not sure this struct adds much. Consider integrating with logshipper

pkg/log/logshipper/logshipper.go
James-Pickett: due to the signature required for rungroup.Execute, wish we could pass ctx there ... I could just do an inline func for execute in the launcher start up, I'll try that and see how it feels

pkg/log/logshipper/logshipper.go
James-Pickett: done

pkg/log/logshipper/logshipper.go
James-Pickett: done

pkg/log/logshipper/logger.go
James-Pickett: done

pkg/log/logshipper/logshipper.go
James-Pickett: ehh.. not sure about this one. We could just create a new context with background as the parent, but would like to also stop if the parent context gets canceled 🤷

James-Pickett: what do you think?

directionless: I usually expect `New` to setup things that cannot be canceled, and `Run` to take a cancelable context.

Eg: new just creates a struct, nothing runtimey.

cmd/launcher/launcher.go
```diff
@@ -144,6 +146,18 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	flagController := flags.NewFlagController(logger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db)
 
+	// Need to set up the log shipper so that we can get the logger early
+	// and pass it to the various systems.
+	var logShipper *logshipper.LogShipper
+	if k.ControlServerURL() != "" {
+		logShipper, err = logshipper.New(k)
```
directionless, line 153: How does this work when log shipping is disabled?

directionless (commented):
I think I'm okay with it. But I have some questions about how this works if we merge it today. (it should be disabled unless the control server sets a target host, or we push a build with a default target host)

cmd/launcher/launcher.go
James-Pickett: updated so we'll only start shipping if we have a host.

directionless: So to be clear...

Can we merge this today and have no effect until we set this from the control server?

cmd/launcher/launcher.go
```diff
@@ -148,8 +148,9 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 
 	// Need to set up the log shipper so that we can get the logger early
 	// and pass it to the various systems.
+	isShippingLogs := k.ObservabilityIngestServerURL() != ""
```
directionless, line 151: Probably okay to start with it. Interesting question is whether it's possible to enable/disable/change the URL from the control server

cmd/launcher/launcher.go
James-Pickett: I missed this comment earlier, I went back and introduced a new "LogShippingEnabled" flag. So the log shipper can be set up and ready to go until we flip the flag. It will still "run" in the background, but just ignore calls to `Log()` until it's enabled. So we can merge it and it will have not effect until we set it from the control server.

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,124 @@
...
+		return "", err
+	}
+
+	return parsedUrl.String(), nil
+}
+
+func (ls *LogShipper) Log(keyvals ...interface{}) error {
+	if !ls.knapsack.LogShippingEnabled() {
+		return nil
+	}
```
directionless, line 104: I cannot remember -- how performant is this? Vs storing a boolean on logshipper and registering it?

James-Pickett: I think the values are cached, but just to be safe, I'll store the boolean on log shipper

cmd/launcher/launcher.go
James-Pickett: dropped the enabled flag, now just enabling based on log_ingest_url presence

RebeccaMahany: Sorry for resolving/unresolving this comment thread like 3 times, this was a little confusing to me. Maybe this section belongs inside the `if k.ControlServerURL() == "" { ... } else { ...` block later, where all the other consumers/subscribers are?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+
+	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval))
+	logger := log.NewJSONLogger(sendBuffer)
+
+	return &LogShipper{
+		sender:     sender,
+		sendBuffer: sendBuffer,
+		logger:     logger,
+		knapsack:   k,
```
RebeccaMahany, line 53: I think you'd want to set `isShippingEnabled: ls.knapsack.LogIngestServerURL() != ""` here now too, otherwise you'd have to wait for a subscriber update to start shipping logs.

pkg/agent/knapsack/knapsack.go
```diff
@@ -362,11 +362,11 @@ func (k *knapsack) TraceSamplingRate() float64 {
...
+func (k *knapsack) SetTraceIngestServerURL(url string) error {
+	return k.flags.SetTraceIngestServerURL(url)
 }
-func (k *knapsack) ObservabilityIngestServerURL() string {
-	return k.flags.ObservabilityIngestServerURL()
+func (k *knapsack) TraceIngestServerURL() string {
+	return k.flags.TraceIngestServerURL()
 }
 
 func (k *knapsack) SetDisableObservabilityIngestTLS(enabled bool) error {
```
RebeccaMahany, line 372: These should probably all get renamed to `SetDisableTraceIngestTLS` too?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/kolide/launcher/pkg/agent/types"
+	"github.com/kolide/launcher/pkg/sendbuffer"
+)
+
+// TODO: consolidate this var, but where?
```
RebeccaMahany, line 15: Agreed but I don't have a good idea either 😁 maybe it could fit somewhere in pkg/agent?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+	isShippingEnabled bool
+}
+
+func New(k types.Knapsack) (*LogShipper, error) {
+	token, _ := k.TokenStore().Get(observabilityIngestTokenKey)
+
+	logEndpoint, err := logEndpoint(k)
+	if err != nil {
+		return nil, err
+	}
```
RebeccaMahany, line 37: I do think we should check the endpoint for validity, but with how this is set up right now I'm not sure we should be returning an error here. If we start off with an invalid/malformed URL, the control server will never be able to update it to a valid one. Maybe there's another/later place we could parse and check the endpoint, or maybe we could log + ignore the error here?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+
+	sender := newAuthHttpSender(logEndpoint, string(token))
+
+	sendInterval := time.Minute * 1
+	if k.Debug() {
+		sendInterval = time.Second * 1
```
RebeccaMahany, line 43: Maybe worth making these `const debugSendInterval = time.Second * 1` and `const defaultSendInterval = time.Minute * 1`?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+		logger:     logger,
+		knapsack:   k,
+	}, nil
+}
+
+func (ls *LogShipper) Ping() {
+	token, _ := ls.knapsack.TokenStore().Get(observabilityIngestTokenKey)
+	ls.sender.authtoken = string(token)
+
+	endpoint, err := logEndpoint(ls.knapsack)
```
RebeccaMahany, line 61: Wouldn't you need to implement `FlagsChanged(flagKeys ...keys.FlagKey)` and call `knapsack.RegisterChangeObserver` to get agent flags (i.e. `LogIngestServerURL`) updates on time?

pkg/log/logshipper/logshipper.go
```diff
@@ -0,0 +1,127 @@
...
+func (ls *LogShipper) Stop(_ error) {
+	ls.stopFunc()
+}
+
+func logEndpoint(k types.Knapsack) (string, error) {
+	endpoint := k.LogIngestServerURL()
+
+	if !strings.HasPrefix(endpoint, "http") {
+		scheme := "https"
+		if k.DisableObservabilityIngestTLS() {
```
RebeccaMahany, line 90: Hm I'm not sure we should be reusing this flag? If we don't think the log vs trace ingest URL will be the same in the future, we are going to want to be able to disable TLS for them separately.

cmd/launcher/launcher.go
RebeccaMahany: Ah sorry, I missed the comment explaining why it has to happen here -- I think I just got confused toggling between this comment thread/references to older versions of this code/logshipper.go, and I don't think this is actually confusing or requires comment updates or anything.

pkg/log/logshipper/logshipper.go
directionless: I am unsure we should have this flag at all. It seems better baked into the scheme on the endpoint. Though maybe we need one for whether or not to validate the certificates.

pkg/log/logshipper/logshipper.go
James-Pickett: nice catch!

pkg/agent/knapsack/knapsack.go
James-Pickett: done

pkg/log/logshipper/logshipper.go
James-Pickett: done

pkg/log/logshipper/logshipper.go
James-Pickett: I kinda took the lazy rout and just subscribed to all agent_flags updates, felt easier to just handle all the control server updates in Ping(), however if I'm willing to change it if you think it would be cleaner.

https://github.com/James-Pickett/launcher/blob/james/log-shipping/cmd/launcher/launcher.go#L324

pkg/log/logshipper/logshipper.go
James-Pickett: dropped usage of flag and just rely on endpoint given by server

pkg/log/logshipper/logshipper.go
James-Pickett: put it in pkg/agent/storage/keys.go

pkg/log/logshipper/logshipper.go
James-Pickett: doing this now, log and ignore, led to some consolidation of is enabled logic

directionless (approved):
Let's try!

pkg/log/logshipper/logshipper.go
directionless: I think you're right that we should come up with some patterns for doing this consistently. But I don't know what it'll be yet.

PR #1229, opened 2023-06-20
Add flag for trace sampling rate

RebeccaMahany:
Adds a new flag value type for float64s to support new flag `trace_sampling_rate`, which controls the fraction of traces that are sampled and exported.

Also starts backfilling some tests.


cmd/launcher/options.go
```diff
@@ -83,6 +83,7 @@ func parseOptions(subcommandName string, args []string) (*launcher.Options, erro
 		flCompactDbMaxTx         = flagset.Int64("compactdb-max-tx", 65536, "Maximum transaction size used when compacting the internal DB")
 		flConfigFilePath         = flagset.String("config", defaultConfigFilePath, "config file to parse options from (optional)")
 		flExportTraces           = flagset.Bool("export_traces", false, "Whether to export traces")
+		flTraceSamplingRate      = flagset.Float64("trace_sampling_rate", 0.0, "What fraction of traces should be sampled")
```
directionless, line 86: My sense is that doing this as an integer is going to be less code, and that there's very little gain from a float. OTOH, you wrote the code, so 🤷

directionless (approved):
Generally feels fine. Not sure we really need the float

cmd/launcher/options.go
RebeccaMahany: Unless I missed it, I don't think we support int flags yet in the `flags` package either, so it'd be the same amount of work: https://github.com/kolide/launcher/tree/main/pkg/agent/flags

directionless: Ah, indeed.

directionless (approved):

cmd/launcher/options.go
directionless: I just like ints more than floats. 😆

PR #1230, opened 2023-06-20
Backfill tests for tracing packages

RebeccaMahany:
Adds tests for the exporter package; makes some small updates and bugfixes.

### DIFF
--- a/pkg/traces/exporter/client_authenticator.go
+++ b/pkg/traces/exporter/client_authenticator.go
@@ -7,12 +7,14 @@ import (
 
 // Implements google.golang.org/grpc/credentials.PerRPCCredentials interface
 type clientAuthenticator struct {
-	token string
+	token      string
+	disableTLS bool
 }
 
-func newClientAuthenticator(token string) *clientAuthenticator {
+func newClientAuthenticator(token string, disableTLS bool) *clientAuthenticator {
 	return &clientAuthenticator{
-		token: token,
+		token:      token,
+		disableTLS: disableTLS,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/traces/exporter/client_authenticator.go
```diff
@@ -22,6 +24,12 @@ func (c *clientAuthenticator) setToken(token string) {
 	c.token = token
 }
 
+// setDisableTLS updates the return value of RequireTransportSecurity. TLS should
+// be disabled only for local testing.
+func (c *clientAuthenticator) setDisableTLS(disableTLS bool) {
+	c.disableTLS = disableTLS
+}
```
directionless, line 31: If we only need this for tests, we can define this in the `_test.go` file.

pkg/traces/exporter/client_authenticator.go
```diff
@@ -7,12 +7,14 @@ import (
 
 // Implements google.golang.org/grpc/credentials.PerRPCCredentials interface
 type clientAuthenticator struct {
-	token string
+	token      string
+	disableTLS bool
 }
 
-func newClientAuthenticator(token string) *clientAuthenticator {
+func newClientAuthenticator(token string, disableTLS bool) *clientAuthenticator {
```
directionless, line 14: With `setDisableTLS` do you need to change the signature here? It feels weird to change it for something only in CI

pkg/traces/exporter/client_authenticator.go
RebeccaMahany: I actually don't want it only for tests! `RequireTransportSecurity` is part of the `google.golang.org/grpc/credentials.PerRPCCredentials` interface and I'd been just hardcoding the return to false for initial testing. But now that we've got a proper flag for `disable_ingest_tls` I wanted to plumb that through here as well.

RebeccaMahany: (By "local testing" in the comment I mean that we would only set `disable_ingest_tls` to false when testing against e.g. a local collector running in docker.)

directionless: okay

directionless (approved):

PR #1231, opened 2023-06-21
Log more information about unknown process locking osquery pidfile

RebeccaMahany:
This PR is an update to https://github.com/kolide/launcher/pull/1130, attempting to collect more information about https://github.com/osquery/osquery/issues/7796.

Adds a log any time we start up osqueryd to note the PID, so we can have a historical record in the logs of previous launcher-owned osqueryd processes.

```
{"caller":"runner.go:320","msg":"launched osquery process","osqueryd_pid":76455,"severity":"info","ts":"2023-06-21T19:11:04.744815Z"}
```

Runs `ps` and `lsof` to attempt to discover more information about the process and what's holding the osquery.pid file.

```
{"caller":"log.go:178","msg":"ran ps on non-osqueryd process using pidfile","output":"USER   PID  PPID  PGID STAT      TIME COMMAND\nroot 76761 76758 76758 S+     0:00.00 ./a.out\n","pid":76761,"severity":"debug","ts":"2023-06-21T19:14:14.771631Z"}
```

```
{"caller":"log.go:202","msg":"ran lsof on non-osqueryd process using pidfile","output":"COMMAND   PID  PPID USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME\na.out   76761 76758 root  cwd    DIR   1,18     2176   256095 /Users/rebeccamahany-horton\na.out   76761 76758 root  txt    REG   1,18    52018 11084464 /Users/rebeccamahany-horton/a.out\na.out   76761 76758 root    0u   CHR   16,1   0t2358     1611 /dev/ttys001\na.out   76761 76758 root    1u   CHR   16,1   0t2358     1611 /dev/ttys001\na.out   76761 76758 root    2u   CHR   16,1   0t2358     1611 /dev/ttys001\na.out   76761 76758 root    3u   REG   1,18        5 13099727 /private/var/kolide-k2/k2device-preprod.kolide.com/osquery.pid\n","pid":76761,"severity":"debug","ts":"2023-06-21T19:14:14.816044Z"}
```

```
{"caller":"log.go:226","msg":"ran lsof on osqueryd pidfile","output":"COMMAND   PID  PPID USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME\na.out   76761 76758 root    3u   REG   1,18        5 13099727 /private/var/kolide-k2/k2device-preprod.kolide.com/osquery.pid\n","pidfile":"/private/var/kolide-k2/k2device-preprod.kolide.com/osquery.pid","severity":"debug","ts":"2023-06-21T19:14:15.002012Z"}
```

Test notes are same as in previous PR.

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -215,7 +215,7 @@ func (l *OsqueryLogAdapter) runAndLogLsofOnPidfile() {
 		return
 	}
 
-	fullPidfile := filepath.Join("/private", l.rootDirectory, "osquery.pid")
+	fullPidfile := filepath.Join(l.rootDirectory, "osquery.pid")
 	cmd := exec.Command("lsof", "-R", fullPidfile)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -85,7 +85,7 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
 	// information here about the process locking the pidfile.
 	// See: https://github.com/osquery/osquery/issues/7796
 	if bytes.Contains(p, []byte("Refusing to kill non-osqueryd process")) {
-		l.logInfoAboutUnrecognizedProcessLockingPidfile(p)
+		go l.logInfoAboutUnrecognizedProcessLockingPidfile(p)
 	}
 
 	msg := strings.TrimSpace(string(p))
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -4,7 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
+	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -17,9 +20,10 @@ import (
 // OsqueryLogAdapater creates an io.Writer implementation useful for attaching
 // to the osquery stdout/stderr
 type OsqueryLogAdapter struct {
-	logger       kitlog.Logger
-	levelFunc    func(kitlog.Logger) kitlog.Logger
-	extraKeyVals []interface{} // log.With expects an interface, not string
+	logger        kitlog.Logger
+	levelFunc     func(kitlog.Logger) kitlog.Logger
+	extraKeyVals  []interface{} // log.With expects an interface, not string
+	rootDirectory string
 }
 
 type Option func(*OsqueryLogAdapter)
@@ -44,11 +48,12 @@ func extractOsqueryCaller(msg string) string {
 	return strings.TrimSuffix(callerRegexp.FindString(msg), "]")
 }
 
-func NewOsqueryLogAdapter(logger kitlog.Logger, opts ...Option) *OsqueryLogAdapter {
+func NewOsqueryLogAdapter(logger kitlog.Logger, rootDirectory string, opts ...Option) *OsqueryLogAdapter {
 	l := &OsqueryLogAdapter{
-		logger:       logger,
-		levelFunc:    level.Debug,
-		extraKeyVals: []interface{}{},
+		logger:        logger,
+		levelFunc:     level.Debug,
+		extraKeyVals:  []interface{}{},
+		rootDirectory: rootDirectory,
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


pkg/log/log.go
```diff
@@ -149,6 +158,82 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
...
 }
 
+// runAndLogPs runs ps filtering on the given PID, and logs the output.
+func (l *OsqueryLogAdapter) runAndLogPs(pid int) {
+	if runtime.GOOS == "windows" {
+		return
+	}
+
+	pidStr := strconv.Itoa(pid)
+	cmd := exec.Command("ps", "-p", pidStr, "-o", "user,pid,ppid,pgid,stat,time,command")
```
directionless, line 168: Do you think this makes more sense than `ps -eaf` or `ps aux`? I don't know how macOS handles the threads here.

directionless, line 168: Execs should have a context with a timeout

pkg/log/log.go
```diff
@@ -149,6 +158,82 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
...
+}
+
+// runAndLogLsofByPID runs lsof filtering on the given PID, and logs the output.
+func (l *OsqueryLogAdapter) runAndLogLsofByPID(pid int) {
+	if runtime.GOOS == "windows" {
+		return
+	}
+
+	pidStr := strconv.Itoa(pid)
+	cmd := exec.Command("lsof", "-R", "-p", pidStr)
```
directionless, line 193: Should have a context and timeout

pkg/log/log.go
```diff
@@ -149,6 +158,82 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
...
+}
+
+// runAndLogLsofOnPidfile runs lsof filtering by the osquery pidfile, and logs
+// the output.
+func (l *OsqueryLogAdapter) runAndLogLsofOnPidfile() {
+	if runtime.GOOS == "windows" {
+		return
+	}
+
+	fullPidfile := filepath.Join("/private", l.rootDirectory, "osquery.pid")
```
directionless, line 218: Why `/private`? I don't think you need that

pkg/log/log.go
directionless, line 193: Probably add `-n`

pkg/log/log.go
```diff
@@ -149,6 +158,82 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
...
+	cmd := exec.Command("lsof", "-R", fullPidfile)
```
directionless, line 219: Same. (context, timeout, maybe `-n`)

directionless (commented):
Seems pretty reasonable. I think execs should get timeouts (no need to tie the context to a parent, but we should timeout)

pkg/log/log.go
RebeccaMahany: Oh weird, I assumed I needed it because that's how it shows up in the output, but I do not. I will remove.

pkg/log/log.go
RebeccaMahany:
```
       -n   inhibits the conversion of network numbers to host names for
            network files.  Inhibiting conversion may make lsof run faster.
            It is also useful when host name lookup is not working properly.
```

Happy to add it but would like to know why -- just for the performance impact, or another reason?

pkg/log/log.go
RebeccaMahany: -e and -a seemed incompatible with filtering by PID with -p, but I might've missed something!

Re threads, `H/-L` don't seem to be doing it on macOS for me.

pkg/log/log.go
directionless: On macOS `/var` is a symlink to `/private/var` but you can access it via `/var`. So we shouldn't assume that link is always present.

pkg/log/log.go
directionless: General performance, yes. We don't need name resolution and the old skool unix habit is to always turn it off.

pkg/log/log.go
```diff
@@ -112,6 +119,10 @@ func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []by
 		return
 	}
 
+	l.runAndLogPs(pid)
+	l.runAndLogLsofByPID(pid)
+	l.runAndLogLsofOnPidfile()
+
 	unknownProcess, err := process.NewProcess(int32(pid))
```
directionless, line 126: Thank you for adding the timeouts. It occurs to me -- this is in the logging path, and we're blocking logs on execs. Thoughts about a goroutine for something in here? (Maybe just the call to `logInfoAboutUnrecognizedProcessLockingPidfile`)

RebeccaMahany: I was wondering about that -- yeah, that makes sense to me! I'll update and test.

directionless (approved):

PR #1234, opened 2023-06-23
Don't attempt to use secure enclave

RebeccaMahany:
We can reliably reproduce an issue on macOS Sonoma beta where an attempt to call secureenclave.CreateKey will cause launcher to hang indefinitely, halting startup.

We don't expect this call to work at the moment anyway because we still need to sort out running it in the correct user context, so for now we will disable this call.


directionless (dismissed):
I am sad, but this seems correct.

@James-Pickett So you know

directionless (approved):

directionless (approved):

PR #1238, opened 2023-06-28
Don't return errors for `falconctl`

directionless:
As we keep finding, if there's an exec error, we should not return a table error. Instead, we should just move on.

This is an attempt at doing that, but _also_ passing along a `_error` field.

### DIFF
--- a/pkg/osquery/tables/crowdstrike/falconctl/table.go
+++ b/pkg/osquery/tables/crowdstrike/falconctl/table.go
@@ -3,6 +3,7 @@ package falconctl
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


pkg/osquery/tables/crowdstrike/falconctl/table.go
```diff
@@ -89,13 +90,26 @@ OUTER:
 		output, err := t.execFunc(ctx, t.logger, 30, falconctlPaths, args, false)
 		if err != nil {
 			level.Info(t.logger).Log("msg", "exec failed", "err", err)
-			return nil, err
+			synthisized_data := map[string]string{
```
RebeccaMahany, line 93: Typo -- I think we probably want `synthesizedData` for consistent formatting

RebeccaMahany (dismissed):
LGTM, just the one nitpick

pkg/osquery/tables/crowdstrike/falconctl/table.go
directionless: hahahaha. You're totally right. I've been writing too much ruby.

RebeccaMahany (approved):

PR #1239, opened 2023-06-29
copy input in send buffer to stop data loss, add ulid and caller to log shipper logs

James-Pickett:


pkg/log/logshipper/logshipper.go
```diff
@@ -39,7 +40,7 @@ func New(k types.Knapsack, baseLogger log.Logger) *LogShipper {
 	}
 
 	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval))
-	shippingLogger := log.NewJSONLogger(sendBuffer)
+	shippingLogger := log.With(log.NewJSONLogger(sendBuffer), "caller", log.Caller(6), "session_ulid", ulid.New())
```
James-Pickett, line 43: wonder if we should send the querier to get serial # or some kind of machine ID

directionless: I'd kick that forward. I think we want it, but I'm not sure where it'll make the most sense to add

directionless: What's the ulid for?

directionless (dismissed):
Approved, though I don't know what the ulid is for.

pkg/log/logshipper/logshipper.go
James-Pickett: The ulid is just uniquely identify the session, I think while the PID is sufficient for local logs to debug.json. We will have overlap when we aggregating 10,000s of logs.

directionless (approved):

PR #1242, opened 2023-06-30
add device identifying info to the log shipper

James-Pickett:


pkg/log/logshipper/logshipper.go
```diff
@@ -118,3 +120,31 @@ func filterResults(keyvals ...interface{}) {
...
 	}
 }
+
+// addDeviceIdentifyingAttributesToLogger gets device identifiers from the server-provided
+// data and adds them as attributes on the logger.
+func (ls *LogShipper) addDeviceIdentifyingAttributesToLogger() {
+	if deviceId, err := ls.knapsack.ServerProvidedDataStore().Get([]byte("device_id")); err != nil {
+		level.Debug(ls.baseLogger).Log("msg", "could not get device id", "err", err)
+	} else {
+		ls.shippingLogger = log.With(ls.shippingLogger, "k2_device_id", string(deviceId))
```
James-Pickett, line 130: when I copied this from traces attributes I changed it from `k2.device_id` to `k2_device_id` since we don't typically use dots in logging, but I wonder if we should make exception for easier correlation

directionless: We should probably pick something. I'm not sure I have strong opinions about what

directionless (approved):

PR #1248, opened 2023-07-06
Subscribe to power events on Windows

RebeccaMahany:
Uses EvtSubscribe to receive notifications when we enter or leave modern standby.

When looking at the sleep report for my computer, I noticed that on lid shut or sleep, the device entered Modern Standby immediately (event ID 506). On resume, it would exit Modern Standby (507).

After many hours in Modern Standby, the device would exit Modern Standby (507) with `Reason: Sleep, Hibernate, or Shutdown`, and immediately enter sleep (42) with reason `Hibernate from Sleep - Standby Battery Budget Exceeded`. On wake, it would re-enter (506) and then exit (507) Modern Standby.

```mermaid
stateDiagram-v2
    Awake --> ModernStandby: 506
    ModernStandby --> Awake: 507
    ModernStandby --> Sleep: 507 => 42
    Sleep --> ModernStandby: 506
```


RebeccaMahany:
> If I read this correctly, it's subscribing to the windows events.

Yes -- I found this was more reliable/useful than registering the service to receive [power events](https://pkg.go.dev/golang.org/x/sys@v0.10.0/windows#SERVICE_CONTROL_POWEREVENT).
 
> Any thoughts about that approach vs setting up a callback on DAM actions? The bottom of https://learn.microsoft.com/en-us/windows/compatibility/desktop-activity-moderator has a solution section about how to do this with windows notifications.

A couple places I read (comment on [this answer](https://stackoverflow.com/a/11775489), [this answer](https://stackoverflow.com/a/22509218) that links to [this official post](https://learn.microsoft.com/en-us/archive/blogs/winsdk/how-to-get-notified-when-going-in-and-out-of-connected-standby-from-a-windows-service)) indicated that `PowerRegisterSuspendResumeNotification` maybe would not work -- I hadn't given it a shot given that and given how much time it took to get _this_ option to work. I can try it this week to confirm, though -- it looks like all of those links are fairly old so maybe something has changed in the meantime. (I did try the option linked in the official post and couldn't get that to work either.)

cmd/launcher/svc_watcher_windows.go
```diff
@@ -0,0 +1,165 @@
+//go:build windows
+// +build windows
+
+package main
```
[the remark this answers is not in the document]
RebeccaMahany: I agree -- let me know if you have thoughts on where!

directionless:
> > Any thoughts about that approach vs setting up a callback on DAM actions? The bottom of https://learn.microsoft.com/en-us/windows/compatibility/desktop-activity-moderator has a solution section about how to do this with windows notifications.
> 
> A couple places I read (comment on [this answer](https://stackoverflow.com/a/11775489), [this answer](https://stackoverflow.com/a/22509218) that links to [this official post](https://learn.microsoft.com/en-us/archive/blogs/winsdk/how-to-get-notified-when-going-in-and-out-of-connected-standby-from-a-windows-service)) indicated that `PowerRegisterSuspendResumeNotification` maybe would not work -- I hadn't given it a shot given that and given how much time it took to get _this_ option to work. I can try it this week to confirm, though -- it looks like all of those links are fairly old so maybe something has changed in the meantime. (I did try the option linked in the official post and couldn't get that to work either.)

I don't know enough to know which approach is more "supported". I do not currently have strong opinions.

cmd/launcher/svc_watcher_windows.go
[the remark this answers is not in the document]
RebeccaMahany: Moved to pkg/windows/eventwatcher

pkg/windows/eventwatcher/power_event_watcher_windows.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// subscribeToPowerEvents sets up a subscription to relevant power events with a callback to `onPowerEvent`.
+func (p *powerEventWatcher) SubscribeToPowerEvents() {
+	// WINEVENT_CHANNEL_GLOBAL_SYSTEM is "System"
+	channelPath, err := syscall.UTF16PtrFromString("System")
+	if err != nil {
+		level.Debug(p.logger).Log("msg", "error creating pointer to channel path", "err", err)
+		return
+	}
```
directionless, line 62: Hrm. I wonder if we should log and return, or return the error. Feels a little TBD. 

This feels a bit like our usual go actors -- and we should return the error. Though we might end up ignoring it at the caller level

directionless (approved):
seems okay to try

pkg/windows/eventwatcher/power_event_watcher_windows.go
RebeccaMahany: I was also thinking this should maybe just be a rungroup actor, and that's why I was waffling on the error handling here -- because I didn't want to return an error on the equivalent of `Execute` and force a launcher restart, since this is just for logging right now.

I think the option that feels the most consistent to me is to move this setup work into the `New` function, and return an error on failure to set up. If `New` returns an error, log it and move on; if it doesn't, add the rungroup with a no-op `Execute`. Does that sound reasonable?

RebeccaMahany: Pushed this change, lmk what you think!

James-Pickett (dismissed):

James-Pickett:
so we just plan to log the events for now? not actually acting on them?

pkg/windows/eventwatcher/power_event_watcher_windows.go
directionless: Yes. I think that pattern is good.

`New` returns an error, and the caller decides if it's fatal. And after that, it's in it's own little world.

directionless:
> so we just plan to log the events for now? not actually acting on them?

Yeah. Partly because I want to get this to customer to see if it's what's happening. And I think actioning them might be tricky. So get this out, then work on what's next

pkg/windows/eventwatcher/power_event_watcher_other.go
```diff
@@ -0,0 +1,27 @@
+//go:build !windows
+// +build !windows
+
+package eventwatcher
```
directionless, line 4: feels weird to have this in a package called `windows/eventwatcher`. Also feels like "eventwatcher" is "powereventwatcher"

pkg/windows/eventwatcher/power_event_watcher_windows.go
```diff
@@ -0,0 +1,170 @@
...
+	p.subscriptionHandle = subscriptionHandle
+
+	return p, nil
+}
+
+// Execute is a no-op, since we've already registered our subscription
+func (p *powerEventWatcher) Execute() error {
+	<-p.interrupt
+	return nil
+}
```
directionless, line 99: Feels like a lot of New should move here. But probably doesn't matter yet

directionless (dismissed):
I'd take it

pkg/windows/eventwatcher/power_event_watcher_other.go
RebeccaMahany: Happy to rename whatever! I'll update the package to `powereventwatcher` so it's more consistent -- let me know if there's an update you want for this no-op one too.

pkg/windows/eventwatcher/power_event_watcher_windows.go
```diff
@@ -0,0 +1,170 @@
...
+		return ret
+	}
+
+	if e.System.EventID == eventIdEnteringModernStandby || e.System.EventID == eventIdEnteringSleep {
+		level.Debug(p.logger).Log("msg", "system is sleeping", "event_id", e.System.EventID)
+	} else if e.System.EventID == eventIdExitingModernStandby {
+		level.Debug(p.logger).Log("msg", "system is waking", "event_id", e.System.EventID)
+	} else {
+		level.Debug(p.logger).Log("msg", "received unexpected event ID in log", "event_id", e.System.EventID, "raw_event", string(utf8bytes))
+	}
```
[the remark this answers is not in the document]
RebeccaMahany: Will update!

pkg/windows/eventwatcher/power_event_watcher_other.go
directionless: Nothing specific to the noop. Just pondering naming and locations.

PR #1249, opened 2023-07-10
Add debian important updates table 

Micah-Kolide:
To add linux distribution version support there were a couple of approaches that I was thinking of.

1. Generate a table with current distribution version info, then find external sources of truth for distribution releases. A check could see if a newer update version is applicable.

2. Generate a table with current upgradeable packages, then determine a level of importance for those packages. A check could see if any upgradeable package is above the set minimum importance level.

This PR is to provide a table with upgradeable packages that have a calculated importance level.

I figured I'd attempt this way first for a couple of assumed reasons:

- 1 would need a source of truth of release versions for each distribution. This will require a lot of work to set up and maintain over time depending on how many distributions we want to support.
- 2 I believe would be more sensitive since it will look for any package that can upgrade, which is determined an important package. Whether being more sensitive in this case is better or not is debatable.

I know that 1 follows the windows and macos approach we have, but with how many distributions linux has I'm not sure if it makes sense to do the same. I think that 2 will be harder to approach for some other distributions, so maybe a mix will be required.

I need to add validation to the input data as well clean the output, but I first wanted to get thoughts on the general approach here.

I felt somewhat stuck trying to think of the best way to determine which packages are important. I'm using the apt upgradeable list merged to the dpkg package version info to try to establish how important an available package update is for the system.

Here is some output from a test ubuntu machine:

```+-----------------------------------+-----------------------------------------------------------+----------------------------------------+-------------------------------------+-----------+-----------+--------------+-------------------------+-----------------+--------------------+
| package                           | sources                                                   | update_version                         | current_version                     | essential | priority  | section      | task                    | build_essential | assumed_importance |
+-----------------------------------+-----------------------------------------------------------+----------------------------------------+-------------------------------------+-----------+-----------+--------------+-------------------------+-----------------+--------------------+
| apt                               | jammy-updates                                             | 2.4.9                                  | 2.4.8                               |           | important | admin        | minimal, server-minimal | yes             | 3                  |
| apt-transport-https               | jammy-updates,jammy-updates                               | 2.4.9                                  | 2.4.8                               |           |           |              |                         |                 | 0                  |
| apt-utils                         | jammy-updates                                             | 2.4.9                                  | 2.4.8                               |           | important | admin        | minimal                 |                 | 2                  |
| avahi-autoipd                     | jammy-updates,jammy-security                              | 0.8-5ubuntu5.1                         | 0.8-5ubuntu5                        |           |           |              |                         |                 | 1                  |
| avahi-daemon                      | jammy-updates,jammy-security                              | 0.8-5ubuntu5.1                         | 0.8-5ubuntu5                        |           |           |              |                         |                 | 1                  |
| avahi-utils                       | jammy-updates,jammy-security                              | 0.8-5ubuntu5.1                         | 0.8-5ubuntu5                        |           |           |              |                         |                 | 1                  |
| base-files                        | jammy-updates                                             | 12ubuntu4.3                            | 12ubuntu4.2                         | yes       | required  | admin        | minimal                 |                 | 4                  |
| binutils                          | jammy-updates,jammy-security                              | 2.38-4ubuntu2.2                        | 2.38-4ubuntu2.1                     |           |           |              |                         |                 | 1                  |
| binutils-common                   | jammy-updates,jammy-security                              | 2.38-4ubuntu2.2                        | 2.38-4ubuntu2.1                     |           |           |              |                         |                 | 1                  |
| binutils-x86-64-linux-gnu         | jammy-updates,jammy-security                              | 2.38-4ubuntu2.2                        | 2.38-4ubuntu2.1                     |           |           |              |                         |                 | 1                  |
| brave-browser                     | stable                                                    | 1.52.130                               | 1.47.171                            |           |           |              |                         |                 | 0                  |
+-----------------------------------+-----------------------------------------------------------+----------------------------------------+-------------------------------------+-----------+-----------+--------------+-------------------------+-----------------+--------------------+
```


pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/parser.go
```diff
@@ -0,0 +1,93 @@
...
+	scanner := bufio.NewScanner(bytes.NewReader(rawdata))
+	for scanner.Scan() {
+		line := scanner.Text()
+		pair := strings.Split(line, "/")
+		if len(pair) < 2 {
+			continue
+		}
+
+		package_name := strings.ToLower(strings.TrimSpace(pair[0]))
+		if len(package_name) > 0 {
```
directionless, line 26: Prefer early return
```
if len(package_name) < 1 {
  continue
}
```

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/parser.go
```diff
@@ -0,0 +1,93 @@
...
+			values := strings.Split(pair[1], " ")
+			if len(values) >= 6 {
```
directionless, line 28: Prefer early return

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/parser.go
```diff
@@ -0,0 +1,93 @@
+package linux_important_updates_debian
```
directionless, line 1: Several comments about names and locations and files, somewhat overlapping.

`pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/` feels off. I think nesting things under `linux_update_tables` helps, and `linux_important_updates_debian` is pretty long. I'd probably `pkg/osquery/tables/apt_updates`

But, I'd also consider putting the parsers into `pkg/osquery/tables/execparsers` and using the wrappers. Kinda depends. Comments over in the `generate`

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/parser.go
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+	}
+
+	return data, nil
+}
+
+func dpkgOutput(rawdata []byte) (map[string]map[string]string, error) {
+	var package_name string
+	var allowedKeys = []string{
```
directionless, line 43: small preference for package vars. (eg: move it out of the function)

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/parser.go
```diff
@@ -0,0 +1,93 @@
...
+		kv := strings.Split(line, ": ")
+		if len(kv) < 2 {
+			continue
+		}
+
+		// dpkg retutns a lot of unused data, so lets filter it.
+		for _, allowedKey := range allowedKeys {
+			var key = strings.ReplaceAll(strings.ToLower(strings.TrimSpace(kv[0])), "-", "_")
+			var value = strings.TrimSpace(kv[1])
+			if key == allowedKey {
```
directionless, line 81: Use `https://pkg.go.dev/golang.org/x/exp/slices#Contains`

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/table.go
```diff
@@ -0,0 +1,125 @@
...
+	pac_man_out, err := tablehelpers.Exec(ctx, t.logger, 5, pacManPath, pacManCmds, false)
+	if err != nil {
+		level.Info(t.logger).Log("msg", "apt list failed", "err", err)
+	}
+	pac_inst_out, err := tablehelpers.Exec(ctx, t.logger, 5, pacInstPath, pacInstCmds, false)
+	if err != nil {
+		level.Info(t.logger).Log("msg", "dpkg failed", "err", err)
+	}
+
+	aptout, err := aptOutput(pac_man_out)
```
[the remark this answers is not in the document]
directionless: eg: check the error

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/table.go
```diff
@@ -0,0 +1,125 @@
...
+	dpkgout, err := dpkgOutput(pac_inst_out)
+
+	for dk, dv := range dpkgout {
+		if _, ok := aptout[dk]; ok {
+			for dvk, dvv := range dv {
+				aptout[dk][dvk] = dvv
+			}
+		}
+	}
```
directionless, line 74: I don't have a linux box to quickly test this on, but I'm curious why why these are being stitched together. 

Do we need to? Are there times we want them independently? Does it make more sense to manipulate in SQL?

If so, we should be using the exec helpers.

pkg/osquery/tables/linux_update_tables/linux_important_updates_debian/table.go
```diff
@@ -0,0 +1,125 @@
...
+	for  _, row := range aptout {
+		importance := 0
+
+		// Figuring out how to tell when an update *should*
+		// be applied has been difficult. I've been trying
+		// to keep system execs low, and for debian I hope
+		// something along these lines works.
+		//
+		// Merging the upgradable apt sources with dpkg's
+		// version info seems to give a good starting point.
```
directionless, line 85: I'm not sure if this logic should be here, vs in sql on the underlying tables.

My current thinking, is that launcher tables should be fairly lightweight. Heavier logic should land in SQL. This is mostly because I find we need to iterate and adjust that logic fairly often, and launcher has a long lead time. 

So launcher presents tables as small consistent wrappers, and deeper works happens in sql

directionless (commented):
Despite all my comments, the underlying parts look pretty solid. 

Test files for the parsers would be nice. They'd also make it easier to understand how the apt and dpkg things combine.

I'm a bit unsure about doing the join here, vs two tables

Micah-Kolide:
After discussing the implementation of these tables, I've changed the approach to make new exec parsers and keeping the merging/importance calculations to the check sql.

pkg/osquery/tables/execparsers/apt/parser_test.go
```diff
@@ -0,0 +1,132 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+		})
+
+		p := New()
+		result, err := p.Parse(bytes.NewReader(tt.input))
+		require.NoError(t, err, "unexpected error parsing input")
+
+		require.ElementsMatch(t, tt.expected, result)
```
directionless, line 130:
```suggestion
		t.Run(tt.name, func(t *testing.T) {
			t.Parallel()
		p := New()
		result, err := p.Parse(bytes.NewReader(tt.input))
		require.NoError(t, err, "unexpected error parsing input")

		require.ElementsMatch(t, tt.expected, result)
		})

```

(the tests should be inside the `t.Run` function

pkg/osquery/tables/execparsers/apt/parser.go
```diff
@@ -0,0 +1,41 @@
+package apt
+
+import (
+	"bufio"
+	//"fmt"
```
directionless, line 5:
```suggestion
```

pkg/osquery/tables/execparsers/apt/parser.go
```diff
@@ -0,0 +1,41 @@
...
+	"strings"
+)
+
+func aptParse(reader io.Reader) (any, error) {
+	results := make([]map[string]string, 0)
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		pair := strings.Split(line, "/")
```
directionless, line 16: I'll take this. But FYI it might be slightly cleaner using a regex match on the line

pkg/osquery/tables/execparsers/apt/parser_test.go
```diff
@@ -0,0 +1,132 @@
...
+					"package":         "xxd",
+					"sources":         "jammy-updates,jammy-security",
+					"update_version":  "2:8.2.3995-1ubuntu2.9",
+					"current_version": "2:8.2.3995-1ubuntu2.3",
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
```
directionless, line 120: Note that you only have one test. So the table test pattern is a little bit of overkill. It's fine though -- it makes it easy to add more, if we need that

pkg/osquery/tables/execparsers/dpkg/parser_test.go
```diff
@@ -0,0 +1,152 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+		})
+
+		p := New()
+		result, err := p.Parse(bytes.NewReader(tt.input))
+		require.NoError(t, err, "unexpected error parsing input")
+
+		require.ElementsMatch(t, tt.expected, result)
```
directionless, line 150: same

pkg/osquery/tables/execparsers/dpkg/parser.go
```diff
@@ -0,0 +1,47 @@
...
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if len(line) == 0 {
+			results = append(results, row)
+			row = make(map[string]string)
+			continue
+		}
+
+		kv := strings.Split(line, ": ")
```
directionless, line 35: `strings.SplitN` feels more correct here.

pkg/osquery/tables/execparsers/dpkg/parser.go
```diff
@@ -0,0 +1,47 @@
...
+		if len(kv) < 2 {
+			continue
+		}
+
+		var key = strings.ReplaceAll(strings.ToLower(strings.TrimSpace(kv[0])), "-", "_")
```
directionless, line 40: Why replace `-` with `_`?

pkg/osquery/tables/execparsers/apt/parser.go
```diff
@@ -0,0 +1,41 @@
...
+	"io"
+	"strings"
+)
+
+func aptParse(reader io.Reader) (any, error) {
+	results := make([]map[string]string, 0)
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
```
RebeccaMahany, line 15: Could you add some quick documentation here with an example, to make it easier to follow the code below? E.g. 

```
We expect a line in the following format:
`ubuntu-standard/jammy-updates 1.481.1 amd64 [upgradable from: 1.481]`.
We split on the forward slash to extract the package name.
We expect the rest of the line to be in the format `<source> <update version> <arch> [upgradable from: <current version>]`.
```

pkg/osquery/tables/execparsers/apt/parser.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		package_name := strings.ToLower(strings.TrimSpace(pair[0]))
+		if len(package_name) < 1 {
+			continue
+		}
+
+		values := strings.Split(pair[1], " ")
+		if len(values) < 6 {
+			continue
```
RebeccaMahany, line 28: Would we find it helpful to log a malformed line? Or do we expect to see a couple malformed lines already and that would just be noisy?

pkg/osquery/tables/execparsers/dpkg/parser.go
```diff
@@ -0,0 +1,47 @@
+package dpkg
+
+import (
+	"bufio"
+	//"fmt"
```
RebeccaMahany, line 5: I think you can remove this 🙂

pkg/osquery/tables/execparsers/dpkg/parser.go
```diff
@@ -0,0 +1,47 @@
...
+	"package",
+	"essential",
+	"priority",
+	"section",
+	"version",
+	"description",
+	"build_essential",
+}
+
+func dpkgParse(reader io.Reader) (any, error) {
```
RebeccaMahany, line 22: As in the other, could you document the format of what we're parsing to make it easier to reference below and understand what the parsing is doing? The output is a lot more verbose so you don't have to put an entire example, but just mentioning that each package block is separated by newlines and is key-value pairs separated by commas would help.

pkg/osquery/tables/execparsers/dpkg/parser.go
```diff
@@ -0,0 +1,47 @@
...
+	results := make([]map[string]string, 0)
+	row := make(map[string]string)
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if len(line) == 0 {
+			results = append(results, row)
```
RebeccaMahany, line 30: Could you add a comment saying that a empty newline means we're done processing data for the current package, so add that row and make a new one for the next package?

pkg/osquery/tables/execparsers/dpkg/parser.go
RebeccaMahany: ➕ 1️⃣ a comment would be helpful here

RebeccaMahany (commented):
A couple small comments -- overall approach looks good to me!

pkg/osquery/tables/execparsers/apt/parser.go
directionless: I always just read them out of the test files. But comments are good too.

pkg/osquery/tables/execparsers/apt/parser_test.go
James-Pickett: I also prefer table pattern even with just one case. As other said, it makes it easier to add stuff in the future, but it is also just our consistent pattern. It now throws me off a little when I come across a non-table test =).

pkg/osquery/tables/execparsers/dpkg/parser_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test-data/dpkg_info.txt
+var dpkg_info []byte
+
+func TestParse(t *testing.T) {
+	t.Parallel()
+
+	var tests = []struct {
```
James-Pickett, line 17: same as other parser, would like to see test cases for malformed data, missing data, no data, etc

James-Pickett (commented):
Nice work! The code looks solid, my most significant comments are asking for more error test cases.

pkg/osquery/tables/execparsers/apt/parser.go
RebeccaMahany: I know I can reference that file, but someone who's new to the codebase might not know that it exists since it's not mentioned here and is nested in a separate directory. Just a little easier to have the info we need here, rather than knowing to check elsewhere (or boot up ubuntu and run the command, which is what I did when reviewing)

pkg/osquery/tables/execparsers/apt/parser_test.go
Micah-Kolide: Added more tests :)

pkg/osquery/tables/execparsers/dpkg/parser.go
Micah-Kolide: Done

pkg/osquery/tables/execparsers/dpkg/parser.go
Micah-Kolide: Old remnant. I removed it.

pkg/osquery/tables/execparsers/apt/parser.go
```diff
@@ -0,0 +1,41 @@
...
+
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		pair := strings.Split(line, "/")
+		if len(pair) < 2 {
+			continue
+		}
+
+		package_name := strings.ToLower(strings.TrimSpace(pair[0]))
```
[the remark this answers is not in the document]
Micah-Kolide: Done

pkg/osquery/tables/execparsers/apt/parser.go
Micah-Kolide: I added some comments similar to the example text you put here :)

pkg/osquery/tables/execparsers/apt/parser.go
Micah-Kolide: I added a malformed test line

pkg/osquery/tables/execparsers/dpkg/parser.go
Micah-Kolide: Added

pkg/osquery/tables/execparsers/apt/parser_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test-data/apt_upgradeable.txt
+var apt_upgradeable []byte
+
+func TestParse(t *testing.T) {
+	t.Parallel()
+
+	var tests = []struct {
```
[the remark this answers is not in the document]
Micah-Kolide: Added

RebeccaMahany (approved):
LGTM, thanks for making those updates!

PR #1250, opened 2023-07-11
james/accelerate control via control update

James-Pickett:
- adds action queue layer to handle actions from control server
- updates notification consumer to use action queue
- adds actor for handling control server acceleration actions


ee/control/actionqueue/actionqueue.go
```diff
@@ -0,0 +1,241 @@
...
+const (
+	ActionsSubsystem       = "actions"
+	defaultCleanupInterval = time.Hour * 12
+	// about 6 months, long enough to ensure that K2 no longer has the message
+	// and will not send a duplicate
+	actionRetentionPeriod = time.Hour * 24 * 30 * 6
+)
+
+type updater interface {
+	Update(data io.Reader) error
```
RebeccaMahany, line 27: I think the reuse of updater/`Update` is a little overloaded (since it's the same function as for consumers in the control system) and isn't as precise as it could be here. Action "updaters" aren't really updaters because they're not performing generic data updates -- they're taking actions. What about `actor interface` and `Action(data io.Reader)` or `TakeAction(data io.Reader)`? I understand if we feel like that might be overloading run group stuff, though -- maybe something with "execute" or "perform" instead?

ee/control/actionqueue/actionqueue.go
```diff
@@ -0,0 +1,241 @@
...
+			aq.cleanupActions()
+		}
+	}
+}
+
+func (aq *actionqueue) StopCleanup(err error) {
+	aq.cancel()
+}
+
+func (aq *actionqueue) storeActionRecord(action action) {
```
RebeccaMahany, line 160: Can we rename the variable (to `actionToStore` or `a`) so that it's not the same as the type? Reusing type or package identifiers as variable names re-defines that keyword for the current scope, which can result in [confusing behavior](https://play.golang.com/p/dCoezMdRnwY).

ee/control/actionqueue/actionqueue.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return nil
+	}); err != nil {
+		level.Debug(aq.logger).Log("msg", "could not iterate over bucket items to determine which are expired", "err", err)
+	}
+
+	// Delete all old keys
+	if err := aq.store.Delete(keysToDelete...); err != nil {
+		level.Debug(aq.logger).Log("msg", "could not delete old notifications from bucket", "err", err)
```
RebeccaMahany, line 239:
```suggestion
		level.Debug(aq.logger).Log("msg", "could not delete old actions from bucket", "err", err)
```

ee/control/consumers/acceleratecontrolconsumer/acceleratecontrolconsumer.go
```diff
@@ -0,0 +1,57 @@
...
+		overrider: overrider,
+	}
+}
+
+func (c *AccelerateControlConsumer) Update(data io.Reader) error {
+	if c.overrider == nil {
+		return errors.New("control request interval overrider is nil")
+	}
+
+	accelerate_data := struct {
```
RebeccaMahany, line 35: Nit -- `accelerateData` for camelcase

RebeccaMahany (commented):
I like how you've expanded this! 🔥

ee/control/actionqueue/actionqueue.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if err := aq.store.Set([]byte(action.ID), rawAction); err != nil {
+		level.Debug(aq.logger).Log("msg", "could not mark notification sent", "err", err)
+	}
+}
+
+func (aq *actionqueue) isActionNew(id string) bool {
+	sentNotificationRaw, err := aq.store.Get([]byte(id))
+	if err != nil {
+		level.Error(aq.logger).Log("msg", "could not read action from bucket", "err", err)
```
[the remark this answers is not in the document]
James-Pickett: I agree there should be a return here, updated.

ee/control/actionqueue/actionqueue.go
James-Pickett: yea, I agree it's overloaded, went with `actor.Do`

ee/control/actionqueue/actionqueue.go
James-Pickett: done

ee/control/actionqueue/actionqueue.go
James-Pickett: done

ee/control/actionqueue/actionqueue_test.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func TestCleanup(t *testing.T) {
+	t.Parallel()
+
+	mockUpdater := mocks.NewUpdater(t)
+	store := setupStorage(t)
+	var logBytes threadsafebuffer.ThreadSafeBuffer
+	logger := log.NewLogfmtLogger(&logBytes)
+	amw := New(
```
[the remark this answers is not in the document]
James-Pickett: left over acronym from another approach, updated

ee/control/actionqueue/actionqueue_test.go
```diff
@@ -0,0 +1,255 @@
...
+			ValidUntil: getValidUntil(),
+			Type:       testUpdaterType,
+		},
+		{
+			ID:         testId,
+			ValidUntil: getValidUntil(),
+			Type:       testUpdaterType,
+		},
+	}
+	testNotificationsRaw, err := json.Marshal(testActions)
```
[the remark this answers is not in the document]
James-Pickett: renamed

ee/control/consumers/acceleratecontrolconsumer/acceleratecontrolconsumer.go
James-Pickett: fixed

pkg/agent/storage/stores.go
```diff
@@ -15,6 +15,7 @@ const (
 	StatusLogsStore             Store = "status_logs"              // The store used for buffered status logs.
 	ServerProvidedDataStore     Store = "server_provided_data"     // The store used for pushing values from server-backed tables.
 	TokenStore                  Store = "token_store"              // The store used for holding bearer auth tokens, e.g. the ones used to authenticate with the observability ingest server.
+	ControlServerActionsStore   Store = "action_store"             // The store used for storing actions sent by control server.
```
James-Pickett, line 18: so this will take the place of the sent notifications store which means we update to this it will be a fresh store and we will possibly send old notifications form K2.

should we ...

1. add a migration that transfers the sent notifications to actions store
2. have action queue look in both stores temporarily
3. something else ...

ee/control/actionqueue/actionqueue.go
RebeccaMahany: Ah that's way better/simpler than Action/TakeAction, I like it! Thanks for updating

ee/control/actionqueue/actionqueue.go
```diff
@@ -0,0 +1,241 @@
...
+	logger                log.Logger
+	actionCleanupInterval time.Duration
+	cancel                context.CancelFunc
+}
+
+type actionqueueOption func(*actionqueue)
+
+func WithLogger(logger log.Logger) actionqueueOption {
+	return func(aq *actionqueue) {
+		aq.logger = logger
```
RebeccaMahany, line 50:
```suggestion
		aq.logger = log.With(logger, "component", "actionqueue")
```

I've found the `With component` we've been doing in other places super helpful for understanding and filtering logs -- could we include it here too? Sorry I didn't notice this on the last pass through!

RebeccaMahany:
LGTM -- happy to approve when you've decided on https://github.com/kolide/launcher/pull/1250#discussion_r1280803546?

pkg/agent/storage/stores.go
James-Pickett: went with 2!

ee/control/actionqueue/actionqueue.go
James-Pickett: done

directionless:
Please hold on merging. I suspect we're going to cut an 1.0.15 RSN.

RebeccaMahany (approved):
Nice work! LGTM

PR #1253, opened 2023-07-14
Updating to a new version of signtool

RebeccaMahany:
Some packaging infrastructure improvements require a new version of signtool; this PR updates to point to the new file location.

### DIFF
--- a/pkg/packagekit/package_wix.go
+++ b/pkg/packagekit/package_wix.go
@@ -38,7 +38,7 @@ var wixTemplateBytes []byte
 var assets embed.FS
 
 const (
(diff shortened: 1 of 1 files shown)


directionless (approved):
Seems fine. Do you think it makes sense to have something that searches for the SDK version?

RebeccaMahany:
> Seems fine. Do you think it makes sense to have something that searches for the SDK version?

@directionless probably, I wouldn't want to have to make this PR again the next time we change signtool versions 😅 -- if I end up with some extra time this week for this project I can come back and take a look?

James-Pickett (approved):

PR #1254, opened 2023-07-17
Pick golang version from go.mod

RebeccaMahany:
Uses https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file -- this should prevent us from having to constantly bump golang patch versions as a response to govulncheck errors.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -18,9 +18,10 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         with:
-          go-version: '1.20.6'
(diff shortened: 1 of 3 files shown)


directionless (approved):

.github/workflows/golangci-lint.yml
```diff
@@ -20,7 +20,8 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
```
directionless, line 21: This might need to be v4

directionless (changes requested):

.github/workflows/golangci-lint.yml
RebeccaMahany: Ah just updated, it needs to be minimum v3!

directionless (approved):

Issue #1144, opened 2023-04-18
Add a `launcher flare` command


directionless:
Tracking for flare (and doctor) work

* [ ] #1255 
* [ ] Add flare support to menu and actions. (see notes in #1255, this needs to run in launcher main not in menu
* [ ] Add upload functions
* [ ] Cleanup stuff around `cmd/launcher/paths.go`

directionless:
This is done by #1255

PR #1255, opened 2023-07-17
Refactor `flare` and `doctor`

directionless:
I picked up #1243 and refactored. I'm not totally sure I love where this ended up. It's a bit less interfacy, and a bit more helpery, but I didn't see a way to get a really clean interface. The way the data works doesn't fit right. 

I also removed a lot of the color changing text. It's neat, but it kept leaving my terminal funny colors, and I think a simple emoji provides the same value, without the confusion.

There are still `checkups` to write, and this does not tackle the upload side. But this feels like the right base. 

Closes: #1243
Relates To: #1144


pkg/debug/checkups/connectivity.go
```diff
@@ -0,0 +1,119 @@
...
+func (c *Connectivity) Status() Status {
+	return c.status
+}
+
+func (c *Connectivity) Summary() string {
+	return c.summary
+}
+
+func (c *Connectivity) Data() any {
+	return nil
```
RebeccaMahany, line 70: Why is this nil?

pkg/debug/checkups/processes.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+func (c *Processes) ExtraFileName() string {
+	return "process.txt"
+}
+
+func (c *Processes) Status() Status {
+	if c.kolideCount > 2 {
```
RebeccaMahany, line 57: Don't we expect 3 for macOS, and 2 for Windows/Linux?

pkg/debug/checkups/root_directory.go
```diff
@@ -0,0 +1,58 @@
...
+		c.status = Passing
+		c.summary = fmt.Sprintf("root directory (%s) contains %d files", c.k.RootDirectory(), len(files))
+
+		fmt.Fprintf(fullFH, "%s\n", files)
+	}
+	return nil
+}
+
+func (c *RootDirectory) ExtraFileName() string {
+	return "file-list"
```
RebeccaMahany, line 45: Other `ExtraFileName`s have an extension -- should this one?

pkg/debug/checkups/root_directory.go
```diff
@@ -0,0 +1,58 @@
...
+		c.summary = fmt.Sprintf("root directory (%s) not present", c.k.RootDirectory())
+	case err != nil:
+		c.status = Erroring
+		c.summary = fmt.Sprintf("listing files in root directory (%s): %s", c.k.RootDirectory(), err)
+	case len(files) == 0:
+		c.status = Warning
+		c.summary = fmt.Sprintf("root directory (%s) empty", c.k.RootDirectory())
+	default:
+		c.status = Passing
+		c.summary = fmt.Sprintf("root directory (%s) contains %d files", c.k.RootDirectory(), len(files))
```
RebeccaMahany, line 37: Would we potentially want to check for the existence of some of the more important files within this directory too?

RebeccaMahany (commented):
I'm happy with the shape of this overall -- the thing that's mostly unclear to me is when we decide to include output in `Data()` versus `ExtraFileName()`. I think clarifying what should go in the data artifact would help with that.

ee/desktop/user/menu/action_flare.go
```diff
@@ -0,0 +1,22 @@
...
+
+func (a actionFlare) Perform(m *menu) {
+	if err := runFlare(); err != nil {
+		level.Error(m.logger).Log("msg", "error creating flare", "err", err)
+	}
+}
+
+func runFlare() error {
+	// TODO: can't create a flare without knapsack, and ideally, being root. This probably is going to be easiest if we use
+	// IPC to send the request for flare back to the parent instead of running it inline here. But it's a bit TBD
```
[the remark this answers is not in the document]
directionless: Yeah. I kicked this forward :)

pkg/debug/checkups/processes.go
directionless: Dunno, I didn't want to think super hard. This felt okay as a high level pass/fail

pkg/debug/checkups/root_directory.go
directionless: Kinda up to us. They're named whatever we think is easy to understand.

Mostly, I started with calling them `full` and decided that was confusing since processes bubbled up a list, while logs wanted to make a zip file. I could decide on an extension here, so I skipped it. posix doesn't require one

pkg/debug/checkups/root_directory.go
RebeccaMahany: I was thinking launcher.db/osquery.db, and maybe it'd be interesting to know if the `updates/` directory exists and what's in it. `updates/` at least sounds like it would belong in another checkup.

pkg/debug/checkups/processes.go
RebeccaMahany: Sorry, I didn't write my comment very clearly -- I meant that I think this will fail on Windows/Linux when it shouldn't, because we only have have the extra (3rd) sudo process on macOS.

cmd/launcher/flare.go
```diff
@@ -42,322 +26,28 @@ func runFlare(args []string) error {
...
-	if err != nil {
-		fatal(b, err)
-	}
-	output(b, stdout, "%v\n", string(jsonVersion))
-
-	logger := log.NewLogfmtLogger(b)
+	logger := log.NewLogfmtLogger(os.Stdout)
 	fcOpts := []flags.Option{flags.WithCmdLineOpts(opts)}
 	flagController := flags.NewFlagController(logger, nil, fcOpts...)
 	k := knapsack.New(nil, flagController, nil)
```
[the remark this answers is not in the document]
directionless: db is single reader -- can't access if it launcher is running. (And honestly, I'm not sure how much of this we want to keep. I think it's going to need to evolve some more. But maybe not in this PR)

pkg/debug/checkups/checkups.go
```diff
@@ -0,0 +1,232 @@
...
+
+type Status string
+
+const (
+	Unknown       Status = "Unknown"
+	Erroring      Status = "Error"         // The checkup had an error generating data
+	Informational Status = "Informational" // Just an FYI
+	Passing       Status = "Passing"       // Checkup passes
+	Warning       Status = "Warning"       // Checkup is warning
+	Failing       Status = "Failing"       // Checkup fails
```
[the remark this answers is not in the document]
directionless: You are correct, and I updated some comments. Maybe more clear now

pkg/debug/checkups/checkups.go
```diff
@@ -0,0 +1,232 @@
...
+	default:
+		return "? "
+	}
+}
+
+// checkupInt is the generalized checkup interface. It is not meant to be exported.
+type checkupInt interface {
+	Name() string
+	Run(ctx context.Context, fullWriter io.Writer) error
+	ExtraFileName() string
```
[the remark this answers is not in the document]
directionless: I think I should purge `detailedlog` from everything. That's more confusing than `extra` (or additional)

pkg/debug/checkups/checkups.go
```diff
@@ -0,0 +1,232 @@
...
+			"status", Erroring,
+		)
+		return
+	}
+
+	level.Debug(logger).Log(
+		"name", c.Name(),
+		"msg", c.Summary(),
+		"status", c.Status(),
+		"data", c.Data(),
```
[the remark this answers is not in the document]
directionless: Hrm. That has a downside that complex objects will serialize poorly. I'm going to note it, and figure it out when we get there

pkg/debug/checkups/checkups.go
```diff
@@ -0,0 +1,232 @@
...
+			fmt.Fprintf(w, "\t* %s", n)
+		}
+	}
+
+	if len(failingCheckups) > 0 {
+		fmt.Fprintf(w, "\nCheckups with failures or warnings: ")
+		for _, n := range failingCheckups {
+			fmt.Fprintf(w, "\t* %s", n)
+		}
+	}
```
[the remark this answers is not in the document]
directionless: We print all the checkups as they happen. This is meant as a small summary at the end.

pkg/debug/checkups/checkups.go
```diff
@@ -0,0 +1,232 @@
...
+
+		summaryFH := io.MultiWriter(summaryFlareFH, combinedSummary)
+
+		summaryFH.Write(summary.Bytes())
+	}()
+
+	fullFH := io.Discard
+	if filename := c.ExtraFileName(); filename != "" {
+		var err error
+		fullFH, err = flare.Create(path.Join(c.Name(), filename))
```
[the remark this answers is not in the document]
directionless: You know, I have no idea whether zip expects `path` or `filepath`. But I commented it.

I dunno, maybe the interface is silly, but it felt right. Partly this is in contrast to how the tar side works, which can't just take a `Create(name)`. so I wanted a simple interface and borrowed this

pkg/debug/checkups/logs.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (c *Logs) Run(_ context.Context, fullFH io.Writer) error {
+
+	debugLog := filepath.Join(c.k.RootDirectory(), "debug.json")
+
+	stat, err := os.Stat(debugLog)
+	if err != nil {
+		c.status = Failing
+		c.summary = fmt.Sprintf("failed to stat debug log: %s", err)
+		return nil
```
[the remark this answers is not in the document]
directionless: Because it represents something like no logs, or no permission, which are detectable as data failures. I think most checkups won't have an error state. Hopefully other comment updates help clarify.

pkg/debug/checkups/connectivity.go
directionless: Because it's not plumbed to logs yet, and I'm not sure how I want to structure it

pkg/debug/checkups/root_directory.go
directionless: updates feels like another checkup.

I was thinking about a checkup specific to the DBs. Do they exist, how big, if we have access, what can we say about it?

pkg/debug/checkups/connectivity.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return fmt.Errorf("parsing url(%s): %w", server, err)
+	}
+
+	response, err := client.Get(parsedUrl.String())
+	if err != nil {
+		return fmt.Errorf("fetching url: %w", err)
+	}
+	defer response.Body.Close()
+
```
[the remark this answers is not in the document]
directionless: Yeah, I should improve this checkup

James-Pickett (commented):
Shape of this looks good, feels a lot the log checkup, wonder if we should just merge the 2.

pkg/debug/checkups/root_directory.go
RebeccaMahany: All of that sounds helpful to me, feel free to punt any of that forward to another PR

pkg/debug/checkups/checkups.go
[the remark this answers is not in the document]
RebeccaMahany: The comments are very helpful, thanks for updating!

directionless:
> Shape of this looks good, feels a lot the log checkup, wonder if we should just merge the 2.

Yes. I'd like this to subsume logcheckup. Probably future PR

pkg/debug/checkups/connectivity.go
```diff
@@ -0,0 +1,119 @@
...
+	hosts := map[string]string{
+		"device server":  c.k.KolideServerURL(),
+		"control server": c.k.ControlServerURL(),
+	}
+
+	for n, v := range hosts {
+		fmt.Fprintf(fullFH, "Response from %s / %s:\n", n, v)
+		if err := checkKolideServer(c.k, httpClient, fullFH, v); err != nil {
+			fmt.Fprintf(fullFH, "\n")
+			return fmt.Errorf("%s(%s): %w", n, v, err)
```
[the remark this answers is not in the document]
directionless: Yes, you're both correct. I update connectivity to be better.

pkg/debug/checkups/processes.go
directionless: Hrm. We should have launcher, desktop, and osquery.

Though in some cases we won't have desktop.

I updated to `>= 2` because it's mostly meant as a very coarse check.

RebeccaMahany: Oh, the thing that confused me was the check -- `if strings.Contains(strings.ToLower(exe), "kolide") {` -- I read too quickly and didn't pick up on that being in the directory part of the filepath. This makes sense to me now. 👍

pkg/debug/checkups/binary_directory.go
```diff
@@ -0,0 +1,77 @@
...
+func (c *BinaryDirectory) Summary() string {
+	return c.summary
+}
+
+func (c *BinaryDirectory) Data() any {
+	return nil
+}
+
+// getBinDir returns the platform default binary directory. It should probably get folded into flags, but I'm not
+// quite sure how yet.
```
RebeccaMahany, line 65: Maybe would be nice to have an option where `BinaryDirectory` takes a path as an arg, or can handle more than one directory, because we'd also probably want to do the same for `/var/kolide-k2/<k2 hostname>/updates` once the new autoupdate work is rolled out?

RebeccaMahany (approved):
Fine with this as is once the pipeline is passing 👍

James-Pickett (dismissed):

RebeccaMahany (approved):

Issue #1257, opened 2023-07-19
Move Build & Packaging Secrets into a key manager


directionless:
Today, most of the build and packaging secrets come in via either a local file or an environmental variable. Neither of these feels _good_. Instead, it would be better if we kept them in GCP's secret manager, and then run things with an appropriate service account. (Or as humans, with human credentials)

This would unify secret management for the build/package things.

Probably also help with the general supply chain security. (Cf: SLSA and whatnot)

Doing this requires:
* [x] Terraform to create / setup GCP secret manager
* [x] Manually populating the secrets
* [x] Creating service accounts for GitHub Actions and MacStadium
* [x] Updating `kworker` to use above
* [ ] Updating `porter` to use above

RebeccaMahany:
Putting this aside for now -- the only work remaining is to update porter to use Secret Manager, which is discrete from the other tasks worked on for this issue so far.

directionless:
I think we can close this. While we should update `porter` that's pretty tied to GHA, and distinct from the packaging. 🎉

PR #1258, opened 2023-07-20
Replace altool with notarytool

RebeccaMahany:
Relates to: https://github.com/kolide/launcher/issues/1257

altool will be deprecated this November; this PR updates to use notarytool instead.


pkg/packagekit/applenotarization/applenotarization.go
```diff
@@ -44,79 +41,10 @@ func New(
...
-		}
-
-		if err == nil {
-			return response.NotarizationUpload.RequestUUID, nil
-		}
-	}
-
-	// Falling out of the for loop means we never succeeded
-	return "", errors.New("Did not successfully submit for notarization")
+	return n.runNotarytool(ctx, filePath)
```
James-Pickett, line 47: using Altool we had some retry logic, is notary tool more stable?

RebeccaMahany: I think so? In test, I didn't see it fail, and I took guidance from endpoint-releaser, where we also don't use a retry with notarytool.

James-Pickett (dismissed):

directionless (approved):
Seems fine. Thank you for picking it up

PR #1259, opened 2023-07-21
Don't use hardcoded path to signtool.exe if we can help it

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1257

Coming back to this suggestion from seph since I have some extra time: https://github.com/kolide/launcher/pull/1253#pullrequestreview-1533472602

### DIFF
--- a/pkg/packagekit/package_wix.go
+++ b/pkg/packagekit/package_wix.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"runtime"
 	"strings"
@@ -38,7 +39,7 @@ var wixTemplateBytes []byte
 var assets embed.FS
 
 const (
-	signtoolPath = `C:\Program Files (x86)\Windows Kits\10\bin\10.0.20348.0\x64\signtool.exe`
+	defaultSigntoolPath = `C:\Program Files (x86)\Windows Kits\10\bin\10.0.20348.0\x64\signtool.exe`
 )
 
 func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, includeService bool) error {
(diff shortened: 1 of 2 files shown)


pkg/packagekit/package_wix.go
```diff
@@ -173,6 +175,34 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
...
+		}
+		return nil
+	}); err != nil {
+		return defaultSigntoolPath
+	}
+
+	if signtoolPath == "" {
+		// Not found -- default to well-known version/location
+		return defaultSigntoolPath
+	}
```
directionless, line 201: I think we should error out, instead of hard coding a default. After all, if we don't find that walking, something is real fishy....

pkg/packagekit/package_wix.go
```diff
@@ -173,6 +175,34 @@ func PackageWixMSI(ctx context.Context, w io.Writer, po *PackageOptions, include
...
+func getSigntoolPath() string {
+	var signtoolPath string
+	root := `C:\Program Files (x86)\Windows Kits` // restrict our lookup to a well-known location
+	fileSystem := os.DirFS(root)
+
+	if err := fs.WalkDir(fileSystem, ".", func(currentPath string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if !d.IsDir() && strings.HasSuffix(currentPath, `x64/signtool.exe`) {
```
directionless, line 190: This finds the first one, do you think we need to try to find the highest version?

pkg/packagekit/package_wix.go
RebeccaMahany: That's reasonable! I will update.

pkg/packagekit/package_wix.go
RebeccaMahany: I think that's a good idea -- I can do that

RebeccaMahany: Installed two versions of signtool locally so I could test -- this appears to work fine. I don't love that it's string comparison, but 🤷‍♀️

James-Pickett (approved):

PR #1260, opened 2023-07-22
Tidy Flare and Doctor

directionless:
Some updates to how flare and doctor work.

1. This unifies the enumeration of the checkups. I opted to put it here, with a bitmask, because it felt reasonably ergonomic.
2. creates a `writeSummary` helper to enforce a consist print output. 
3. adds something to flare so we can tell if it's a command line, or in-situ flare.


pkg/debug/checkups/checkups.go
```diff
@@ -40,7 +43,7 @@ const (
 	Failing       Status = "Failing"       // Checkup is failing
 )
 
-func CharFor(s Status) string {
+func charFor(s Status) string {
```
directionless, line 46: I think this would be cleaner as `func (s Status) Emoji() string`

James-Pickett (approved):

PR #1262, opened 2023-07-24
add random suffix to desktop socket

James-Pickett:


directionless (dismissed):
I'm happy to try, but I think we need some deeper rethink

James-Pickett:
> I'm happy to try, but I think we need some deeper rethink

I updated it to use a random suffix on the socket path.

ee/desktop/runner/runner.go
```diff
@@ -671,7 +672,11 @@ func (r *DesktopUsersProcessesRunner) socketPath(uid string) (string, error) {
 		return "", fmt.Errorf("chowning user folder: %w", err)
 	}
 
-	path := filepath.Join(userFolderPath, "kolide_desktop.sock")
+	rand.Seed(time.Now().UnixNano())
```
RebeccaMahany, line 675: It looks like rand.Seed has updated in 1.20 and is now deprecated:

> The [math/rand](https://tip.golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://tip.golang.org/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).

Should we update to use `rand.New...`?

I was looking at:

https://pkg.go.dev/math/rand@master#Seed
https://tip.golang.org/doc/go1.20

James-Pickett: good catch, since it's seeded automatically no need to call new, can just use top level func

RebeccaMahany (approved):

RebeccaMahany (approved):

PR #1271, opened 2023-07-28
bump timeout on mdfind from 30 to 60 seconds

James-Pickett:
We've had some issues with timeouts using the kolide_spotlight table. This bumps the time out from 30 to 60 seconds. This is a bandaid and we should consider updating checks that are looking for files to use osquery mdfind table that uses a C api instead of execing.

I experimented with piping the cmd output to a scanner to start populating the response when each line came in instead of waiting for the command to fully complete, but there was no notable difference in the time it took.


directionless (approved):
Feels weird, but okay

Issue #1205, opened 2023-05-22
Our run group idiom is confusing


directionless:
Digging through a debug log, I find myself wondering about https://github.com/kolide/launcher/blob/7217b1944ab23a72191bcf1b035b4579e1ff60f7/cmd/launcher/internal/updater/updater.go#L143-L158

This is the interrupt function for the updater. `u.stopExecution()` makes sense, but signaling `u.stopChan <- true` feels weird. That stops the main launcher thread, which should have it's own interrupt function. I might be missing something clever, but my hunch is that it's superfluous.

RebeccaMahany:
Investigated -- `u.stopChan` is a separate channel from the `sigChannel` that gets passed in, so signaling `u.stopChan` only stops the updater from attempting to init an autoupdater if it hasn't yet, and does not stop the main launcher thread.

Discussed offline the need for better logging and reporting for when our many rungroups exit.

RebeccaMahany:
Some specific wishlist items while working through https://github.com/kolide/launcher/pull/1272:

- [ ] The rungroup replacement should be able to identify actors, and log which actor initiated the shutdown and what the error was
- [ ] We should audit all the `Interrupt` functions to remove superfluous logging of the error because they really muck up the logs
- [ ] The rungroup replacement should have a timeout on waiting for all actors to `Interrupt` successfully, so that we aren't stuck forever waiting for a broken `Interrupt => Execute` return
- [ ] The rungroup replacement should log when an actor cannot be interrupted successfully

PR #1273, opened 2023-08-01
Remove superfluous logs in actor.Interrupt

RebeccaMahany:

A bunch of our rungroup actors log errors in their `Interrupt` functions that did not originate from them. This creates a lot of noise in the logs. Launcher already logs the originating error on exit:

```
{"caller":"locallogger.go:43","level":"debug","run service: fake error from signal notify":"run launcher","stack":"run service: fake error from signal notify","ts":"2023-08-01T16:44:11.189798Z"}
{"caller":"locallogger.go:43","level":"info","run service: fake error from signal notify":"run launcher","ts":"2023-08-01T16:44:11.189913Z"}
```

This PR removes all other extra instances of this log to make shutdown a little less noisy and a little easier to understand.

### DIFF
--- a/cmd/launcher/extension.go
+++ b/cmd/launcher/extension.go
@@ -174,9 +174,7 @@ func createExtensionRuntime(ctx context.Context, k types.Knapsack, launcherClien
 					<-ctx.Done()
 					return nil
 				},
-				Interrupt: func(err error) {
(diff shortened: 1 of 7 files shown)


directionless (approved):
I always wonder about the intent behind `Interrupt(err error)` as a signature.

PR #1274, opened 2023-08-01
only lock mutex while accessing auth map on runner server

James-Pickett:


ee/desktop/runner/server/server.go
```diff
@@ -134,6 +133,11 @@ func (ms *RunnerServer) authMiddleware(next http.Handler) http.Handler {
 			}
 		}
 
+		// not using defer here because this function
+		// calls the next function rather than returing,
+		// this blocks mutex for too long
+		ms.mutex.Unlock()
```
directionless, line 139: I think this is okay -- there's no return earlier here. But at 13 lines between the Lock and the Unlock, I feel nervous about it surviving a refactor. Do you think this makes sense as-is, or can we push this section into a new method with a deferred unlock?

James-Pickett: good call, moved to separate func with defer unlock

directionless (approved):

RebeccaMahany (approved):

PR #1278, opened 2023-08-02
add retry logic to getting all target metadata form notary

James-Pickett:
adds retry logic to handle cases where notary doesn't return metadata such as
https://github.com/kolide/launcher/actions/runs/5731222088/job/15531675282


directionless (approved):
I know this is a draft, but it looks right!

RebeccaMahany (approved):

Issue #1276, opened 2023-08-02
Launcher probably broken on m1 machines running Big Sur


directionless:
Customer [report](https://kolide.slack.com/archives/CFPMXV7CZ/p1689635933981329) that launcher isn't working on Big Sur:
```
dyld: Symbol not found: _kIOMainPortDefault
 Referenced from: /usr/local/kolide-k2/bin/launcher (which was built for Mac OS X 12.0)
 Expected in: /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit
zsh: abort /usr/local/kolide-k2/bin/launcher doctor"
```

which I suspect might be the result of #1151

RebeccaMahany:
Related: https://github.com/kolide/launcher/pull/1133, https://github.com/kolide/launcher/pull/1150

directionless:
Hrm. At least a a first blush, macos-11, seems to be able to run this okay:
https://github.com/kolide/launcher/actions/runs/5742081124/job/15563848822

directionless:
And https://github.com/kolide/launcher/actions/runs/5746094914/job/15575064966 is a `doctor` run on macos-11

directionless:
I have not yet been able to replicate this.

https://github.com/kolide/launcher/pull/1279 brings in all the GHA runners for smoke tests, and macos-11 passed fine -- https://github.com/kolide/launcher/actions/runs/5751156346/job/15589880445

I setup a laptop with Big Sur, and running both the Big Sur beta, and 11.7.9, I can run the released launcher:
```
% ./usr/local/kolide-k2/bin/launcher --version 
{"caller":"main.go:30","msg":"Launcher starting up","revision":"6a1bfb213ab66c497645c865d19114c936af14ce","severity":"info","ts":"2023-08-03T14:26:19.275612Z","version":"1.0.12"}
launcher - version 1.0.12
  branch: 	HEAD
  revision: 	6a1bfb213ab66c497645c865d19114c936af14ce
  build date: 	2023-06-23
  build user: 	runner (runner)
  go version: 	go1.20.5
```

directionless:
https://github.com/pulumi/pulumi/issues/11152 discusses a similar problem, and using a suggested otool command I can see:
```
$ otool -l usr/local/kolide-k2/Kolide.app/Contents/MacOS/launcher | grep -A 3 min
    minos 12.0
      sdk 13.1
   ntools 1
     tool 3
--
    minos 12.0
      sdk 13.1
   ntools 1
[...]

PR #1282, opened 2023-08-03
Set `MACOSX_DEPLOYMENT_TARGET` on builds

directionless:
Fixes: #1276

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -64,6 +64,13 @@ jobs:
 
     - name: Build
       run: make -j2 github-build
+      env:
+        MACOSX_DEPLOYMENT_TARGET: 13 # FIXME
+
+    - name: Check macOS minumum Version
+      if: contains(matrix.os, 'macos')
+      # this uses grep's exit code
+      run: otool -l build/launcher | grep "minos 11"
 
     - name: Lipo
       run: make github-lipo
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,6 +5,11 @@ all: build
 GOPATH ?= $(HOME)/go
 PATH := $(GOPATH)/bin:$(PATH)
 
+
+# Set the target version of macOS. This is an ENV that is consumed by go build (and downstream c pieces)
+# It's set here, and not in the GitHub Actions tooling, so that we notice build warnings.
+export MACOSX_DEPLOYMENT_TARGET = 11
```
directionless, line 11: I can't decide if this is better in the Makefile, or the workflow. I opted to put it here, on the theory that our normal builds should follow as much of the same path.

_But_ it's a small lie -- changes to this don't rebuild cached files. So if you _really_ want it, you need to also run `go clean --cache`

pkg/make/builder.go
```diff
@@ -582,6 +588,10 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 			return err
 		}
 
+		if stderr.Len() > 0 {
+			return errors.New("stderr not empty")
```
James-Pickett, line 592: should we add the contents of stderr to the error?

directionless: I decided not too -- the STDERR might be hundreds of lines. And since it's written to `os.Stderr` I don't think it's adds much here

James-Pickett (approved):

PR #1284, opened 2023-08-07
Do not perform osquery healthchecks while system is sleeping

RebeccaMahany:
We occasionally see an issue where launcher shuts itself down on Windows while the device is sleeping, and fails to be restarted successfully. Our best guess is that while in modern standby, launcher is throttled and the osquery healthcheck therefore fails enough to shut down launcher, and the throttling prevents launcher from starting back up. This PR disables the osquery healthcheck while the system is sleeping.


James-Pickett (dismissed):

pkg/agent/flags/keys/keys.go
```diff
@@ -47,6 +47,7 @@ const (
 	LogIngestServerURL         FlagKey = "log_ingest_url"
 	TraceIngestServerURL       FlagKey = "trace_ingest_url"
 	DisableTraceIngestTLS      FlagKey = "disable_trace_ingest_tls"
+	DisableOsqueryHealthchecks FlagKey = "disable_osquery_healthchecks"
```
directionless, line 50: Thoughts about whether should do this as `DisableOsqueryHealthchecks` vs `PowersaveMode`?

directionless (approved):
I'd take it as is, or called `PowersaveMode`. I don't know that I know which feel more correct yet.

pkg/agent/flags/keys/keys.go
RebeccaMahany: I had gone back and forth on similar too, thinking about this versus something like `InModernStandby`. I ended up on this one because the osquery extension doesn't need to know why it should skip healthchecks, just what to do -- but I do not feel that strongly about it.

PR #1285, opened 2023-08-07
Initial investigation into repcli

directionless:
We'd like to build a table to parse Carbon Black status. It sounds like the correct way to do this is to examine the output of `repcli status`. Since we don't know what to expect, ideally we could to capture that output from some devices in the field. Conveniently, we have a table designed to help with exactly that. 

This adds the appropriate `repcli` commands to our allow list. Hopefully we can get some initial data, and then build a table around it


pkg/osquery/tables/dev_table_tooling/commands_windows.go
```diff
@@ -3,9 +3,15 @@
...
 
+import "path/filepath"
+
 var allowedCommands = map[string]allowedCommand{
 	"echo": {
 		binPaths: []string{"echo"},
 		args:     []string{"hello"},
 	},
+	"cb_repcli": {
+		binPaths: []string{filepath.Join("Program Files", "Confer", "repcli")},
```
RebeccaMahany, line 14: This doesn't need `C:` prepended?

RebeccaMahany (approved):
LGTM, just the one question about the path

pkg/osquery/tables/dev_table_tooling/commands_windows.go
directionless: I don't think so? Go seems pretty chill about that

PR #1287, opened 2023-08-08
Set timeouts on notary requests during build

RebeccaMahany:
We added a retry in https://github.com/kolide/launcher/pull/1278, but I saw in a [recent job](https://github.com/kolide/launcher/actions/runs/5799484889/job/15719583154) that we were only trying 1 request:

```
ts=2023-08-08T16:30:47.947216363Z
caller=logutil.go:13
severity=info
msg="Target Failed"
err="bootstrap notary GUN kolide/launcher: timeout after 5m0s (1 attempts): getting all target metadata: unable to reach trust server at this time: 500."
target=generate-tuf
```

It looks like this is because we don't have a timeout set. This PR updates the transport we use when talking to notary to set some timeouts.

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -15,6 +15,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


pkg/make/builder.go
```diff
@@ -441,7 +442,7 @@ func (b *Builder) execBindata(ctx context.Context, dir string) error {
 	return nil
 }
 
-func bootstrapFromNotary(notaryConfigDir, remoteServerURL, localRepo, gun string) error {
+func bootstrapFromNotary(notaryConfigDir, remoteServerURL, localRepo, gun string, retryTimeout time.Duration, bootstrapTimeout time.Duration) error {
```
directionless, line 445: I'd hard code the values instead of passing. But whatev

directionless (approved):

pkg/make/builder.go
RebeccaMahany: I did this so the test wouldn't take 5min to run

PR #1288, opened 2023-08-09
Add checkups for installation troubleshooting + general Windows troubleshooting

RebeccaMahany:
macOS:
* Grabs installation logs

Windows:
* Get all running services
* Get status and config for Launcher service
* Get recent service manager event logs for Launcher service
* Generate systempowerreport

### DIFF
--- a/pkg/debug/checkups/install.go
+++ b/pkg/debug/checkups/install.go
@@ -13,7 +13,7 @@ type installCheckup struct {
 }
 
 func (i *installCheckup) Name() string {
-	return "Installation"
+	return "Package Install Logs"
 }
 
 func (i *installCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
@@ -21,7 +21,7 @@ func (i *installCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
 	defer extraZip.Close()
 
 	if err := gatherInstallationLogs(extraZip); err != nil {
-		return fmt.Errorf("gathering installation logss: %w", err)
+		return fmt.Errorf("gathering installation logs: %w", err)
 	}
 
 	return nil
@@ -49,9 +49,9 @@ func gatherInstallationLogs(z *zip.Writer) error {
 		return nil
 	}
 
-	out, err := z.Create("install.log")
+	out, err := z.Create("macos-var-log-install.log")
 	if err != nil {
-		return fmt.Errorf("creating install.log in zip: %w", err)
+		return fmt.Errorf("creating macos-var-log-install.log in zip: %w", err)
 	}
 
 	installLog, err := os.Open("/var/log/install.log")
@@ -60,12 +60,7 @@ func gatherInstallationLogs(z *zip.Writer) error {
 	}
 	defer installLog.Close()
 
-	installLogRaw, err := io.ReadAll(installLog)
-	if err != nil {
-		return fmt.Errorf("reading /var/log/install.log: %w", err)
-	}
-
-	if _, err := out.Write(installLogRaw); err != nil {
+	if _, err := io.Copy(out, installLog); err != nil {
 		return fmt.Errorf("writing /var/log/install.log contents to zip: %w", err)
 	}
 
--- a/pkg/debug/checkups/install.go
+++ b/pkg/debug/checkups/install.go
@@ -1,0 +1,73 @@
+package checkups
+
+import (
+	"archive/zip"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"runtime"
+)
+
+type installCheckup struct {
+}
+
+func (i *installCheckup) Name() string {
+	return "Installation"
+}
+
+func (i *installCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
+	extraZip := zip.NewWriter(extraWriter)
+	defer extraZip.Close()
+
+	if err := gatherInstallationLogs(extraZip); err != nil {
+		return fmt.Errorf("gathering installation logss: %w", err)
+	}
+
+	return nil
+
+}
+
+func (i *installCheckup) ExtraFileName() string {
+	return "install.zip"
+}
+
+func (i *installCheckup) Status() Status {
+	return Informational
+}
+
+func (i *installCheckup) Summary() string {
+	return "N/A"
+}
+
+func (i *installCheckup) Data() any {
+	return nil
+}
+
+func gatherInstallationLogs(z *zip.Writer) error {
+	if runtime.GOOS == "windows" || runtime.GOOS == "linux" {
+		return nil
+	}
+
+	out, err := z.Create("install.log")
+	if err != nil {
+		return fmt.Errorf("creating install.log in zip: %w", err)
+	}
+
+	installLog, err := os.Open("/var/log/install.log")
+	if err != nil {
+		return fmt.Errorf("opening /var/log/install.log: %w", err)
+	}
+	defer installLog.Close()
+
+	installLogRaw, err := io.ReadAll(installLog)
+	if err != nil {
+		return fmt.Errorf("reading /var/log/install.log: %w", err)
+	}
+
+	if _, err := out.Write(installLogRaw); err != nil {
+		return fmt.Errorf("writing /var/log/install.log contents to zip: %w", err)
+	}
+
+	return nil
+}
--- a/pkg/debug/checkups/power_windows.go
+++ b/pkg/debug/checkups/power_windows.go
@@ -23,9 +23,7 @@ func (p *powerCheckup) Name() string {
 func (p *powerCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
 	// Create a temporary file for powercfg to write its output to
 	tmpFilePath := agent.TempPath("launcher-checkup-spr.html")
-	defer func() {
-		os.Remove(tmpFilePath)
-	}()
+	defer os.Remove(tmpFilePath)
 
 	// See: https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/powercfg-command-line-options#option_systempowerreport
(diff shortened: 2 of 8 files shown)


pkg/debug/checkups/install.go
```diff
@@ -0,0 +1,73 @@
...
+	"io"
+	"os"
+	"runtime"
+)
+
+type installCheckup struct {
+}
+
+func (i *installCheckup) Name() string {
+	return "Installation"
```
directionless, line 16:
```suggestion
	return "Package Install Logs"
```

pkg/debug/checkups/install.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (i *installCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
+	extraZip := zip.NewWriter(extraWriter)
+	defer extraZip.Close()
+
+	if err := gatherInstallationLogs(extraZip); err != nil {
+		return fmt.Errorf("gathering installation logss: %w", err)
```
directionless, line 24:
```suggestion
		return fmt.Errorf("gathering installation logs: %w", err)
```

pkg/debug/checkups/install.go
```diff
@@ -0,0 +1,73 @@
...
+func (i *installCheckup) Data() any {
+	return nil
+}
+
+func gatherInstallationLogs(z *zip.Writer) error {
+	if runtime.GOOS == "windows" || runtime.GOOS == "linux" {
+		return nil
+	}
+
+	out, err := z.Create("install.log")
```
directionless, line 52:
```suggestion
	out, err := z.Create("macos-var-log-install.log")
```

pkg/debug/checkups/install.go
```diff
@@ -0,0 +1,73 @@
...
+		return fmt.Errorf("creating install.log in zip: %w", err)
+	}
+
+	installLog, err := os.Open("/var/log/install.log")
+	if err != nil {
+		return fmt.Errorf("opening /var/log/install.log: %w", err)
+	}
+	defer installLog.Close()
+
+	installLogRaw, err := io.ReadAll(installLog)
```
directionless, line 63: Maybe copy instead of readall

pkg/debug/checkups/services.go
```diff
@@ -0,0 +1,224 @@
...
+const kolideSvcName = "LauncherKolideK2Svc"
+
+type servicesCheckup struct {
+	data         map[string]any
+	serviceState svc.State
+}
+
+func (s *servicesCheckup) Name() string {
+	return "Service Report"
+}
```
directionless, line 23:
```suggestion
func (s *servicesCheckup) Name() string {
	if runtime.GOOS != "windows" {
		return ""
	}
	return "Service Report"
}
```

directionless: Though given that svc only exist on windows, we're going to need to build constrain this anyhow :|

pkg/debug/checkups/services.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return fmt.Errorf("opening service: %w", err)
+	}
+	defer serviceHandle.Close()
+
+	s.data["service_name"] = kolideSvcName
+
+	// Get status
+	if status, err := serviceHandle.Query(); err != nil {
+		s.data["state"] = fmt.Sprintf("error: %v", err)
```
directionless, line 48: Could also populate this into the Summary

pkg/debug/checkups/services.go
```diff
@@ -0,0 +1,224 @@
...
+	} else {
+		// Map state (uint32) to human-readable statuses
+		s.serviceState = status.State
+		s.data["state_raw"] = status.State
+		s.data["state"] = stateHumanReadable(status.State)
+	}
+
+	// Get config
+	if cfg, err := serviceHandle.Config(); err != nil {
+		s.data["config"] = fmt.Sprintf("error: %v", err)
```
directionless, line 58: Same

RebeccaMahany: i'm not including config info in the summary right now, so i'm not sure how to fit this in gracefully. is there config info that you think should go in the summary?

directionless: I wasn't thinking about any info, but failing to parse it seems like it's probably an error condition worth noting.

pkg/debug/checkups/power_windows.go
```diff
@@ -0,0 +1,64 @@
+//go:build windows
+// +build windows
```
directionless, line 2: This is fine, but we don't need to build tag -- can switch in the `Name()`. Mostly depends what feels easier

pkg/debug/checkups/power_windows.go
```diff
@@ -0,0 +1,64 @@
...
+func (p *powerCheckup) Name() string {
+	return "Power Report"
+}
+
+func (p *powerCheckup) Run(ctx context.Context, extraWriter io.Writer) error {
+	// Create a temporary file for powercfg to write its output to
+	tmpFilePath := agent.TempPath("launcher-checkup-spr.html")
+	defer func() {
+		os.Remove(tmpFilePath)
+	}()
```
directionless, line 28:
```suggestion
	defer os.Remove(tmpFilePath)
```

maybe?

pkg/debug/checkups/power_windows.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// See: https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/powercfg-command-line-options#option_systempowerreport
+	powerCfgCmd := exec.CommandContext(ctx, "powercfg.exe", "/systempowerreport", "/output", tmpFilePath)
+	powerCfgCmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true} // prevents spawning window
```
directionless, line 32: nice

pkg/debug/checkups/power_windows.go
```diff
@@ -0,0 +1,64 @@
...
+	"os/exec"
+	"syscall"
+
+	"github.com/kolide/launcher/pkg/agent"
+)
+
+type powerCheckup struct{}
+
+func (p *powerCheckup) Name() string {
+	return "Power Report"
```
directionless, line 20: If this is destined to only be windows, I think we should include that in the name.

pkg/debug/checkups/services_windows.go
```diff
@@ -0,0 +1,285 @@
...
+	default:
+		return fmt.Sprintf("unknown: %d", serviceType)
+	}
+}
+
+func startTypeHumanReadable(startType uint32) string {
+	// For mapping, see https://pkg.go.dev/golang.org/x/sys/windows/svc/mgr#pkg-constants
+	// and https://pkg.go.dev/golang.org/x/sys@v0.11.0/windows#SERVICE_BOOT_START
+	switch startType {
+	case 0:
```
directionless, line 160: These are constants in the windows package, should we use them? `case windows.SERVICE_BOOT_START` or something?

pkg/debug/checkups/services_windows.go
```diff
@@ -0,0 +1,285 @@
...
+
+func gatherServiceManagerEventLogs(ctx context.Context, z *zip.Writer) error {
+	cmdletArgs := []string{
+		"Get-EventLog",
+		"-Newest", "50",
+		"-LogName", "System",
+		"-Source", "\"Service Control Manager\"",
+		"-Message", fmt.Sprintf("*%s*", kolideSvcName),
+		"|",
+		"Format-Table", "-Wrap", "-AutoSize", // ensure output doesn't get truncated
```
directionless, line 237: Does this actually expand the messages? I know I've had trouble with that before.

There may also be a to json tool

pkg/debug/checkups/services_windows.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	getEventLogCmd := exec.CommandContext(ctx, "powershell.exe", cmdletArgs...)
+	getEventLogCmd.SysProcAttr = &syscall.SysProcAttr{HideWindow: true} // prevents spawning window
+	out, err := getEventLogCmd.CombinedOutput()
```
directionless, line 242: You might be able to create the writer, and then set `getEventLogCmd.Stdout` and getEventLogCmd.Stderr` to the writer directly.

pkg/debug/checkups/power_windows.go
RebeccaMahany: We do have to use the build tag for this, unfortunately -- before I separated them out and added the tag: https://github.com/kolide/launcher/actions/runs/5813681186/job/15761773747

pkg/debug/checkups/power_windows.go
RebeccaMahany: I thought maybe we'd want to pull something similar for macos/linux in the future? And we probably don't need to include the OS for the device in the checkup name

pkg/debug/checkups/services_windows.go
RebeccaMahany: It does, yes

pkg/debug/checkups/power_windows.go
RebeccaMahany: Oh yup, no good reason to do it the way I have it now, I'll update

pkg/debug/checkups/services_windows.go
RebeccaMahany: Oh yup that would improve readability a lot -- thank you, I'll update

pkg/debug/checkups/services_windows.go
RebeccaMahany: Yup, that works! Updated

pkg/debug/checkups/services.go
RebeccaMahany: Added

directionless (approved):
Woo! LGTM

PR #1289, opened 2023-08-14
Custom URL handler for Windows

RebeccaMahany:


directionless:
I think we've decided not to go this way. Though we might revisit

PR #1291, opened 2023-08-17
Remove noisy log + prevent panic after parse

RebeccaMahany:
Prevents filling logs with `error parsing log ingest server url, shipping disabled` when shipping is disabled on purpose via sending down an empty URL.

Also prevents a potential panic trying to access `parsedUrl.String()` when parsedUrl can be nil in the event of an actual parse error.

### DIFF
--- a/pkg/log/logshipper/logshipper.go
+++ b/pkg/log/logshipper/logshipper.go
@@ -49,7 +49,7 @@ func New(k types.Knapsack, baseLogger log.Logger) *LogShipper {
 		sender:         sender,
 		sendBuffer:     sendBuffer,
 		shippingLogger: shippingLogger,
-		baseLogger:     baseLogger,
+		baseLogger:     log.With(baseLogger, "component", "logshipper"),
 		knapsack:       k,
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/log/logshipper/logshipper.go
```diff
@@ -64,24 +64,30 @@ func (ls *LogShipper) Ping() {
...
-		shouldEnable = false
-		level.Debug(ls.baseLogger).Log(
-			"msg", "error parsing log ingest server url, shipping disabled",
-			"err", err,
-			"log_ingest_url", ls.knapsack.LogIngestServerURL(),
-		)
+
+	// Only attempt to parse a URL if we have one
+	if shouldEnable {
+		parsedUrl, err := url.Parse(ls.knapsack.LogIngestServerURL())
```
directionless, line 70: I'm surprised you need this. 

https://go.dev/play/p/P2NKDl7t9U_q makes it look like parsing `""` will return empty url object. But I guess I'm not sure how to best handle that. So maybe this is good enough

pkg/log/logshipper/logshipper.go
```diff
@@ -64,24 +64,30 @@ func (ls *LogShipper) Ping() {
...
+			// If we have a bad endpoint, just disable for now.
+			// It will get renabled when control server sends a
+			// valid endpoint.
+			shouldEnable = false
+			level.Debug(ls.baseLogger).Log(
+				"msg", "error parsing log ingest server url, shipping disabled",
+				"err", err,
+				"log_ingest_url", ls.knapsack.LogIngestServerURL(),
+			)
+		} else if parsedUrl != nil {
```
directionless, line 81: `|| u.String() == ""` maybe? But that feels kinda gross

directionless (dismissed):

pkg/log/logshipper/logshipper.go
RebeccaMahany: It will return an empty URL object (and will therefore not panic on parsedUrl.String), but I figured no reason to perform the parse if we don't need to

pkg/log/logshipper/logshipper.go
RebeccaMahany: I didn't want to do that because `parsedUrl.String()` will panic when the parsedUrl is nil, and I believe it's possible for parsing to return nil without also returning an error per the docs:

> Trying to parse a hostname and path without a scheme is invalid but may not necessarily return an error, due to parsing ambiguities. 

So this is a safer check

RebeccaMahany: Oh sorry I misread your comment!

pkg/log/logshipper/logshipper.go
RebeccaMahany: Simplified this

directionless (approved):

pkg/log/logshipper/logshipper.go
```diff
@@ -63,22 +63,22 @@ func (ls *LogShipper) Ping() {
 	token, _ := ls.knapsack.TokenStore().Get(storage.ObservabilityIngestAuthTokenKey)
 	ls.sender.authtoken = string(token)
 
-	shouldEnable := ls.knapsack.LogIngestServerURL() != ""
 	parsedUrl, err := url.Parse(ls.knapsack.LogIngestServerURL())
-	if err != nil || parsedUrl.String() == "" {
+	if err != nil {
```
James-Pickett, line 67: thank you for fixing!

PR #1292, opened 2023-08-17
Make flaky test less flaky

RebeccaMahany:
Addresses test failure: https://github.com/kolide/launcher/actions/runs/5895454742/job/15991262027?pr=1291

As long as we're making a decent number of attempts, this is fine.


pkg/make/builder_test.go
```diff
@@ -275,6 +276,9 @@ func Test_bootstrapFromNotary_retryOnTimeout(t *testing.T) {
 	err := bootstrapFromNotary(t.TempDir(), testNotaryServer.URL, t.TempDir(), testGun, 1*time.Second, 10*time.Second)
 	require.NotNil(t, err, "expected timeout during bootstrap")
 
-	// Confirm we made the expected number of attempts
-	require.Contains(t, err.Error(), "timeout after 10s (10 attempts)")
+	// Confirm we made (more or less) the expected number of attempts
+	eightOrMoreAttempts := strings.Contains(err.Error(), "timeout after 10s (8 attempts)") ||
```
directionless, line 280: Would it be simpler to `strings.Contains(err.Error(), "timeout after 10s")` ?

RebeccaMahany: The motivation for the original change was that we were only ever making 1 attempt (because we didn't have a timeout set on our request to notary), so I wanted to validate in this test that we were actually timing out and making multiple attempts.

directionless: okay then

directionless (approved):

PR #1294, opened 2023-08-18
Add flag to delay osquery healthchecks; set opts on instance after restart

RebeccaMahany:
Delays osquery healthchecks for a default of 10 minutes, in case socket contention causes issues on startup.

Fixes issue where knapsack (and other OsqueryInstanceOptions) would not be applied to the new instance after restart.


directionless (approved):

pkg/osquery/runtime/runner.go
```diff
@@ -111,6 +112,9 @@ func (r *Runner) Start() error {
 			opts := r.instance.opts
 			r.instance = newInstance()
 			r.instance.opts = opts
+			for _, opt := range r.opts {
+				opt(r.instance)
+			}
```
directionless, line 117: hahahahahahaha. oops. 🤦

directionless (approved):

directionless:
This relates to #1211

PR #1302, opened 2023-08-24
adds launcher flags to flare

James-Pickett:
summary.log:

```
✅	Launcher Flags: launcher flags exists and are parsable
```

launcherFlags.log:

```
flags file path: /private/etc/kolide-k2/launcher.flags
flags file is 308 bytes, and was last modified at 2023-07-10 13:31:30.430633404 -0700 PDT

launcher.flags contents:

with_initial_runner
autoupdate
hostname k2device-preprod.kolide.com
root_directory /var/kolide-k2/k2device-preprod.kolide.com
osqueryd_path /usr/local/kolide-k2/bin/osqueryd
enroll_secret_path /etc/kolide-k2/secret
update_channel nightly
transport jsonrpc
autoupdate_interval 1m
autoupdater_initial_delay 1s

```


pkg/debug/checkups/launcher_flags.go
```diff
@@ -0,0 +1,85 @@
...
+func (lf *launcherFlags) ExtraFileName() string {
+	return "launcherFlags.log"
+}
+
+func (lf *launcherFlags) Data() any {
+	return nil
+}
+
+func (lf *launcherFlags) flagsFilePath() string {
+	if runtime.GOOS != "windows" {
```
directionless, line 79: I'd put the windows in the if block, but 🤷

pkg/debug/checkups/launcher_flags.go
```diff
@@ -0,0 +1,85 @@
...
+
+	file, err := os.Open(configFilePath)
+	if err != nil {
+		lf.summary = fmt.Sprintf("failed to open %s: %s", configFilePath, err)
+		return nil
+	}
+
+	fmt.Fprint(extraFh, "Parsed Flags (boolean flags appear as {KEY} TRUE):\n\n")
+
+	if err := ff.PlainParser(file, func(name, value string) error {
```
directionless, line 47: I'm not sure how easy it is to do better -- this might be a good enough for now thing, but this isn't going to be the same parser as [`options.go`](https://github.com/kolide/launcher/blob/main/cmd/launcher/options.go) -- it accepts any options, and won't detect skew

pkg/debug/checkups/launcher_flags.go
```diff
@@ -0,0 +1,85 @@
...
+		// print each flag and value to file handler
+		fmt.Fprint(extraFh, fmt.Sprintf("%s %s\n", name, value))
```
directionless, line 49: This is an interesting approach. Do you know if we lose information vs just copying the file?

pkg/debug/checkups/launcher_flags.go
directionless: Actually, reviewing `parseOptions` I think you can just call it

directionless (commented):
I think it's worth using our options parser

pkg/debug/checkups/launcher_flags.go
directionless: Maybe both'ish? copy the file, and dump the resultant options through data?

pkg/debug/checkups/launcher_flags.go
James-Pickett: parseOptions lives in the main, cannot reference main outside of main, so I'm looking at moving all the options stuff to it's own pkg

pkg/debug/checkups/launcher_flags.go
James-Pickett: now just dumping the file to the log, easy enough

pkg/debug/checkups/launcher_flags.go
James-Pickett: moved to own pkg, now using ParseOptions to verify flags

pkg/debug/checkups/launcher_flags.go
James-Pickett: I agree ... better readability to condition on a positive

pkg/debug/checkups/launcher_flags.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil
+}
+
+func (lf *launcherFlags) flagsFilePath() string {
+	if runtime.GOOS == "windows" {
+		return filepath.Join("C:\\Program Files\\Kolide\\Launcher-kolide-k2\\conf", "launcher.flags")
+	}
+
+	// non-windows
+	return "/private/etc/kolide-k2/launcher.flags"
```
RebeccaMahany, line 84: I don't think you can assume that `/private` exists on Linux (I ran into this here -- https://github.com/kolide/launcher/pull/1231#discussion_r1237760751)

James-Pickett: good catch, updated

RebeccaMahany (approved):
LGTM!

directionless (approved):
I didn't look too hard, but it seems fine. I assume it still works

PR #1303, opened 2023-08-25
Add callback support to `kryptoEcMiddleware`

directionless:
## Problem

Sometimes we see a browser able to make requests to `http://localhost`, but unable to meaningfully work with the response. This breaks agent detection in the device trust flow.

## Solution

One solution is to setup a simple callback. The intent here, is to allow K2 to set a URL to post back to, and then `kryptoEcMiddleware` can post the response to it. 

This should be reasonable secure -- the callback URL is inside the krypto box, and the response is the same as what we'd send via PNG. eg: it's all opaque crypto data. There are a couple of small exceptions...
1. We include the user agent headers
2. we send errors

I first tried implementing this in middleware, but it was messy. First, because the url comes out of the box, so that has to be passed from the _inner_ layer up to the outer layer. Second, because the outer layer works with readers, so there's a bunch of extra io teeing. It's much simpler this way.


directionless:
@James-Pickett This is in draft since the corresponding K2 side isn't merged, but how's the shape look to you?

James-Pickett (dismissed):
This code looks good to me. Would like a little clarification on the goal of this thought, is it just so we can report errors back up to k2?

James-Pickett (approved):

Issue #1295, opened 2023-08-21
Dependency Building Fails on Arch Linux


prikhi:
Hi, with the latest `main` branch, I'm unable to build the the `kolide-launcher-git` AUR package. It fails during the `make deps` command:

```sh
$ go version
go version go1.21.0 linux/amd64
$ make deps
go run cmd/make/make.go -targets=deps-go,install-tools
ts=2023-08-21T03:28:56.354696228Z caller=logutil.go:13 severity=info msg="Target Failed" err="decode go list output: EOF" target=install-tools
exit status 1
make: *** [Makefile:227: deps-go] Error 1
$ go list -tags tools -json ./pkg/tools/
pkg/tools/tools.go:14:2: import "github.com/go-bindata/go-bindata/go-bindata" is a program, not an importable package
```

Changing the above line's URL to include a trailing slash lets `make deps` & `make` complete & for the package to build & install correctly, after applying this diff & building with `makepkg -e`:

```diff
diff --git a/pkg/tools/tools.go b/pkg/tools/tools.go
index 812a8208..123d1491 100644
--- a/pkg/tools/tools.go
+++ b/pkg/tools/tools.go
@@ -11,5 +11,5 @@ For additional documentation on the topic:
 package tools
 
 import (
-	_ "github.com/go-bindata/go-bindata/go-bindata"
+	_ "github.com/go-bindata/go-bindata/go-bindata/"
 )
```

directionless:
Thank you!

PR #1304, opened 2023-08-25
Fix for go 1.21

directionless:
Something in the 1.21 tooling changed how this works, and broke the janky pipeline we use in `pkg/make/builder.go`. But, this is a simple enough fix. See if CI passes it on 1.20

hat tip to @RebeccaMahany who found this last week


James-Pickett (approved):

prikhi:
Fixes #1295

PR #1307, opened 2023-08-29
Run osquery version check immediately on launcher startup

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1211

Operating on the theory that osquery startup may be slow due to the notarization check, this PR adds an earlier execution of the osqueryd binary to kick off the notarization check earlier.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -9,8 +9,10 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):

cmd/launcher/launcher.go
```diff
@@ -466,3 +470,30 @@ func writePidFile(path string) error {
...
+// runOsqueryVersionCheck execs the osqueryd binary in the background when we're running
+// on darwin. Operating on our theory that some startup delay issues for osquery might
+// be due to the notarization check taking too long, we execute the binary here ahead
+// of time in the hopes of getting the check out of the way.
+func runOsqueryVersionCheck(ctx context.Context, logger log.Logger, osquerydPath string) {
+	if runtime.GOOS != "darwin" {
+		return
+	}
+
+	versionCtx, versionCancel := context.WithTimeout(ctx, 10*time.Second)
```
directionless, line 483:
```suggestion
	versionCtx, versionCancel := context.WithTimeout(ctx, 30*time.Second)
```

Maybe?

cmd/launcher/launcher.go
```diff
@@ -70,6 +72,8 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 
 	logger := log.With(ctxlog.FromContext(ctx), "caller", log.DefaultCaller, "session_pid", os.Getpid())
 
+	go runOsqueryVersionCheck(ctx, logger, opts.OsquerydPath)
```
directionless, line 75: Do you think we want this in a routine? Probably okay. I guess... having it async is nice, since it won't block and it will gather info.

But if it turns out to matter, having it sync might help. 

Eh, as is!

directionless (approved):

James-Pickett (approved):

directionless (approved):

PR #1315, opened 2023-08-30
The osquery healthcheck startup delay should not block extension shutdown

RebeccaMahany:
If launcher shuts down within the first 10 minutes, it will fail to shut down completely while waiting for the healthcheck delay (introduced in https://github.com/kolide/launcher/pull/1294) to complete. This PR fixes that issue.


directionless (approved):
I wonder if we snuck this bug in elsewhere. 

I kinda remember writing, and fixing, it for the autoupdater delays

PR #1316, opened 2023-08-30
Rungroup improvements: logging

RebeccaMahany:

Builds on oklog/rungroup to add logging, making it possible to identify which actors are not shutting down in a timely manner.

### DIFF
--- a/pkg/rungroup/rungroup.go
+++ b/pkg/rungroup/rungroup.go
@@ -1,5 +1,10 @@
 package rungroup
 
+// rungroup expands on oklog/run, adding logs to indicate which actor caused
+// the interrupt and which actor, if any, is preventing shutdown. In the
+// future, we would like to add the ability to force shutdown before a given
+// timeout. See: https://github.com/kolide/launcher/issues/1205
+
 import (
 	"fmt"
 
--- a/pkg/rungroup/rungroup.go
+++ b/pkg/rungroup/rungroup.go
@@ -61,13 +61,13 @@ func (g *Group) Run() error {
 	}
 
 	// Wait for the first actor to stop.
-	err := <-errors
-	level.Debug(g.logger).Log("msg", "received interrupt error from first actor -- shutting down other actors", "err", err)
+	initialActorErr := <-errors
+	level.Debug(g.logger).Log("msg", "received interrupt error from first actor -- shutting down other actors", "err", initialActorErr)
 
 	// Signal all actors to stop.
 	for _, a := range g.actors {
 		level.Debug(g.logger).Log("msg", "interrupting actor", "actor", a.name)
-		a.interrupt(err.err)
+		a.interrupt(initialActorErr.err)
 	}
 
 	// Wait for all actors to stop.
@@ -77,7 +77,7 @@ func (g *Group) Run() error {
 	}
 
 	// Return the original error.
-	return err.err
+	return initialActorErr.err
 }
 
 func (a actorError) String() string {
--- a/pkg/rungroup/rungroup.go
+++ b/pkg/rungroup/rungroup.go
@@ -1,0 +1,80 @@
+package rungroup
+
+import (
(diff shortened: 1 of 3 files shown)


James-Pickett (commented):
This is really good! I think it may just need some tests to be shippable.

pkg/rungroup/rungroup.go
```diff
@@ -0,0 +1,85 @@
+package rungroup
```
directionless, line 1: Ah. So this is a _replacement_ for oklog's rungroup, not simple a wrapper. Seems legit

pkg/rungroup/rungroup.go
```diff
@@ -0,0 +1,85 @@
...
+func (g *Group) Run() error {
+	if len(g.actors) == 0 {
+		return nil
+	}
+
+	// Run each actor.
+	level.Debug(g.logger).Log("msg", "starting all actors", "actor_count", len(g.actors))
+	errors := make(chan actorError, len(g.actors))
+	for _, a := range g.actors {
+		go func(a rungroupActor) {
```
directionless, line 53: This is an unusual usage of an anonymous function.

If you're going to declare one that that takes an `actor` you could make a `func (g *group) start(a rungroupActor)`

or if it's anonymous, just use the `a` defined in the forloop

pkg/rungroup/rungroup.go
```diff
@@ -0,0 +1,85 @@
...
+			err := a.execute()
+			errors <- actorError{
+				errorSourceName: a.name,
+				err:             err,
+			}
+		}(a)
+	}
+
+	// Wait for the first actor to stop.
+	err := <-errors
```
directionless, line 64: Is this of type `error`? If not, calling it `err` feels confusing

pkg/rungroup/rungroup.go
```diff
@@ -0,0 +1,85 @@
...
+	level.Debug(g.logger).Log("msg", "received interrupt error from first actor -- shutting down other actors", "err", err)
+
+	// Signal all actors to stop.
+	for _, a := range g.actors {
+		level.Debug(g.logger).Log("msg", "interrupting actor", "actor", a.name)
+		a.interrupt(err.err)
```
directionless, line 70: Do we want to pass `nil` here? or leave that for future work?

pkg/rungroup/rungroup.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	// Wait for all actors to stop.
+	for i := 1; i < cap(errors); i++ {
+		e := <-errors
+		level.Debug(g.logger).Log("msg", "successfully interrupted actor", "actor", e.errorSourceName, "index", i)
+	}
```
directionless, line 77: I know that this is a copy from the original, but this is a weird idiom. Feels like wait groups would be more sensible (if harder to add logging for). Eh, maybe we leave it and adjust later?

directionless (commented):
I think it's fine. I think it's pulling in a lot of weird cruft from the original, but we can iterate in subsequent steps

pkg/rungroup/rungroup.go
RebeccaMahany: You're right, updated!

pkg/rungroup/rungroup.go
RebeccaMahany: I don't feel strongly about it -- I'd been erring on the side of not adjusting any logic for now

James-Pickett (approved):

pkg/rungroup/rungroup.go
directionless: Needing `a := a` is a good point. I dunno, it's not usually what I see in for loops.

pkg/rungroup/rungroup.go
directionless: legit!

PR #1318, opened 2023-08-30
Add parser for CarbonBlack repcli output

zackattack01:


pkg/osquery/tables/execparsers/repcli/test-data/repcli_darwin.txt
```diff
@@ -0,0 +1,77 @@
...
+    First Boot After OS Upgrade: No
+    Service Uptime: 155110500 ms
+    Service Waketime: 37860000 ms
+Cloud Status: 
+    Server Address: https://dev-prod05.example.com
+    Registered: Yes
+    Next Check-In: Now
+    Private Logging: Disabled
+    Next Cloud Upgrade: None
+    MDM Device ID: 9FCF04E4-4C8C-45A0-B3EA-053672776382
```
directionless, line 35: Is this redacted? I'd start it with `99999999` and it'll be obvious

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,171 @@
...
+// https://community.carbonblack.com/t5/Knowledge-Base/Endpoint-Standard-How-to-Verify-Sensor-Status-With-RepCLI/ta-p/62524
+
+import (
+	"bufio"
+	"io"
+	"strings"
+	"unicode"
+)
+
+func NewStatusTemplate() map[string]map[string]any {
```
directionless, line 15: Do you need to export this?

But aside from that, I'm leery of needing to pre-define known fields. It tends change out from under us

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,171 @@
...
+
+// formatKey prepares raw (potentially multi-word) key values by:
+// - stripping all surrounding whitespace
+// - coercing the entire string to lowercase
+// - splitting multiple words and joining them as snake_case
+func formatKey(key string) string {
+	processed := strings.TrimSpace(strings.ToLower(key))
+	words := strings.Fields(processed)
+	// State and Status are used interchangeably in section headers
+	// across device types, so we coerce all to status here
```
directionless, line 81: How confident are you about that? Because this sort of thing means that if they evolve different meaning, we've backed ourselves into a corner and have bad data. I have a strong bias to presenting the information as it is.

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,171 @@
...
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		rawLine := scanner.Text()
+		if len(rawLine) == 0 {
+			continue // blank lines are not expected or meaningful
+		}
+
+		kv := strings.SplitN(rawLine, ":", 2)
+		if len(kv) < 2 {
+			continue // lines without a colon are not expected or meaningful
```
directionless, line 105: Is it worth logging something to the debug logs here? Might not be

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,171 @@
...
+
+		// if this is the case then we're leaving a nested section (and should
+		// re-enter the parent section instead)
+		if currentNestedLevel < nestedLevel && len(sectionHeaders) > 1 {
+			sectionHeaders = sectionHeaders[0 : len(sectionHeaders)-1]
+		}
+
+		if _, ok := results[sectionHeaders[0]]; !ok {
+			continue // we don't handle this section
+		}
```
directionless, line 138: I think this has a lot of assumptions about the data. I think this should create the `map[string]any` if `!ok`. Might also need recursion

directionless: I could maybe be pushed into this, but I've seen too many third parties change their data out from under us.

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: I was torn with this and would love more ideas here- the largest offender was that the linux output has a section header named `Sensor Status` and the same on mac was `Sensor State`. I think if we do what the rest of your comments are getting at (regarding flexibility) it would be less cumbersome to just let them be separate keys on separate distros in general

pkg/osquery/tables/execparsers/repcli/test-data/repcli_darwin.txt
zackattack01: this was just from me running `uuidgen` locally, I will update with 9s for clarity thank you!

pkg/osquery/tables/execparsers/repcli/parser.go
James-Pickett: I think I led @zackattack01 astray by suggesting that we should hard code the data we want.

I was originally worried that counting spaces would be unreliable, but upon reflection and discussion, I think the chance of changing the data keys is higher than the chance of changing the spaces.

Also, we could dynamically determine the length of an indent. Just save the len to a var the first time we encounter a nested header and use that to determine nesting level.

pkg/osquery/tables/execparsers/repcli/parser.go
directionless: Oh that's a good question... Hrm. Somewhat, more data needed.

I don't know if they'll be a time when `Sensor Status` and `Sensor State` mean different things. Or if a platform has both. Etc. I can say that in SQL, if we know about the differences we can say `select * where key in ('sensor', 'status')`, so it's easy to handle at that layer.

pkg/osquery/tables/execparsers/repcli/parser.go
directionless: Yes, I think you don't have to assume the _length_ of whitespace, just that there is whitespace and it increases when you go in a level. (Python style, I guess?)

I would put the likelihood of data changes much higher than the change at reformatting. And if it reformats, we'll need a new parser anyhow

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,171 @@
...
+	for idx, word := range words {
+		if word == "state" {
+			words[idx] = "status"
+		}
+	}
+
+	return strings.Join(words, "_")
+}
+
+func repcliParse(reader io.Reader) (any, error) {
```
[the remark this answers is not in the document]
zackattack01: absolutely! thank you

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,129 @@
...
+		// handle any cases where there is already a value set for key
+		if existingValue, ok := results[key]; ok {
+			switch existingValue.(type) {
+			case []string:
+				results[key] = append(results[key].([]string), value)
+			case string:
+				results[key] = []string{results[key].(string), value}
+			default:
+				// if additional nested types are supported they should be added to the switch here
+				continue
```
directionless, line 82: We should do something here, otherwise this branch is invisible to us. `fmt.Sprintf("Unknown type %T", existingValue)` isn't wrong. Is there a way to tell if it's stringable?

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,129 @@
...
+	return formatKey(kv[0]), strings.TrimSpace(kv[1]), nestedDepth
+}
+
+// parseSection works recursively to handle nested sections. There is slight duplication
+// between the outer scanner.Scan() call here and in the main repcliParse function to ensure
+// this would still function as expected with arbitrary levels of nesting while supporting
+// standard sections at the top level.
+// This returns the given results for the section, along with a boolean flag indicating
+// that the previous line should be re-read when the parsing determines we've moved
+// back outside the target depth.
```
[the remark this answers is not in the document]
directionless: Oh, I see, it's so you can:
```
		if nextDepth <= currentDepth {
			// we must pass skipScan here because we'll need to re-read this line on the next pass
			return results, true
		}
```

I'm not sure I can quite wrap my head around the recursion, but I'd probably reach for something involving
```
skipScan = true
continue
```

I realize that needs something to to hint we need to move up a level. So maybe it's not worth it. But you shouldn't need two methods

directionless (commented):
Need to handle the unknown case somehow. The rests are nits / commentary.

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,129 @@
...
+
+		if key == "" {
+			continue
+		}
+
+		isSectionHeader := len(value) == 0
+		if nextDepth <= currentDepth {
+			// we must pass skipScan here because we'll need to re-read this line on the next pass
+			return results, true
+		}
```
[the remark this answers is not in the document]
zackattack01: correct! that's the general idea here. i had started exploring more explicit management of the file handle/read pointer but that seemed messier than necessary for this because i haven't found a need to keep more than one line

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: that makes sense, i'll look into checking for stringable!

pkg/osquery/tables/execparsers/repcli/parser.go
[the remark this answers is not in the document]
zackattack01: yeah this was the duplication i had mentioned bothering me when we discussed offline, I will spend a little more time trying to clean this up. thank you!

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		currentKeyPaths = updatedKeyPaths(currentKeyPaths, line)
+
+		if line.isSectionHeader {
+			continue
+		}
+
+		setNestedValue(results, currentKeyPaths)
```
zackattack01, line 146: forgot to surface any potential error here, will add

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(line) == 0 {
+		return "", "", 0 // blank lines are not expected or meaningful
+	}
+
+	kv := strings.SplitN(line, ":", 2)
+	if len(kv) < 2 {
+		return "", "", 0 // lines without a colon are not expected or meaningful
+	}
+
+	nestedDepth := len(kv[0]) - len(strings.TrimLeftFunc(kv[0], unicode.IsSpace))
```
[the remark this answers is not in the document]
zackattack01: It has been fine in testing- that was specifically to ensure we clean up all white space for the keys in the key value pairs presented, but are able to read the leading whitespace for indent level

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: Going to resolve this (no longer relevant), thank you both for the feedback

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: that makes sense to me! I've updated to remove the coercion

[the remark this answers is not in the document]
zackattack01: it does! but the coercion is removed now (per discussion in another thread [here](https://github.com/kolide/launcher/pull/1318#discussion_r1310823554))

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: I ended up just returning an error here after the full cleanup- as far as I can tell it would have to mean something is very wrong because we're always the ones setting the value type in the first place. let me know if that seems alright

directionless (commented):
I can't easily follow this, but the tests look okay. I'd approve it, but I think we need to handle the unknown differently.

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,152 @@
...
+// - coercing the entire string to lowercase
+// - splitting multiple words and joining them as snake_case
+func formatKey(key string) string {
+	processed := strings.TrimSpace(strings.ToLower(key))
+	words := strings.Fields(processed)
+	return strings.Join(words, "_")
+}
+
+// parseLine reads the next line from a scanner and attempts to pull out the
+// key, value, and key depth (level of nesting) into a repcliLine struct.
```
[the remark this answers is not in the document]
zackattack01: good catch. applied, thank you!

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,152 @@
...
+		switch knownValue := results[key].(type) {
+		case []string:
+			results[key] = append(knownValue, value)
+		case string:
+			results[key] = []string{knownValue, value}
+		case resultMap, interface{}, nil:
+			results[key] = value
+		default:
+			// if additional nested types are required they should be added above
+			return fmt.Errorf("unknown type %T requested for nested set", knownValue)
```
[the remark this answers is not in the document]
zackattack01: makes sense, updated!

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: discussed offline- we've opted to include the info we have with the record to prevent info loss here

pkg/osquery/tables/execparsers/repcli/parser.go
```diff
@@ -0,0 +1,165 @@
...
+			results[key] = fmt.Sprintf("unknown type %T requested on value %v", knownValue, value)
+		}
+
+		return results
+	}
+
+	_, ok := results[key]
+	if !ok {
+		results[key] = make(resultMap, 0)
+	}
```
James-Pickett, line 123:
```suggestion
	if _, ok := results[key]; !ok {
		results[key] = make(resultMap, 0)
	}
```

James-Pickett (commented):
Nice! Just one small nit.

pkg/osquery/tables/execparsers/repcli/parser.go
zackattack01: nice, thank you!

directionless (approved):
Let's try!

James-Pickett (approved):

PR #1322, opened 2023-08-31
Log Routing

directionless:
Replaces #1183

This PR is an attempt at setting up logging. The goal is to create something like:
```mermaid
flowchart LR
    subgraph inputs
        direction TB
        systemLogger>systemLogger]
        logger>logger]

        %% Force formatting
        systemLogger ~~~ logger
    end

    subgraph destinations
        direction TB
        destSys[(System Log Store)]
        destDebugLog[(debug.json)]
        destMem[(inmemory ring)]
        destStore[(persisted ring)]
    end

    systemLogger --> slTee{tee} --> logger
    slTee --> slFilter[[filter out debug]] --> destSys

    %% force formatting
    logger ~~~ slTee ~~~ systemLogger

    logger --> lTee{tee}
    lTee{tee} --> destDebugLog
    lTee --> destMem
    lTee --> f2[[filter: info higher]] --> destStore

    osquery -. kolide_launcher_logs .-> destMem
    osquery -. kolide_launcher_crit_logs .-> destStore

```

To achieve that, this PR adds:
1. a simple ring buffer, backed by a `store`
2. building on that, it creates a `ringlogger`
3. a log thicket, which uses these to create that routing.

TODO:
* replace the logger in main
* setup osquery table access


pkg/agent/logrouter/logrouter.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (lr *logRouter) Log(keyvals ...interface{}) error {
+	return lr.loggerTee.Log(keyvals...)
+}
+
+func (lr *logRouter) Logger() log.Logger {
+	return lr.loggerTee
+}
+
+func (lr *logRouter) SystemLogger() log.Logger {
```
James-Pickett, line 97: How does the system logger know to write to the system log? Or rather, how do the other logs know not to write to the system log? I just want to understand how launcher make that distinction.

pkg/agent/logrouter/logrouter.go
```diff
@@ -0,0 +1,203 @@
...
+// TODO: This should probably move to the underlying ringlogger, but we should also add an iterator for that.
+func (lr *logRouter) Replay(logger log.Logger) error {
+	logs, err := lr.logRing.GetAll()
+	if err != nil {
+		return fmt.Errorf("gettings logs to replay: %w", err)
+	}
+	for _, l := range logs {
+		var levelFn func(logger log.Logger) log.Logger
+
+		pairs := make([]any, 2+len(l))
```
James-Pickett, line 125: Preallocating the pairs array here and then appending leaves the the first few elements in the array nil and appends to the end of it.

So we should either keep track of the index and insert into the array or just make an empty array and append to it.

Here what the debugger shows when debugging the tests:

image

pkg/agent/logrouter/logrouter.go
directionless: Somewhat -- it doesn't. 

There are two ingestion points, Logger, and SystemLogger.

SystemLogger is a tee that routes to Logger and to the provided system log facility. There may be filtering on the system log facility. 

The caller needs to decide whether to call Logger or SystemLogger. And practically speaking, I think nothing outside of main will call SystemLogger.

pkg/agent/logrouter/logrouter.go
directionless: Thanks for calling that out. I always _want_ to hint at length, to avoid extra allocations, but it probably doesn't work. I've removed it,

directionless:
While this has some neat things in it, we've moved to `slogger` (sweet!) we can pick parts out if we want them

Issue #1320, opened 2023-08-30
Launcher should report errors from execs in `kolide_dev_table_tooling`


directionless:
The purpose of `kolide_dev_table_tooling` is to execute a command and allow us developers to gather data for building a parser. _But_ we don't report errors back. They get logged, but not reported back. We should report errors.

Instead of https://github.com/kolide/launcher/blob/1594781a56a6da9ed0cc1f77d97a00dfe8094218/pkg/osquery/tables/dev_table_tooling/table.go#L57-L67

we should do something like:
```
		results = append(results, map[string]string{
			"name":   name,
			"args":   strings.Join(cmd.args, " "),
			"output": base64.StdEncoding.EncodeToString(output),
			"error": err,
		})
```

PR #1323, opened 2023-08-31
ensure that dev table exec calls include error output

zackattack01:
Currently calls to `tablehelpers.Exec` can pass an `includeStderr bool` flag to pipe stderr into stdout- but if the `cmd.Run` call returns an actual error then stdout is omitted from the return value of `Exec`.

To ensure we can access the stderr data, we can include an additional error column in the expected dev table output, and populate this value as appropriate.

Fixes: #1320

### DIFF
--- a/pkg/osquery/tables/dev_table_tooling/table.go
+++ b/pkg/osquery/tables/dev_table_tooling/table.go
@@ -27,6 +27,7 @@ func TablePlugin(logger log.Logger) *table.Plugin {
 		table.TextColumn("name"),
 		table.TextColumn("args"),
 		table.TextColumn("output"),
+		table.TextColumn("error"),
 	}
 
 	tableName := "kolide_dev_table_tooling"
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):
nice

RebeccaMahany (approved):

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -54,17 +55,21 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
+			"args":   strings.Join(cmd.args, " "),
+			"output": "",
+			"error":  "",
+		}
+
+		if output, err := tablehelpers.Exec(ctx, t.logger, 30, cmd.binPaths, cmd.args, false); err != nil {
 			level.Info(t.logger).Log("msg", "execution failed", "name", name, "err", err)
-			continue
+			result["error"] = err.Error()
+		} else {
```
directionless, line 68: You could strike the `if` here, but this is fine too.

directionless: I think I'm wrong, at least, sometimes. If err is nil, some ways of casting it to a string work, and some fail. So probably best to keep it like this

Issue #1325, opened 2023-09-01
Launcher needs a checkup for gnome extensions


directionless:
I'm only 75% confident here....

Both systray and notifications need a specific gnome extension. `ubuntu-appindicators@ubuntu.com`. And these extensions can possible make trouble for us.

We should have a checkup that:
1. Uses the presence of the required extension as status -- `gnome-extensions show ubuntu-appindicators@ubuntu.com`
2. For flare, includes the full extensions list `gnome-extensions list`
3. Caveat -- gnome extension commands are going to be per user. And I don't know how we want to handle that

directionless:
It looks like this extension might [log somewhere](https://github.com/hovancik/stretchly/issues/812). We should find the logs and include them

PR #1328, opened 2023-09-02
Add checkup for gnome extensions

directionless:
Fixes: #1325

This adds a checkup for the gnome extensions. It has some extra for loops because it's per user. But, I figured out how to interface with the users gnome session. So that's cool


James-Pickett (approved):
Nice!

PR #1331, opened 2023-09-05
Don't perform unnecessary restarts when not using new autoupdater

RebeccaMahany:

If launcher is not on nightly or alpha channel, it should not restart after performing a download to the new autoupdate library.


directionless (approved):
Feels like a bit of a bandaid, but the transition is kinda messy

PR #1340, opened 2023-09-06
add system time flare check

James-Pickett:


pkg/debug/checkups/time.go
```diff
@@ -0,0 +1,103 @@
...
+func getTimeFromURL(url string) (time.Time, error) {
+	// Make an HTTP GET request to the specified URL.
+	resp, err := http.Get(url)
+	if err != nil {
+		return time.Time{}, err
+	}
+	defer resp.Body.Close()
+
+	// Parse the "Date" header from the response.
+	dateHeader := resp.Header.Get("Date")
```
James-Pickett, line 75: is this a legit way to get the time from a "reputable" source?

pkg/debug/checkups/time.go
```diff
@@ -0,0 +1,103 @@
...
+
+		diff := systemTime.Sub(serverTime)
+		if diff < 0 {
+			diff = -diff
+		}
+
+		maxDiff := 5 * time.Minute
+		if diff > maxDiff {
+			st.summary = fmt.Sprintf("system time off by more than %f minutes, delta = %f minutes", maxDiff.Minutes(), diff.Minutes())
+			st.status = Failing
```
directionless, line 51:
```suggestion
			st.status = Warning
```

I think we're handling, it so maybe just warning

pkg/debug/checkups/time.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		st.summary = fmt.Sprintf("system time is within %f minutes, delta = %f minutes", maxDiff.Minutes(), diff.Minutes())
+		st.status = Passing
+		return nil
+	}
+
+	// if we made it here, we never got valid server time
+	st.summary = "could not get valid server time"
+	st.status = Failing
```
directionless, line 62:
```suggestion
	st.status = Erroring
```

erroring Indicates a failing in the checkup, vs failing which indicates a bad install

pkg/debug/checkups/time.go
directionless: Good question! It feels a little janky, but also clever? I'd be cautious about relying on it, the server time on a random internet server may not be correct, but it's probably fine for here. Just kinda know this is the sort of thing that might be flakey when you least expect it

directionless: It would be "more correct" to use a time serving protocol. Whether it's NTP, rfc3161, or something else I don't know. But I really don't want to pull ntp code into go.

RebeccaMahany: Sure, then I'd rather we document above with the URLs that we're using the date header returned by the servers, not NTP or timestamping, to pull current time.

Or I wonder if we'd want an `https://app.kolide.com/time` (like we have `https://app.kolide.com/version` now). If the point is verifying time agreement for krypto, then that would be our most reliable option.

James-Pickett: I switched to using `https://app.kolide.com/version` as the first attempt to get the time. It's still using the date header though.

Given that this flare is a "warning" at best, I don't think the juice is worth the squeeze of setting up an time endpoint or using NTP protocol. However, if anyone feels strongly, I'm willing to pursue one of those.

pkg/debug/checkups/time.go
James-Pickett: done

pkg/debug/checkups/time.go
James-Pickett: done

pkg/debug/checkups/system_time.go
```diff
@@ -20,21 +20,22 @@ func (st *systemTime) Name() string {
 func (st *systemTime) Run(_ context.Context, extraFh io.Writer) error {
 	var (
 		urls = []string{
+			"https://app.kolide.com/version",
```
directionless, line 23:
```suggestion
			"https://k2control.kolide.com/version",
```
I'm willing to try it, but we have fairly limited concurrency on our servers

directionless (approved):
Let's try

directionless (approved):

PR #1342, opened 2023-09-07
Make it safe to call rungroup actor Interrupt multiple times

RebeccaMahany:

While investigating an autoupdate shutdown issue, noticed that shutdown could block on runner.Interrupt being called multiple times.

### DIFF
--- a/ee/desktop/runner/runner.go
+++ b/ee/desktop/runner/runner.go
@@ -108,7 +108,8 @@ type DesktopUsersProcessesRunner struct {
 	menuRefreshInterval time.Duration
 	interrupt           chan struct{}
 	// uidProcs is a map of uid to desktop process
-	uidProcs map[string]processRecord
+	uidProcs      map[string]processRecord
+	uidProcsMutex sync.Mutex
 	// procsWg is a WaitGroup to wait for all desktop processes to finish during an interrupt
 	procsWg *sync.WaitGroup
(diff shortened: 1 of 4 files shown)


ee/desktop/runner/runner.go
```diff
@@ -222,8 +224,15 @@ func (r *DesktopUsersProcessesRunner) Execute() error {
...
 // It also signals the execute loop to exit, so new desktop processes cease to spawn.
 func (r *DesktopUsersProcessesRunner) Interrupt(_ error) {
-	// Tell the execute loop to stop checking, and exit
-	r.interrupt <- struct{}{}
+	// Non-blocking send to interrupt channel
+	select {
+	case r.interrupt <- struct{}{}:
+		// First time we've received an interrupt, so we've notified r.Execute.
+	default:
+		// Execute loop is no longer running, so there's nothing to interrupt
```
directionless, line 232: Should we log this? Or is it so common it'd just be noise

RebeccaMahany: It's not common but I'm also not sure it'd be useful?

RebeccaMahany:
Closing in favor of https://github.com/kolide/launcher/pull/1344

directionless:
> Closing in favor of #1344

This one handles more of the actors. Do you think we still need it?

PR #1343, opened 2023-09-07
add appindicator check for linux before running desktop

James-Pickett:
closed in favor of https://github.com/kolide/launcher/pull/1545

We've seen issues where a user does not have a supported appindicator enabled on linux and when the desktop runs it consumes excessive CPU. This PR adds a check before running desktop to ensure that a supported appindicator is enabled. Also adds a flare to check for supported appindicators.


pkg/debug/checkups/appindicator.go
```diff
@@ -0,0 +1,50 @@
+package checkups
```
directionless, line 1: This duplicates https://github.com/kolide/launcher/blob/main/pkg/debug/checkups/gnome-extensions.go I don't think we need both?

ee/desktop/runner/runner.go
```diff
@@ -466,6 +467,12 @@ func (r *DesktopUsersProcessesRunner) runConsoleUserDesktop() error {
 		return nil
 	}
 
+	if runtime.GOOS == "linux" && !IsAppindicatorEnabled(context.Background()) {
+		level.Info(r.logger).Log("msg", "no appindicator enabled, not running desktop")
+		r.killDesktopProcesses()
```
directionless, line 472: This feels like we need more research. In my little testing, it seemed like when the screen is locked, the extension is disabled. I'm not sure if we want to _kill_ on that condition, or simply not start

ee/desktop/runner/runner.go
```diff
@@ -800,3 +819,42 @@ func removeFilesWithPrefix(folderPath, prefix string) error {
...
+		// appindicator that ships with NixOS
+		"appindicatorsupport@rgcjonas.gmail.com",
+	}
+
+	for _, extension := range extensions {
+		for _, uid := range uids {
+			ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+			defer cancel()
+
+			cmd := exec.CommandContext(ctx, "/usr/bin/gnome-extensions", "show", extension)
```
directionless, line 842: Does `gnome-extensions list --enabled` let us do this with a single exec?

ee/desktop/runner/runner.go
```diff
@@ -800,3 +819,42 @@ func removeFilesWithPrefix(folderPath, prefix string) error {
...
+
+			// gnome seems to do things through this env
+			cmd.Env = append(cmd.Env, fmt.Sprintf("XDG_RUNTIME_DIR=%s", fmt.Sprintf("/run/user/%s", uid)))
+
+			if err := runas.SetCmdToExecAsUser(ctx, uid, cmd); err != nil {
```
directionless, line 847: I don't think you need this with `XDG_RUNTIME_DIR` set

ee/desktop/runner/runner.go
James-Pickett: yes! single exec per user

ee/desktop/runner/runner.go
James-Pickett: after some testing, this is required when running as root, in face the gnome-extension check fails for me on my ubuntu machine when run as root

ee/desktop/runner/runner.go
James-Pickett: it's possibly belt and suspenders, but if someone does disable or install some funky appindicator, I would think we want to turn off the desktop menu to make sure we don't eat CPU

ee/desktop/runner/runner.go
James-Pickett: updated

ee/desktop/runner/runner.go
James-Pickett: when running as user, we don't need to set the XDG_RUNTIME_DIR

pkg/debug/checkups/appindicator.go
James-Pickett: I was just leaving it there for a bit to compare, it's removed now.

ee/desktop/runner/runner.go
RebeccaMahany: What if someone's running cinnamon/mate/xfce4/kde/something else besides gnome?

James-Pickett: hmm ... I don't know

We could check to see if the user is running gnome and then do the check. Letting non-gnome machines continue to run. 

What do you think?

RebeccaMahany: It's been a while since I've tested (and I only tested in multipass VMs, so who knows how accurate of a test), but the desktop seemed to behave okay on non-gnome machines. Maybe worth a quick validation that that's still true? And if so, letting non-gnome machines continue to run desktop.

ee/desktop/runner/runner.go
directionless: This confuses me. At least my testing found the opposite -- that if I ran as root, with `XDG_RUNTIME_DIR` it was enough. I think it's worth figuring out which of us is correct, since if you're right, we'll need to update the flare i wrote.

James-Pickett: I get this error on ubuntu. Could you check again or maybe @RebeccaMahany could you give it a shot?

```

/tmp/flare 
❯  sudo /usr/local/kolide-k2/bin/launcher flare
{"caller":"main.go:31","msg":"Launcher starting up","revision":"dba729447ed40e56783695a2cbf0ece230cdb885","severity":"info","ts":"2023-09-12T20:41:33.046304966Z","version":"1.0.16"}
{"caller":"main.go:33","msg":"Launcher starting up","revision":"81def1ba72095207174c5cf91e7ae7290619b312","severity":"info","ts":"2023-09-12T20:41:33.057998277Z","version":"1.0.16-25-g81def1b"}
{"caller":"main.go:188","err":"not rolling out new TUF to channel beta that should still use legacy autoupdater","msg":"could not check out latest launcher, will fall back to old autoupdate library","severity":"error","ts":"2023-09-12T20:41:33.058059875Z"}

/tmp/flare 
❯  unzip -q ./kolide_agent_flare_report_01HA5J1F4229MQG19MB8YNX7J7.zip 

/tmp/flare 
❯  cat ./Gnome\ Extensions/summary.log 
❌	Gnome Extensions: uid:1000: error running gnome-extensions: running "/usr/bin/gnome-extensions show ubuntu-appindicators@ubuntu.com", err is: : exit status 2

```

RebeccaMahany: That looks like it's running the wrong version of launcher for what you're trying to test, since it's running the gnome extensions checkup that you deleted?

I pulled the current commit af2106... and I ran `sudo LAUNCHER_SKIP_UPDATES=true ./build/launcher flare` -- my appindicator summary.log reports `No appindicator is enabled, cannot display desktop`, which isn't true.

James-Pickett: @RebeccaMahany , apologies for not being more specific. I was wondering if the gnome extension check on the current nighlty build worked for you.

James-Pickett: but also yes, the appindicator check is broken ... fixing it

directionless: Huh. I'm testing again and I cannot recreate my path. I cannot get this to work without having user permissions, so I think you're right. Empirically I also need to set `XDG_RUNTIME_DIR`, but that's not enough.

James-Pickett: added `XDG_RUNTIME_DIR` env var back to cmd and working as expected (when run as user)

ee/desktop/runner/runner.go
James-Pickett: There seems to be lots of ways to check if a machine is using gnome with lots of edge cases, so I installed KDE plasma on a vm and ran `/usr/bin/gnome-extensions list`. It will error with something like "failed to connect to GNOME shell". So I updated the code to just assume that gnome is not running when the cmd fails. I'm not sure if this is good enough

directionless: I don't know a perfect way to tell if we're running gnome. I think checking for the gnome session, or gdm, probably.

James-Pickett: I tried a couple of different ways

One machine is just default ubunutu mostly ....
One machine is ubuntu with KDE (plasma) installed after starting up ...

`loginctl`: works in both gnome and kde
`ls /usr/bin/*session`: will list both gnome-session and plasma-session (KDE)
`gdm3 --version`: works on both kde and gnome
`/usr/bin/gnome-extensions list`: fails on KDE works on gnome ... most reliable way I've found so far

directionless: I was expecting to need to check the process list. 

But, I think just execing `gnome-extensions` might be okay. We're going to exec that anyhow, and if we can discern "not running gnome" than we can call it good enough

James-Pickett:
closed in favor of https://github.com/kolide/launcher/pull/1545

PR #1344, opened 2023-09-07
Allow desktop runner Interrupt to be called multiple times

RebeccaMahany:

While investigating an autoupdate shutdown issue, noticed that shutdown could block on runner.Interrupt being called multiple times. I experimented with a couple different solutions, but found this was ultimately the simplest solution that didn't result in data races in tests.


zackattack01 (approved):
nice! love the test

Issue #1338, opened 2023-09-06
Another possible source of contention -- tables that query osquery


directionless:
As part of the general search for what might be causing contention on the osquery socket, there are launcher tables, that query osquery. This can be seen in the tables that use `osquery.ExtensionManagerClient`

A quick grep makes me think it might only be:
* https://github.com/kolide/launcher/blob/6e7dce3a993e0e8a20755f1b357452f62fa173fd/pkg/osquery/table/airdrop_darwin.go#L58
* https://github.com/kolide/launcher/blob/6e7dce3a993e0e8a20755f1b357452f62fa173fd/pkg/osquery/tables/macos_software_update/software_update_table.go#L87

Removing these will remove some of the contention, and make our code a bit cleaner. 

We do see this pop up as errors in our logs sometimes. https://app.bugsnag.com/kolide-inc/k2-device/errors/64fb30784a725e0007b9a307?event_id=64fb307800c4f7f539c20000&i=sk&m=nw

PR #1347, opened 2023-09-12
get build prefix for upgrade table from sysctl instead of osquery

zackattack01:
This is part 1 to begin addressing https://github.com/kolide/launcher/issues/1338 (reducing osquery socket contention)

This replaces the osquery `os_version` table with a `unix.Sysctl()` call to get the build train prefix used to gather the rest of the macos software update table data. This change should only impact macos


James-Pickett (approved):

directionless (approved):
I think this is okay, but I wonder if there are places we can test to be sure.

PR #1348, opened 2023-09-12
remove kolide_airdrop_preferences table

zackattack01:
The `kolide_airdrop_preferences` table has historically been unreliable and not widely used. The generation process creates  unnecessary contention on the osquery socket while attempting to gather the primary user to determine the appropriate settings.

For now we've decided to remove the table until a more reliable approach to gather this information is available.


RebeccaMahany (approved):
LGTM!

directionless (approved):

PR #1353, opened 2023-09-14
[TUF autoupdater] Fix prerelease sorting

RebeccaMahany:
The semver package expects that the semver parts will be separated by a period, and ours are separated by a hyphen. This means it won't correctly recognize that versions like "1.0.16-30-ge34c9a0" should be larger than "1.0.16-9-g613d85c", because it's using string comparison instead of splitting out 30 and 9 and comparing those as ints instead.

I'd previously decided this wasn't a big deal because we don't publish tags with pre-releases, but it's been enough of a pain that I think we should just correct the behavior so that it's what we expect.


zackattack01 (approved):
nice! 🚀

James-Pickett (approved):

PR #1355, opened 2023-09-14
[TUF autoupdater] Bump lookup logs up to Info level

RebeccaMahany:
At this point in main.go, we haven't set the `Debug` flag yet on our logger, so we lose all logs at the debug level. This PR bumps up the log level to `Info` for the logs we might care about when debugging autoupdater version selection.

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -163,11 +163,11 @@ func runNewerLauncherIfAvailable(ctx context.Context, logger log.Logger) {
 	}
 
 	if newerBinary == "" {
-		level.Debug(logger).Log("msg", "nothing newer")
+		level.Info(logger).Log("msg", "nothing newer")
 		return
 	}
 
-	level.Debug(logger).Log(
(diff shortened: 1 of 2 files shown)


zackattack01 (approved):

cmd/launcher/main.go
```diff
@@ -163,11 +163,11 @@ func runNewerLauncherIfAvailable(ctx context.Context, logger log.Logger) {
 	}
 
 	if newerBinary == "" {
-		level.Debug(logger).Log("msg", "nothing newer")
+		level.Info(logger).Log("msg", "nothing newer")
```
directionless, line 166: Hrm. 

Worth noting that all Info logs land in the _system_ log files. So this is potentially noisy for customers that have centralized windows event logging.

RebeccaMahany: We can probably revert at least some of these back down to the debug level, now that I've worked through the issue I was trying to troubleshoot 🙂

PR #1358, opened 2023-09-14
add cache clean to make build and test

zackattack01:
This has stopped the macOS version linking warnings on a few dev environments and is intended to prevent us from needing to bypass the stderr checks during the build process

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ build_%: ARCHARG = $(if $(ARCH), --arch $(ARCH))
 build_%: GOARG = $(if $(CROSSGOPATH), --go $(CROSSGOPATH))
 build_%: GOBUILD = $(if $(CROSSGOPATH), $(CROSSGOPATH), go)
 build_%: .pre-build
+	$(GOBUILD) clean -cache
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):

Makefile
```diff
@@ -44,6 +44,7 @@ build_%: ARCHARG = $(if $(ARCH), --arch $(ARCH))
 build_%: GOARG = $(if $(CROSSGOPATH), --go $(CROSSGOPATH))
 build_%: GOBUILD = $(if $(CROSSGOPATH), $(CROSSGOPATH), go)
 build_%: .pre-build
+	$(GOBUILD) clean -cache
```
directionless, line 47: I don't think this is a good idea. The go cache is why builds are fast, and that we need this is a real bug

zackattack01: yeah was just discussing more with James, would you be open to something more like: 
`if [[ "${GOCLEANCACHE}" -eq 1 ]]; then $(GOBUILD) clean -cache; fi`
or the simplest for now, just telling people the can run cache clean until we get to the bottom of the linking (I was under the impression this was an ongoing issue because it's happened since I started)

James-Pickett: i can live with just remembering to run `go clean -cache` until we figure it out

zackattack01: that makes sense to me, I will close this. Thanks for the feedback all!

PR #1359, opened 2023-09-15
reduce early localserver -> osquery contention

zackattack01:
Localserver currently utilizes a backoff routine to repeatedly hit osquery early during startup, and then refreshes information less frequently after the initial load.

The information gathered here is no longer functionally required (is used only for debugging purposes), so we can remove the initial backoff routine and prevent localserver from even attempting to hit osquery for 15 minutes instead.

### DIFF
--- a/ee/localserver/server.go
+++ b/ee/localserver/server.go
@@ -213,24 +213,22 @@ func (ls *localServer) runAsyncdWorkers() time.Time {
 }
 
 func (ls *localServer) Start() error {
-	// Spawn background workers. This loop is a bit weird on startup. We want to populate this data as soon as we can, but because the underlying launcher
-	// run group isn't ordered, this is likely to happen before querier is ready. So we wait for <pollInterval> and allow attempting at this more frequent
-	// interval until the first run is successful, then adhere to a much longer recalculation interval.
+	// Spawn background workers. This loop is a bit weird on startup. We want to populate this data as soon as we can,
+	// this is likely to happen before querier is ready. So we wait for <pollInterval> and before starting and then only
+	// rerun if the previous run was unsuccessful, or has been greater than <recalculateInterval>.
 	// Note that this polling is merely a check against time, we don't repopulate this data nearly so often. (But we poll
 	// frequently to account for the difference between wall clock time, and sleep time)
 	const (
-		pollInterval        = 10 * time.Minute
+		pollInterval        = 15 * time.Minute
 		recalculateInterval = 24 * time.Hour
 	)
 
 	go func() {
 		var lastRun time.Time
 
+		// note that this will trigger the check for the first time after pollInterval (not immediately)
 		for range time.Tick(pollInterval) {
-			// every pollInterval, run if:
-			//  - we've never run successfully, or the last run was unsuccessful
-			//  - we haven't run successfully in over <recalculateInterval>
-			if lastRun.IsZero() || time.Since(lastRun) > recalculateInterval {
+			if time.Since(lastRun) > recalculateInterval {
 				lastRun = ls.runAsyncdWorkers()
 				if lastRun.IsZero() {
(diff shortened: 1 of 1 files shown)


ee/localserver/server.go
```diff
@@ -214,36 +213,26 @@ func (ls *localServer) runAsyncdWorkers() time.Time {
 }
 
 func (ls *localServer) Start() error {
-	// Spawn background workers. This loop is a bit weird on startup. We want to populate this data as soon as we can, but because the underlying launcher
-	// run group isn't ordered, this is likely to happen before querier is ready. So we retry at a frequent interval for a couple of minutes, then we drop
-	// back to a slower poll interval. Note that this polling is merely a check against time, we don't repopulate this data nearly so often. (But we poll
+	// Spawn background workers. This loop is a bit weird on startup. We want to populate this data as soon as we can,
```
directionless, line 216: Comment is out of date -- we're not doing this as soon as we can

ee/localserver/server.go
```diff
@@ -214,36 +213,26 @@ func (ls *localServer) runAsyncdWorkers() time.Time {
...
-		}
 
-		// Now that we're done with the initial population, fall back to a periodic polling
+		// note that this will trigger the check for the first time after pollInterval (not immediately)
 		for range time.Tick(pollInterval) {
-			if time.Since(lastRun) > (recalculateInterval) {
+			if time.Since(lastRun) > recalculateInterval {
 				lastRun = ls.runAsyncdWorkers()
+				if lastRun.IsZero() {
+					level.Info(ls.logger).Log("message", "runAsyncdWorkers unsuccessful, will retry in the future.")
```
RebeccaMahany, line 234: I think we probably want Debug level for this one?

RebeccaMahany: Oh sorry, I see this was Info before. I think it probably doesn't need to be at that level, though

zackattack01: will do, thanks!

ee/localserver/server.go
zackattack01: I'll update this a bit more, thanks!

RebeccaMahany (approved):

ee/localserver/server.go
```diff
@@ -214,36 +213,26 @@ func (ls *localServer) runAsyncdWorkers() time.Time {
...
-			initialPollInterval,
-		); err != nil {
-			level.Info(ls.logger).Log("message", "Initial async runs unsuccessful. Will retry in the future.", "err", err)
-		}
 
-		// Now that we're done with the initial population, fall back to a periodic polling
+		// note that this will trigger the check for the first time after pollInterval (not immediately)
 		for range time.Tick(pollInterval) {
-			if time.Since(lastRun) > (recalculateInterval) {
+			if time.Since(lastRun) > recalculateInterval {
```
James-Pickett, line 231: Kind of a style nit I suppose, but I don't like the time comparison here. I would prefer something like this:

```suggestion
	go func() {
		pollTicker := time.NewTicker(pollInterval)
		recalcTicker := time.NewTicker(recalculateInterval)
		recalculate := false

		for {
			select {
			case <-pollTicker.C:
				if !recalculate {
					recalculate = ls.runAsyncdWorkers()
				}
			case <-recalcTicker.C:
				recalculate = false
			}
		}
	}()
```

James-Pickett (approved):

ee/localserver/server.go
James-Pickett: but I'm okay with leaving it as is, just couldn't resist commenting =)

zackattack01: that is nice thank you! I'm going to note this as preferred style going forward but leave for now to avoid retesting it all if that's okay?

zackattack01: discussed with @James-Pickett offline - moving forward for now!

directionless:
This relates to #1211

PR #1360, opened 2023-09-19
Add windows-specific test cases to findNew

RebeccaMahany:


directionless (commented):
What's the state of this PR?

RebeccaMahany:
@directionless not needed -- I think I added this when we were exploring using an env var to set the launcher installation path, and seeing some weirdness on Windows.

PR #1363, opened 2023-09-25
Add `access-control-allow-private-network` to the preflight

directionless:
See https://developer.chrome.com/blog/private-network-access-preflight/

### DIFF
--- a/ee/localserver/server.go
+++ b/ee/localserver/server.go
@@ -310,6 +310,7 @@ func (ls *localServer) preflightCorsHandler(next http.Handler) http.Handler {
 
 		// Some modern chrome and derivatives use Access-Control-Allow-Private-Network
 		// https://developer.chrome.com/blog/private-network-access-preflight/
+		// Though it's unclear if this is still needed, see https://developer.chrome.com/blog/private-network-access-update/
 		w.Header().Set("Access-Control-Allow-Private-Network", "true")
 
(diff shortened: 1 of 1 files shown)


ee/localserver/server.go
```diff
@@ -308,6 +308,10 @@ func (ls *localServer) preflightCorsHandler(next http.Handler) http.Handler {
 		w.Header().Set("Access-Control-Allow-Headers",
 			"Accept, Content-Type, Content-Length, Accept-Encoding, X-CSRF-Token, Authorization")
 
+		// Some modern chrome and derivatives use Access-Control-Allow-Private-Network
+		// https://developer.chrome.com/blog/private-network-access-preflight/
```
RebeccaMahany, line 312: https://developer.chrome.com/blog/private-network-access-update/ newer docs say this is going to be deprecated -- do we want to add a note about that?

directionless:
```suggestion
		// https://developer.chrome.com/blog/private-network-access-preflight/
		// Though it's unclear if this is still needed, see https://developer.chrome.com/blog/private-network-access-update/
```

directionless: Maybe like that?

RebeccaMahany: Works for me!

RebeccaMahany (approved):

PR #1368, opened 2023-09-26
Fix `exec: already started` error in uninstall on Linux

RebeccaMahany:
Keeps us from attempting to run the systemctl disable command twice.

### DIFF
--- a/cmd/launcher/uninstall_linux.go
+++ b/cmd/launcher/uninstall_linux.go
@@ -23,6 +23,7 @@ func removeLauncher(ctx context.Context, identifier string) error {
 	// Stop and disable launcher service
 	cmd := exec.CommandContext(ctx, "systemctl", []string{"disable", "--now", serviceName}...)
 	if out, err := cmd.CombinedOutput(); err != nil {
+		// Don't exit. Log and move on to the next uninstall command
 		fmt.Printf("error occurred while stopping/disabling launcher service, systemctl output %s: err: %s\n", string(out), err)
 		return err
 	}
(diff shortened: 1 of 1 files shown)


directionless (dismissed):

cmd/launcher/uninstall_linux.go
```diff
@@ -22,12 +22,8 @@ func removeLauncher(ctx context.Context, identifier string) error {
...
 	// Stop and disable launcher service
 	cmd := exec.CommandContext(ctx, "systemctl", []string{"disable", "--now", serviceName}...)
-	if out, err := cmd.Output(); err != nil {
-		fmt.Printf("error occurred while stopping/disabling launcher service, systemctl output %s: err: %s\n", out, err)
-		return err
-	}
-
-	if err := cmd.Run(); err != nil {
+	if out, err := cmd.CombinedOutput(); err != nil {
+		fmt.Printf("error occurred while stopping/disabling launcher service, systemctl output %s: err: %s\n", string(out), err)
```
directionless, line 26:
```suggestion
	if out, err := cmd.CombinedOutput(); err != nil {
		// Don't exit. Log and move on to the next uninstall command
		fmt.Printf("error occurred while stopping/disabling launcher service, systemctl output %s: err: %s\n", string(out), err)
```

RebeccaMahany: Wait sorry, applied before I checked the full diff -- do you want me to also remove the `return err` below this, then?

directionless: I mean, I also approved it before I added the suggestion. 😆 

Reading the file, I think all the `return err` should probably be removed. The ones around `dpkg` and `rpm` feel suspect.

RebeccaMahany: Done!

directionless (approved):

PR #1378, opened 2023-10-02
restart desktop on macos update

James-Pickett:


ee/desktop/runner/runner.go
```diff
@@ -213,6 +225,24 @@ func (r *DesktopUsersProcessesRunner) Execute() error {
...
+			}
+
+			osVersion, err := osversion()
+			if err != nil {
+				level.Error(r.logger).Log("msg", "getting os version", "err", err)
+				continue
+			}
+
+			if osVersion != r.osVersion {
+				r.osVersion = osVersion
```
zackattack01, line 243: would we want some logging here to indicate why the restart happens?

zackattack01 (dismissed):
nice, LGTM! just the one question about logging

ee/desktop/runner/runner.go
```diff
@@ -200,6 +210,8 @@ func (r *DesktopUsersProcessesRunner) Execute() error {
 	defer updateTicker.Stop()
 	menuRefreshTicker := time.NewTicker(r.menuRefreshInterval)
 	defer menuRefreshTicker.Stop()
+	osUpdateCheckTicker := time.NewTicker(5 * time.Second)
```
RebeccaMahany, line 213: Maybe a little less frequently than this? I could see once a minute being tolerable for the end user

zackattack01 (approved):
🔥

ee/desktop/runner/runner.go
James-Pickett: added some more logging, put in own func

ee/desktop/runner/runner.go
James-Pickett: bumped to 1 min

RebeccaMahany (approved):

directionless (approved):
neat. I wonder if this crops up on window.

PR #1381, opened 2023-10-04
Ensure debug logs are written to debug.json on Windows

RebeccaMahany:
This PR makes the logging behavior on Windows consistent with the behavior on macOS and Linux: all logs, regardless of level, go to debug.json; the debug flag only controls whether debug logs go to the event logs (stderr logs on macOS/Linux).


James-Pickett (approved):

zackattack01 (approved):
nice find 🚀

PR #1382, opened 2023-10-05
Update Mac OS instructions

fatso83:
the current instructions refer to old folders that do not exist. fixed the paths


directionless:
Hrm. This document is very out of date. And this PR correctly updates it to the Kolide SaaS product, but that does not cover the opensource users, which I have no information about

fatso83:
I am not signing anything if this is not going places.

PR #1388, opened 2023-10-10
Give rungroup actors enough time to shut down on service shutdown

RebeccaMahany:
I was consistently seeing only 8 of 13 actors get a chance to shut down on windows on service stop -- waiting slightly longer before we return and exit the service gives the other actors a chance to shut down gracefully as well.

As a drive-by, fixes the log when recovering from a panic to include info about the panic.


directionless (approved):
Make a bigger bandaid. Great.

PR #1389, opened 2023-10-10
Add utf16 decoding to json dataflatten

Micah-Kolide:
This fixes cases we've seen where json files containing application config data isn't readable by `kolide_json`. From the case I was able to examine, this is because the file was encoded with `utf-16le`.

This limitation was noted in a [go issue](https://github.com/golang/go/issues/36686) a while back, but it didn't really go anywhere (hah). As noted from that, go's json unmarshal is expecting `utf-8` data, so I added some text transformation to achieve this.

```
osquery> SELECT * FROM magic WHERE path = '/Users/micahsorenson/Downloads/UserConfig.json';
+------------------------------------------------+-----------------------------------------------------+-----------+------------------+---------------+
| path                                           | magic_db_files                                      | data      | mime_type        | mime_encoding |
+------------------------------------------------+-----------------------------------------------------+-----------+------------------+---------------+
| /Users/micahsorenson/Downloads/UserConfig.json | /usr/share/file/magic.mgc:/usr/share/misc/magic.mgc | JSON data | application/json | us-ascii      |
+------------------------------------------------+-----------------------------------------------------+-----------+------------------+---------------+

osquery> SELECT * FROM magic WHERE path = '/Users/micahsorenson/Downloads/config/UserConfig.json';
+-------------------------------------------------------+-----------------------------------------------------+--------------------------------------------------------------------+------------+---------------+
| path                                                  | magic_db_files                                      | data                                                               | mime_type  | mime_encoding |
+-------------------------------------------------------+-----------------------------------------------------+--------------------------------------------------------------------+------------+---------------+
| /Users/micahsorenson/Downloads/config/UserConfig.json | /usr/share/file/magic.mgc:/usr/share/misc/magic.mgc | Unicode text, UTF-16, little-endian text, with no line terminators | text/plain | utf-16le      |
+-------------------------------------------------------+-----------------------------------------------------+--------------------------------------------------------------------+------------+---------------+
```

Before commit:
```
osquery> SELECT * FROM kolide_json WHERE path = '/Users/micahsorenson/Downloads/UserConfig.json';
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+
| fullkey          | parent | key              | value                             | query | path                                           |
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+
| mgmtServer       |        | mgmtServer       | https://usea1-016.sentinelone.net | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| newConfiguration |        | newConfiguration | false                             | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| proxy            |        | proxy            |                                   | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| site             |        | site             | 0000000000000000                  | *     | /Users/micahsorenson/Downloads/UserConfig.json |
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+

osquery> SELECT * FROM kolide_json WHERE path = '/Users/micahsorenson/Downloads/config/UserConfig.json';
```

After commit:
```
osquery> SELECT * FROM kolide_json WHERE path = '/Users/micahsorenson/Downloads/UserConfig.json';
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+
| fullkey          | parent | key              | value                             | query | path                                           |
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+
| proxy            |        | proxy            |                                   | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| site             |        | site             | 0000000000000000                  | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| mgmtServer       |        | mgmtServer       | https://usea1-016.sentinelone.net | *     | /Users/micahsorenson/Downloads/UserConfig.json |
| newConfiguration |        | newConfiguration | false                             | *     | /Users/micahsorenson/Downloads/UserConfig.json |
+------------------+--------+------------------+-----------------------------------+-------+------------------------------------------------+

osquery> SELECT * FROM kolide_json WHERE path = '/Users/micahsorenson/Downloads/config/UserConfig.json';
+------------------+--------+------------------+-----------------------------------+-------+-------------------------------------------------------+
| fullkey          | parent | key              | value                             | query | path                                                  |
+------------------+--------+------------------+-----------------------------------+-------+-------------------------------------------------------+
| mgmtServer       |        | mgmtServer       | https://usea1-016.sentinelone.net | *     | /Users/micahsorenson/Downloads/config/UserConfig.json |
| newConfiguration |        | newConfiguration | false                             | *     | /Users/micahsorenson/Downloads/config/UserConfig.json |
| proxy            |        | proxy            |                                   | *     | /Users/micahsorenson/Downloads/config/UserConfig.json |
| site             |        | site             | 0000000000000000                  | *     | /Users/micahsorenson/Downloads/config/UserConfig.json |
+------------------+--------+------------------+-----------------------------------+-------+-------------------------------------------------------+
```

### DIFF
--- a/pkg/dataflatten/json.go
+++ b/pkg/dataflatten/json.go
@@ -22,7 +22,7 @@ func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 	// We don't have valid json data, so try to convert possible utf16 data to utf8.
 	rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
 	if err != nil {
-		return nil, fmt.Errorf("invalid json. attempt to transform from utf16 to utf8: %w", err)
+		return nil, fmt.Errorf("invalid json. Attempted transform from utf16 to utf8: %w", err)
 	}
 
 	return Json(rawdata, opts...)
--- a/pkg/dataflatten/json.go
+++ b/pkg/dataflatten/json.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
 )
 
 func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
@@ -11,6 +14,16 @@ func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Attempt to decode utf16 data.
+	valid_json := json.Valid(rawdata)
+	if !valid_json {
+		rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return Json(rawdata, opts...)
 }
 
--- a/pkg/dataflatten/json.go
+++ b/pkg/dataflatten/json.go
@@ -15,13 +15,14 @@ func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 		return nil, err
 	}
 
-	// Attempt to decode utf16 data.
-	valid_json := json.Valid(rawdata)
-	if !valid_json {
-		rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
-		if err != nil {
-			return nil, err
-		}
+	if json.Valid(rawdata) {
+		return Json(rawdata, opts...)
+	}
+
+	// We don't have valid json data, so try to convert possible utf16 data to utf8.
+	rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
+	if err != nil {
+		return nil, fmt.Errorf("invalid json. attempt to transform from utf16 to utf8: %w", err)
 	}
 
 	return Json(rawdata, opts...)
--- a/pkg/dataflatten/json.go
+++ b/pkg/dataflatten/json.go
@@ -2,6 +2,7 @@ package dataflatten
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 
@@ -22,7 +23,7 @@ func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 	// We don't have valid json data, so try to convert possible utf16 data to utf8.
 	rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
 	if err != nil {
-		return nil, fmt.Errorf("invalid json. Attempted transform from utf16 to utf8: %w", err)
+		return nil, errors.New("invalid json. (Despite attempted transform from utf16 to utf8")
 	}
 
 	return Json(rawdata, opts...)


pkg/dataflatten/json.go
```diff
@@ -4,13 +4,26 @@ import (
...
 )
 
 func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 	rawdata, err := os.ReadFile(file)
 	if err != nil {
 		return nil, err
 	}
+
+	// Attempt to decode utf16 data.
+	valid_json := json.Valid(rawdata)
```
RebeccaMahany, line 19: `validJson` for camel case please

RebeccaMahany (commented):
LGTM, just the one nitpick

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,26 @@ import (
...
+	if !valid_json {
+		rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return Json(rawdata, opts...)
 }
```
James-Pickett, line 28: Code is fine, but I have a style nit. We typically try to reduce nesting as much as possible, so you could do something like this:

```suggestion
        if json.Valid(rawdata) {
            return Json(rawdata, opts...)
        }
        
        // invalid json
        rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
        if err != nil {
	    return nil, err
	}

	return Json(rawdata, opts...)
}
```

James-Pickett (approved):

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,26 @@ import (
...
 
 func JsonFile(file string, opts ...FlattenOpts) ([]Row, error) {
 	rawdata, err := os.ReadFile(file)
 	if err != nil {
 		return nil, err
 	}
+
+	// Attempt to decode utf16 data.
+	valid_json := json.Valid(rawdata)
+	if !valid_json {
```
directionless, line 20:
```suggestion
	if !json.Valid(rawdata) {
	  // Attempt utf16 conversion
```

For this kind of boolean, `ok` is normal, but we can just drop the variable here

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,26 @@ import (
...
+		rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
```
directionless, line 21: Is there a way to detect this, or is just trying it the simplest pattern?

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,26 @@ import (
...
+		if err != nil {
+			return nil, err
```
directionless, line 23: I think this may end up as the wrong error. We probably want to return something like "not json" and not whatever the transform error is

pkg/dataflatten/json.go
directionless: this is also a good suggestion

pkg/dataflatten/json.go
Micah-Kolide: Best I could find at the moment is https://pkg.go.dev/github.com/saintfish/chardet, but I didn't dig very deep yet since there's not much info on it.

directionless: `json.Valid` seems reasonable. Pulling in another library is always mixed.

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,27 @@ import (
...
 	}
+
+	if json.Valid(rawdata) {
+		return Json(rawdata, opts...)
+	}
+
+	// We don't have valid json data, so try to convert possible utf16 data to utf8.
+	rawdata, _, err = transform.Bytes(unicode.UTF16(unicode.LittleEndian, unicode.UseBOM).NewDecoder(), rawdata)
+	if err != nil {
+		return nil, fmt.Errorf("invalid json. Attempted transform from utf16 to utf8: %w", err)
```
directionless, line 25:
```suggestion
		return nil, errors.New("invalid json. (Despite attempted transform from utf16 to utf8")
```

pkg/dataflatten/json.go
```diff
@@ -4,13 +4,27 @@ import (
 	"encoding/json"
```
directionless, line 4:
```suggestion
	"encoding/json"
	"errors"
```

directionless (approved):

pkg/dataflatten/json.go
directionless: I tweaked your errors. Lets see if I got the go code right in a PR suggest window 😆

PR #1391, opened 2023-10-11
[TUF] Roll out new autoupdater to nightly channel

RebeccaMahany:

Rolls out new autoupdater to nightly channel.

<details>
<summary>Future work</summary>

1. Devices on beta channel use new autoupdate library
1. Devices on stable channel use new autoupdate library
1. Ship a more up-to-date TUF repo
1. Report new version to K2 after update

</details>


directionless (approved):
Looks pretty clean. I think the difference is that you're trying to disable the old updater?

RebeccaMahany:
> Looks pretty clean. I think the difference is that you're trying to disable the old updater?

@directionless 

Yes, correct! It's a little riskier rollout-wise, but the changes are a lot cleaner this way, and I think are better for the launcher runtime overall -- no possibility of both autoupdaters trying to restart at close to the same time and getting the rungroup shutdown stuck.

PR #1392, opened 2023-10-11
add uptime to hostinfo checkup and tests

zackattack01:
This change adds uptime to the Host Info checkup. 

It will appear as informational output for doctor/flare as follows:
```
Host Info: hostname: example-macbook.localdomain, uptime: 7 days, 23 hours, 8 minutes, 38 seconds
```

It will be logged within checkpoints as follows:
```
zack:launcher❱ lastcheck 'Host Info' | jq '{uptime, uptime_friendly}'
{
  "uptime": 687982,
  "uptime_friendly": "7 days, 23 hours, 6 minutes, 22 seconds"
}
```

### DIFF
--- a/pkg/debug/checkups/host.go
+++ b/pkg/debug/checkups/host.go
@@ -9,10 +9,12 @@ import (
 	"os"
 	"runtime"
 	"strconv"
+	"strings"
 
 	"github.com/kolide/launcher/ee/desktop/runner"
 	"github.com/kolide/launcher/pkg/agent"
 	"github.com/kolide/launcher/pkg/agent/types"
+	"github.com/shirou/gopsutil/v3/host"
 )
 
 type (
@@ -38,12 +40,21 @@ func (hc *hostInfoCheckup) Run(ctx context.Context, extraFH io.Writer) error {
 	desktopProcesses := runner.InstanceDesktopProcessRecords()
 	hc.data["user_desktop_processes"] = desktopProcesses
 
+	uptimeRaw, err := host.Uptime()
+	if err != nil {
+		hc.data["uptime_friendly"] = fmt.Sprintf("ERROR: %s", err.Error())
+	} else {
+		hc.data["uptime_friendly"] = formatUptime(uptimeRaw)
+	}
+
+	hc.data["uptime"] = uptimeRaw
+
 	if runtime.GOOS == "windows" {
 		hc.data["in_modern_standby"] = hc.k.InModernStandby()
 	}
 
 	hc.status = Informational
-	hc.summary = fmt.Sprintf("hostname: %s", hc.data["hostname"])
+	hc.summary = fmt.Sprintf("hostname: %s, uptime: %s", hc.data["hostname"], hc.data["uptime_friendly"])
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/debug/checkups/host_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	tests := []struct {
+		name   string
+		uptime uint64
+		want   string
+	}{
+		{name: "1 day", uptime: 86400, want: "1 day"},
+		{name: "less than a day", uptime: 82860, want: "23 hours, 1 minute"},
+		{name: "just booted", uptime: 0, want: "0 seconds"},
+		{name: "you should reboot", uptime: 34559999, want: "399 days, 23 hours, 59 minutes, 59 seconds"},
```
RebeccaMahany, line 16: lol

RebeccaMahany (approved):

pkg/debug/checkups/host.go
```diff
@@ -71,6 +82,34 @@ func hostName() string {
...
+	}
+
+	formattedUptime := make([]string, 0)
+	// unitDivisors is a map of time units => seconds per unit
+	unitDivisors := map[string]uint64{"days": 86400, "hours": 3600, "minutes": 60, "seconds": 1}
+	// iterate over an ordered list of the keys above
+	for _, unit := range []string{"days", "hours", "minutes", "seconds"} {
+		divisor := unitDivisors[unit]
+		if uptime >= divisor {
+			if uptime == divisor { // cut plural ending if the quantity will be singular
```
zackattack01, line 99: i see the issue here, fixing

RebeccaMahany (approved):

pkg/debug/checkups/host.go
```diff
@@ -38,12 +40,21 @@ func (hc *hostInfoCheckup) Run(ctx context.Context, extraFH io.Writer) error {
...
 	hc.data["user_desktop_processes"] = desktopProcesses
 
+	uptimeRaw, err := host.Uptime()
+	if err != nil {
+		hc.data["uptime_friendly"] = fmt.Sprintf("ERROR: %s", err.Error())
+	} else {
+		hc.data["uptime_friendly"] = formatUptime(uptimeRaw)
+	}
+
+	hc.data["uptime"] = uptimeRaw
```
directionless, line 50: Shouldn't this be inside that `err == nil`

pkg/debug/checkups/host.go
```diff
@@ -71,6 +82,34 @@ func hostName() string {
 	return hostname
 }
 
+// formatUptime takes a raw uptime value in seconds
+// and returns a human friendly relative time string
+func formatUptime(uptime uint64) string {
```
directionless, line 87: Fancy! I'd take fractional days.

pkg/debug/checkups/host_test.go
```diff
@@ -0,0 +1,28 @@
...
+		{name: "just over 1 day", uptime: 86401, want: "1 day, 1 second"},
+		{name: "less than a day", uptime: 82860, want: "23 hours, 1 minute"},
+		{name: "just booted", uptime: 0, want: "0 seconds"},
+		{name: "you should reboot", uptime: 34559999, want: "399 days, 23 hours, 59 minutes, 59 seconds"},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			if got := formatUptime(tt.uptime); got != tt.want {
```
directionless, line 23: This is fine, but we usually use `expected` and `actual`

directionless (approved):

pkg/debug/checkups/host.go
zackattack01: i assumed it would just be zero value and the friendly would show the error. I can leave unset in that case if you'd like

pkg/debug/checkups/host_test.go
zackattack01: 👍 thank you!

pkg/debug/checkups/host.go
directionless: I'm never really sure if the return is defined if there's an error. So my bias would be to drop it. (it should be zero, but...)

Issue #1345, opened 2023-09-09
Launcher should support remote uninstall (or at least disable)


directionless:
> I want the ability for K2 to send a command to launcher to cause it to uninstall itself to solve the “we don’t want this device to check in anymore”

The general goal here, is to allow K2 to remotely disable an agent. Not just disable, but to remove the authentication keys effectively destroying the agent. One important caveat, is that we do _not_ want the agent to re-enroll. This means, we:
* MUST delete the enrollment secret
* MUST delete the databases
* SHOULD remove the startup file
* MAY delete the entire install

I think which of those we do depends a bit on safe we think we can implement it. `rm` is pretty safe, but we wouldn't want to accidentally delete the wrong thing.

We can build on https://github.com/kolide/launcher/pull/924

directionless:
I _expect_ we'll want to call this from both the extension, and from the control server. 

I'm imagining it's some underlying agent library. 

I'm not about about removing the startup file. Maybe okay for now, need to think harder long term

directionless:
Closed by https://github.com/kolide/launcher/pull/1393

PR #1393, opened 2023-10-11
remote uninstall / disable

James-Pickett:
relates to https://github.com/kolide/launcher/issues/1345
relates to https://github.com/kolide/k2/pull/8933

adds action and code for remote uninstall / disable


pkg/launcher/uninstall/uninstall_darwin.go
```diff
@@ -0,0 +1,35 @@
...
+	cmd := exec.CommandContext(ctx, launchCtlPath, launchCtlArgs...)
+	if out, err := cmd.Output(); err != nil {
+		return fmt.Errorf("unloading launcher daemon: %w: %s", err, out)
+	}
+
+	return nil
+}
+
+func removeInstallation() error {
+	cmd := exec.Command("/usr/sbin/pkgutil", "--forget", "com.kolide.k2.launcher")
```
James-Pickett, line 30: use CommandContext

pkg/launcher/uninstall/uninstall_windows.go
```diff
@@ -0,0 +1,34 @@
...
+		return fmt.Errorf("opening launcher service: %w", err)
+	}
+	defer launcherSvc.Close()
+
+	cfg, err := launcherSvc.Config()
+	if err != nil {
+		return fmt.Errorf("getting launcher service config: %w", err)
+	}
+
+	cfg.StartType = mgr.StartManual
```
James-Pickett, line 28: should we set this to disabled?

pkg/launcher/uninstall/uninstall.go
```diff
@@ -0,0 +1,71 @@
+package uninstall
```
James-Pickett, line 1: should we call this "disable" ?

ee/agent/reset.go
```diff
@@ -340,16 +340,13 @@ func getLocalPubKey(k types.Knapsack) ([]byte, error) { // nolint:unused
 	return pubKeyBytes, nil
 }
 
-// wipeDatabase iterates over all stores in the database, deleting all keys from
+// WipeDatabase iterates over all stores in the database, deleting all keys from
 // each one.
-func wipeDatabase(ctx context.Context, k types.Knapsack) error { // nolint:unused
-	return errors.New("currently not supported")
-	/*
-		for storeName, store := range k.Stores() {
```
directionless, line 348: Why was this commented out? Was it an additional safety for `DetectAndRemediateHardwareChange`, which looks like it's not calling anyhow? (@RebeccaMahany ?)

pkg/launcher/uninstall/uninstall.go
directionless: TBH I don't feel strongly. I think we're likely to iterate and shift and we figure out all the parts.

pkg/launcher/uninstall/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// TODO: remove start up files
+	// TODO: remove installation
+
+	os.Exit(0)
+}
+
+// uninstallNoExit just removes the enroll secret file and wipes the database without
+// exiting the process. This is so that we can test a portion of the uninstall process.
```
directionless, line 30:
```suggestion
// uninstallNoExit just removes the enroll secret file and wipes the database without
// exiting the process. This is so that we can test a portion of the uninstall process. This
// logs errors, but does not return them, because we want to try each uninstall independently
```

directionless: I think?

pkg/launcher/uninstall/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+	slogger := k.Slogger().With("component", "uninstall")
+
+	if err := removeEnrollSecretFile(k); err != nil {
+		slogger.Log(ctx, slog.LevelError,
+			"removing enroll secret file",
+			"err", err,
+		)
+	}
+
+	agent.WipeDatabase(ctx, k)
```
directionless, line 41:
```suggestion
	if err := agent.WipeDatabase(ctx, k); err != nil {
	    slogger(...)
	}
```

pkg/launcher/uninstall/uninstall.go
```diff
@@ -0,0 +1,54 @@
...
+	"os"
+
+	"github.com/kolide/launcher/ee/agent"
+	"github.com/kolide/launcher/ee/agent/types"
+)
+
+func Uninstall(ctx context.Context, k types.Knapsack) {
+	uninstallNoExit(ctx, k)
+
+	if err := disableAutoStart(ctx); err != nil {
```
directionless, line 16: why isn't this in `uninstallNoExit`?

pkg/launcher/uninstall/uninstall_linux.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/kolide/launcher/ee/allowedcmd"
+)
+
+func disableAutoStart(ctx context.Context) error {
+	// the --now flag will disable and stop the service
+	cmd, err := allowedcmd.Systemctl(ctx, "disable", "--now", "launcher.kolide-k2.service")
```
directionless, line 12: This will only work with systemd. Which is _most_ but not all of our systems. OTOH, I think this this disable function isn't going to last for more than a release so 🤷

pkg/launcher/uninstall/uninstall_windows.go
directionless: Not sure it matters. I think the the usual recovery path is to install/repair an MSI over it

directionless (commented):
Looks pretty good. Probably move to `ee`?

pkg/launcher/uninstall/uninstall.go
James-Pickett: done, except replaced "try each uninstall" with "try each step"

pkg/launcher/uninstall/uninstall.go
James-Pickett: done

pkg/launcher/uninstall/uninstall.go
directionless: Hrm... This raises some interesting questions.

I think you're saying that testing this locally, triggers the uninstall actions, which might uninstall/disable your prod launcher. 

Does it make sense to allow passing the identifier in, or some other test mode that uses different files?

James-Pickett: we could do that or create something like an `autoStartDisabler` interface that we pass in, the test could at least verify that a func like `disableAutoStart` was called

ee/agent/reset.go
```diff
@@ -340,16 +340,13 @@ func getLocalPubKey(k types.Knapsack) ([]byte, error) { // nolint:unused
...
-			if err := store.DeleteAll(); err != nil {
-				return fmt.Errorf("deleting keys in store %s: %w", storeName, err)
-			}
+func WipeDatabase(ctx context.Context, k types.Knapsack) error { // nolint:unused
```
directionless, line 345: Do we still need the nolint?

ee/uninstall/uninstall.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if err := disableAutoStart(ctx); err != nil {
+		k.Slogger().Log(ctx, slog.LevelError,
+			"disabling auto start",
+			"err", err,
+		)
+	}
+
+	// TODO: remove start up files
+	// TODO: remove installation
```
directionless, line 24: TODO implies we want that. I think it's not clear. Are you planning that?

pkg/launcher/uninstall/uninstall.go
directionless: Or maybe have it take a boolean that CI passes differently. Or use the compiled constant trick.

But it feels weird to have it in two methods.

directionless (commented):
I think this is basically mergable. I think we should figure out the method split, and then it's good enough.

I think we'll need to figure out how to call it outside the action queue. (I have a sneaky plan) but that can be a followup.

ee/agent/reset.go
James-Pickett: removed

ee/uninstall/uninstall.go
James-Pickett: no, removed

pkg/launcher/uninstall/uninstall.go
James-Pickett: Combined the functions

directionless (approved):
wooo!

PR #1394, opened 2023-10-11
Allow actor Interrupt functions to be called multiple times without blocking

RebeccaMahany:
Updates remaining actors' `Interrupt` functions to ensure all are safe to be called multiple times without blocking. Ensures all actors have tests verifying same.

Fixes issue where trace exporter could leave a goroutine running without being able to interrupt it if launcher shuts down within the first 3 minutes.


pkg/log/logshipper/logshipper.go
```diff
@@ -87,9 +92,7 @@ func (ls *LogShipper) Ping() {
 }
 
 func (ls *LogShipper) Run() error {
-	ctx, cancel := context.WithCancel(context.Background())
-	ls.stopFunc = cancel
-	return ls.sendBuffer.Run(ctx)
+	return ls.sendBuffer.Run(ls.ctx)
```
RebeccaMahany, line 95: I had to move setting `ls.stopFunc` into `New` to avoid a data race in the test, where `Run` and `Stop` could be reading and setting `ls.stopFunc` at the same time.

directionless: This feels wrong to me. I'd put a mutex on it if needed.

(It's not high performance, so...)

pkg/log/logshipper/logshipper.go
```diff
@@ -92,10 +90,19 @@ func (ls *LogShipper) Ping() {
...
+	ls.stopFuncMutex.Unlock()
+
+	return ls.sendBuffer.Run(ctx)
 }
 
 func (ls *LogShipper) Stop(_ error) {
+	ls.stopFuncMutex.Lock()
+	defer ls.stopFuncMutex.Unlock()
+
 	ls.stopFunc()
```
directionless, line 106: Probably okay, but I think this will panic if Stop is called before Run

directionless (dismissed):

directionless (approved):

PR #1399, opened 2023-10-12
fails checkup when no kolide process running as root

James-Pickett:


pkg/debug/checkups/processes.go
```diff
@@ -56,6 +57,15 @@ func (c *Processes) Run(ctx context.Context, fullWriter io.Writer) error {
...
 			c.data[fmt.Sprintf("%d", p.Pid)] = pMap
+
+			if c.kolideRunningAsRoot {
+				continue
+			}
+
+			username := pMap["username"].(string)
+			if username == "root" || username == "NT AUTHORITY\\SYSTEM" {
+				c.kolideRunningAsRoot = true
+			}
```
directionless, line 68:
```suggestion

			if !c.kolideRunningAsRoot {
				username := pMap["username"].(string)
				if username == "root" || username == "NT AUTHORITY\\SYSTEM" {
					c.kolideRunningAsRoot = true
				}
			}
```

In this case, I think _not_ early return is more clear.

directionless (dismissed):
Sure! Thank you

pkg/debug/checkups/processes.go
James-Pickett: agree, updated

directionless (approved):

RebeccaMahany (approved):

PR #1400, opened 2023-10-12
add running user in debug upload request, unexport flare env type

James-Pickett:


RebeccaMahany (approved):

directionless (approved):
Seems reasonable!

Do you think this should go into a checkup, or stay here in the shipper? Approved if you want it here

James-Pickett:
> Seems reasonable!
> 
> Do you think this should go into a checkup, or stay here in the shipper? Approved if you want it here

I will leave this in the shipper as this is the data sent to server when requesting a url for flare upload.

PR #1401, opened 2023-10-13
Improve TUF test reliability and speed

RebeccaMahany:
I've noticed a lot of flaky errors with the TUF tests recently, and I think it's because they make heavier use of `CopyBinary`, which copies the test executable. (Example failure: https://github.com/kolide/launcher/actions/runs/6508335416/job/17677493575.) This PR tries out symlinking instead of copying to see if we experience fewer errors that way.

Also, the TUF tests are super slow. I determined this was because of the compression step -- locally, it takes 8 seconds per compression for me, but I suspect longer in the CI pipeline. Each usage of the TUF CI server (six separate instances) performs three compressions, so 18 compressions per test run. This PR embeds pre-compressed test binaries to significantly speed up the TUF tests.

Prior to this change, TUF autoupdater tests in the pipeline were taking an average of 209 seconds, or about 3.5 minutes (averaged over the most recent 5 CI jobs). In this PR, they took an average of just under 10 seconds.


pkg/autoupdate/tuf/ci/tuf_server.go
```diff
@@ -17,6 +15,34 @@ import (
 	"github.com/theupdateframework/go-tuf"
 )
 
+var (
+	//go:embed testdata/darwin_launcher.tar.gz
+	darwinLauncher []byte
```
directionless, line 20: Feels like this might have been cleaner with the embed.FS

pkg/autoupdate/tuf/ci/tuf_server.go
```diff
@@ -17,6 +15,34 @@ import (
...
+	//go:embed testdata/darwin_osqueryd.tar.gz
+	darwinOsqueryd []byte
+	//go:embed testdata/linux_launcher.tar.gz
+	linuxLauncher []byte
+	//go:embed testdata/linux_osqueryd.tar.gz
+	linuxOsqueryd []byte
```
directionless, line 26: These were compiled from the same source? Does it make sense to embed the same thing twice, and slightly reduce the git checkout size?

directionless (dismissed):
Sure! Faster tests! 🚀

pkg/autoupdate/tuf/ci/tuf_server.go
RebeccaMahany: You mean embedding the entire testdata dir with embed.FS, or something else? I can take a look!

pkg/autoupdate/tuf/ci/tuf_server.go
RebeccaMahany: Yes, compiled from the same source, but I renamed the executable for each (launcher/launcher.exe vs osqueryd/osqueryd.exe) -- the executable's name has to match the given TUF target. Not sure there's a simple way around that in the tests, but maybe there's something I'm not thinking of?

pkg/autoupdate/tuf/ci/tuf_server.go
RebeccaMahany: Updated and I like it better! That also sped up the tests even more

pkg/autoupdate/tuf/ci/tuf_server.go
directionless: I'm not sure it's worth it, but embed is just about stashing the binary bits in a go structure. How you extract them is up to you. You have:
https://github.com/kolide/launcher/pull/1401/files#diff-114d276c4d96fce71399f71f2f007b77eaddd1a3b4109e1b8430cc9d7f4c8ec0R21-R28
```go
func getTarballContents(t *testing.T, binary string) []byte {
	tarballName := fmt.Sprintf("testdata/%s_%s.tar.gz", runtime.GOOS, binary)

	contents, err := testTarballs.ReadFile(tarballName)
	require.NoError(t, err)

	return contents
}
```

But if you wanted to merge binaries, I think you'd do something like:
```
func getTarballContents(t *testing.T) []byte {
	tarballName := fmt.Sprintf("testdata/%s.tar.gz", runtime.GOOS)

	contents, err := testTarballs.ReadFile(tarballName)
	require.NoError(t, err)

	return contents
}
```

All the naming is in `f.Write(getTarballContents(t, b))`

directionless (approved):

pkg/autoupdate/tuf/ci/tuf_server.go
RebeccaMahany: It's not just the tarball that has to be renamed, though -- the binary inside the tarball that needs to be renamed also?

directionless: _oh_. That makes sense. And the whole point of this is to pre-generate the tarballs. :hides:

PR #1405, opened 2023-10-16
Bump the otel version

directionless:
Closes #1403

Bump otel, as dependabot noticed in #1403

And fix the import statement that was breaking `go mod tidy`


pkg/tools/tools.go
```diff
@@ -11,5 +11,5 @@ For additional documentation on the topic:
 package tools
 
 import (
-	_ "github.com/go-bindata/go-bindata/go-bindata/"
+	_ "github.com/go-bindata/go-bindata/go-bindata"
```
RebeccaMahany, line 14: When I make this change locally, `make deps` fails for me, which I think is also breaking the pipeline build -- I think this started after the upgrade to 1.21. (I've been removing this slash every time I run `go mod tidy` and then adding it back so that `make -j2 github-build` doesn't fail for me, which is...an annoying workflow.)

directionless: You know... It might be fine to drop this from `go.mod`

PR #1411, opened 2023-10-17
[TUF] If a downgrade occurs, restart the given binary

RebeccaMahany:
As I started to look at automated tests for autoupdate, I noticed that if the release version for a binary switches to a version already in the local library, the binary won't be restarted to run the appropriate version. This PR corrects that oversight, ensuring that we don't perform unnecessary restarts when running dev builds.


zackattack01 (approved):
LGTM 🚀

PR #1416, opened 2023-10-18
Add `pkgutil --forget` to the darwin uninstall command

directionless:


cmd/launcher/uninstall_darwin.go
```diff
@@ -36,13 +39,27 @@ func removeLauncher(ctx context.Context, identifier string) error {
...
+	if !removeErr {
+		pkgutiltCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		defer cancel()
+
+		cmd := exec.CommandContext(pkgutiltCtx, "/usr/sbin/pkgutil", "--forget", fmt.Sprintf("com.%s.launcher", identifier))
+		if out, err := cmd.Output(); err != nil {
+			fmt.Printf("error occurred while forgetting package: output %s: err: %s\n", out, err)
+		}
+	}
+
```
James-Pickett, line 62: I think we should not print a success message at the end unless it all worked?

```suggestion
        if removeErr {
                return nil
        }
        
        pkgutiltCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
	defer cancel()      
        cmd := exec.CommandContext(pkgutiltCtx, "/usr/sbin/pkgutil", "--forget", fmt.Sprintf("com.%s.launcher", identifier))
        
        if out, err := cmd.Output(); err != nil {
                fmt.Printf("error occurred while forgetting package: output %s: err: %s\n", out, err)
                return nil
	}

```

James-Pickett (dismissed):

James-Pickett (approved):

Issue #1404, opened 2023-10-16
Write metadata to root directory


RebeccaMahany:
Write a metadata.json and metadata.plist file to launcher's root directory, including device ID, organization ID, launcher version, and time of file generation. We can do this at the same time that we write out the launcher version file (see [here](https://github.com/kolide/launcher/blob/main/cmd/launcher/internal/record_launcher_version.go)).

PR #1417, opened 2023-10-19
write out metadata json and plist files to root install directory

zackattack01:
addresses https://github.com/kolide/launcher/issues/1404

### DIFF
--- a/cmd/launcher/internal/record_metadata.go
+++ b/cmd/launcher/internal/record_metadata.go
@@ -49,7 +49,7 @@ func (mw *metadataWriter) Ping() {
 	}
 
 	if err := mw.recordMetadata(metadata); err != nil {
-		level.Error(mw.logger).Log("msg", "unable to write out metadata files", "err", err)
+		level.Debug(mw.logger).Log("msg", "unable to write out metadata files", "err", err)
 	}
 }
 
--- a/cmd/launcher/internal/record_metadata.go
+++ b/cmd/launcher/internal/record_metadata.go
@@ -1,0 +1,67 @@
+package internal
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"time"
+
+	"github.com/groob/plist"
+	"github.com/kolide/kit/version"
+	"github.com/kolide/launcher/pkg/agent/types"
+	"github.com/kolide/launcher/pkg/debug/checkups"
+)
+
+type metadata struct {
+	DeviceId       string `json:"device_id" plist:"device_id"`
+	OrganizationId string `json:"organization_id" plist:"organization_id"`
+	Timestamp      string `json:"timestamp" plist:"timestamp"`
+	Version        string `json:"version" plist:"version"`
+}
+
+func RecordMetadata(rootDir string, ctx context.Context, knapsack types.Knapsack) error {
+	metadataJSONFile := filepath.Join(rootDir, "metadata.json")
+	sdc := checkups.NewServerDataCheckup(knapsack)
+	if err := sdc.Run(ctx, io.Discard); err != nil {
+		return fmt.Errorf("unable to gather metadata, error: %w", err)
+	}
+
+	metadata := metadata{
+		DeviceId:       sdc.DeviceId,
+		OrganizationId: sdc.OrganizationId,
+		Timestamp:      time.Now().String(),
+		Version:        version.Version().Version,
+	}
+
+	metadataJSON, err := json.MarshalIndent(metadata, "", "  ")
+	if err != nil {
+		return fmt.Errorf("unable to JSON marshal metadata, error: %w", err)
+	}
+
+	err = os.WriteFile(metadataJSONFile, metadataJSON, 0644)
+	if err != nil {
+		return fmt.Errorf("unable to write JSON metadata, error: %w", err)
+	}
+
+	if runtime.GOOS != "darwin" {
+		return nil
+	}
+
+	metadataPlistFile := filepath.Join(rootDir, "metadata.plist")
+	metadataPlist, err := plist.MarshalIndent(metadata, "  ")
+
+	if err != nil {
+		return fmt.Errorf("unable to Plist marshal metadata, error: %w", err)
+	}
+
+	err = os.WriteFile(metadataPlistFile, metadataPlist, 0644)
+	if err != nil {
+		return fmt.Errorf("unable to write plist metadata, error: %w", err)
+	}
+
+	return nil
+}
--- a/cmd/launcher/internal/record_metadata.go
+++ b/cmd/launcher/internal/record_metadata.go
@@ -28,9 +28,8 @@ type metadata struct {
 
 // RecordMetadata writes out both a json and plist (for darwin) file including all information
 // in the metadata struct to the root install directory
-func RecordMetadata(rootDir string, ctx context.Context, knapsack types.Knapsack, logger log.Logger) {
-	logger = log.With(logger, "method")
-	metadataJSONFile := filepath.Join(rootDir, "metadata.json")
+func RecordMetadata(ctx context.Context, logger log.Logger, knapsack types.Knapsack) {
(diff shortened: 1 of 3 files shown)


cmd/launcher/launcher.go
```diff
@@ -193,6 +193,9 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	// we expect we're live. Record the version for osquery to
 	// pickup
 	internal.RecordLauncherVersion(rootDirectory)
+	if err = internal.RecordMetadata(rootDirectory, ctx, k); err != nil {
+		level.Error(logger).Log("msg", "unable to write metadata", "error", err.Error())
```
James-Pickett, line 197: we typically just use "err" as the key for error and don't need to add the `.String()`
```suggestion
		level.Error(logger).Log("msg", "unable to write metadata", "err", err)
```

cmd/launcher/launcher.go
```diff
@@ -193,6 +193,9 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	// we expect we're live. Record the version for osquery to
 	// pickup
 	internal.RecordLauncherVersion(rootDirectory)
+	if err = internal.RecordMetadata(rootDirectory, ctx, k); err != nil {
```
James-Pickett, line 196: Are we confident we'll have all this data on a fresh install since it has to come down via control server? Maybe that's okay and it will get written on the next launcher start up? or perhaps we would to do some retry logic in a goroutine.

zackattack01: discussed offline^ my testing may have been flawed here- adding some async retry, thank you!

James-Pickett (approved):
Nice

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	metadataJSON, err := json.MarshalIndent(metadata, "", "  ")
+	if err != nil {
+		level.Error(logger).Log("msg", "unable to JSON marshal metadata", "err", err)
+		return
+	}
+
+	err = os.WriteFile(metadataJSONFile, metadataJSON, 0644)
+	if err != nil {
```
directionless, line 69:
```suggestion
	if err := os.WriteFile(metadataJSONFile, metadataJSON, 0644); err != nil
```

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	metadataPlistFile := filepath.Join(rootDir, "metadata.plist")
+	metadataPlist, err := plist.MarshalIndent(metadata, "  ")
+
+	if err != nil {
+		level.Error(logger).Log("msg", "unable to plist marshal metadata", "err", err)
+		return
+	}
+
+	err = os.WriteFile(metadataPlistFile, metadataPlist, 0644)
+	if err != nil {
```
directionless, line 87:
```suggestion
	if err := os.WriteFile(metadataPlistFile, metadataPlist, 0644); err != nil {
```

cmd/launcher/launcher.go
directionless: Hrm... if we're really concerned about that, I wonder if we should move the write into the server data consumer.

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,90 @@
...
+	DeviceId       string `json:"device_id" plist:"device_id"`
+	OrganizationId string `json:"organization_id" plist:"organization_id"`
+	Timestamp      string `json:"timestamp" plist:"timestamp"`
+	Version        string `json:"version" plist:"version"`
+}
+
+// RecordMetadata writes out both a json and plist (for darwin) file including all information
+// in the metadata struct to the root install directory
+func RecordMetadata(rootDir string, ctx context.Context, knapsack types.Knapsack, logger log.Logger) {
+	logger = log.With(logger, "method")
```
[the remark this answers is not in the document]
zackattack01: 🤦 will fix this

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,90 @@
...
+type metadata struct {
+	DeviceId       string `json:"device_id" plist:"device_id"`
+	OrganizationId string `json:"organization_id" plist:"organization_id"`
+	Timestamp      string `json:"timestamp" plist:"timestamp"`
+	Version        string `json:"version" plist:"version"`
+}
+
+// RecordMetadata writes out both a json and plist (for darwin) file including all information
+// in the metadata struct to the root install directory
+func RecordMetadata(rootDir string, ctx context.Context, knapsack types.Knapsack, logger log.Logger) {
```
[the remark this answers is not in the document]
zackattack01: cleaning this up, yes it should be just making sure it is initialized by then

cmd/launcher/launcher.go
James-Pickett: yea, this is better

directionless: actually, a subscriber makes more sense

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/groob/plist"
+	"github.com/kolide/kit/version"
+	"github.com/kolide/launcher/pkg/agent/types"
+	"github.com/kolide/launcher/pkg/backoff"
+	"github.com/kolide/launcher/pkg/debug/checkups"
+)
+
+type metadata struct {
+	DeviceId       string `json:"device_id" plist:"device_id"`
+	OrganizationId string `json:"organization_id" plist:"organization_id"`
```
directionless, line 24: Can you add OrganizationMunemo too?

zackattack01: yep!

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,129 @@
...
+	if store == nil {
+		return errors.New("ServerProvidedDataStore is uninitialized")
+	}
+
+	deviceId, err := store.Get([]byte("device_id"))
+	if err != nil {
+		return err
+	} else if string(deviceId) == "" {
+		return errors.New("device_id is not yet present in ServerProvidedDataStore")
+	}
```
James-Pickett, line 75: done need the else if here since you're returning
```suggestion
	if err != nil {
		return err
	}
	
	if string(deviceId) == "" {
		return errors.New("device_id is not yet present in ServerProvidedDataStore")
	}
```

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,129 @@
...
+
+	metadata.DeviceId = string(deviceId)
+
+	organizationId, err := store.Get([]byte("organization_id"))
+	if err != nil {
+		return err
+	} else if string(organizationId) == "" {
+		return errors.New("organization_id is not yet present in ServerProvidedDataStore")
+	}
```
James-Pickett, line 84: same
```suggestion
	if err != nil {
		return err
	}
	
	if string(organizationId) == "" {
		return errors.New("organization_id is not yet present in ServerProvidedDataStore")
	}
```

cmd/launcher/internal/record_metadata.go
```diff
@@ -0,0 +1,129 @@
...
+
+	metadata.OrganizationId = string(organizationId)
+
+	munemo, err := store.Get([]byte("munemo"))
+	if err != nil {
+		return err
+	} else if string(munemo) == "" {
+		return errors.New("munemo is not yet present in ServerProvidedDataStore")
+	}
```
James-Pickett, line 93: same ... starting to feel like this should be a little helper func
```suggestion
	if err != nil {
		return err
	}
	
	if string(munemo) == "" {
		return errors.New("munemo is not yet present in ServerProvidedDataStore")
	}
```

zackattack01: agreed, I will pull these into a helper. thank you!

James-Pickett (approved):

directionless (approved):

PR #1419, opened 2023-10-20
adds std lib slog logger to knapsack writing to existing log destinations

James-Pickett:


cmd/launcher/launcher.go
```diff
@@ -168,13 +170,34 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
 	flagController := flags.NewFlagController(logger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db)
 
+	slogLevel := slog.LevelInfo
+	if k.Debug() {
+		slogLevel = slog.LevelDebug
+	}
```
RebeccaMahany, line 176: Unless I'm missing something, I think this departs from our current expectation for how levels work (where the debug flag is respected when writing to stderr, but we ignore it when writing to debug.json so that all logs regardless of level end up there). Do we plan on starting to log a lot more at the info level and higher with slog? Or if we want to keep current behavior, I think we could alter the level we pass in via `slogHandlerOpts` per call to `k.AddLogHandler`.

James-Pickett: yep, updated!

pkg/log/logshipper/logshipper.go
```diff
@@ -117,6 +118,10 @@ func (ls *LogShipper) Log(keyvals ...interface{}) error {
 	return ls.shippingLogger.Log(keyvals...)
 }
 
+func (ls *LogShipper) Writer() io.Writer {
```
directionless, line 121: I assume this is exposed for `slog` -- why expose the Writer, and not an `slog.Handler`? (honest question -- I don't know why one over the other)

directionless: I guess one reason to expose a Writer, and not Handler, is that changing the level is very tied to the handler. I think I'm going to drop a pile of comments about that elsewhere in this review

pkg/log/locallogger/locallogger.go
```diff
@@ -18,7 +19,7 @@ type localLogger struct {
 	logger log.Logger
 }
 
-func NewKitLogger(logFilePath string) log.Logger {
+func NewKitLogger(logFilePath string) (log.Logger, io.Writer) {
```
directionless, line 22: This feels like a weird change.

cmd/launcher/launcher.go
```diff
@@ -168,13 +170,49 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+			Key:   slog.TimeKey,
+			Value: slog.AnyValue(time.Now().UTC()),
+		}
+	}
+
+	k.AddLogHandler(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		AddSource:   true,
+		Level:       slogLevel,
+		ReplaceAttr: slogAttrReplacementFunc,
+	}))
```
directionless, line 193: I'm not sure if we want that, and TBH, it reminds me of a thing I never really solved in #1322. To wit:

In today's world, we log info the os facilities, and debug to internal logs. _And_ when running from the command line, during devel, we usually want debug to stderr. _But_ because on macOS and linux stderr is the system log, this leaves us with a bit of a gordian knot.

I think my idea world has something like:
- all logs written to internal debug logs
- all logs sent to logshipper, though maybe only shipped if the control server says so
- Some startup errors sent to system logs

But that leaves an question about how to handle dev usage.

I wonder if we should slightly repurpose the `debug` flag. When it's present, we copy all logs to the stderr. And when it's not, only the startup logs. Eg, kinda making it something like "I am a developer running this on the command line"

cmd/launcher/launcher.go
```diff
@@ -168,13 +170,49 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+
+	k.AddLogHandler(slog.NewJSONHandler(logFileWriter, &slog.HandlerOptions{
```
directionless, line 195: I _think_ but I'm not sure, this might be cleaner if the `locallogger` took a `*slog. Level`, and returned a handler. Though maybe too many of the options should be configured up here?

And maybe we should be standardizing on io.Writers instead of log.Loggers?

directionless (commented):
(early thoughts)

cmd/launcher/launcher.go
directionless: I often run launcher in the foreground when debugging k2. (And I bet k2 folks do as well) And in that use case, I very much want my logs in my terminal.

cmd/launcher/main.go
```diff
@@ -97,9 +98,16 @@ func main() {
 	// recreate the logger with  the appropriate level.
 	logger = logutil.NewServerLogger(opts.Debug)
 
+	// It's kind of weird exposing the log writers like this, but at this time
+	// were evaluating transitioning to std lib slog and making it accessible
+	// via knapsack. So we have to pass the writer around so multiple loggers
+	// can log to same file.
```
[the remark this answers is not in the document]
James-Pickett: The primary issue I see with that is that we would loose control over the handler(s). There is a func hanging on slog that returns the handler, but if you take that handler and wrap with more, you'll get some weird behavior.

```go
func main() {
	logger := slog.New(slog.NewTextHandler(os.Stdout, nil)).With("logger", "first_logger")
	logger.Info("one handler")
	// time=2023-10-23T09:00:27.306-07:00 level=INFO msg="one handler" logger=first_logger

	multiHandler := slogmulti.Fanout(logger.Handler(), slog.NewJSONHandler(os.Stdout, nil))
	logger = slog.New(multiHandler).With("logger", "second_logger")
	logger.Info("two handlers")
	// notice our first handler now has 2 "logger" attributes
	//
	// time=2023-10-23T09:00:27.306-07:00 level=INFO msg="two handlers" logger=first_logger logger=second_logger
	// {"time":"2023-10-23T09:00:27.306788-07:00","level":"INFO","msg":"two handlers","logger":"second_logger"}
}
```

Maybe we should create a pkg that wraps slog and allows us to manage handlers?

cmd/launcher/launcher.go
James-Pickett: not sure if leveling is the right approach, but we could create a systemLevel and add a handler that write to stderr only when someone uses system level

cmd/launcher/launcher.go
James-Pickett: > And maybe we should be standardizing on io.Writers instead of log.Loggers?
 Maybe? I think it's just weird because we have two now, I would think we would gradually migrate toward slog.

I added a `multislogger` stuct that tracks the handlers and now passing that around in the ctx from main.go -> launcher.go. I think it's cleaner, still had to expose the writer on locallogger, but passing it around.

cmd/launcher/launcher.go
James-Pickett: I went ahead and added this, just to see how it felt, was simple

pkg/log/locallogger/locallogger.go
James-Pickett: yea, i droped it and just expose via Writer() func

pkg/log/logshipper/logshipper.go
James-Pickett: yes, it is tied to the handler

cmd/launcher/launcher.go
```diff
@@ -166,7 +168,26 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+	k.AddLogHandler(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
+		AddSource:   true,
+		Level:       slogLevel,
+		ReplaceAttr: slogAttrReplacementFunc,
+	}))
+
+	// set up system log
+	const levelSystem = slog.Level(20)
+	k.AddLogHandler(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		Level: levelSystem,
```
James-Pickett, line 189:
```suggestion
	k.AddLogHandler(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
		Level: slog.Level(20),
```

pkg/contexts/ctxlog/ctxlog.go
```diff
@@ -4,17 +4,25 @@ import (
...
+	"github.com/kolide/launcher/pkg/log/multislogger"
 	"go.opencensus.io/trace"
 )
 
 type key int
 
-const loggerKey key = 0
+const (
+	loggerKey  key = 0
+	sloggerKey key = 1
```
directionless, line 15: I'm not sure we should pass slogger through context. 

If the contextual cues come from using `DebugContext(ctx, ...)` then carrying slogger in context feels off.

I might be wrong. Was there somewhere that felt better with this?

pkg/log/logshipper/logshipper.go
directionless: I think this should expose a Handler. Probably just by pulling in `slog.NewJSONHandler(logShipper.Writer(), &slog.HandlerOptions{...})`. 

And then we can also expose `shipper.SetSloggerLevel(...)` with something akin to:
```
        loggerLevel := new(slog.LevelVar)
        loggerLevel.Set(slog.LevelInfo)
        logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: loggerLevel })
```

(yoinked from some personal project code)

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+func (m *MultiSlogger) AddReplaceHandler(name string, handler slog.Handler, matchers ...func(ctx context.Context, r slog.Record) bool) *MultiSlogger {
+	m.handlers[name] = MultiSloggerHandler{
+		handler:  handler,
+		matchers: matchers,
+	}
+
+	router := slogmulti.Router()
+	for _, handler := range m.handlers {
+		router = router.Add(handler.handler, handler.matchers...)
+	}
```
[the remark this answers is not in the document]
directionless: I suspect we're incurring complexity to do `AddReplace` and not simply `Add`, and I wonder if there's a usecase

pkg/agent/types/flags.go
```diff
@@ -181,6 +181,10 @@ type Flags interface {
 	SetLogIngestServerURL(url string) error
 	LogIngestServerURL() string
 
+	// LogShippingLevel is the level at which logs should be shipped to the server
+	SetLogShippingLevel(level string) error
+	LogShippingLevel() string
```
directionless, line 186: Flags implies command line / config presence, right? I'm not sure we need to expose `LogShippingLevel` at that level. Default it to Error, and allow the control server to override.

directionless: (eg: can logshipper be a subscriber of the control system)

cmd/launcher/main.go
```diff
@@ -96,10 +98,19 @@ func main() {
 
 	// recreate the logger with  the appropriate level.
 	logger = logutil.NewServerLogger(opts.Debug)
+	var multiSlogger *multislogger.MultiSlogger
```
directionless, line 101: Why not `multiSlogger := multislogger.New()` here?

cmd/launcher/svc_windows.go
```diff
@@ -62,7 +62,8 @@ func runWindowsSvc(args []string) error {
 
 	// Create a local logger. This logs to a known path, and aims to help diagnostics
 	if opts.RootDirectory != "" {
-		logger = teelogger.New(logger, locallogger.NewKitLogger(filepath.Join(opts.RootDirectory, "debug.json")))
+		ll := locallogger.NewKitLogger(filepath.Join(opts.RootDirectory, "debug.json"))
+		logger = teelogger.New(logger, ll)
```
directionless, line 66: no slogger here?

pkg/contexts/ctxlog/ctxlog.go
directionless: If it's just to pass it to runLauncher, I'd change the signature.

cmd/launcher/launcher.go
```diff
@@ -166,7 +169,24 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
+	if k.Debug() {
+		// if were in debug mode, we want to log everything including debug messages
+		// to the system log (stderr). When not in debug mode, were only logging logs
+		// with attribute system_log=true to the system log (stderr).
+		slogLevel = slog.LevelDebug
+		sysLogMatcher = nil
+	}
+
+	// system log
+	k.AddReplaceSlogHandler("system", slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
```
directionless, line 186: This probably won't work for windows.

cmd/launcher/launcher.go
```diff
@@ -166,7 +169,24 @@ func runLauncher(ctx context.Context, cancel func(), opts *launcher.Options) err
...
-	k := knapsack.New(stores, flagController, db)
+	slogger := ctxlog.FromContextWithSlogger(ctx)
+	k := knapsack.New(stores, flagController, db, slogger)
+
+	slogLevel := slog.LevelInfo
+	sysLogMatcher := multislogger.SystemLogMatcher
+	if k.Debug() {
+		// if were in debug mode, we want to log everything including debug messages
+		// to the system log (stderr). When not in debug mode, were only logging logs
+		// with attribute system_log=true to the system log (stderr).
```
directionless, line 180: I think this will work, but I wonder if it's simpler to have something like:
```go
	k.AddReplaceSlogHandler("system", ...)
	
	// If we're in debug mode, then we assume we want to echo _all_ logs to stderr. 
        if k.Debug() {
  	    k.AddReplaceSlogHandler("debugcli", slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{...}))
        }
```

directionless: This makes the split very clean, and I think handles the windows case more easily.

directionless (commented):
This feels like it's taking a reasonable shape

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+			return false
+		}
+		// continue iteration
+		return true
+	})
+
+	return ok
+}
+
+func utcTimeMiddleware(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
```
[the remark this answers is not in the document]
James-Pickett: the baked time is local time and we currently use UTC, I prefer UTC and was just trying to keep parody

pkg/log/logshipper/logshipper.go
James-Pickett: nice, I didn't realize you could do that, I was trying to reassign the level of the handler and it wasn't working so I assumed you had to replace the handler, this is way better! thank you

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
directionless: Legit. UTC does make more sense

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+
+type MultiSlogger struct {
+	*slog.Logger
+	handlers map[string]MultiSloggerHandler
+}
+
+type MultiSloggerHandler struct {
+	handler  slog.Handler
+	matchers []func(ctx context.Context, r slog.Record) bool
+}
```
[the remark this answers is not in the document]
James-Pickett: It does not need to be exported, it's just a bag to hold handlers and their matchers. IMO, it's cleaner to have a collection of structs instead of the dual maps.

We need to store the matchers because each time a handler is added, we have to rebuild the router. I'm using a the router I found [here](https://github.com/samber/slog-multi/blob/main/router.go), but we could role our own.

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
James-Pickett: The primary reason is that slogmulit.router is unexported. We could fork it / roll our own, but in terms of overall complexity, this seems the lowest of the options. Unless we find some other features or conveniences we want.

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+	m.Logger = slog.New(
+		slogmulti.
+			Pipe(slogmulti.NewHandleInlineMiddleware(utcTimeMiddleware)).
+			Pipe(slogmulti.NewHandleInlineMiddleware(ctxValuesMiddleWare)).
+			Handler(router.Handler()),
+	)
+	return m
+}
+
+func SystemLogMatcher(ctx context.Context, r slog.Record) bool {
```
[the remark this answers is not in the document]
James-Pickett: I was not sure about this one. Seemed easier to have the same func available to test and to cmd/launcher to do the routing. I'll leave a comment.

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+	ok := false
+	r.Attrs(func(attr slog.Attr) bool {
+		if attr.Key == "system_log" && attr.Value.Kind() == slog.KindBool && attr.Value.Bool() == true {
```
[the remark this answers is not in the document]
James-Pickett: I can't think of a better way at the moment, but we could cut down on the string comparison, which is probably the most expensive part. Something like:

```go
// SystemLogMatcher is a matcher that matches records that have the system_log=true attribute.
func SystemLogMatcher(ctx context.Context, r slog.Record) bool {
	isMatch := false
	r.Attrs(func(attr slog.Attr) bool {
		if attr.Value.Kind() != slog.KindBool {
			// continue iteration
			return true
		}

		// it's a boolean value
		if attr.Key == "system_log" {
			isMatch = attr.Value.Bool()

			// stop iteration
			return false
		}

		// continue iteration
		return true
	})

	return isMatch
}
```

pkg/log/logshipper/logshipper.go
James-Pickett: updated this, exposed as handler and added version as log attribute

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,86 @@
...
+	for _, key := range ctxValueKeysToAdd {
+		v := ctx.Value(key)
+		if v == nil {
+			continue
+		}
+
+		record.AddAttrs(slog.Attr{
+			Key:   key,
+			Value: slog.AnyValue(v),
+		})
```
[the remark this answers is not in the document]
James-Pickett: updated

pkg/log/multislogger/multislogger_test.go
```diff
@@ -0,0 +1,104 @@
...
+	clearBufsFn()
+
+	multislogger.Logger.Info("system_log_msg", "system_log", true)
+	require.Contains(t, debugLogBuf.String(), "system_log_msg", "should be in debug log since it's info level and that's higher than debug level")
+	require.Contains(t, shipperBuf.String(), "system_log_msg", "should be in shipper log since it's info level")
+	require.Contains(t, systemLogBuf.String(), "system_log_msg", "should be in system log since there was system_log=true attr")
+	clearBufsFn()
+
+	// replace shipper with a debug level handler
+	multislogger.AddReplaceHandler("shipper", slog.NewJSONHandler(&shipperBuf, &slog.HandlerOptions{Level: slog.LevelDebug}))
```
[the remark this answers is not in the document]
James-Pickett: now re-leveling via variable

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
James-Pickett: ehhh .... I'm not sure if the types switch is less expensive than the string compare ... but, at least then we can check the bool value before the string compare

I'm honestly not really sure whats better or if we need to find a different mechanism

cmd/launcher/main.go
James-Pickett: fixed

cmd/launcher/launcher.go
James-Pickett: yea, that's cleaner, done

pkg/agent/flags/flag_controller.go
```diff
@@ -474,6 +474,16 @@ func (fc *FlagController) LogIngestServerURL() string {
 	).get(fc.getControlServerValue(keys.LogIngestServerURL))
 }
 
+// LogShippingLevel is the level at which logs should be shipped to the server
+func (fc *FlagController) SetLogShippingLevel(level string) error {
+	return fc.setControlServerValue(keys.LogShippingLevel, []byte(level))
+}
+func (fc *FlagController) LogShippingLevel() string {
+	return NewStringFlagValue(
+		WithDefaultString("info"),
```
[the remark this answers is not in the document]
James-Pickett: added, thank you

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
directionless: This, and the prior, are really asking why we're rebuilding the router, and not simply exposing an `Add(...)` that wraps the `router.Add`

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
directionless: Right, but why do that? what circumstances lead to wanting to change a handler?

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
directionless: I think if we want to trigger system logs via KV pair, this is probably the best we can do. I remain unsure if we want to carry two sloggers, one for the system logs and one for the internal logs. Or try this.

pkg/agent/types/flags.go
directionless: Okay! I still haven't totally absorbed how the flags worked out

pkg/log/multislogger/multislogger.go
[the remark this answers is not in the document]
directionless: James walked me through this. The router doesn't work how I expect -- while it has an`Add`, the router (a) isn't an exported interface, so we can't save it. And (b) under the hood it's creating a new router anyhow.

I think we should leave a comment about why this does what it does

pkg/agent/knapsack/knapsack.go
```diff
@@ -26,23 +30,45 @@ type knapsack struct {
...
-		flags:  flags,
-		stores: stores,
+		db:      db,
+		flags:   flags,
+		stores:  stores,
+		slogger: slogger,
 	}
 
+	// TODO: figure out windows thang
+	k.systemSlogger = slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
```
directionless, line 50: Not sure if it'll make sense to pass the system slogger in at creation time, or create it here and add to it

directionless (commented):
this is looking pretty good.

pkg/log/logshipper/logshipper.go
```diff
@@ -81,6 +89,17 @@ func (ls *LogShipper) Ping() {
 		ls.sender.endpoint = parsedUrl.String()
 	}
 
+	level := slog.LevelError
+	switch ls.knapsack.LogShippingLevel() {
```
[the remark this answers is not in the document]
directionless: Okay, we can leave it as a string. 

I think you should leave the sanitization in -- it helps enforce things sooner.

I think it's still worth adding a `default` branch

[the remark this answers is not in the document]
James-Pickett: added default to switch statement

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,51 @@
...
+// AddHandler adds a handler to the multislogger
+func (m *MultiSlogger) AddHandler(handler slog.Handler) *MultiSlogger {
+	m.handlers = append(m.handlers, handler)
+
+	m.Logger = slog.New(
+		slogmulti.
+			Pipe(slogmulti.NewHandleInlineMiddleware(utcTimeMiddleware)).
+			Pipe(slogmulti.NewHandleInlineMiddleware(ctxValuesMiddleWare)).
+			// we have to rebuild the handler everytime because the slogmulti package we're
+			// using doesn't support adding handlers after the Fanout handler has been created
```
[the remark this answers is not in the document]
James-Pickett: done

pkg/log/multislogger/multislogger.go
```diff
@@ -0,0 +1,51 @@
...
+	slogmulti "github.com/samber/slog-multi"
+)
+
+type MultiSlogger struct {
+	*slog.Logger
+	handlers []slog.Handler
+}
+
+// AddHandler adds a handler to the multislogger
+func (m *MultiSlogger) AddHandler(handler slog.Handler) *MultiSlogger {
```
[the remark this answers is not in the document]
James-Pickett: done

pkg/agent/types/slogger.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"log/slog"
+)
+
+type Slogger interface {
+	// Logging interface methods
+	Slogger() *slog.Logger
+	SystemSlogger() *slog.Logger
+	AddSlogHandler(handler slog.Handler, matchers ...func(ctx context.Context, r slog.Record) bool)
```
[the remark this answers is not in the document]
James-Pickett: yes, I agree, now handlers also get added to system logger

pkg/agent/knapsack/knapsack.go
```diff
@@ -26,23 +30,45 @@ type knapsack struct {
 	// remove this field and prevent "leaking" bbolt into places it doesn't need to.
 	db *bbolt.DB
 
+	slogger       *multislogger.MultiSlogger
+	systemSlogger *slog.Logger
```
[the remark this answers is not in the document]
James-Pickett: agree, now adding handlers to system log as well

pkg/agent/knapsack/knapsack.go
James-Pickett: Updated to create in main and pass to run launcher

cmd/launcher/svc_windows.go
James-Pickett: added

cmd/launcher/main.go
```diff
@@ -97,9 +99,30 @@ func main() {
...
 	logger = logutil.NewServerLogger(opts.Debug)
 
+	// set up system slogger to write to os logs
+	systemSlogger := new(multislogger.MultiSlogger)
+	systemSlogger.AddHandler(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+	}))
+
+	// set up local slogger to write to debug.json
+	localSlogger := new(multislogger.MultiSlogger)
```
directionless, line 109:
```suggestion
	// set up slogger for internal launcher logging
	slogger := new(multislogger.MultiSlogger)
```

This should be the one people reach for first

pkg/agent/knapsack/knapsack.go
```diff
@@ -26,23 +29,43 @@ type knapsack struct {
...
+	return k.slogger.Logger
+}
+
+func (k *knapsack) SystemSlogger() *slog.Logger {
+	return k.systemSlogger.Logger
+}
+
+func (k *knapsack) AddSlogHandler(handler ...slog.Handler) {
+	k.slogger = k.slogger.AddHandler(handler...)
+	k.slogger.Logger = k.slogger.Logger.With("logger", "knapsack")
```
directionless, line 62: Do we need this?

directionless (commented):
I think we could ship this. Sweet!

pkg/log/logshipper/logshipper.go
```diff
@@ -81,6 +88,19 @@ func (ls *LogShipper) Ping() {
...
+	var level slog.Level
+	switch ls.knapsack.LogShippingLevel() {
+	case "debug":
+		level = slog.LevelDebug
+	case "info":
+		level = slog.LevelInfo
+	case "warn":
+		level = slog.LevelWarn
+	default:
+		level = slog.LevelError
```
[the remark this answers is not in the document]
James-Pickett: done

cmd/launcher/main.go
James-Pickett: done

zackattack01 (approved):

pkg/log/logshipper/logshipper.go
```diff
@@ -81,6 +88,22 @@ func (ls *LogShipper) Ping() {
...
+	case "debug":
+		ls.slogLevel.Set(slog.LevelDebug)
+	case "info":
+		ls.slogLevel.Set(slog.LevelInfo)
+	case "warn":
+		ls.slogLevel.Set(slog.LevelWarn)
+	case "error":
+		ls.slogLevel.Set(slog.LevelError)
+	case "default":
+		ls.knapsack.Slogger().Error("unrecognized flag value for log shipping level",
```
directionless, line 101: Ha! I love that we can use knapsack.Slogger here.

directionless (approved):
Woo! Looks like a good base

PR #1420, opened 2023-10-23
[Flare] Update a couple things in how flares are shipped

directionless:
Now that the SaaS side supports flare shipping, I've found a couple of issues, and tweaks:

- The most important is that we need to set a `Content-Type` header. Otherwise rails won't parse the data and we don't get the extra stuff.
- Rails will return the file name the flare is saved as. I've added that into the json struct.
- Update the output to and unify the call paths
- Set flare to upload by default

I considered, but did not, adjust flare to _also_ print doctor output. I'm on the fence about output in general. So it's still TBD

### DIFF
--- a/cmd/launcher/flare.go
+++ b/cmd/launcher/flare.go
@@ -33,7 +33,7 @@ func runFlare(args []string) error {
 		flagset            = flag.NewFlagSet("flare", flag.ExitOnError)
 		flSave             = flagset.String("save", "upload", "local | upload")
 		flOutputDir        = flagset.String("output_dir", ".", "path to directory to save flare output")
-		flUploadRequestURL = flagset.String("upload_request_url", "", "URL to request a signed upload URL")
+		flUploadRequestURL = flagset.String("upload_request_url", "https://api.kolide.com/api/agent/flare", "URL to request a signed upload URL")
 	)
 
 	if err := ff.Parse(flagset, args); err != nil {
--- a/cmd/launcher/flare.go
+++ b/cmd/launcher/flare.go
@@ -4,11 +4,13 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/pkg/agent/flags"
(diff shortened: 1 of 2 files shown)


James-Pickett (approved):

cmd/launcher/flare.go
```diff
@@ -28,32 +30,16 @@ func runFlare(args []string) error {
...
-		)
-		flUploadRequestURL = flagset.String(
-			"upload_request_url",
-			"",
-			"URL to request a signed upload URL",
-		)
+		flagset            = flag.NewFlagSet("flare", flag.ExitOnError)
+		flSave             = flagset.String("save", "upload", "local | upload")
+		flOutputDir        = flagset.String("output_dir", ".", "path to directory to save flare output")
+		flUploadRequestURL = flagset.String("upload_request_url", "", "URL to request a signed upload URL")
```
James-Pickett, line 36: should we populate this with the default URL or do we want to send down from control server? But sending from control server would mean that a non-enrolled launcher would still need this as a cmd line flag.

cmd/launcher/flare.go
```diff
@@ -67,23 +53,41 @@ func runFlare(args []string) error {
...
-		return fmt.Errorf("creating flare file (%s): %w", reportPath, err)
 	}
-	defer flareFile.Close()
 
-	return checkups.RunFlare(ctx, k, flareFile, checkups.StandaloneEnviroment)
+	if err := checkups.RunFlare(ctx, k, flareDest, checkups.StandaloneEnviroment); err != nil {
+		return err
+	}
+
+	level.Info(logger).Log("msg", successMessage, "file", flareDest.Name())
```
RebeccaMahany, line 91: I like how tidy adding `flareDest.Name()` is

RebeccaMahany (dismissed):

zackattack01 (dismissed):

cmd/launcher/flare.go
directionless: Oh! I hadn't even realized it didn't default to something. Yeah, let me update that. Thank you

directionless: How's that?

cmd/launcher/flare.go
directionless: Tiny inline interface FTW. (Not sure if inline is correct, but 🤷 )

James-Pickett (approved):

PR #1422, opened 2023-10-25
Export traces less frequently

RebeccaMahany:
The exporter already has a maximum batch size and maximum queue size configured, but we were never hitting the maximum batch size because the batch timeout is set very low by default, enforcing shipping batches every 5 seconds. This probably isn't necessary for our usecase, so I've increased the batch timeout to a minute.


directionless (dismissed):
Do we want control server adjustability here?

Or the pattern around uptime?

James-Pickett (dismissed):

RebeccaMahany:
> Do we want control server adjustability here?
> 
> Or the pattern around uptime?

I've updated so it's a control server flag!

pkg/agent/flags/keys/keys.go
```diff
@@ -45,6 +45,7 @@ const (
 	UpdateDirectory                FlagKey = "update_directory"
 	ExportTraces                   FlagKey = "export_traces"
 	TraceSamplingRate              FlagKey = "trace_sampling_rate"
+	TraceBatchTimeout              FlagKey = "trace_batch_timeout"
```
directionless, line 48: Do we need this as an exported flag?

directionless (approved):
If you think this should be exposed as a flag, okay.

pkg/agent/flags/keys/keys.go
RebeccaMahany: You're right -- removed from pkg/launcher/options.go, which will keep it from being a command-line/config flag. (We still need it here for all the control server setting stuff to work as expected -- this is the key it's stored under for the control server.)

directionless: I feel like we don't have great idioms yet.

directionless (approved):

PR #1423, opened 2023-10-25
Ignore duplicate libraries warning during build

RebeccaMahany:
Temporary fix for an annoying build issue

PR #1424, opened 2023-10-25
Fix Sonoma issues with the build

directionless:
We've been seeing build issues with modern Xcode. In theory they should have been fixed with go 1.21, but it's not clear that code merged. So, this is a simple hammer.

It defines some acceptable warnings, and then doesn't exit of those are the only things in stderr.

Closes  #1423

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -602,7 +602,7 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 		// _mostly_ effect our developer build process. In the interest in not breaking the build, we ignore them.
 		// https://github.com/golang/go/issues/62597#issuecomment-1733893918 is some of them, and some seem related to the zig cross compiling
 		stderrStr := stderr.String()
-		if os.Getenv("GITHUB_ACTIONS") != "" {
+		if os.Getenv("GITHUB_ACTIONS") == "" {
 			stderrStr = strings.ReplaceAll(stderrStr, "ld: warning: ignoring duplicate libraries: '-lobjc'\n", "")
 			stderrStr = strings.ReplaceAll(stderrStr, "# github.com/kolide/launcher/cmd/launcher\n", "")
 
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -600,13 +600,14 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 
 		// With the Sonoma-era Xcode binaries, we've started seeing a bunch of spurious warnings. They appear to
 		// _mostly_ effect our developer build process. In the interest in not breaking the build, we ignore them.
-		// https://github.com/golang/go/issues/62597#issuecomment-1733893918 is some of them, and some seem related to the zig cross compiling
+		// https://github.com/golang/go/issues/62597#issuecomment-1733893918 is some of them, and some seem related to
+		// the zig cross compiling
 		stderrStr := stderr.String()
 		if os.Getenv("GITHUB_ACTIONS") == "" {
 			stderrStr = strings.ReplaceAll(stderrStr, "ld: warning: ignoring duplicate libraries: '-lobjc'\n", "")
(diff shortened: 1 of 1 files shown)


pkg/make/builder.go
```diff
@@ -598,7 +598,19 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
...
-		if stderr.Len() > 0 {
+		// With the Sonoma-era Xcode binaries, we've started seeing a bunch of spurious warnings. They appear to
+		// _mostly_ effect our developer build process. In the interest in not breaking the build, we ignore them.
+		// https://github.com/golang/go/issues/62597#issuecomment-1733893918 is some of them, and some seem related to the zig cross compiling
+		stderrStr := stderr.String()
+		if os.Getenv("GITHUB_ACTIONS") != "" {
+			stderrStr = strings.ReplaceAll(stderrStr, "ld: warning: ignoring duplicate libraries: '-lobjc'\n", "")
+			stderrStr = strings.ReplaceAll(stderrStr, "# github.com/kolide/launcher/cmd/launcher\n", "")
+
+			re := regexp.MustCompile(`ld: warning: object file \(.*\) was built for newer 'macOS' version \(14.0\) than being linked \(11.0\)\n`)
```
zackattack01, line 609: do we want to make this one any more flexible with the version numbers? just wondering if it will still work for those who haven't upgraded yet

pkg/make/builder.go
```diff
@@ -598,7 +598,19 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 			return err
 		}
 
-		if stderr.Len() > 0 {
+		// With the Sonoma-era Xcode binaries, we've started seeing a bunch of spurious warnings. They appear to
+		// _mostly_ effect our developer build process. In the interest in not breaking the build, we ignore them.
+		// https://github.com/golang/go/issues/62597#issuecomment-1733893918 is some of them, and some seem related to the zig cross compiling
+		stderrStr := stderr.String()
+		if os.Getenv("GITHUB_ACTIONS") != "" {
```
RebeccaMahany, line 605:
```suggestion
		if os.Getenv("GITHUB_ACTIONS") == "" {
```

directionless: oops. Thank you

pkg/make/builder.go
```diff
@@ -598,7 +598,19 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
...
+			stderrStr = strings.ReplaceAll(stderrStr, "ld: warning: ignoring duplicate libraries: '-lobjc'\n", "")
+			stderrStr = strings.ReplaceAll(stderrStr, "# github.com/kolide/launcher/cmd/launcher\n", "")
```
RebeccaMahany, line 607:
```suggestion
			stderrStr = strings.ReplaceAll(stderrStr, fmt.Sprintf("# github.com/kolide/launcher/cmd/%s\n", filepath.Base(src)), "")
```

I build with `make -j2 github-build` locally -- this change ensures that `# github.com/kolide/launcher/cmd/launcher.ext` will also match.

directionless: Might be simpler just to move it to a regex?

pkg/make/builder.go
directionless: 🤷 We could. Though I hope it's fixed by then.

pkg/make/builder.go
RebeccaMahany: 🤷‍♀️ whichever you prefer, I'm happy as long as github-build works!

pkg/make/builder.go
zackattack01: same 🙏 I just wasn't sure if we were seeing 13.0 before this, I thought mine started earlier than the upgrade

James-Pickett: I still get the error sometimes ad I'm on Ventura still, I don't know how to repro it

pkg/make/builder.go
directionless: I used your code.

RebeccaMahany (approved):

PR #1425, opened 2023-10-25
prevent duplicate connectivity checks and only check /version endpoint

zackattack01:
Our connectivity log checkpointer currently appends `/version` to the end of various URLs stored in knapsack (device, control, trace, log) before issuing a GET request to determine connectivity status. This typically works but in some cases the URL is already configured with a path (e.g. `/log/`) causing a GET to `/log/version`. This changes two things:
- reconstruct the full URLs into `<scheme>://<host>:<port>/version` to ensure we hit the correct path
- track the URLs that we do hit so we can avoid hitting the same endpoint multiple times while the host is the same (currently trace and log hosts)


RebeccaMahany (approved):
🔥

James-Pickett (approved):

PR #1427, opened 2023-10-26
Add traces to important startup behavior

RebeccaMahany:
Adds traces for launching osquery instance (including choosing osqueryd binary from library) and starting up desktop process. Removes an unused function from the TUF autoupdate code.


James-Pickett (approved):
🔥

PR #1436, opened 2023-11-02
use knapsack and slogger in kolide service

James-Pickett:


cmd/grpc.ext/grpc.go
```diff
@@ -100,8 +78,51 @@ func main() {
 		logutil.Fatal(logger, "err", fmt.Errorf("creating stores: %w", err), "stack", fmt.Sprintf("%+v", err))
 	}
 	f := flags.NewFlagController(logger, stores[storage.AgentFlagsStore])
+
+	slogLevel := slog.LevelInfo
+	if *flVerbose {
+		slogLevel = slog.LevelDebug
+	}
+
+	multiSlogger := new(multislogger.MultiSlogger)
```
James-Pickett, line 87: need to actually pass this to knapsack

cmd/grpc.ext/grpc.go
James-Pickett, line 1: are we still using / building this at all or can it go away?

cmd/grpc.ext/grpc.go
```diff
@@ -100,8 +78,51 @@ func main() {
...
+	multiSlogger.AddHandler(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
+		Level: slogLevel,
+	}))
+
 	k := knapsack.New(stores, f, db, nil, nil)
 
+	if err := k.SetKolideServerURL(serverURL); err != nil {
+		logutil.Fatal(logger, "err", fmt.Errorf("setting kolide server url: %w", err), "stack", fmt.Sprintf("%+v", err))
```
James-Pickett, line 95: use slogger

pkg/service/client_grpc.go
```diff
@@ -101,39 +99,31 @@ func NewGRPCClient(conn *grpc.ClientConn, logger log.Logger) KolideService {
...
+	if k.InsecureTransportTLS() {
 		grpcOpts = append(grpcOpts, grpc.WithInsecure())
 	} else {
-		host, _, err := net.SplitHostPort(serverURL)
-		if err != nil {
-			return nil, fmt.Errorf("split grpc server host and port: %s: %w", serverURL, err)
-		}
-
-		creds := &tlsCreds{credentials.NewTLS(makeTLSConfig(host, insecureTLS, certPins, rootPool, logger))}
+		creds := &tlsCreds{credentials.NewTLS(makeTLSConfig(k, rootPool))}
```
James-Pickett, line 120: so the old code was splitting the host form port in the server url, is that still needed? will test against local k2 with port

James-Pickett: yes, this is a thing ... working on it

James-Pickett: now extracting port inside makeTlsConfig

pkg/service/publish_results.go
```diff
@@ -183,12 +182,8 @@ func (mw logmw) PublishResults(ctx context.Context, nodeKey string, results []di
...
 		resJSON, _ := json.Marshal(results)
 		uuid, _ := uuid.FromContext(ctx)
-		logger := level.Debug(mw.logger)
-		if err != nil {
-			logger = level.Info(mw.logger)
-		}
-		logger.Log(
-			"method", "PublishResults",
+
+		mw.knapsack.Slogger().Log(ctx, levelForError(err), "publish results",
```
directionless, line 186: This is not quite right. -- the first argument should be the `message` from line 189. The `method` should remain a KV pair. (I think true for all of these)

pkg/service/tls.go
```diff
@@ -6,28 +6,42 @@ import (
...
-	"github.com/go-kit/kit/log/level"
+	"github.com/kolide/launcher/pkg/agent/types"
 )
 
-func makeTLSConfig(host string, insecureTLS bool, certPins [][]byte, rootPool *x509.CertPool, logger log.Logger) *tls.Config {
+func makeTLSConfig(k types.Knapsack, rootPool *x509.CertPool) *tls.Config {
+
+	hostname := k.KolideServerURL()
+	if k.Transport() == "grpc" {
+		// If we're using gRPC, we need to strip the port from the host
```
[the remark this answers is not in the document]
James-Pickett: updated

pkg/service/publish_results.go
James-Pickett: Okay, I've added the message val as the `msg` for the logs.

Should we drop the kvp with the `message` key (now it would be `msg`) ... I don't think anything should rely on log output, but I have an urge to leave it in to maintain parity.

pkg/log/multislogger/multislogger.go
```diff
@@ -12,6 +13,12 @@ type MultiSlogger struct {
 	handlers []slog.Handler
 }
 
+func NewNoop() *MultiSlogger {
+	ms := new(MultiSlogger)
+	ms.AddHandler(slog.NewTextHandler(io.Discard, nil))
```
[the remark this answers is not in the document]
James-Pickett: you need something or the call to MultiSlogger.Logger will give nil pointer. But you must pass a handler to slog.New(), so I've set it up now so it writes to io.Discard if you do not provide a handler

pkg/service/publish_results.go
directionless: I think there's no value in duplicating the `message` KV. It's fine that it changed from `message` to `msg`. (I just don't want to have `method` become `msg` which is semantically different)

eg: You should strike `message`.

zackattack01 (dismissed):
nice! 🔥

pkg/service/publish_logs.go
```diff
@@ -171,16 +170,12 @@ func (s *grpcServer) PublishLogs(ctx context.Context, req *pb.LogCollection) (*p
 func (mw logmw) PublishLogs(ctx context.Context, nodeKey string, logType logger.LogType, logs []string) (message, errcode string, reauth bool, err error) {
 	defer func(begin time.Time) {
 		uuid, _ := uuid.FromContext(ctx)
-		logger := level.Debug(mw.logger)
-		if err != nil {
-			logger = level.Info(mw.logger)
-		}
-		logger.Log(
+
+		mw.knapsack.Slogger().Log(ctx, levelForError(err), message,
```
RebeccaMahany, line 174: For this one and PublishResults, it looks like the message is empty (at least in the successful case) -- do we want to do `"publish logs"` and `"publish results"` for the log message instead, like how you've got it for e.g. RequestConfig?

RebeccaMahany (approved):
LGTM, my suggestion is optional

pkg/service/publish_logs.go
James-Pickett: I'm not quite sure what is correct. Possibly using `"publish logs"` when message is empty. see previous conversation on subject https://github.com/kolide/launcher/pull/1436#discussion_r1380905299

directionless: So these cases are a successful completion of the thing, right? And we have `method` there. So maybe just "success"?

RebeccaMahany: I'm fine with `success` or `publish logs`!

directionless: The case on "publish logs" feels wrong. Could be "published logs" or even "published"

Consider something like scoping the view down to just showing the publish logs method.

James-Pickett: I've updated all the publish methods to either have success of failure as the message

RebeccaMahany (approved):

directionless (approved):
Good enough!

PR #1437, opened 2023-11-02
Remove old `grpc.ext`

directionless:
We haven't use `grpc` in many years, and we don't maintain this as a standalone extension. Even though it probably still works, it's better to remove this, than maintain it. 

If people miss it, we can investigate bringing it back


RebeccaMahany (approved):
🔥

Issue #1433, opened 2023-10-31
launcher doctor output looks malformed on Windows


RebeccaMahany:
Emojis are not properly encoded when running `& "C:\Program Files\Kolide\Launcher-kolide-k2\bin\launcher.exe" doctor | Out-String` in PowerShell on Windows. This issue is to make the output look reasonable. Dealing with encoding in PowerShell when working on the notifications was a pain -- I think it's probably not worth it to pursue emoji support just for doctor on Windows, and write out `Passing|Warning|Failing` or similar instead of the emoji on Windows.

If possible, it would be nice to be able to run `launcher.exe doctor` without having to pipe to `Out-String` to view the results, as well.

directionless:
> If possible, it would be nice to be able to run `launcher.exe doctor` without having to pipe to `Out-String` to view the results, as well.

Yeah... IIRC this is fallout from https://github.com/kolide/launcher/pull/1007 and our choice to have a single binary

directionless:
Well well.. [`AttachConsole`](https://learn.microsoft.com/en-us/windows/console/attachconsole?redirectedfrom=MSDN) might do it. And seems to have some go libraries. https://pkg.go.dev/github.com/minux/windows/attachconsole

RebeccaMahany:
AttachConsole looks like it should work -- I couldn't get it working with that exact package, but doing similar plus redirecting stdout seems to do the trick

Issue #1387, opened 2023-10-10
Fairly regular and unexplained terminations of launcher service on Windows


RebeccaMahany:
Event logs on Windows contain `The LauncherKolideK2Svc service terminated unexpectedly` fairly frequently. These are sometimes expected -- for example, when launcher autoupdates -- but I still see one or two a day that don't have a corresponding explanation in the launcher logs.

We had thought that these could be explained by throttling issues with modern standby (more details [here](https://github.com/kolide/launcher/pull/1284)), but I saw one happen while regression testing last week while my laptop was awake, so I think something else is going on.

We should try to write corresponding launcher logs when this event occurs -- maybe there's somewhere we're missing a call to `recover`?

The relevant event logs can be found by filtering Windows Logs => System with:

```
<QueryList>
  <Query Id="0" Path="System">
    <Select Path="System">*[System[Provider[@Name='Service Control Manager'] and (EventID=7031)]]</Select>
  </Query>
</QueryList>
```

directionless:
Ha. This issue was prescient. I think it's fixed by #1699 and the other recent panic work

PR #1438, opened 2023-11-03
Call AttachConsole for better output on Windows

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1433

Fixes writing output to the terminal for:
* doctor
* flare
* version
* svc-fg


cmd/launcher/doctor.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func runDoctor(args []string) error {
+	attachConsole()
+	defer detachConsole()
```
directionless, line 17: I think we should do this everywhere but desktop. I'm not sure if it's easier to put this into main with an exception for dekstop, or in each piece

cmd/launcher/doctor.go
```diff
@@ -13,6 +13,9 @@ import (
 )
 
 func runDoctor(args []string) error {
+	attachConsole()
```
James-Pickett, line 16: should we log any returned errors?

directionless, line 16: This ignores the returned error. I think it's reasonable to continue execution, but we should either log the error here, or log it in `attachConsole` and return nothing

RebeccaMahany: If `attachConsole` fails, nothing will be printed without the user piping to `Out-String`, so I wasn't sure if printing the error message would be very useful. And since this is end user-facing, I wasn't sure if printing error messages about this failure would be appropriate, either. Thoughts?

cc @directionless since you left a similar comment.

RebeccaMahany: Consolidating discussion to https://github.com/kolide/launcher/pull/1438#discussion_r1382079870 🙂

cmd/launcher/doctor.go
RebeccaMahany: Looking through --

* I can add it to `flare` -- that gets us "Flare uploaded successfully" + the file name, so that seems useful
* Adding it to `interactive` breaks interactive
* Adding it to `version` makes sense
* I'm not sure why we'd want it for launcher proper stuff (runLauncher, windows service, windows service foreground, extension) and I am a little nervous that adding it there would break something
* `uninstall` isn't implemented on Windows so it doesn't hurt to add it there

directionless: I think we want it in `runLauncher` -- that's mostly going to be used by developers in a console window

`windows service foreground` is an interesting one. That's designed to run in the foreground to debug issues, it'll never run through the service manager. Probably needs it

cmd/launcher/doctor.go
directionless: Log them? I mean, sure, probably no one is going to look. But if someone ever files a bug report they'll be there...

cmd/launcher/doctor.go
RebeccaMahany: `runLauncher` is called by the Windows service and that's why I'm hesitant to add it there -- I'll try it out and see how it behaves.

I'll try adding it to foreground and see how it behaves.

cmd/launcher/doctor.go
RebeccaMahany: The logs for doctor only ever get written to stdout (unless someone happens to specify the root dir argument when running doctor), so they won't be available in the log file. They'll only be printed if the user is piping to `Out-String`, hence why I'm not sure if it's useful or appropriate.

cmd/launcher/doctor.go
RebeccaMahany: It works fine in foreground and I've added it there.

It also seems fine for `runLauncher`, tentatively, but I'm really not sure we gain a lot there for the risk that it affects something in an unexpected way.

directionless: Hrm. I guess that's a good question. I think there's value in getting console output for developers. But it's hard to quantify the risk. We used to run launcher without windowsgui, and that was fine. But this is different. I guess we can go either way

RebeccaMahany: I agree there's value in getting console output for developers -- I think my preference is to not add it for `runLauncher` now to avoid unnecessary risk on Windows since [this issue](https://github.com/kolide/launcher/issues/1387) is on my mind (knowing we're thinking of solving the output issue more cleanly and thoroughly in the future with a separate desktop binary).

directionless: Okay. Changing it is the status quo, so let's kick it forward.

directionless (approved):
Okay, happy to try!

Issue #1327, opened 2023-09-01
Launcher should have a table to parse jwts


directionless:
I think it would be worth having a table that can parse jwts. I imagine it'll require `path` be passed in. Maybe an optional secret. It should be pretty simple to parse out whatever data is there, and pass it along. 

Goals:
- `kolide_jwt` that uses dataflatten to return information about a jwt
- `path` column requirements
- It looks like there are a couple of different "claims". The go library seems to have `MapClaims` and `RegisteredClaims` I don't know enough if we should have a column for specifying that, or if there's a general claim we can write to process the data. Or maybe both. If claims is unset we use some simple data snarf?
- no idea what `parts` is. Probably "things that didn't fit in a claim"
- I don't know if we want to support validation. It'd mean a column to pass the secret, which feels icky

Use Cases:
- general jwt parsing
- CrowdStrike's `data.zta` file. I can't decide if we should make CrowdStrike table as well, or expect people to build that on `kolide_jwt`
- can we parse the kolide enrollment secret? It's a jwt on disk...

## Old

CrowdStrike leaves a `data.zta` file on disk. I think this is a jwt with some information about the stance of the client. I know some vendors parse that, we should figure out if we can.

If it's a jwt, we can parse it, but we'd need key to be able to validate it.

Also, check out this nice doc I found on the internet -- [zta crowdstrike sensor reading.docx](https://github.com/kolide/launcher/files/12500658/zta.crowdstrike.sensor.reading.docx). There might also be something in the python tools -- https://github.com/CrowdStrike/falconpy

directionless:
I do have prior art in jwt parsing in [`pkg/debug/checkups/enroll-secret.go`](https://github.com/kolide/launcher/blob/main/pkg/debug/checkups/enroll-secret.go)

Might be interesting to extract that to a library, support parsing unvalidated or validated, and make a table for it.

directionless:
Path locations here are:
* `%ProgramData%\CrowdStrike\ZeroTrustAssessment\data.zta`
* `/Library/Application Support/Crowdstrike/ZeroTrustAssessment/data.zta`

PR #1440, opened 2023-11-06
add kolide_jwt parsing table

zackattack01:
This adds a `kolide_jwt` table to parse claims from a JWT when queried against a file path containing one. The general functionality is similar to the `kolide_json` table, requiring a valid path to an existing file, and parsing the contents where available:
```
osquery> select * from kolide_jwt where 1=1;
Error: error generating table: The kolide_jwt table requires that you specify a single constraint for path
osquery> select * from kolide_jwt where path='/tmp/data.zta';
+--------------------------+------------+---------------+----------------------------------+-------+---------------+
| fullkey                  | parent     | key           | value                            | query | path          |
+--------------------------+------------+---------------+----------------------------------+-------+---------------+
| iat                      |            | iat           | 1614791923                       | *     | /tmp/data.zta |
| platform                 |            | platform      | Windows 10                       | *     | /tmp/data.zta |
| serial_number            |            | serial_number |                                  | *     | /tmp/data.zta |
| sub                      |            | sub           | 99999c489c1a499999ef5b213ee99999 | *     | /tmp/data.zta |
| typ                      |            | typ           | crowdstrike-zta+jwt              | *     | /tmp/data.zta |
| assessment/overall       | assessment | overall       | 48                               | *     | /tmp/data.zta |
| assessment/os            | assessment | os            | 77                               | *     | /tmp/data.zta |
| assessment/sensor_config | assessment | sensor_config | 30                               | *     | /tmp/data.zta |
| assessment/version       | assessment | version       | 2.0.1                            | *     | /tmp/data.zta |
| exp                      |            | exp           | 1616001523                       | *     | /tmp/data.zta |
+--------------------------+------------+---------------+----------------------------------+-------+---------------+
```
This should address the issue [here](https://github.com/kolide/launcher/issues/1327)


pkg/dataflatten/jwt.go
```diff
@@ -0,0 +1,49 @@
...
+	defer jwtFH.Close()
+
+	tokenRaw, err := io.ReadAll(jwtFH)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read JWT: %w", err)
+	}
+
+	// attempt decode into the generic (default) MapClaims struct to ensure we capture
+	// any claims data that might be useful
+	token, _, err := new(jwt.Parser).ParseUnverified(string(tokenRaw), jwt.MapClaims{})
```
directionless, line 34: What's the second return value?

pkg/dataflatten/jwt.go
```diff
@@ -0,0 +1,49 @@
...
+func JWTFile(file string, opts ...FlattenOpts) ([]Row, error) {
+	return flattenJWT(file, opts...)
+}
+
+func flattenJWT(path string, opts ...FlattenOpts) ([]Row, error) {
+	results := map[string]interface{}{}
+	jwtFH, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("unable to access file: %w", err)
+	}
```
directionless, line 23: I forget how all the layers here work, if you pass the table a path that doesn't exist, will it error out or continue?

pkg/dataflatten/jwt.go
```diff
@@ -0,0 +1,49 @@
...
+
+	claims, ok := token.Claims.(jwt.MapClaims)
+	if !ok {
+		return nil, fmt.Errorf("JWT has no parseable claims")
+	}
+
+	for k, v := range claims {
+		results[k] = v
+	}
```
directionless, line 46: Do we need `results` or can we pass `claims` straight to Flatten?

And is there any more data we should pass back? Like, this is `claims`, but is there more? I kinda want to return
```
struct {
  Claims: map[string]any
}
```

Because then if we add validate, or file metadata, or ???, we won't need to change the structure.

directionless (commented):
Doing it in `pkg/dataflatten` is a little weird, but I think okay. Mostly, I think of things down in here as low level, and we can build atop them. And I'm not sure what's going to want to build on this. But :shrug: it's no different than plist. So it's probably fine?

OTOH, if we wanted to bring in validation, we'd need to refactor. OTOH, we could refactor if/when we care.

pkg/dataflatten/jwt.go
zackattack01: this is `parts = strings.Split(tokenString, ".")` - we are essentially just base64 decoding `parts[1]` in the remaining bits

pkg/dataflatten/jwt.go
zackattack01: it will continue, giving zero rows from the missing path. e.g. `where path='/tmp/missing' or path='/tmp/present'` will just return the values found in /tmp/present

directionless: Cool. Then we handle it up a layer.

pkg/dataflatten/jwt.go
directionless: Is there anything to extract from the other parts?

pkg/dataflatten/jwt.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+// JWTFile adds support for the kolide_jwt table, which allows parsing
+// a file containing a JWT. Note that the kolide_jwt table does not handle
+// verification - this is a utility table for convenience.
```
RebeccaMahany, line 13: Docs are nice, thank you

pkg/dataflatten/jwt.go
zackattack01: I can try that out! we could additionally include the header (likely containing the signing algo) but in general its just:
| Header | Payload | Signature |
|--------|--------|--------|
| algo | claims | signature |
| type (=JWT) |  | |

directionless: Algo seems interesting. An issued timestamp if one exists...

And someday I think it would be super interesting to get a validation in, and that would give us a place to hang it. I guess we could preemptively add `validated: false` or similar.

RebeccaMahany (approved):
I'm good with this as-is, but also like seph's suggestion https://github.com/kolide/launcher/pull/1440/files#r1383951951 -- happy to reapprove if you update

pkg/dataflatten/jwt.go
zackattack01: we will get the header now too per the conversation below. I can include the raw signature as well if you think that would be helpful?

directionless: I don't think we should return the raw signature. Or the contents of the file

pkg/dataflatten/jwt.go
zackattack01: we can't pass claims to flatten directly but that will work for the header and signature. the registered claim for timestamp will be at `claims->iat` if included. and I will add a validated `false` for clarity here, thank you!

zackattack01: PR description is updated with the new format discussed here

directionless (approved):
Sweet!

As a side note, GitHub has special syntax for how to link PRs to Issues and have them mark them as closed and whatnot. https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

RebeccaMahany (approved):

PR #1441, opened 2023-11-07
Add some logs to the remote flare consumer

directionless:
Note much to say here -- add some logs. (And convert to using the slog internal logger as well)

### DIFF
--- a/ee/control/consumers/flareconsumer/flareconsumer.go
+++ b/ee/control/consumers/flareconsumer/flareconsumer.go
@@ -6,10 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"time"
-
-	"github.com/go-kit/kit/log"
-	"github.com/go-kit/kit/log/level"
 
 	"github.com/kolide/launcher/pkg/agent/types"
 	"github.com/kolide/launcher/pkg/debug/checkups"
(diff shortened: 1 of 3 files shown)


ee/control/consumers/flareconsumer/flareconsumer.go
```diff
@@ -41,23 +38,24 @@ func (f *FlareRunner) RunFlare(ctx context.Context, k types.Knapsack, flareStrea
...
 		knapsack: knapsack,
-		logger:   logger,
 		newFlareStream: func(note, uploadRequestURL string) (io.WriteCloser, error) {
 			return shipper.New(knapsack, shipper.WithNote(note), shipper.WithUploadRequestURL(uploadRequestURL))
 		},
 	}
 }
 
 func (fc *FlareConsumer) Do(data io.Reader) error {
+	// slog needs a ctx
```
RebeccaMahany, line 52: Do we prefer this over the option without a ctx -- `fc.knapsack.Slogger().Info("skipping flare, run too recently"...`?

ee/control/consumers/flareconsumer/flareconsumer.go
```diff
@@ -81,6 +79,10 @@ func (fc *FlareConsumer) Do(data io.Reader) error {
 		return fmt.Errorf("failed to decode key-value json: %w", err)
 	}
 
+	fc.knapsack.Slogger().Log(ctx, slog.LevelInfo, "Recieved remote flare request",
```
James-Pickett, line 82:
```suggestion
	fc.knapsack.Slogger().Log(ctx, slog.LevelInfo, "received remote flare request",
```

James-Pickett (approved):

ee/control/consumers/flareconsumer/flareconsumer.go
RebeccaMahany: That works for me! If we want to standardize on this, then we can also enforce it via sloglint's [context-only](context-only) setting.

James-Pickett (approved):

Issue #833, opened 2022-04-30
Hardcoded paths to binaries everywhere


znewman01:
I'm on [NixOS](https://nixos.org), which doesn't really put binaries in `/bin`:

```shell
$ ls /bin/
sh
$
```

Additionally, there's no `/sbin`. This is causing all sorts of problems, since the `launcher` code base has hardcoded paths in a bunch of places. The one that's really giving me trouble at the moment is [zfs](https://github.com/kolide/launcher/blob/adaddbd7ec1743f4fc020ced139853eb72836e19/pkg/osquery/tables/zfs/tables.go#L19).

This is failing even when the appropriate binaries are on the `$PATH`:

```shell
$ for cmd in zfs gsettings nmcli; do
       eval $(sudo strings /proc/$(pgrep launcher)/environ | grep '^PATH') command -v $cmd
  done
/nix/store/yw1by07874clir1jqlnvc9xz7k27q3vh-zfs-user-2.1.4/bin/zfs
/nix/store/4ml4higjpkgjzpzm1qa0rabn7qjdcbn4-glib-2.72.0-bin/bin/gsettings
/nix/store/akig9n0zp09cq7rvc91dhlz2jyhi52rv-networkmanager-1.36.4/bin/nmcli
```

So I guess I have two(ish) questions:

1. Why do we hardcode paths rather than respecting the `$PATH` of the launcher process? Would it be okay to just use the binary names instead?
2. If not, can we provide an alternate way of specifying paths to each of the binaries that the launcher shells out to?

directionless:
My historic experience is that it's best practice to specify full paths. Partly because system daemons often have minimal paths, and can't always find things. And sometimes because people sneak malicious binaries with shared names into the path.

It's not clear to me that either of these holds for NixOS, and I'm amenable to finding some additional patterns.

What do you think makes sense? I could imagine looking on the path and verifying that it's in `/nix/store/`

znewman01:
The following seem unlikely to be true at the same time:

> Partly because system daemons often have minimal paths, and can't always find things.

> And sometimes because people sneak malicious binaries with shared names into the path.

I wonder whether we can just add a flag like `--use-PATH` (name WIP) to allow using the `$PATH` like normal if a user requests it, or having a separate `KOLIDE_PATH` variable.

directionless:
I've been thinking about this.

I'm currently somewhat against a `--use-path` flag, because I think it would just end up set all the time.

[...]

Issue #896, opened 2022-09-19
Nix Support (And maybe Fedora Silverblue) (due 2024-02)


directionless:
* [x] https://github.com/kolide/launcher/issues/833
* [x] https://github.com/kolide/launcher/issues/770
* [x] https://github.com/kolide/launcher/issues/638
* [ ] https://github.com/kolide/launcher/issues/868
* [x] https://github.com/kolide/launcher/issues/1200
* [x] Autoupdate path needs to be configurable (See TUF work)
* [ ] https://github.com/kolide/launcher/issues/1473

znewman01:
Hi all! Thanks for your patience with our weird OS 😄 I've just filed [an issue](https://github.com/NixOS/nixpkgs/issues/195965) for adding the launcher to the Nix package repository. We'd love to work with you on getting support for NixOS in a good place.

I can send a PR (no concrete timeline; let me know if it becomes urgent) for #833. I'll also chime in on a couple of the other issues. Let me know if I can help otherwise (especially testing things out).

directionless:
Bouncing to this ticket... If Nix is like the other distros, getting osquery in the formal packages will be quite hard. Osquery is designed to build with its own third party libraries and versions. And most distros really want software to build with the shipped libraries. I've seen a couple of distros ship and then drop osquery because of this.

Kolide ships our own packages for things, because we don't need to follow all the os packaging rules. So we can just ship an osquery binary. 

I can do a little to try to help get these into Nix itself, but I think that's a hard road. I'd love to get a Kolide distributed nix package for launcher. But I know very little about how that would work. (Like, what's a package here? Is there something like a build thing, like a deb, or is it closer to only distributing source manifests?)

terlar:
Regarding having specifically pinned third party libraries and versions for a specific package is not something that would be a problem for a nix package as far as I understand. If needed the package set can contain several different versions of the same package, especially if required to build another package. What kind of libraries are we talking about here?

directionless:
You're asking about the osquery build? It links to things via git submodules. Nose around https://github.com/osquery/osquery/tree/master/libraries/cmake

terlar:
[...]

PR #1443, opened 2023-11-07
Add package to perform path lookups and validations for commands

RebeccaMahany:
Update all usages of `exec.Command` and `exec.CommandContext` to no longer perform path lookups, instead checking against an allowlist of known paths. In the case where the path cannot be allowlisted (e.g. when using Nix, the install path will be somewhere in `/nix/store`), check for known path prefixes (like `/nix/store`).

Closes https://github.com/kolide/launcher/issues/833

Relates to https://github.com/kolide/launcher/issues/896

### DIFF
--- a/pkg/allowedpaths/cmd.go
+++ b/pkg/allowedpaths/cmd.go
@@ -1,0 +1,85 @@
+package allowedpaths
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// CommandWithLookup should be used when the full path to the command is not known and it is
+// likely to be found in PATH.
+func CommandWithLookup(name string, arg ...string) (*exec.Cmd, error) {
+	fullPathToCommand, err := exec.LookPath(name)
+	if err != nil {
+		return nil, fmt.Errorf("looking up path to %s: %w", name, err)
+	}
+
+	return CommandWithPath(fullPathToCommand, arg...)
+}
+
+// CommandWithPath should be used when the full path to the command is known, or unlikely to be
+// found in PATH.
+func CommandWithPath(fullPathToCommand string, arg ...string) (*exec.Cmd, error) {
+	fullPathToCommand = filepath.Clean(fullPathToCommand)
+
+	if err := pathIsAllowed(fullPathToCommand); err != nil {
+		return nil, fmt.Errorf("path is not allowed: %w", err)
+	}
+
+	return exec.Command(fullPathToCommand, arg...), nil
+}
+
+// CommandContextWithLookup should be used when the full path to the command is not known and it is
+// likely to be found in PATH.
+func CommandContextWithLookup(ctx context.Context, name string, arg ...string) (*exec.Cmd, error) {
+	fullPathToCommand, err := exec.LookPath(name)
+	if err != nil {
+		return nil, fmt.Errorf("looking up path to %s: %w", name, err)
+	}
+
+	return CommandContextWithPath(ctx, fullPathToCommand, arg...)
+}
+
+// CommandContextWithPath should be used when the full path to the command is known, or unlikely to be
+// found in PATH.
+func CommandContextWithPath(ctx context.Context, fullPathToCommand string, arg ...string) (*exec.Cmd, error) {
+	fullPathToCommand = filepath.Clean(fullPathToCommand)
+
+	if err := pathIsAllowed(fullPathToCommand); err != nil {
+		return nil, fmt.Errorf("path is not allowed: %w", err)
+	}
+
+	return exec.CommandContext(ctx, fullPathToCommand, arg...), nil
+}
+
+// pathIsAllowed validates the path to the command against our allowlist.
+func pathIsAllowed(fullPathToCommand string) error {
+	cmdName := strings.ToLower(filepath.Base(fullPathToCommand))
+
+	// We trust the autoupdate libraries to select the correct paths
+	if cmdName == "launcher" || cmdName == "launcher.exe" || cmdName == "osqueryd" || cmdName == "osqueryd.exe" {
+		return nil
+	}
+
+	// Check that we have known paths for the given command
+	knownCmdPaths, ok := knownPaths[cmdName]
+	if !ok {
+		return fmt.Errorf("no known paths for command %s, cannot validate path %s", cmdName, fullPathToCommand)
+	}
+
+	// Check if this path is registered
+	if _, ok := knownCmdPaths[fullPathToCommand]; ok {
+		return nil
+	}
+
+	// Check to make sure the command is at least in a known directory
+	for _, knownPathPrefix := range knownPathPrefixes {
+		if strings.HasPrefix(fullPathToCommand, knownPathPrefix) {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("%s not in known paths and does not have known prefix", fullPathToCommand)
+}
--- a/pkg/allowedpaths/cmd_darwin.go
+++ b/pkg/allowedpaths/cmd_darwin.go
@@ -1,0 +1,81 @@
+//go:build darwin
+// +build darwin
+
+package allowedpaths
+
+var knownPaths = map[string]map[string]bool{
+	"bioutil": {
+		"/usr/bin/bioutil": true,
+	},
+	"diskutil": {
+		"/usr/sbin/diskutil": true,
+	},
+	"falconctl": {
+		"/Applications/Falcon.app/Contents/Resources/falconctl": true,
+	},
+	"firmwarepasswd": {
+		"/usr/sbin/firmwarepasswd": true,
+	},
+	"ifconfig": {
+		"/sbin/ifconfig": true,
+	},
+	"launchctl": {
+		"/bin/launchctl": true,
+	},
+	"lsof": {
+		"/usr/sbin/lsof": true,
+	},
+	"mdfind": {
+		"/usr/bin/mdfind": true,
+	},
+	"netstat": {
+		"/usr/sbin/netstat": true,
+	},
+	"open": {
+		"/usr/bin/open": true,
+	},
+	"pkgutil": {
+		"/usr/sbin/pkgutil": true,
+	},
+	"powermetrics": {
+		"/usr/bin/powermetrics": true,
+	},
+	"profiles": {
+		"/usr/bin/profiles": true,
+	},
+	"ps": {
+		"/bin/ps": true,
+	},
+	"pwpolicy": {
+		"/usr/bin/pwpolicy": true,
+	},
+	"remotectl": {
+		"/usr/libexec/remotectl": true,
+	},
+	"repcli": {
+		"/Applications/VMware Carbon Black Cloud/repcli.bundle/Contents/MacOS/repcli": true,
+	},
+	"scutil": {
+		"/usr/sbin/scutil": true,
+	},
+	"softwareupdate": {
+		"/usr/sbin/softwareupdate": true,
+	},
+	"system_profiler": {
+		"/usr/sbin/system_profiler": true,
+	},
+	"tmutil": {
+		"/usr/bin/tmutil": true,
+	},
+	"zerotier-cli": {
+		"/usr/local/bin/zerotier-cli": true,
+	},
+}
+
+var knownPathPrefixes = []string{
+	"/bin",
+	"/usr/bin",
+	"/usr/libexec",
+	"/usr/local/bin",
+	"/usr/sbin",
+}
--- a/pkg/allowedpaths/cmd_linux.go
+++ b/pkg/allowedpaths/cmd_linux.go
@@ -1,0 +1,60 @@
+//go:build linux
+// +build linux
+
+package allowedpaths
+
+var knownPaths = map[string]map[string]bool{
+	"dpkg": {
+		"/usr/bin/dpkg": true,
+	},
+	"lsof": {
+		"/usr/bin/lsof": true,
+	},
+	"gnome-extensions": {
+		"/usr/bin/gnome-extensions": true,
+	},
+	"gsettings": {
+		"/usr/bin/gsettings": true,
+	},
+	"ifconfig": {
+		"/usr/sbin/ifconfig": true,
+	},
+	"ip": {
+		"/usr/sbin/ip": true,
+	},
+	"loginctl": {
+		"/usr/bin/loginctl": true,
+	},
+	"notify-send": {
+		"/usr/bin/notify-send": true,
+	},
+	"ps": {
+		"/usr/bin/ps": true,
+	},
+	"rpm": {
+		"/bin/rpm": true,
+	},
+	"systemctl": {
+		"/usr/bin/systemctl": true,
+	},
+	"x-www-browser": {
+		"/usr/bin/x-www-browser": true,
+	},
+	"xdg-open": {
+		"/usr/bin/xdg-open": true,
+	},
+	"xrdb": {
+		"/usr/bin/xrdb": true,
+	},
+	"zerotier-cli": {
+		"/usr/local/bin/zerotier-cli": true,
+	},
+}
+
+var knownPathPrefixes = []string{
+	"/bin",
+	"/nix/store", // NixOS support
+	"/usr/bin",
+	"/usr/local/bin",
+	"/usr/sbin",
+}
--- a/pkg/allowedpaths/cmd_test.go
+++ b/pkg/allowedpaths/cmd_test.go
@@ -1,0 +1,201 @@
+package allowedpaths
+
+import (
+	"context"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCommandWithLookup(t *testing.T) {
+	t.Parallel()
+
+	var cmdName string
+	if runtime.GOOS == "windows" {
+		cmdName = "powershell.exe"
+	} else {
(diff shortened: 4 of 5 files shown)


.golangci.yml
```diff
@@ -32,6 +33,10 @@ linters-settings:
     ignore: github.com/go-kit/kit/log:Log
   gofmt:
     simplify: false
+  forbidigo:
+    forbid:
+      - p: ^exec\.Command.*$
+        msg: use pkg/allowedpaths functions instead
```
directionless, line 39: _nice_. Really glad to see a linter for this

pkg/allowedpaths/cmd_darwin.go
```diff
@@ -0,0 +1,96 @@
+//go:build darwin
+// +build darwin
+
+package allowedpaths
+
+var knownPaths = map[string]map[string]bool{
```
directionless, line 6: One callout here... This will _require_ we define per platform. There's no shared mechanism. I think that's okay, but I did want to note it.

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// CommandWithLookup should be used when the full path to the command is not known and it is
+// likely to be found in PATH.
+func CommandWithLookup(name string, arg ...string) (*exec.Cmd, error) {
```
directionless, line 13: I wonder if we need this, or if we can push everything to using `CommandContextWithLookup`

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := pathIsAllowed(fullPathToCommand); err != nil {
+		return nil, fmt.Errorf("path is not allowed: %w", err)
+	}
+
+	return exec.CommandContext(ctx, fullPathToCommand, arg...), nil //nolint:forbidigo
+}
+
+// pathIsAllowed validates the path to the command against our allowlist.
+func pathIsAllowed(fullPathToCommand string) error {
+	cmdName := strings.ToLower(filepath.Base(fullPathToCommand))
```
directionless, line 59: I can't decide if this is clever, or too clever. And maybe I want to understand the callers.

directionless: (I moved this conversation to slack)

pkg/allowedpaths/cmd_darwin.go
RebeccaMahany: I _think_ that's my preference -- I figured since we're switching from hardcoded paths, we'd want the narrowest possible definitions, and per-platform gives us that. But since we have to do a prefix allowlist (at least for linux) anyway, maybe it doesn't matter.

pkg/allowedpaths/cmd.go
RebeccaMahany: There were a handful of places using Command instead of CommandContext and I didn't want to update them unnecessarily, given how much regression testing this change is already going to require.

But I guess we could do `CommandContextWithLookup(context.TODO()...` for those instead. And I do think that it'd be better to push us towards using contexts for commands in the future.

pkg/allowedpaths/cmd_darwin.go
directionless: We do share some posix things, but I suspect it's just not common enough to matter. Most of the mac execs are really mac specific.

Dunno though, I think with our slack conversation we'll have more rope

pkg/allowedpaths/cmd.go
directionless: Yeah. I think if we want context, we should just `context.TODO()`

RebeccaMahany: (I've gone ahead and required a context for the newest iteration of these changes)

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,784 @@
...
+	"path/filepath"
+	"runtime"
+)
+
+type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
+
+func Airport(ctx context.Context, arg ...string) (*exec.Cmd, error) {
+	if runtime.GOOS != "darwin" {
+		return nil, errors.New("airport supported on darwin only")
+	}
```
directionless, line 19: This loses the compile time enforcement -- I think we should define this in an `_darwin.go` file. Nothing on another platform should call it.

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,784 @@
...
+
+	fullPathToCmdValidated, err := validatedPath("/System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport")
+	if err != nil {
+		return nil, fmt.Errorf("airport not found: %w", err)
+	}
+	return newCmd(ctx, fullPathToCmdValidated, arg...), nil
```
directionless, line 25: I would probably store an array of possible paths, and make this more of a helper. I think that'll give us a single place to adjust nix (or similar).

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,784 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"runtime"
+)
+
+type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
```
directionless, line 14: I have no idea if this should be `type AllowedCommand func` or 
```
type Command interface {
  Ctx(ctx context.Context, arg ...string) (*exec.Cmd, error)
}
```

and if it's an interface, is `ExpectedPaths()` or `Basename()` belong there

pkg/allowedpaths/cmd.go
```diff
@@ -0,0 +1,784 @@
...
+	// We expect to know the exact location for allowlisted commands on all
+	// OSes except for a few Linux distros.
+	if runtime.GOOS != "linux" {
+		return "", fmt.Errorf("not found: %s", knownPath)
+	}
+
+	cmdName := filepath.Base(knownPath)
+	if foundPath, err := exec.LookPath(cmdName); err == nil {
+		return foundPath, nil
+	}
```
directionless, line 781: I don't think we want this for all linuxes

pkg/allowedpaths/cmd.go
RebeccaMahany: Yeah, agreed -- I'll go through and audit to see what we can move. I know there are a couple places where we call a darwin- or posix-specific func in a file that's still compiled for all platforms, so I think I'll have to do a little bit of rearranging.

pkg/allowedpaths/cmd.go
RebeccaMahany: I think the single place to adjust for nix will already be in `validatedPath`. I liked this approach because it made it easier to adjust for one-offs (e.g. like echo, which doesn't have a path on Windows). But it does certainly feel very repetitive.

RebeccaMahany: Oh, I think I misunderstood -- if you mean combining so that we have a function like --

```
func newValidatedCmd(ctx context.Context, knownPath string, arg ...string) *exec.Cmd {
        // do all the validatePath stuff here
	return exec.CommandContext(ctx, fullPathToCmd, arg...) //nolint:forbidigo
}
```

then sure, that seems fine to me

pkg/allowedpaths/cmd.go
RebeccaMahany: I thought about both, and I think both would probably be fine -- this option ultimately felt a little easier/less cumbersome to use on the consuming side.

pkg/allowedpaths/cmd.go
RebeccaMahany: Not as bad as I thought it'd be -- just a couple checkups and a couple tables to adjust.

pkg/allowedpaths/cmd.go
RebeccaMahany: Updated!

pkg/allowedpaths/cmd.go
directionless: I'm not even sure what I'm suggesing. 😆 

This blurs with the func vs interface conversation. I'm thinking about how to represent things. We have a short name, a set of possible paths, maybe some extra logic. I can imaging hanging that on an object with some interface glued on, or what you have now. I guess it's pretty easy to swap back and forth.

But I was chewing on something like:
```
type knownCommand struct {
  possiblePaths []string
}

func (kc knownCommand) Ctx(ctx context.Context, args ...string) (exec.Command, error) {
 ...
}

var Airport = knownCommand{
  possiblePaths: []string{"/System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport"}
}
```

pkg/allowedpaths/cmd.go
directionless: I'm good either way. I'm pretty sure I've always found interfaces better in the long run, but not totally clear why

pkg/allowedcmd/cmd.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil, fmt.Errorf("%s not found at %s and could not be located elsewhere", cmdName, knownPath)
+}
+
+func allowSearchPath() bool {
+	if runtime.GOOS != "linux" {
+		return false
+	}
+
+	// We only allow searching for binaries in PATH on NixOS
+	if _, err := os.Stat("/etc/NIXOS"); err == nil {
```
directionless, line 46: I'm not sure if we want to autodetect vs making a config file flag. 

Config file is slightly more effort, but also let's people twiddle things. OTOH, I can almost imagine allowing someone to specify an entire dictionary of where to find commands.

Talking about it, I guess what's here seems fine. But there's some chance we'll iterate. It being in this one place makes that pretty smooth

pkg/allowedcmd/cmd_linux.go
```diff
@@ -0,0 +1,145 @@
...
+
+func Falconctl(ctx context.Context, arg ...string) (*exec.Cmd, error) {
+	return validatedCommand(ctx, "/opt/CrowdStrike/falconctl", arg...)
+}
+
+func Falconkernelcheck(ctx context.Context, arg ...string) (*exec.Cmd, error) {
+	return validatedCommand(ctx, "/opt/CrowdStrike/falcon-kernel-check", arg...)
+}
+
+func Gnomeextensions(ctx context.Context, arg ...string) (*exec.Cmd, error) {
```
directionless, line 49:
```suggestion
func GnomeExtensions(ctx context.Context, arg ...string) (*exec.Cmd, error) {
```

Not really important, but I'd probably use camel case

pkg/osquery/tables/dataflattentable/exec.go
```diff
@@ -29,18 +28,13 @@ func WithKVSeparator(separator string) ExecTableOpt {
...
-}
-
-func TablePluginExec(logger log.Logger, tableName string, dataSourceType DataSourceType, execArgs []string, opts ...ExecTableOpt) *table.Plugin {
+func TablePluginExec(logger log.Logger, tableName string, dataSourceType DataSourceType, cmd allowedcmd.AllowedCommand, execArgs []string, opts ...ExecTableOpt) *table.Plugin {
 	columns := Columns()
 
 	t := &Table{
 		logger:            level.NewFilter(logger, level.AllowInfo()),
 		tableName:         tableName,
+		cmd:               cmd,
```
directionless, line 37:
```suggestion
		cmdGen:               cmdGen,
```

Not really important, but this isn't a cd. It's a generator.

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -41,7 +41,10 @@ func (t *touchIDSystemConfigTable) generate(ctx context.Context, queryContext ta
 
 	// Read the security chip from system_profiler
 	var stdout bytes.Buffer
-	cmd := exec.CommandContext(ctx, "/usr/sbin/system_profiler", "SPiBridgeDataType")
+	cmd, err := allowedcmd.Systemprofiler(ctx, "SPiBridgeDataType")
+	if err != nil {
+		return nil, fmt.Errorf("creating system_profiler command: %w", err)
```
directionless, line 46: This is a table -- it'll be better to log the error and return no rows. (Though I see from a couple lines down, that's not what we've done prior)

pkg/osquery/tables/dev_table_tooling/table.go
```diff
@@ -7,15 +7,27 @@ import (
...
 // allowedCommand encapsulates the possible binary path(s) of a command allowed to execute
 // along with a strict list of arguments.
 type allowedCommand struct {
-	binPaths []string
-	args     []string
+	bin  allowedcmd.AllowedCommand
+	args []string
+}
+
+var allowedCommands = map[string]allowedCommand{
```
directionless, line 22: Unifying this works today, but may not work in the future -- I sorta expect we'll have totally different command here. (Not just different paths)

directionless (commented):
With the caveat that I did not review _all_ the calls. This looks pretty good. I'd accept it, though I wanted to let other people take a look before approving.

There's a couple of small nits I saw, I think they're ignorable. 

I really like the compile safety

pkg/allowedcmd/cmd_linux.go
RebeccaMahany: Ah yeah, I have a bunch of these to update, then. I was waffling on how to capitalize anything with abbreviations in it (e.g. `bputil` -- `Bputil` or `BPUtil` or `BpUtil`) and this was my solution for not having to decide 😅 .

pkg/osquery/table/touchid_system_darwin.go
RebeccaMahany: I'm happy to update -- I'll go through and review anywhere I've touched a table to make sure behavior is consistent!

pkg/allowedcmd/cmd.go
RebeccaMahany: This seems like it should be reliable (per [this](https://discourse.nixos.org/t/detect-that-bash-script-is-running-under-nixos/11402/2)).

Maybe leave this as-is for now, and we can iterate once we're further into NixOS support and know better whether this vs a config file flag makes more sense?

pkg/osquery/table/touchid_system_darwin.go
RebeccaMahany: This is actually the only place with that issue -- I've updated.

pkg/allowedcmd/cmd_linux.go
RebeccaMahany: I have, at least, now adjusted casing for any command with punctuation in it (so `falcon-kernel-check/Falconkernelcheck` is now `FalconKernelCheck` and `gnome-extensions/Gnomeextensions` is now `GnomeExtensions`.

pkg/allowedcmd/cmd.go
directionless: Yah. I think leaving for now, and then we can iterate when we know more.

pkg/debug/checkups/gnome_extensions_linux.go
```diff
@@ -27,10 +30,6 @@ const (
 )
 
 func (c *gnomeExtensions) Name() string {
-	if runtime.GOOS != "linux" {
```
zackattack01, line 30: do we need to get rid of this pattern because it won't compile without `GnomeExtensions` for other distros? Is it maybe cleaner to make the alternate versions of offending methods in pkg/allowedcmd and just have them return nil/error since they wouldn't ever run anyway

RebeccaMahany: I went back and forth on this a bit in earlier iterations of this PR, but landed here for compile safety -- if a command isn't available on a particular OS, then we don't want to be able to use it.

There were actually relatively few places I had to update to make this work -- I think that putting OS-specific commands in `_<os>.go` is actually closer to the pattern we use frequently, and `pkg/debug/checkups` was a departure from that pattern. There were just a few one-offs in the tables I had to update, plus here, plus one in the osquery log processing (which was my fault anyway 😅 ).

zackattack01: ahh okay i see what you mean. this makes more sense to me then, thanks!

pkg/packaging/packaging.go
```diff
@@ -1,3 +1,4 @@
+//nolint:forbidigo
```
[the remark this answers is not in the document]
RebeccaMahany: Oh fair, let me update these!

Looks like we could also pull in [nolintlint](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/nolintlint/README.md) in the future.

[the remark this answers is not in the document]
zackattack01: oh nice! changes look great, thank you

zackattack01 (dismissed):
this looks understandable/useable and much safer, appreciate all the comments and the new linter! looks great 🔥

pkg/debug/checkups/gnome_extensions_linux.go
directionless: Compiler enforced restrictions is the best. When I wrote the checkups, I did the lazy thing and added the ability to shortcut without needing the build tags. And given their use, it seemed okay. But now that we have allowed commands, we need to push that through here

directionless (dismissed):

directionless (approved):

zackattack01 (approved):

James-Pickett (approved):
Wow this is pretty cool, much better. I did not review every single cmd that was updated, but went over a few to understand how it works.

PR #1446, opened 2023-11-07
use slogger on local server, add span_id id to logs, add kolide session id to logs

James-Pickett:
This PR:

* updates localserver to use slogger
* adds span id as ctx value for all spans which is picked up by multislogger middleware and added to logs
* add launcher_run_id to logs, ulid to identify single run of launcher
* add kolide session (saml id) to local server requests
* update multislogger middleware to use unexported types for ctx keys (per golang documentation)

### DIFF
--- a/ee/localserver/krypto-ec-middleware.go
+++ b/ee/localserver/krypto-ec-middleware.go
@@ -233,7 +233,7 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 			},
 		}
 
-		// setting the newReq context to the current request context context
+		// setting the newReq context to the current request context
 		// allows the trace to continue to the inner request,
 		// maintains the same lifetime as the original request,
 		// allows same ctx values such as session id to be passed to the inner request
--- a/ee/localserver/krypto-ec-middleware.go
+++ b/ee/localserver/krypto-ec-middleware.go
@@ -36,7 +36,7 @@ type v2CmdRequestType struct {
 	CallbackHeaders map[string][]string
 }
 
-func (cmdReq v2CmdRequestType) CallbackReq() (*http.Request, error) {
+func (cmdReq v2CmdRequestType) CallbackReq(ctx context.Context) (*http.Request, error) {
 	if cmdReq.CallbackUrl == "" {
 		return nil, nil
 	}
@@ -46,6 +46,7 @@ func (cmdReq v2CmdRequestType) CallbackReq() (*http.Request, error) {
 		return nil, fmt.Errorf("making http request: %w", err)
 	}
 
+	req = req.WithContext(ctx)
 	req.Header.Set("Content-Type", "application/json")
 
 	// Iterate and deep copy
@@ -103,7 +104,7 @@ func (e *kryptoEcMiddleware) sendCallback(req *http.Request, data *callbackDataS
 
 	b, err := json.Marshal(data)
 	if err != nil {
-		e.slogger.Log(req.Context(), slog.LevelDebug,
+		e.slogger.Log(req.Context(), slog.LevelError,
 			"unable to marshal callback data",
 			"err", err,
 		)
@@ -118,7 +119,7 @@ func (e *kryptoEcMiddleware) sendCallback(req *http.Request, data *callbackDataS
 
 	resp, err := client.Do(req)
 	if err != nil {
-		e.slogger.Log(req.Context(), slog.LevelDebug,
+		e.slogger.Log(req.Context(), slog.LevelError,
 			"got error in callback",
 			"err", err,
 		)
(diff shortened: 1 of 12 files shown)


ee/localserver/krypto-ec-middleware.go
```diff
@@ -190,14 +232,19 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 				Path:   cmdReq.Path,
 			},
 		}
-		newReq = newReq.WithContext(ctx) // allows the trace to continue to the inner request
+
+		// setting the newReq context to the current request context context
```
James-Pickett, line 236:
```suggestion
		// setting the newReq context to the current request context
```

ee/localserver/krypto-ec-middleware.go
```diff
@@ -113,20 +119,28 @@ func (e *kryptoEcMiddleware) sendCallback(req *http.Request, data *callbackDataS
...
+		"response-status", resp.Status,
+	)
 }
 
 func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		ctx, span := traces.StartSpan(r.Context())
+
+		spanCtx, span := traces.StartSpan(r.Context())
+		r = r.WithContext(context.WithValue(spanCtx, multislogger.SpanIdKey, span.SpanContext().SpanID().String()))
```
James-Pickett, line 142: wonder if there is a way to make multislogger middleware just inspect the context for the otel span info and pull it out instead of having to manually set it when ever we start a trace

James-Pickett: figured it out!

ee/localserver/krypto-ec-middleware.go
```diff
@@ -113,20 +118,27 @@ func (e *kryptoEcMiddleware) sendCallback(req *http.Request, data *callbackDataS
...
 		return
 	}
 
 	if resp != nil && resp.Body != nil {
 		resp.Body.Close()
 	}
 
-	level.Debug(e.logger).Log("msg", "Finished callback", "response-status", resp.Status)
+	e.slogger.Log(req.Context(), slog.LevelDebug, "finished callback",
+		"response-status", resp.Status,
```
RebeccaMahany, line 133:
```suggestion
		"response_status", resp.Status,
```

RebeccaMahany (dismissed):

RebeccaMahany (approved):

ee/localserver/krypto-ec-middleware.go
```diff
@@ -153,20 +176,44 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+			e.slogger.Log(r.Context(), slog.LevelError,
+				"unable to unmarshal cmd request",
+				"err", err,
+			)
+
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
 
+		// set the kolide session id if it exists, this also the saml session id
```
directionless, line 188: I'm not sure kolide session id and saml session id are the same.

pkg/log/multislogger/multislogger.go
```diff
@@ -8,6 +8,18 @@ import (
...
+type contextKey string
+
+func (c contextKey) String() string {
+	return string(c)
+}
+
+const (
+	// KolideSessionIdKey this the also the saml session id
+	KolideSessionIdKey contextKey = "X-Kolide-Session"
+	SpanIdKey          contextKey = "span_id"
```
directionless, line 20: This feels wrong. If these are context keys, they should probably be iotas. I don't think we should encode the header name here.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -153,20 +176,44 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+		kolideSessionId, ok := cmdReq.CallbackHeaders[multislogger.KolideSessionIdKey.String()]
+		if ok && len(kolideSessionId) > 0 {
+			span.SetAttributes(attribute.String(multislogger.KolideSessionIdKey.String(), kolideSessionId[0]))
+			r = r.WithContext(context.WithValue(r.Context(), multislogger.KolideSessionIdKey, kolideSessionId[0]))
```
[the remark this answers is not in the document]
directionless: I think it would be less awkward if the expected header name was a constant here, it should not overload `multislogger.KolideSessionIdKey`. That brings in unneeded coupling

directionless (changes requested):

ee/localserver/krypto-ec-middleware.go
James-Pickett: I ask Chris and he confirmed it was the saml id

ee/localserver/krypto-ec-middleware.go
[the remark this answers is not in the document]
James-Pickett: I've updated it so that we define the header key here as a constant. However, we still use the contextKey iota in multislogger to set context value

pkg/log/multislogger/multislogger.go
James-Pickett: no using iota, now the string of KolideSessionIdKey is `kolide_session_id`

ee/localserver/krypto-ec-middleware.go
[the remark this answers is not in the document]
directionless: Yah. That part feels awkward, but I don't know enough to know how to fix it. We are, ultimately, trying to set internal state in context that multislogger uses. So it has to come from there.

But we should not stash http headers there

ee/localserver/krypto-ec-middleware.go
directionless: They might be the same now, but I'm not sure they'll be the same later. I'm suggesting we drop the comment about saml session id.

pkg/log/multislogger/multislogger.go
```diff
@@ -8,6 +8,25 @@ import (
 	slogmulti "github.com/samber/slog-multi"
 )
 
+type contextKey int
+
+func (c contextKey) String() string {
```
directionless, line 13: Do we need these strings at all?

James-Pickett: yea, the middle ware needs to know what to set the key of the attribute to in the log

directionless (approved):
Let's try

PR #1448, opened 2023-11-08
Add revive linter and enable superfluous-else rule

RebeccaMahany:
Adds https://github.com/mgechev/revive -- https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#superfluous-else

Example error message:

```
% golangci-lint run      
cmd/launcher/main.go:78:11: superfluous-else: if block ends with call to panic function, so drop this else and outdent its block (revive)
			} else {
				level.Debug(logger).Log("msg", "Nothing new")
			}
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,7 @@ linters:
     - sqlclosecheck
     - unconvert
     - paralleltest
+    - revive
   disable:
     - errcheck
     - gosec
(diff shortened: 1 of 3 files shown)


zackattack01 (dismissed):
🔥 🚀

James-Pickett (approved):

zackattack01 (approved):

PR #1452, opened 2023-11-13
Include logs in test failure when TUF autoupdater doesn't shut down/restart within 5 seconds

RebeccaMahany:
I've seen this test failure fairly frequently: https://github.com/kolide/launcher/actions/runs/6852148852/job/18630008978

Not sure why it's happening -- it doesn't happen locally -- so I'm including the autoupdater logs in the test failure message in order to figure it out.

### DIFF
--- a/pkg/autoupdate/tuf/autoupdate_test.go
+++ b/pkg/autoupdate/tuf/autoupdate_test.go
@@ -116,7 +116,7 @@ func TestExecute_launcherUpdate(t *testing.T) {
 	shutdownDeadline := time.Now().Add(5 * time.Second).Unix()
 	for {
 		if time.Now().Unix() > shutdownDeadline {
-			t.Error("autoupdater did not shut down within 5 seconds")
+			t.Error("autoupdater did not shut down within 5 seconds -- logs: ", logBytes.String())
 			t.FailNow()
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/autoupdate/tuf/autoupdate_test.go
```diff
@@ -116,7 +116,7 @@ func TestExecute_launcherUpdate(t *testing.T) {
 	shutdownDeadline := time.Now().Add(5 * time.Second).Unix()
 	for {
 		if time.Now().Unix() > shutdownDeadline {
-			t.Error("autoupdater did not shut down within 5 seconds")
+			t.Error("autoupdater did not shut down within 5 seconds -- logs: ", logBytes.String())
```
directionless, line 119: I wonder if this is going to be readable, or kind of a mess. I wonder about just copying it to `os.Stderr`

directionless (approved):

pkg/autoupdate/tuf/autoupdate_test.go
RebeccaMahany: It'll definitely be a little messy, but it's hopefully not permanent! I don't think we'd see output if we copy to `os.Stderr` -- I find that output from tests tends to get swallowed in CI, and if I write something to `os.Stderr` in a test I don't see that output when running `make test` locally.

PR #1463, opened 2023-11-16
dont delete logs until after successful send

James-Pickett:
This _may_ be a first step in getting more observability into enrollment. This updates the send buffer so that it doesn't just drop logs it cant send. Now it will accumulate them and send them when as long as they don't go over the max storage threshold. So it's a bit of a race, but writing the the db would be a race as well ... but we would better be able to predict the winner.

Next steps:
* populate the initial accumulated logs with the device data once received from control server
* turn on debug logging for for X minutes of new launcher run
* look into tracing

### DIFF
--- a/pkg/sendbuffer/sendbuffer.go
+++ b/pkg/sendbuffer/sendbuffer.go
@@ -16,7 +16,7 @@ type sender interface {
 }
 
 var (
-	defaultMaxSize     = 128 * 1024
+	defaultMaxSize     = 512 * 1024
 	defaultMaxSendSize = 8 * 1024
 )
 
@@ -161,7 +161,8 @@ func (sb *SendBuffer) sendAndPurge() error {
 	defer sb.sendMutex.Unlock()
 
 	toSendBuff := &bytes.Buffer{}
-	if err := sb.flushToWriter(toSendBuff); err != nil {
+	lastKey, err := sb.copyLogs(toSendBuff, sb.maxSendSize)
+	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/sendbuffer/sendbuffer.go
```diff
@@ -16,7 +16,7 @@ type sender interface {
 }
 
 var (
-	defaultMaxSize     = 128 * 1024
+	defaultMaxSize     = 512 * 1024
```
James-Pickett, line 19: too much?

directionless: What's the unit?

pkg/sendbuffer/sendbuffer.go
```diff
@@ -161,7 +162,8 @@ func (sb *SendBuffer) sendAndPurge() error {
 	defer sb.sendMutex.Unlock()
 
 	toSendBuff := &bytes.Buffer{}
-	if err := sb.flushToWriter(toSendBuff); err != nil {
+	lastKey, err := sb.copyLogs(toSendBuff, sb.maxSendSize)
```
directionless, line 165: if `toSendBuff` is larger than `maxSendSize` will this cause logs to be split mid-line?

pkg/sendbuffer/sendbuffer.go
```diff
@@ -170,34 +172,46 @@ func (sb *SendBuffer) sendAndPurge() error {
...
 	}
+	// TODO: populate logs with device data (id, serial, munemo, orgid) when we
+	// get first set of control data with device info before shipping
+
+	// testing on a new enrollment in debug mode, log size hit 130K bytes
+	// before enrollment completed and was able to ship logs
+	// 2023-11-16
+	sb.writeMutex.Lock()
+	defer sb.writeMutex.Unlock()
+	sb.deleteLogs(lastKey)
```
directionless, line 186: Should these mutexes be in deleteLogs?

directionless (dismissed):
I think this is keeping a buffer, and if the buffer is exceeded the entire thing is flushed, correct? There is no FIFO style ring here? I think that's okay, this is probably good enough!

pkg/sendbuffer/sendbuffer.go
James-Pickett: bytes, I updated all the size vars to have bytes in their name

pkg/sendbuffer/sendbuffer.go
James-Pickett: the write method may also call deleteLogs if at capacity it also locks the mutex

pkg/sendbuffer/sendbuffer.go
James-Pickett: no, copyLogs peeks at the next logs size before adding and returns if adding it would exceed max size, I updated the copy logs comment.

James-Pickett:
> I think this is keeping a buffer, and if the buffer is exceeded the entire thing is flushed, correct? There is no FIFO style ring here? I think that's okay, this is probably good enough!

@directionless , I was considering doing this in a follow on PR, it's a little more involved, need to update unit tests and such, but there are other improvements I think are higher priority (populating device data, force tracing for enrollment, turn debugging on for first X minutes)

directionless (approved):

Issue #1200, opened 2023-05-19
Nix will need some kind of `finalization` step on downloaded osqueryd binaries


directionless:
Chatting with some nix folks, one of the issues that's going to bite us with autoupdate is that every time we download an osquery, we're going to need to run `patchelf` so it can find the various libraries it needs. This feels tractable as some kind of `finalization` step to me. 

Caveat, `patchelf` isn't going to make nix aware of this new, hidden, dependancy. So there are some dangers there.

Some interesting threads
- https://discourse.nixos.org/t/runtime-alternative-to-patchelf-set-interpreter/3539/20
- https://unix.stackexchange.com/questions/522822/different-methods-to-run-a-non-nixos-executable-on-nixos

I'm currently wondering if launcher can declare the dependancies, and then invoke osquery appropriately. Maybe LD paths can help

PR #1468, opened 2023-11-21
[NixOS support] Run patchelf after autoupdate download

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1200

Relates to https://github.com/kolide/launcher/issues/896

Relates to https://github.com/kolide/launcher/issues/954

<details>

Tested on a NixOS 23.05 VM with a bespoke launcher installation.

Uploaded the binary and used `patchelf --set-interpreter` to patch it. Updated the launcher installation so that launcher.flag's `localdev_path` pointed to the updated binary. Removed all updates from the launcher updates directory, and adjusted the autoupdate interval to something short.

Waited for launcher to download a new update. Verified with `patchelf --print-interpreter` that the downloaded binary had been patched appropriately.

</details>


James-Pickett (approved):

zackattack01 (approved):

ee/tuf/finalize_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	"debug/elf"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"time"
+
+	"github.com/kolide/launcher/pkg/allowedcmd"
+)
+
+// On NixOS, we have to set the interpreter for any non-NixOS executable we want to
```
directionless, line 18: Comment should be godoc style

ee/tuf/finalize_linux.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	trimmedInterpData := bytes.TrimRight(interpData, "\x00")
+
+	// interpData should look something like "/lib64/ld-linux-x86-64.so.2" -- grab just the filename
+	return filepath.Base(string(trimmedInterpData)), nil
+}
+
+func findInterpreterInNixStore(interpreter string) (string, error) {
+	storeLocationPattern := filepath.Join("/nix/store/*glibc*/lib", interpreter)
```
directionless, line 74: Hrm. I'm not sure this is correct.

I _think _ this is grabbing whatever glibc we find, but if that glibc was installed for some other package it might be removed. Is there a way to get the glibc that launcher has a dependency on?

directionless (changes requested):
Structure looks okay, but I don't think that's the way we should be working with the nix store

ee/tuf/finalize_linux.go
RebeccaMahany: Whoops, thanks, will update

ee/tuf/finalize_linux.go
RebeccaMahany: Hmm, okay -- I think we could run `patchelf --print-interpreter <path to current executable>` and assume that whatever `autoPatchelfHook` originally used is appropriate.

directionless: I might be missing something... Does launcher declare a dependancy on glibc?

directionless: Talking to Becca... I think this is still wrong, but I think solving it corretly might need more than we know today. We probably need to figure out how to declare runtime dependancies. To that end, I'm good merging and iterating

pkg/allowedcmd/cmd.go
```diff
@@ -38,14 +37,25 @@ func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*ex
...
-	if runtime.GOOS != "linux" {
-		return false
+	return IsNixOS()
+}
+
+// Save results of lookup so we don't have to stat for /etc/NIXOS every time
+// we want to know.
+var (
+	checkedIsNixOS = false
+	isNixOS        = false
```
directionless, line 47: Could also use a pointer to a bool. But this is fine

directionless (approved):
I'm still not sure we're doing the patching correctly, but I think we can try and iterate

PR #1471, opened 2023-11-22
Fall back to finding DISPLAY from display server process

RebeccaMahany:
This PR provides menu bar app support for Debian 11/12 + Cinnamon.

At least over Chrome Remote Desktop, `loginctl show-session` doesn't have a `Display` property for Debian 11/12 + Cinnamon -- but we can do similar to what we do for Wayland, and find the x display from the Xorg or Xvfb process.


James-Pickett (dismissed):
NICE!

zackattack01 (dismissed):

directionless (approved):

PR #1472, opened 2023-11-22
populate device id attributes to buffered logs before turning on shipping

James-Pickett:
* log shipper waits until it has all device id attributes before shipping
* populates buffered logs with device id attributes before sending

### DIFF
--- a/pkg/log/logshipper/logshipper.go
+++ b/pkg/log/logshipper/logshipper.go
@@ -206,7 +206,12 @@ func (ls *LogShipper) updateDevideIdentifyingAttributes() error {
 		deviceInfo[key] = string(v)
 	}
 
-	ls.shippingLogger = log.With(ls.shippingLogger, deviceInfo)
+	var deviceInfoKvps []any
+	for k, v := range deviceInfo {
+		deviceInfoKvps = append(deviceInfoKvps, k, v)
+	}
+
+	ls.shippingLogger = log.With(ls.shippingLogger, deviceInfoKvps)
 
 	var additionalSlogAttrs []slog.Attr
 	for k, v := range deviceInfo {
--- a/pkg/log/logshipper/logshipper.go
+++ b/pkg/log/logshipper/logshipper.go
@@ -203,9 +203,10 @@ func (ls *LogShipper) updateDevideIdentifyingAttributes() error {
 			return fmt.Errorf("no value for %s in server provided data", key)
 		}
 
-		ls.shippingLogger = log.With(ls.shippingLogger, key, string(v))
 		deviceInfo[key] = string(v)
 	}
+
+	ls.shippingLogger = log.With(ls.shippingLogger, deviceInfo)
 
 	var additionalSlogAttrs []slog.Attr
 	for k, v := range deviceInfo {
--- a/pkg/log/logshipper/logshipper.go
+++ b/pkg/log/logshipper/logshipper.go
@@ -120,6 +120,10 @@ func (ls *LogShipper) Ping() {
 func (ls *LogShipper) Run() error {
 	<-ls.startShippingChan
 
+	ls.knapsack.Slogger().Log(context.Background(), slog.LevelInfo,
+		"starting log shipping",
+	)
+
 	ctx, cancel := context.WithCancel(context.Background())
 
 	ls.stopFuncMutex.Lock()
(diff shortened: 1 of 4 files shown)


pkg/log/logshipper/logshipper.go
```diff
@@ -192,42 +173,142 @@ func filterResults(keyvals ...interface{}) {
...
+	deviceInfo["version"] = versionInfo.Version
 	ls.shippingLogger = log.With(ls.shippingLogger, "version", versionInfo.Version)
-	additionalSlogAttrs = append(additionalSlogAttrs, slog.String("version", versionInfo.Version))
 
-	if deviceId, err := ls.knapsack.ServerProvidedDataStore().Get([]byte("device_id")); err != nil {
-		level.Debug(ls.baseLogger).Log("msg", "could not get device id", "err", err)
-	} else {
-		ls.shippingLogger = log.With(ls.shippingLogger, "k2_device_id", string(deviceId))
-		additionalSlogAttrs = append(additionalSlogAttrs, slog.String("k2_device_id", string(deviceId)))
+	for _, key := range []string{"device_id", "munemo", "organization_id", "serial_number"} {
```
James-Pickett, line 185: this drops the `k2` prefix from some of the attributes, is that okay?

pkg/sendbuffer/sendbuffer_test.go
```diff
@@ -232,6 +233,183 @@ func requireStoreSizeEqualsHttpBufferReportedSize(t *testing.T, sb *SendBuffer)
...
+			initialLogs: [][]byte{
+				[]byte("0"),
+				[]byte("1"),
+				[]byte("2"),
+				[]byte("3"),
+				[]byte("4"),
+				[]byte("5"),
+			},
+			updateFunction: func(in io.Reader, out io.Writer) error {
+				// add 10 to the even numbers
```
James-Pickett, line 282: fix comment, this is not happening

pkg/log/logshipper/logshipper.go
directionless: I don't feel strongly. 

I often like `k2` because it makes it very clear what I'm looking at. But I'm not sure we need that disambiguation

directionless (commented):
Some nits/questions, but I think this is mostly okay.

If I understand correctly, the way this works is to have the log shipper keep a set of attributes it uses to apply to things shipped through it. 

And when it receives a `Ping`, it does two things:
1. it rebuilds the saved shipper with the new attributes
2. it iterates over the logs pending and updates them

That seems reasonable

zackattack01 (approved):
looks great 🔥

pkg/log/logshipper/logshipper.go
```diff
@@ -192,42 +182,136 @@ func filterResults(keyvals ...interface{}) {
...
+		}
+
+		if len(v) == 0 {
+			return fmt.Errorf("no value for %s in server provided data", key)
+		}
+
+		deviceInfo[key] = string(v)
+	}
+
+	ls.shippingLogger = log.With(ls.shippingLogger, deviceInfo)
```
directionless, line 209: Okay, I'm somewhat wrong -- you have to splat it out. Calling like this has the key set to `deviceInfo` and the value nil. Take a look at https://go.dev/play/p/bJMUWw28Ekp (Or building the array as you go)

pkg/log/logshipper/logshipper.go
```diff
@@ -192,42 +182,144 @@ func filterResults(keyvals ...interface{}) {
...
-	} else {
-		ls.shippingLogger = log.With(ls.shippingLogger, "k2_device_id", string(deviceId))
-		additionalSlogAttrs = append(additionalSlogAttrs, slog.String("k2_device_id", string(deviceId)))
+	for _, key := range []string{"device_id", "munemo", "organization_id", "serial_number"} {
+		v, err := ls.fetchServerProvidedData(key)
+		if err != nil {
+			return err
+		}
+
+		ls.shippingLogger = log.With(ls.shippingLogger, key, string(v))
```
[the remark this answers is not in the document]
directionless: Needs to be uglier. See https://go.dev/play/p/bJMUWw28Ekp

pkg/log/logshipper/logshipper.go
James-Pickett: ahh, good catch, updated

directionless (approved):

PR #1475, opened 2023-11-27
preserve ctx when dialing unix socket for desktop

James-Pickett:
we see a lot of errors communicating over desktop socket, I don't know if properly passing the context will help with that, but seems correct either way


zackattack01 (dismissed):

ee/desktop/user/client/dialer_posix.go
```diff
@@ -9,7 +9,8 @@ import (
 )
 
 func dialContext(socketPath string) func(_ context.Context, _, _ string) (net.Conn, error) {
-	return func(_ context.Context, _, _ string) (net.Conn, error) {
-		return net.Dial("unix", socketPath)
+	return func(ctx context.Context, _, _ string) (net.Conn, error) {
+		d := net.Dialer{}
+		return d.DialContext(ctx, "unix", socketPath)
```
James-Pickett, line 14: wonder if we should reuse dialer?

```suggestion
        d := net.Dialer{}
	return func(ctx context.Context, _, _ string) (net.Conn, error) {
		return d.DialContext(ctx, "unix", socketPath)
```

James-Pickett: updated to do the same thing the net pkg does

ee/desktop/user/client/dialer_posix.go
```diff
@@ -9,7 +9,8 @@ import (
 )
 
 func dialContext(socketPath string) func(_ context.Context, _, _ string) (net.Conn, error) {
```
directionless, line 11: Does the return need `ctx`? I'm not really sure how named vars work here

directionless (commented):
Seems reasonable. one question

ee/desktop/user/client/dialer_posix.go
James-Pickett: I'm not sure, so I added just incase, it shouldn't hurt anything

James-Pickett: actually, nvm, it's not needed

https://go.dev/play/p/BPUDbYv-slf

directionless (approved):

ee/desktop/user/client/dialer_posix.go
directionless: Thank you for checking

RebeccaMahany (approved):

Issue #1442, opened 2023-11-07
Windows sometimes hangs


directionless:
More and more I'm seeing occasions where the windows agent hangs. 

Symptomatically, I see things like distributed queries hanging, while scheduled ones continue. Inside K2, we sometimes see last seen remain current, but live query stops working. Etc.

I _suspect_ there's a query that hangs, somewhere. Might be osquery, might be launcher. And what we see is the osquery distributed thread hang, but the scheduled one remains. My hunch is that it's somehow WMI related, but I don't have any evidence to back that up.

I wonder if we can:
- still get a flare, that's a _different_ code path
- add more debug logs to capture ???
- detect and remediate?

PR #1477, opened 2023-11-28
Add querier to knapsack; expand osquery checkup

RebeccaMahany:
This PR expands the osquery checkup:

* Adds instance history to the checkup, giving us the ten most recent osqueryd restarts with reasons and timestamps
* Adds an osquery extension healthcheck to the checkup -- we already run this healthcheck periodically within the osquery runner, but I thought that adding it here would give us a better snapshot of the state of osquery in one place
* Since the above tests launcher <=> osquery communication, removes the exec to `launcher interactive` as it's redundant
* Marks the osquery checkup as log-supported, so we can get a full snapshot of osquery state in the logs regularly

To accomplish the healthchecking, I added the querier interface into knapsack.


directionless:
#1180 is related work, but driven using callbacks. I don't know that we need the callbacks. but I wanted to link the PRs

RebeccaMahany:
Closing since we don't need this at the moment

PR #1479, opened 2023-11-28
Remove os.Exit in favor of more graceful shutdown when osquery instance can't be restarted

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1387

The osquery runner currently calls os.Exit when the osquery instance dies and it cannot start a new one. In this case, the rungroup actors will never get a chance to shut down gracefully.

I've updated so that the osquery runner instead signals the osquery extension rungroup to exit its `Execute` function, allowing for a regular shutdown of all rungroup actors.


zackattack01 (approved):
nice 🚀

PR #1481, opened 2023-11-29
Add timeout to rungroup shutdown

RebeccaMahany:
Ensure that the rungroup will still exit in a timely manner even if a rungroup actor a) doesn't interrupt in a timely manner or b) doesn't terminate its execution function in a timely manner. This prevents any rungroup actor from holding launcher in a suspended state, where it isn't doing anything but also isn't exiting in order to reload itself.

### DIFF
--- a/pkg/rungroup/rungroup.go
+++ b/pkg/rungroup/rungroup.go
@@ -6,10 +6,13 @@ package rungroup
 // timeout. See: https://github.com/kolide/launcher/issues/1205
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"golang.org/x/sync/semaphore"
 )
 
 type (
@@ -28,6 +31,11 @@ type (
 		errorSourceName string
 		err             error
 	}
+)
+
+const (
+	interruptTimeout     = 5 * time.Second // How long for all actors to return from their `interrupt` function
+	executeReturnTimeout = 5 * time.Second // After interrupted, how long for all actors to exit their `execute` functions
 )
 
 func NewRunGroup(logger log.Logger) *Group {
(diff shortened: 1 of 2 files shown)


pkg/rungroup/rungroup.go
```diff
@@ -65,15 +73,39 @@ func (g *Group) Run() error {
...
+	}
+
+	// Wait for all other actors to stop, but only until we hit our executeReturnTimeout
+	timeoutTimer := time.NewTimer(executeReturnTimeout)
+	defer timeoutTimer.Stop()
 	for i := 1; i < cap(errors); i++ {
-		e := <-errors
-		level.Debug(g.logger).Log("msg", "successfully interrupted actor", "actor", e.errorSourceName, "index", i)
+		select {
+		case <-timeoutTimer.C:
```
directionless, line 101: I _think_ this case potentially leaks goroutines, which might leave a zombie process?

pkg/rungroup/rungroup.go
```diff
@@ -65,15 +73,39 @@ func (g *Group) Run() error {
 	level.Debug(g.logger).Log("msg", "received interrupt error from first actor -- shutting down other actors", "err", initialActorErr)
 
 	// Signal all actors to stop.
+	numActors := int64(len(g.actors))
+	interruptWait := semaphore.NewWeighted(numActors)
```
directionless, line 77: `semaphore.NewWeighted` is interesting as compared to a waitgroup. I think it's good, though a little harder to think about all the edges. 

Maybe cleaner than the channels I used for https://github.com/osquery/osquery-go/pull/108, less sure about perf. But that's not a real concern here.

directionless (commented):
I think this is okay. Need to chew on it a little, I think it has surprising complexity

directionless (approved):
I think it's good to try.

It feels a little weird, because this rungroup code has all this shutdown/cleanup logic hiding in the start method. But that's where it's always been, and it seems fine

PR #1484, opened 2023-11-30
Increase the default osquery wait time

directionless:
We see some logs that a 200ms causes occasional timeouts. So, let's bump this to 1s.


RebeccaMahany (dismissed):
LGTM, think lint is just complaining about the spaces in `1*time.Second`

James-Pickett (dismissed):

PR #1485, opened 2023-11-30
Add logs to track osquery errgroup shutdowns

RebeccaMahany:

We see the osquery extension manager occasionally fail to register a plugin, and osquery fail to start -- but launcher fails to restart the osquery instance. I think it's possible that the restart loop is blocking waiting for an errgroup member to finish. This PR adds logging to confirm that theory.


directionless (approved):
Seems good enough

PR #1486, opened 2023-11-30
Add log when the rungroup has completed shutdown

directionless:
I'm not totally sure this isn't duplicated in a caller, but....

### DIFF
--- a/pkg/rungroup/rungroup.go
+++ b/pkg/rungroup/rungroup.go
@@ -76,6 +76,8 @@ func (g *Group) Run() error {
 		level.Debug(g.logger).Log("msg", "successfully interrupted actor", "actor", e.errorSourceName, "index", i)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/rungroup/rungroup.go
```diff
@@ -76,6 +76,8 @@ func (g *Group) Run() error {
 		level.Debug(g.logger).Log("msg", "successfully interrupted actor", "actor", e.errorSourceName, "index", i)
 	}
 
+	level.Debug(g.logger).Log("msg", "done shutting own actors", "actor_count", len(g.actors), "initial_err", initialActorErr.err)
```
RebeccaMahany, line 79:
```suggestion
	level.Debug(g.logger).Log("msg", "done shutting down actors", "actor_count", len(g.actors), "initial_err", initialActorErr.err)
```

?

directionless: Already fixed this. And moved it to a defer

PR #1487, opened 2023-11-30
Add a timeout for killing osquery process group

RebeccaMahany:

Impose a timeout on running taskkill on Windows; returns error from running taskkill so that it can be logged by the errgroup.

### DIFF
--- a/pkg/osquery/runtime/runtime_helpers_windows.go
+++ b/pkg/osquery/runtime/runtime_helpers_windows.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os/exec"
 	"syscall"
+	"time"
 
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/pkg/allowedcmd"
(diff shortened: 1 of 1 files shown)


directionless (approved):

directionless (changes requested):
Actually... Can we log something if the context is canceled? Or maybe that will be in the `err` anyhow

RebeccaMahany:
@directionless updated, log looks like this in the case of timeout now:

```
{"caller":"runner.go:437","err":"running taskkill: context err: context deadline exceeded, err: exit status 1","level":"info","msg":"killing osquery process","session_pid":6092,"ts":"2023-11-30T19:12:20.7923216Z"}
```

And here's what it looks like with a taskkill error:

```
{"caller":"runner.go:437","err":"running taskkill: output: ERROR: Invalid argument/option - '/QQQ'.\r\nType \"TASKKILL /?\" for usage.\r\n, err: exit status 1","level":"info","msg":"killing osquery process","session_pid":6024,"ts":"2023-11-30T19:17:55.5194436Z"}
```

PR #1488, opened 2023-11-30
use slogger for kolide log adaptor

James-Pickett:


RebeccaMahany (approved):
nice

ee/log/osquerylogs/log.go
```diff
@@ -96,21 +90,23 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
 func (l *OsqueryLogAdapter) logInfoAboutUnrecognizedProcessLockingPidfile(p []byte) {
 	matches := pidRegex.FindAllStringSubmatch(string(p), -1)
 	if len(matches) < 1 || len(matches[0]) < 2 {
-		level.Debug(l.logger).Log(
-			"msg", "could not extract PID of non-osqueryd process using pidfile from log line",
+		l.slogger.Log(context.TODO(), slog.LevelDebug,
```
RebeccaMahany, line 93: Oh, debug logs don't get shipped except during startup, right? If that's correct, would you please update all of the logs in `logInfoAboutUnrecognizedProcessLockingPidfile` and the functions it calls to be info or error level? We haven't been able to reproduce that error at all, and it would be super cool if we can ship it to the cloud logs and find it!

ee/log/osquerylogs/log_posix.go
```diff
@@ -19,24 +19,26 @@ func (l *OsqueryLogAdapter) runAndLogPs(pidStr string) {
 
 	cmd, err := allowedcmd.Ps(ctx, "-p", pidStr, "-o", "user,pid,ppid,pgid,stat,time,command")
 	if err != nil {
-		level.Debug(l.logger).Log(
-			"msg", "error creating command to run ps on osqueryd pidfile",
+		l.slogger.Log(context.TODO(), slog.LevelDebug,
+			"error creating command to run ps on osqueryd pidfile",
```
directionless, line 23: This whole section should probably be level info. It's about detecting some obscure and weird error

ee/log/osquerylogs/log.go
```diff
@@ -64,9 +56,9 @@ func NewOsqueryLogAdapter(logger kitlog.Logger, rootDirectory string, opts ...Op
 func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
 	// Work around osquery being overly verbose with it's logs
 	// see: https://github.com/osquery/osquery/pull/6271
-	lf := l.levelFunc
+	level := l.level
 	if bytes.Contains(p, []byte("Executing scheduled query pack")) {
-		lf = level.Debug
+		level = slog.LevelDebug
```
directionless, line 61: I think this is okay, but I could imagine some future work where we parsed the osquery log to determine the level.

ee/log/osquerylogs/log.go
James-Pickett: updated to log an before we start dumping info about locking process (the info dump is info level or error if we encounter error getting info)

ee/log/osquerylogs/log_posix.go
James-Pickett: updated to log an before we start dumping info about locking process (the info dump is info level or error if we encounter error getting info)

cmd/launcher/extension.go
```diff
@@ -4,17 +4,18 @@ import (
...
 	"fmt"
+	"log/slog"
 	"os"
 	"path/filepath"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/kit/actor"
 	"github.com/kolide/launcher/cmd/launcher/internal"
+	kolidelog "github.com/kolide/launcher/ee/log/osquerylogs"
```
directionless, line 15: I don't think we need to call this `kolidelog` anymore

directionless (approved):
I think it's probably good

PR #1491, opened 2023-12-04
buffer initial otel spans, update enrollment logging

James-Pickett:
* buffers initial otel spans in memory until launcher gets all the info it needs to set up tracing, then passes buffered spans to span processor
* adds overrides for trace sampling rate and traces exported knspsack flags
* adds span events to enrollment
* updates enrollment to use slogger

A possible follow on to this would be to add attributes to the buffered spans, but the otel library itself doesn't have a way to do this that I have found. The only way I could get that to work was to pass our own ReadOnlySpan interface to otel that overrides the Attributes() func. However, I felt this PR had grown large enough already, so I left it out.


pkg/osquery/extension.go
```diff
@@ -375,9 +376,12 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
 	defer span.End()
 
-	logger := log.With(e.logger, "method", "enroll")
+	slogger := e.knapsack.Slogger().With("method", "enroll")
 
-	level.Debug(logger).Log("msg", "checking enrollment")
+	slogger.Log(ctx, slog.LevelInfo,
+		"checking enrollment",
+	)
+	span.AddEvent("checking_enrollment")
```
RebeccaMahany, line 384: Nice! The events will be helpful to have in the span

pkg/traces/bufspanprocessor/bufspanprocessor_test.go
```diff
@@ -0,0 +1,97 @@
...
+			t.Parallel()
+
+			maxBufSpans := 10
+
+			// create the buf span processor
+			bsp := &BufSpanProcessor{
+				MaxBufferedSpans: maxBufSpans,
+			}
+
+			// callling these for shameless code coverage
```
RebeccaMahany, line 38: 😄

pkg/traces/exporter/exporter.go
```diff
@@ -262,10 +255,46 @@ func (t *TraceExporter) setNewGlobalProvider() {
...
 		t.provider.Shutdown(t.ctx)
 	}
 
 	t.provider = newProvider
+
+	// now that we have set up new provider, see if we need a new exporter and therefore
+	// a new child processor for the buf span processor
+
+	if t.bufSpanProcessor.HasProcessor() && t.ingestUrl == t.knapsack.TraceIngestServerURL() {
+		// already have processor pointing at the same url, so no need to recreate
```
RebeccaMahany, line 270: I think we'd also want to recreate if `t.disableIngestTLS` or `t.batchTimeout` changed?

pkg/traces/exporter/exporter.go
```diff
@@ -303,6 +332,12 @@ func (t *TraceExporter) Ping() {
 	// to the new token in place.
 	t.ingestAuthToken = string(newToken)
 	t.ingestClientAuthenticator.setToken(t.ingestAuthToken)
+
+	// if the bufspanprocessor doesn't have a child processor
+	// then we don't have an exporter
+	if !t.bufSpanProcessor.HasProcessor() {
+		t.setNewGlobalProvider()
```
RebeccaMahany, line 339: Why are we doing this?

RebeccaMahany (commented):
Super cool!

pkg/traces/exporter/exporter.go
James-Pickett: I left a comment to try and explain better. If I understand correctly, we can't create otel batchSpanProcessor unless we have a working client and we have to wait until we get an auth token to get a working client

pkg/traces/exporter/exporter.go
James-Pickett: yep, you're correct ... been trying to find a good way to do this, now I'm thinking of adding another func for setNewBatchProvider or something along those lines, I'll let you know when I update

RebeccaMahany: Maybe instead of just `needsNewProvider` bool in `FlagsChanged`, you could have `needsNewProvider` and `needsNewExporter` and pass both of those in to this function? And then you could still set `t.ingestUrl` inside of `FlagsChanged` instead of having to remember to do it here

pkg/traces/exporter/exporter.go
RebeccaMahany: I don't think that was the case before -- I was able to set up a processor with a client with a bad auth token, and it would start working once the good auth token came in -- but maybe something has changed in this update.

Either way I think we can't call `t.setNewGlobalProvider()` here without a check against `t.enabled` first.

pkg/traces/exporter/exporter.go
James-Pickett: added a bool param to set new provider for rebuilding the exporter, the only time we explicitly don't to rebuild the exporter is when we set the osquery attributes, since it's async and could whomp the buffered spans if exporter is rebuilt

pkg/traces/exporter/exporter.go
James-Pickett: your correct, my mistake, it was not having the url that was causing it to fail to create exporter, not a bad / missing auth token

pkg/osquery/extension.go
```diff
@@ -400,13 +406,18 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 	}
 
 	if key != "" {
-		level.Debug(logger).Log("msg", "found stored node key, skipping enrollment")
+		slogger.Log(ctx, slog.LevelInfo,
```
directionless, line 409: This should probably be debug

pkg/osquery/extension.go
```diff
@@ -387,7 +391,9 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 	// If we already have a successful enrollment (perhaps from another
 	// thread), no need to do anything else.
 	if e.NodeKey != "" {
-		level.Debug(logger).Log("msg", "node key exists, skipping enrollment")
+		slogger.Log(ctx, slog.LevelInfo,
```
directionless, line 394: THis should probably be debug

pkg/osquery/extension.go
```diff
@@ -417,20 +428,32 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 	// it seems less likely. Try a couple times, but backoff fast.
 	var enrollDetails service.EnrollmentDetails
 	if osqPath := e.knapsack.LatestOsquerydPath(ctx); osqPath == "" {
-		level.Info(logger).Log("msg", "Cannot get additional enrollment details without an osqueryd path. This is probably CI")
+		slogger.Log(ctx, slog.LevelInfo,
+			"sipping enrollment details, no osqueryd path, this is probably CI",
```
directionless, line 432:
```suggestion
			"skipping enrollment details, no osqueryd path, this is probably CI",
```
_maybe_

directionless (commented):
I didn't look too much, but it seems okay. I might tweak a couple log levels, but 🤷

RebeccaMahany (dismissed):

RebeccaMahany (dismissed):

ee/agent/flags/flag_value_bool_test.go
```diff
@@ -3,13 +3,17 @@ package flags
 import (
 	"testing"
 
+	"github.com/kolide/launcher/pkg/agent/flags/mocks"
```
RebeccaMahany, line 6:
```suggestion
	"github.com/kolide/launcher/ee/agent/flags/mocks"
```

?

RebeccaMahany (approved):

PR #1493, opened 2023-12-05
Ensure localserver's runAsyncdWorkers shuts down on rungroup shutdown

RebeccaMahany:
Cleans up goroutine on rungroup shutdown.


zackattack01 (approved):

directionless (approved):
I'm game. I think embedding ctx is an anti-pattern, but I'm hard pressed to think of a simpler one just now

PR #1500, opened 2023-12-07
dont use array for log with, set os as log attribute

James-Pickett:


pkg/log/logshipper/logshipper.go
```diff
@@ -206,15 +209,9 @@ func (ls *LogShipper) updateDevideIdentifyingAttributes() error {
...
-	var deviceInfoKvps []any
-	for k, v := range deviceInfo {
-		deviceInfoKvps = append(deviceInfoKvps, k, v)
-	}
-
-	ls.shippingLogger = log.With(ls.shippingLogger, deviceInfoKvps)
-
 	var additionalSlogAttrs []slog.Attr
 	for k, v := range deviceInfo {
+		ls.shippingLogger = log.With(ls.shippingLogger, k, v)
```
James-Pickett, line 214: I'm not sure why the array wasn't working, this is kind of a guess. I can't consistently repro the issue where the log line sets the entire array as the key with no value

James-Pickett: ahh okay, thanks seph! updated

RebeccaMahany (approved):

Issue #1490, opened 2023-12-04
We should have a flare for pulling some of the MDM related logs


directionless:
See https://kolide.slack.com/archives/CFPMXV7CZ/p1701470617108589?thread_ts=1701458652.880249&cid=CFPMXV7CZ

I'm stuck on naming. And whether or not this is platform specific

PR #1501, opened 2023-12-08
Add checkup for Intune

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1490


ee/debug/checkups/intune_windows.go
```diff
@@ -0,0 +1,95 @@
...
+		i.summary += fmt.Sprintf("Failed to collect Intune install logs: %v. ", err)
+	}
+
+	i.summary = strings.TrimSpace(i.summary)
+
+	return nil
+}
+
+func agentLogs(zipWriter *zip.Writer) error {
+	agentLogsPathPattern := filepath.Join(os.Getenv("SYSTEMROOT"), "ProgramData", "Microsoft", "IntuneManagementExtension", "Logs", "*")
```
directionless, line 44: Do we want `*` or `*.log` here?

ee/debug/checkups/intune_windows.go
```diff
@@ -0,0 +1,95 @@
...
+	matches, err := filepath.Glob(agentLogsPathPattern)
+	if err != nil {
+		return fmt.Errorf("globbing for agent logs at %s: %w", agentLogsPathPattern, err)
+	}
+	if len(matches) == 0 {
+		return fmt.Errorf("no intune agent logs found at %s", agentLogsPathPattern)
+	}
```
directionless, line 51: Is this an error? (It will run on non-intune machines)

ee/debug/checkups/intune_windows.go
RebeccaMahany: https://learn.microsoft.com/en-us/mem/intune/remote-actions/collect-diagnostics says `%ProgramData%\Microsoft\IntuneManagementExtension\Logs\*.*` and I haven't gotten set up enough to be able to determine if we can narrow it down by extension further

ee/debug/checkups/intune_windows.go
RebeccaMahany: I think I just added this to make it more obvious what was going on when testing -- I'll update to remove here + below

ee/debug/checkups/intune_windows.go
directionless: I guess okay. I'm slightly worried we'll find a 500meg crash dump. But sure!

ee/debug/checkups/intune_windows.go
directionless: I think I was mostly quibbling over `Errorf` vs some debug/info log line, I don't really want to see the error on every flare...

ee/debug/checkups/intune_windows.go
```diff
@@ -0,0 +1,97 @@
...
+
+	if err := agentLogs(zipWriter); err != nil {
+		i.summary += fmt.Sprintf("Failed to collect Intune agent logs: %v. ", err)
+	}
+
+	if err := installLogs(zipWriter); err != nil {
+		i.summary += fmt.Sprintf("Failed to collect Intune install logs: %v. ", err)
+	}
+
+	i.summary = strings.TrimSpace(i.summary)
```
zackattack01, line 46: do we have a way to easily tell if the device is intune managed from the resulting flare? just wondering if it seems worth it to add an affirmative summary when we do find logs, vs needing to unzip and check out if we got anything

RebeccaMahany: Reasonable! Added ✅ 

Summary now looks like this:

```
Intune: Intune agent logs not found. Intune install logs not found.
```

PR #1507, opened 2023-12-12
[DO NOT MERGE] secure enclave user ctx

James-Pickett:
This PR sets up the ground for execing secure enclave in user ctx.

The broad strokes are:

ee/secureenclavesigner is a struct that implements the cyrpto.Signer interface. Its implementation of the the Public() and Sign() functions will exec out to new launcher secure-enclave subcommand when needed. Since the secureEnclaveSigner implements crypto.Signer it can be passed to the krypto/challenge package to sign responses.

TODO:
* have the secure-enclave subcommand verify the challenge sit from server before signing stuff or creating a key
* follow on PRs will update local server to initiate secureEnclaveSigner(s) and pass them to krypto lib

related krypto PR: https://github.com/kolide/krypto/pull/35


cmd/launcher/secure_enclave_other.go
```diff
@@ -0,0 +1,12 @@
...
+// +build !darwin
+
+package main
+
+import (
+	"fmt"
+)
+
+func runSecureEnclave(args []string) error {
+	return fmt.Errorf("not implemented on non darwin platforms")
```
James-Pickett, line 11: just use errors.New()

ee/secureenclavesigner/secureenclavesigner.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+type secureEnclaveSigner struct {
+	uid                  string
+	challenge            []byte
+	pubKey               *ecdsa.PublicKey
+	pathToLauncherBinary string
+}
+
+func New(uid string, challenge []byte, opts ...opt) (*secureEnclaveSigner, error) {
```
James-Pickett, line 49: should we go ahead and create / verify the key exists in secure enclave in the new func? Fail faster at the expense of a possible extra exec?

ee/secureenclavesigner/secureenclavesigner.go
```diff
@@ -0,0 +1,197 @@
...
+	"os/exec"
+	"os/user"
+	"strings"
+	"time"
+
+	"github.com/kolide/krypto/pkg/echelper"
+	"github.com/vmihailenco/msgpack/v5"
+)
+
+type SignRequest struct {
```
James-Pickett, line 23: Kind a weird that 2 / 3 members are expected to be []byte(B64) ... kind a wonder if they should all be B64 strings or all be expect to be in straight []bytes (without B64)

PR #1511, opened 2023-12-13
Enable more revive lint rules

RebeccaMahany:
Happy to add other revive lint rules if there are ones that people would like -- list is here: https://github.com/mgechev/revive#available-rules, with the caveat that I would rather not add the `comment-spacings` rule because it doesn't let me exclude cgo `//export` comments from linting.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -65,6 +65,8 @@ linters-settings:
         disabled: false
       - name: early-return
         disabled: false
+      - name: confusing-naming
+        disabled: false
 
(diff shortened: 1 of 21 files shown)


zackattack01 (approved):
love it, thanks!

zackattack01 (dismissed):

James-Pickett (approved):

directionless (approved):

zackattack01 (approved):

PR #1512, opened 2023-12-14
[TUF] Pull autoupdate config values from config file OR command-line args

RebeccaMahany:
Currently, the new autoupdater only pulls autoupdate config values (most notably the root dir and the update channel) from the config flag file, either pulling this from the command-line args or the well-known location. However, we should support setting those config values via command-line args as well; this PR implements that change.

### DIFF
--- a/ee/tuf/library_lookup.go
+++ b/ee/tuf/library_lookup.go
@@ -81,8 +81,12 @@ func getAutoupdateConfig(args []string) (*autoupdateConfig, error) {
 		return nil, fmt.Errorf("parsing command-line flags: %w", err)
 	}
 
-	// If the config file wasn't set AND the other flags weren't set, fall back
-	// to looking in the default config flag file location
+	// If the config file wasn't set AND the other critical flags weren't set, fall back
+	// to looking in the default config flag file location. (The update directory and local
+	// development path are both optional flags and not critical to library lookup
+	// functionality.) We expect all the flags to be set either via config flag (flConfigFilePath
+	// is set) or via command line (flRootDirectory and flUpdateChannel are set), but do not
+	// support a mix of both for this usage.
 	if flConfigFilePath == "" && flRootDirectory == "" && flUpdateChannel == "" {
 		return getAutoupdateConfigFromFile(launcher.ConfigFilePath(args))
 	}
--- a/ee/tuf/library_lookup.go
+++ b/ee/tuf/library_lookup.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kolide/launcher/pkg/launcher"
 	"github.com/kolide/launcher/pkg/traces"
 	"github.com/peterbourgon/ff/v3"
+	"github.com/spf13/pflag"
 )
 
 type BinaryUpdateInfo struct {
@@ -38,7 +39,7 @@ var channelsUsingNewAutoupdater = map[string]bool{
 // For now, it is only available when launcher is on the nightly update channel.
 func CheckOutLatestWithoutConfig(binary autoupdatableBinary, logger log.Logger) (*BinaryUpdateInfo, error) {
 	logger = log.With(logger, "component", "tuf_library_lookup")
-	cfg, err := getAutoupdateConfig()
+	cfg, err := getAutoupdateConfig(os.Args[1:])
 	if err != nil {
 		return nil, fmt.Errorf("could not get autoupdate config: %w", err)
 	}
@@ -52,7 +53,7 @@ func CheckOutLatestWithoutConfig(binary autoupdatableBinary, logger log.Logger) 
 }
 
 func UsingNewAutoupdater() bool {
-	cfg, err := getAutoupdateConfig()
+	cfg, err := getAutoupdateConfig(os.Args[1:])
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


ee/tuf/library_lookup.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/kolide/launcher/pkg/launcher"
 	"github.com/kolide/launcher/pkg/traces"
 	"github.com/peterbourgon/ff/v3"
+	"github.com/spf13/pflag"
```
directionless, line 16: I feel mixed about bringing in a new library, and I think we should be somewhat intentional about it.

I think the problem we have with `ff` and the standard `flags` is that they throw errors on unknown flags. I assume `pflag` handles that. Does pflag also support reading from a config file?

I think I'm game for us moving everything to pflag, and starting here. Or maybe developing our own. But I think we should _not_ use both ff and pflag in this same file...

ee/tuf/library_lookup.go
```diff
@@ -38,7 +39,7 @@ var channelsUsingNewAutoupdater = map[string]bool{
 // For now, it is only available when launcher is on the nightly update channel.
 func CheckOutLatestWithoutConfig(binary autoupdatableBinary, logger log.Logger) (*BinaryUpdateInfo, error) {
 	logger = log.With(logger, "component", "tuf_library_lookup")
-	cfg, err := getAutoupdateConfig()
+	cfg, err := getAutoupdateConfig(os.Args[1:])
```
directionless, line 42: You think os.Args, or should we be passing it in?

ee/tuf/library_lookup.go
```diff
@@ -52,21 +53,57 @@ func CheckOutLatestWithoutConfig(binary autoupdatableBinary, logger log.Logger)
...
+	pflagSet.StringVar(&flUpdateChannel, "update_channel", "", "")
+	pflagSet.StringVar(&flLocalDevelopmentPath, "localdev_path", "", "")
+
+	if err := pflagSet.Parse(args); err != nil {
+		return nil, fmt.Errorf("parsing command-line flags: %w", err)
+	}
+
+	// If the config file wasn't set AND the other flags weren't set, fall back
+	// to looking in the default config flag file location
+	if flConfigFilePath == "" && flRootDirectory == "" && flUpdateChannel == "" {
```
zackattack01, line 86: should we also be checking for `flLocalDevelopmentPath` here?

zackattack01: or if it's set, seems like we need a way to add that to what's returned from `getAutoupdateConfigFromFile`

ee/tuf/library_lookup.go
RebeccaMahany: Correct, I had to bring in [pflag](https://pkg.go.dev/github.com/spf13/pflag) because `flag`, and `ff` because it uses `flag` for command-line args, will not parse args if they contain unknown flags.

`pflag` does not support reading from a config file. The other option I considered was [go-flags](https://pkg.go.dev/github.com/jessevdk/go-flags), but this also doesn't have support for config file. There weren't any other packages I could find that would handle unknown flags appropriately, let alone do that plus support a config file. (We actually already had pflag in go.sum, as well.)

ee/tuf/library_lookup.go
RebeccaMahany: Mostly just moved `os.Args[1:]` up from where it previously was in `getAutoupdateConfig()` so that I could unit test it. I don't have a strong opinion.

ee/tuf/library_lookup.go
directionless: I can't decide if we should look at rolling our own. Or finding some other way to read a config and merge, or what. I don't love any of it, really.

Maybe the hacky way is to read the config into an array, and append flags, and parse the whole thing.

directionless: I guess I could live with this, but using two flag libraries definitely feels like it raises the messy factor

ee/tuf/library_lookup.go
RebeccaMahany: We have to have either `--config`, or `--root_directory` and `--update_channel` at the minimum, so that's what I'm checking for here -- if none of them are set, then we need to check in the well-known config file location for a config instead. `--localdev_path` and `--update_directory` are both optional so I didn't check for them here.

I think I'm not super concerned with the edge case where someone runs launcher in a weird way where they pass in `--config` with some of the values set, but specifies `--update_directory` or something via command line separately -- defaulting to the well-known config file location in that case seems fine.

zackattack01: 👍 makes sense to me, thanks for the explanation!

ee/tuf/library_lookup.go
RebeccaMahany: I'm game for moving everything to pflag or another option at some point, but I am hesitant to do that now since changing command-line flag parsing is risky and I'd rather avoid adding more risk to https://github.com/kolide/launcher/issues/954.

zackattack01 (dismissed):

ee/tuf/library_lookup.go
directionless: Yeah... I agree we shouldn't do it in _this_ PR. But I think we should figure out where we want to go, and slowly move there. 

I haven't enjoyed the various tools from spf13, so I'm initially skeptical of this one

ee/tuf/library_lookup.go
RebeccaMahany: I will update the documentation to address this better as well!

ee/tuf/library_lookup.go
RebeccaMahany: Wrote up a quick issue: https://github.com/kolide/launcher/issues/1513 -- feel free to edit.

I chose this one over `go-flags` because it's more widely used and has been updated more recently. I'm happy to trial `go-flags` for this usage if you'd like.

directionless: Meh. We can evaluate under #1513. This is probably fine here

ee/tuf/library_lookup.go
RebeccaMahany: Done

zackattack01 (approved):

ee/tuf/library_lookup.go
RebeccaMahany: @directionless I'm going to merge this as-is to move this forward, but I'm going to be doing another PR in this same area and I'm can update this there if you'd like

PR #1515, opened 2023-12-15
Add sqlite database for storing shared agent data; store and monitor startup data

RebeccaMahany:
The new `startup` package monitors the flags that it's registered to observe, and updates them in a shared sqlite database on change. It also allows for querying for these flags outside of the launcher root context. This will be useful on launcher startup, when we need to pull autoupdate-related settings before we have `knapsack` available.

The new `agentsqlite` database is a partial implementation of a KVStore -- I only added Get/Set/Update, since that was what was required for the new `startup` package.


ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return s, nil
+}
+
+// dbConn returns a connection to the database in the given rootDirectory.
+// It will create a database there if one does not yet exist.
+func dbConn(ctx context.Context, rootDirectory string) (*sql.DB, error) {
+	// Ensure db exists
+	startupDbFilepath := dbLocation(rootDirectory)
+	if _, err := os.Stat(startupDbFilepath); os.IsNotExist(err) {
```
zackattack01, line 51: wondering if there's a way that we can also check for corruption or at least an empty file here? Just thinking if this is something we want cleaned up on uninstall, we will likely have to touch an empty file here for windows on install (and i'm not sure what happens from `sql.Open` then, might be fine anyway)

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,213 @@
...
+		f, err := os.Create(startupDbFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("creating file at %s: %w", startupDbFilepath, err)
+		}
+		f.Close()
+	}
```
zackattack01, line 57: are we swallowing all non-`IsNotExist` errors

RebeccaMahany: Oh yup you're right -- I don't remember why I decided to do this but I don't think it makes sense anymore. I will update!

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
RebeccaMahany: Added some tests for this -- it looks like it already appropriately handles the case where the file is empty.

In the case where the database file is not valid, we hit the error during migration: `migrating the database: creating migrate instance: file is not a database (26)`. This is a little later than I expected -- I thought the call to `PingContext` below should've done that validation.

For now in the test I'm just requiring an error, but maybe we want to do some mitigation when we notice the db is corrupt, like delete+recreate?

zackattack01: okay nice! I was more concerned about the empty file and unsure how likely the corruption scenario is, but I definitely agree if that's something we want to handle now it would be nice to have a delete/recreate mechanism handy

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE IF NOT EXISTS keyvalue_pairs (
```
directionless, line 1: In my head, I think of the KV store as sqlite:table::boltdb:bucket. In that vein, this should probably be named something else. (probably also true up in the go code)

cmd/launcher/launcher.go
```diff
@@ -216,6 +217,13 @@ func runLauncher(ctx context.Context, cancel func(), slogger, systemSlogger *mul
 		}
 	}
 
+	s, err := startup.NewStartupDatabase(ctx, k)
+	if err != nil {
+		k.Slogger().Log(ctx, slog.LevelWarn, "could not create startup db", "err", err)
```
directionless, line 222: We might want to bail out if we can't make this. Not sure

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,94 @@
+// Package startup provides access to and manages storage of startup data:
+// flags/config values/settings that launcher needs during initialization,
+// before the knapsack is available.
+package startup
```
directionless, line 4: I _suspect_ this is broader than `startup`. I'm thinking `shared`.

Though I think it might make more sense to interface with this through knapsack. storage/sqlite conforms to our store interface right? (Or do you think we need this before we have knapsack?)

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
```diff
@@ -0,0 +1,5 @@
...
+    pair_id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
```
directionless, line 2: Do you need this? Isn't key_name enough?

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
```diff
@@ -0,0 +1,5 @@
...
+    key_name TEXT NOT NULL UNIQUE,
+    key_value TEXT
```
directionless, line 4:
```suggestion
    key TEXT NOT NULL UNIQUE,
    value TEXT
```
would be more traditional

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (s *SqliteStore) Get(key []byte) (value []byte, err error) {
+	if s == nil {
+		return nil, errors.New("store is nil")
+	}
+
+	var keyValue string
+	if err := s.conn.QueryRow(`SELECT key_value FROM keyvalue_pairs WHERE key_name = ?;`, string(key)).Scan(&keyValue); err != nil {
+		return nil, fmt.Errorf("querying flag value: %w", err)
```
directionless, line 117:
```suggestion
		return nil, fmt.Errorf(`querying key "%s": %w`,key, err)
```

maybe?

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,219 @@
...
+
+//go:embed migrations/*.sqlite
+var migrations embed.FS
+
+type SqliteStore struct {
+	conn          *sql.DB
+	rootDirectory string
+}
+
+func NewStore(ctx context.Context, rootDirectory string) (*SqliteStore, error) {
```
directionless, line 28: I think this should take a `tableName` parameter.

ee/agent/startup/startup.go
directionless: Thinking about it, an argument _against_ knapsack, is that something like `startup.GetValue()` might be cleaner than `knapsack.GetStartupValue` Maybe.

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,94 @@
...
+	_ "github.com/golang-migrate/migrate/v4/database/sqlite"
+	"github.com/kolide/launcher/ee/agent/flags/keys"
+	agentsqlite "github.com/kolide/launcher/ee/agent/storage/sqlite"
+	"github.com/kolide/launcher/ee/agent/types"
+	_ "modernc.org/sqlite"
+)
+
+// GetStartupValue retrieves the value for the given flagKey from the startup database
+// located in the given rootDirectory. It wraps creation and closing of the sqlite store.
+func GetStartupValue(ctx context.Context, rootDirectory string, flagKey string) (string, error) {
```
directionless, line 20:
```suggestion
// GetValue retrieves the value for the given flagKey from the startup database
// located in the given rootDirectory. It wraps creation and closing of the sqlite store.
func GetValue(ctx context.Context, rootDirectory string, flagKey string) (string, error) {
```

I'm not yet convinced we want to expose this here, vs through knapsack, but if we do we can use a simpler signature. Even just `Value` I guess

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: Do you have suggestions for another name? I've renamed this a couple times and haven't landed on anything that seemed great. It was `startup_flags` before, which was too specific.

ee/agent/startup/startup.go
RebeccaMahany: > I suspect this is broader than startup. I'm thinking shared.

I know we talked on slack about this too -- I do like `shared` better but I think we both felt like it was a little broad. Maybe -- `shared data`? `shared settings`?

> storage/sqlite conforms to our store interface right?

It mostly does -- I didn't implement `Delete`, `DeleteAll`, or `ForEach` since I didn't need them yet.

> Or do you think we need this before we have knapsack?

Yup, exactly -- we'll need to be able to fetch values out of this in main.go, before the knapsack exists.

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: I didn't want to do `key` because that's a reserved keyword in sqlite

ee/agent/startup/startup.go
RebeccaMahany: I will rename these when I rename the package 🙂

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: I think I'm used to avoiding varchar primary keys for performance reasons that are not at all relevant to this table 🙂 getting rid of this, making the `key_name` the primary key, and using `WITHOUT ROWID` seems most appropriate?

cmd/launcher/launcher.go
RebeccaMahany: Since it's not critical yet, I figured don't bail out now, and update in the future when it becomes critical -- but it probably makes more sense to just bail out now so I don't have to remember to update it later

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,94 @@
...
+			keys.UpdateChannel: func() string { return knapsack.UpdateChannel() },
+		},
+	}
+
+	// Attempt to ensure flags are up-to-date
+	if err := s.setFlags(ctx); err != nil {
+		s.knapsack.Slogger().Log(ctx, slog.LevelWarn, "could not set flags", "err", err)
+	}
+
+	s.knapsack.RegisterChangeObserver(s, keys.UpdateChannel)
```
James-Pickett, line 64: maybe
```suggestion
        for k := range s.storedFlags {
            s.knapsack.RegisterChangeObserver(s, k)
         }
```

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,94 @@
...
+		kvStore:  store,
+		knapsack: knapsack,
+		storedFlags: map[keys.FlagKey]func() string{
+			keys.UpdateChannel: func() string { return knapsack.UpdateChannel() },
+		},
+	}
+
+	// Attempt to ensure flags are up-to-date
+	if err := s.setFlags(ctx); err != nil {
+		s.knapsack.Slogger().Log(ctx, slog.LevelWarn, "could not set flags", "err", err)
```
James-Pickett, line 61: just a nit, but when ever we add a kvp beyond msg, I prefer it to look like this ... but that is just my preference, can live with either way

```suggestion
		s.knapsack.Slogger().Log(ctx, slog.LevelWarn, 
		    "could not set flags", 
		    "err", err,
		  )
```

James-Pickett (commented):
Nice

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
RebeccaMahany: I'm on the fence about this. I think taking a `tableName` parameter makes this store more flexible. On the other hand, it requires consumers to know info that should be abstracted from them, and it means that it's now possible for a consumer to request a store for a table that doesn't exist.

ee/agent/startup/startup.go
RebeccaMahany: Oh yup absolutely! Will update.

ee/agent/startup/startup.go
RebeccaMahany: I prefer that as well! I'll update.

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: Updated like ☝️

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
RebeccaMahany: So something like this would address the drawbacks above:

```
const (
	TableKeyValuePairs = "keyvalue_pairs"
)

var supportedTables = map[string]struct{}{
	TableKeyValuePairs: struct{}{},
}

...


func NewStore(ctx context.Context, rootDirectory string, tableName string) (*SqliteStore, error) {
	if _, ok := supportedTables[tableName]; !ok {
		return nil, fmt.Errorf("unsupported table %s", tableName)
	}
...
```

...but now we've either got to standardize the two column names for all possible tables, or take and validate those as arguments too. Feels maybe like overoptimizing at this point?

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
directionless: huh. Okay!

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
directionless: `startup_settings` isn't bad.

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: I also figure that once I rename this table per https://github.com/kolide/launcher/pull/1515#discussion_r1433386798, I can update the column names to something less clunky 🙂

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
directionless: I'll grant that this might be premature optimization, but here's how I structure this in my head.
- We want to store information
- For historical reasons, our information is stored in KV stores, broken up by area/type etc
- (This is implemented in bolt, using buckets, which provide kv stores)
- We don't love bbolt. The store interface is a newish layer on top of bbolt to allow us to play with swapping out the underlying implementation
- We now need to content with multi-accessor
- And also not-private data
- To me, this fits with our existing model -- segment that into "buckets" and treat them like KV stores.
- Yes, that's not as powerful as using all of what SQL can provide, but there's value in sticking with our interfaces
- Which leads me to a place of wanting
   + underlying backing stores (like bbolt and sqlite)
   + consistent store layer build on ^^^
   + deciding which "bucket" uses which backing store. 
  
I see the `ee/agent/storage/sqlite` as being that backing store. But I think it needs that bucket layer on top of it. 

The startup settings stuff feels a natural "bucket" to me.

In terms of signatures, I think `map[string]struct` would be weird, but I can imagine `type supportedStores string` and an list of known ones. I think it's correct to use the same column names -- any KV using sqlite should have the same schema.

RebeccaMahany: This makes sense to me, and I'll add it!

> In terms of signatures, I think `map[string]struct` would be weird, but I can imagine `type supportedStores string` and an list of known ones.

I like a map over a list for ease of checking whether a table is supported or not. `map[string]struct{}` is a little better than `map[string]bool` IMO because it's clearer that you're just checking for presence in the map, instead of maybe checking for presence in the map but maybe also needing to check if the bool is true or false.

RebeccaMahany: I've updated this!

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: I like that!

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
RebeccaMahany: Went with `name` and `value`

ee/agent/startup/startup.go
RebeccaMahany: Now that we've named the database table `startup_settings` I feel like switching to `shared` here is weird...do we want to do `shared_settings` or `startup_settings` in both places?

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
RebeccaMahany: I'm not sure how likely the corruption scenario is either, but probably better to handle it now than wait for it to be a problem. It looks like [pragma quick_check](https://www.sqlite.org/pragma.html#pragma_quick_check) is a good option for verifying -- I'll give that a shot.

directionless (commented):
Mostly nits. (and our future selves, we're having a slack discussion about the call semantics of startup)

ee/agent/storage/sqlite/migrations/000001_kv.up.sqlite
directionless: Yeah, `name` is good

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/kolide/launcher/ee/agent/types"
+	_ "modernc.org/sqlite"
+)
+
+// GetStartupValue retrieves the value for the given flagKey from the startup database
+// located in the given rootDirectory. It wraps creation and closing of the sqlite store.
+func GetStartupValue(ctx context.Context, rootDirectory string, flagKey string) (string, error) {
+	store, err := agentsqlite.NewStore(ctx, rootDirectory, agentsqlite.TableStartupSettings)
+	if err != nil {
+		return "", fmt.Errorf("opening startup db in %s: %w", rootDirectory, err)
```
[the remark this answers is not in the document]
RebeccaMahany: If we wipe the database, then there's nothing to retry here -- the database is wiped so there is no value to retrieve.

ee/agent/startup/startup.go
```diff
@@ -0,0 +1,101 @@
...
+// flags/config values/settings that launcher needs during initialization,
+// before the knapsack is available.
+package startup
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	_ "github.com/golang-migrate/migrate/v4/database/sqlite"
```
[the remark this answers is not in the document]
RebeccaMahany: Ah yup, thank you!

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+var supportedTables = map[string]struct{}{
+	TableStartupSettings: {},
+}
+
+//go:embed migrations/*.sqlite
+var migrations embed.FS
+
+type SqliteStore struct {
```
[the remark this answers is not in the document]
RebeccaMahany: I've unexported this now that I can use `types.GetterCloser` and `types.GetterUpdaterCloser` to refer to this store. I think the name is fine now that it's unexported -- it's equivalent to the others e.g. `inMemoryKeyValueStore`.

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+const (
+	// Supported tables -- all tables should have the same `name` and `value` columns
+	TableStartupSettings = "startup_settings"
+)
+
+var supportedTables = map[string]struct{}{
+	TableStartupSettings: {},
+}
```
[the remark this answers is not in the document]
RebeccaMahany: Is this really safer? A consumer could still pass in an invalid value -- `OpenRO(ctx, rootDirectory, 100000001)`.

(We also still need the store consts to be exported for other packages to pass in their desired table names.)

cmd/launcher/launcher.go
```diff
@@ -216,6 +217,12 @@ func runLauncher(ctx context.Context, cancel func(), slogger, systemSlogger *mul
 		}
 	}
 
+	s, err := startup.NewStartupDatabase(ctx, k)
+	if err != nil {
+		return fmt.Errorf("creating startup db: %w", err)
```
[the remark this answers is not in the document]
RebeccaMahany: I can swap it back, but I think we'll want it this way at some point anyway

James-Pickett (approved):
LGTM!

cmd/launcher/launcher.go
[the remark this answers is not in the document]
James-Pickett: we do the same with launcher.db ... we could just log it for now

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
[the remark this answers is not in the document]
directionless: True -- go's enums are kinda fake, so you _could_ pass an int. But that's a much harder mistake to make than `OpenRO(ctx, rootDirectory, "startup_setings") (Or "StartupSettings" vs "startupSettings" vs "startup_settings")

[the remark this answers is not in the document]
RebeccaMahany: Yeah, that's a good point -- I will update!

ee/agent/startup/startup.go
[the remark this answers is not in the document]
RebeccaMahany: Done!

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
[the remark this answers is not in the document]
RebeccaMahany: Done!

cmd/launcher/launcher.go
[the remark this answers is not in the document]
RebeccaMahany: I think we should leave this as-is. We'll want to require it soon enough anyway.

James-Pickett (dismissed):

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,300 @@
...
+	query := fmt.Sprintf(`SELECT value FROM %s WHERE name = ?;`, s.tableName)
+
+	var keyValue string
+	if err := s.conn.QueryRow(query, string(key)).Scan(&keyValue); err != nil {
+		return nil, fmt.Errorf("querying key `%s`: %w", string(key), err)
+	}
+	return []byte(keyValue), nil
+}
+
+func (s *sqliteStore) Set(key, value []byte) error {
```
directionless, line 196: Consider something like `s.readOnly` as a check around `OpenRO` creating the DB and allowing writes.

directionless (approved):
I think this is okay. I like the underlying functions, but I feel a little nervous about error handling

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -0,0 +1,292 @@
...
+// not perform database creation or migration.
+func OpenRO(ctx context.Context, rootDirectory string, tableName string) (*sqliteStore, error) {
+	if _, ok := supportedTables[tableName]; !ok {
+		return nil, fmt.Errorf("unsupported table %s", tableName)
+	}
+
+	conn, err := sql.Open("sqlite", dbLocation(rootDirectory))
+	if err != nil {
+		return nil, fmt.Errorf("opening startup db in %s: %w", rootDirectory, err)
+	}
```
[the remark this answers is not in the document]
RebeccaMahany: You'd asked for the RO option to be more lightweight and to skip the validation when we'd discussed on slack, since we need to use the RO option early on in launcher startup.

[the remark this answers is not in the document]
directionless: I was thinking about skipping migration -- I'm less sure what should happen if the DB file isn't present. Or doesn't match expected schema. Bailing early and getting nothing isn't wrong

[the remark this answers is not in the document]
directionless: I guess that's probably the right way to look at it -- If the DB is wrong, we should bail out and get nothing. Maybe it doesn't exist yet, maybe it hasn't been udpated to the current schema. So whatever we need to catch that and move on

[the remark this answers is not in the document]
RebeccaMahany: 👍 I will handle that in the next PR that starts to use the RO connection

directionless (approved):

James-Pickett (approved):

PR #1517, opened 2023-12-18
Use embedded flags interface, and not wrapper functions

directionless:
The other day James pointed out that we didn't need all these methods that simply called to the underlying `flags`. Instead, we could simply use the embedded methods from the type.

I'm not wholly sure how to test this. It compiles... Suggestions?


James-Pickett:
this looks correct to me, I think compiling / soaking on nightly for a while is a good enough test

James-Pickett:
closed for https://github.com/kolide/launcher/pull/1575

PR #1525, opened 2023-12-29
update desktop to use slogger

James-Pickett:


ee/desktop/runner/server/server.go
```diff
@@ -36,21 +35,21 @@ type controlRequestIntervalOverrider interface {
...
 		listener:                        listener,
-		logger:                          logger,
+		slogger:                         slogger,
 		desktopProcAuthTokens:           make(map[string]string),
 		controlRequestIntervalOverrider: controlRequestIntervalOverrider,
 	}
 
-	if rs.logger != nil {
-		rs.logger = log.With(rs.logger, "component", "desktop_runner_root_server")
+	if rs.slogger != nil {
```
RebeccaMahany, line 51: Do we need to enforce that rs.slogger is not nil, or will calls to rs.slogger.Log below fail gracefully when nil instead of panicking?

RebeccaMahany (dismissed):
LGTM, just the one question -- if you decide to update, ping me and I can reapprove 🙂

ee/desktop/runner/server/server.go
James-Pickett: good catch, now returning error if slogger is nil

RebeccaMahany (approved):

PR #1534, opened 2024-01-08
Modify flags when needed to accommodate pflag

RebeccaMahany:
pflag, while mostly great for our usecase in the tuf package, expects getopt-style flags, which means it doesn't support the Golang standard of using single and double dashes interchangeably for flags. (e.g., pflag cannot parse `-config`, but Golang treats `-config` the same as `--config`.)

This PR transforms all single-dash args to double-dashes so that pflag can parse them as expected.

### DIFF
--- a/ee/tuf/library_lookup.go
+++ b/ee/tuf/library_lookup.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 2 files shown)


ee/tuf/library_lookup.go
```diff
@@ -64,6 +65,21 @@ func ShouldUseNewAutoupdater(ctx context.Context) bool {
 // getAutoupdateConfig pulls the configuration values necessary to work with the autoupdate library
 // from either the given args or from the config file.
 func getAutoupdateConfig(args []string) (*autoupdateConfig, error) {
+	// pflag, while mostly great for our usecase here, expects getopt-style flags, which means
+	// it doesn't support the Golang standard of using single and double dashes interchangeably
+	// for flags. (e.g., pflag cannot parse `-config`, but Golang treats `-config` the same as
+	// `--config`.) This transforms all single-dash args to double-dashes so that pflag can parse
+	// them as expected.
+	argsToParse := make([]string, len(args))
```
directionless, line 73: seems fine (well, as much as any hack) but why not modify `args` in place?

directionless (approved):

ee/tuf/library_lookup.go
RebeccaMahany: It felt a little more cautious to not modify `os.Args`

directionless: Is it os.Args? I guess I'm not sure how the references/values work for the `args []string` in the signature

RebeccaMahany: It is `os.Args`, and modifying `args` here would modify `os.Args`

RebeccaMahany: (https://play.golang.com/p/IcOt9XjGcvb if you'd like to see)

RebeccaMahany: Want me to add a comment in the code re: why I'm not modifying in place?

directionless: Well, I guess arrays in Go are all references, so there we go. 

I don't know that it needs comment.  🤷

PR #1541, opened 2024-01-10
Remove some unused code

RebeccaMahany:
Removes launcher-pummel, simulator, and legacy v1 autoupdate tests -- we know we don't use any of these.


directionless (approved):
Byyyeeeeeee

PR #1542, opened 2024-01-11
Don't use reserved version key in logs

RebeccaMahany:
The logshipper adds a `version` key for current launcher version to logs, so using `version` keys elsewhere means we miss out on that logshipper data.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -577,7 +577,7 @@ func runOsqueryVersionCheck(ctx context.Context, slogger *slog.Logger, osquerydP
 
 	slogger.Log(ctx, slog.LevelDebug,
 		"checked osqueryd version",
-		"version", outTrimmed,
+		"osqueryd_version", outTrimmed,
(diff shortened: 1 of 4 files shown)


directionless (dismissed):

James-Pickett (approved):

James-Pickett:
alternatively, we could update our current "version" keyword to "launcher_version"

RebeccaMahany:
@James-Pickett I'm fine either way -- do you have a preference? I didn't know if maybe the ingest server or something in Cloud Log relied on `version` -- but after looking into it, it doesn't seem like it does.

James-Pickett:
> @James-Pickett I'm fine either way -- do you have a preference? I didn't know if maybe the ingest server or something in Cloud Log relied on `version` -- but after looking into it, it doesn't seem like it does.

feel like using `version` again in a log message is a possibility and we would need to lint for it or something, but if we switch to `launcher_version` we kind of future proof ourselves from making same error in the future

James-Pickett (approved):

PR #1551, opened 2024-01-16
add job to make sure version match

James-Pickett:
adds github action job between ci and exec tests to grab the version from a single runner and make sure all other runners match


directionless (approved):
Seems okay to try

PR #1557, opened 2024-01-22
control server sends message on start

James-Pickett:
sends a message to the control server when launcher starts up

### DIFF
--- a/ee/control/client_http.go
+++ b/ee/control/client_http.go
@@ -150,23 +150,15 @@ func (c *HTTPClient) GetSubsystemData(hash string) (io.Reader, error) {
 	return reader, nil
 }
 
-// serverMessageMethod is a method that can be sent to the server, it is
-// set as the method in json-rpc format
-type serverMessageMethod string
-
-const (
-	recheck serverMessageMethod = "recheck"
-)
-
 // MessageServer sends a message to the server using JSON-RPC format
-func (c *HTTPClient) MessageServer(method serverMessageMethod, params interface{}) error {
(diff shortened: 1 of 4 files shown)


RebeccaMahany (dismissed):

ee/control/client_http.go
```diff
@@ -150,6 +150,42 @@ func (c *HTTPClient) GetSubsystemData(hash string) (io.Reader, error) {
 	return reader, nil
 }
 
+// MessageServer sends a message to the server using JSON-RPC format
+func (c *HTTPClient) MessageServer(method string, params interface{}) error {
```
directionless, line 154: I'd name this `MessageSender` "server" sounds like it's the receiver portion

directionless (approved):
I might quibble over the name,  but I think the code is fine.

This does use the control token, which I think is fine, but also means the control system has to be up. How does that work with not yet enrolled clients?

ee/control/client_http.go
James-Pickett: How about just `Message`? `MessageSender` feels strange to me because this in http client and the client does the "sending".

James-Pickett:
@directionless 

> This does use the control token, which I think is fine, but also means the control system has to be up. How does that work with not yet enrolled clients?

If the control server issues a token, we'll attempt to use it to send "startup message". If we get a non-200 back on the "startup" message, we'll try again the next time we fetch a new token and keep trying until we get a success.

RebeccaMahany (approved):

directionless (approved):

PR #1559, opened 2024-01-23
Don't select launcher version under v1.4.1 on stable

RebeccaMahany:
If a user has a recent launcher version (e.g. 1.4.4) in their legacy autoupdate library _and_ this recent launcher version is not the current stable (e.g. we roll back from 1.4.4 to 1.4.2) _and_ their new autoupdate library has versions that are only _older_ than 1.4.1 (e.g. 1.3.2), they can get stuck in a loop on startup:

1. The older installed version of launcher bootstraps itself to the newest version in the legacy library v1.4.4
2. v1.4.4 recognizes that it's not on the the release version v1.4.2, so falls back to loading the newest version in the new autoupdate library, v1.3.2
3. v1.3.2 does not know it should use the new autoupdate library yet, so it goes back through and bootstraps itself to the newest version in the legacy library, v1.4.4
4. Repeat 2-3

### DIFF
--- a/ee/tuf/library_lookup.go
+++ b/ee/tuf/library_lookup.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/launcher/ee/agent/flags"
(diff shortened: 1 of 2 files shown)


James-Pickett (approved):

directionless (commented):
This is a good theory! Should we cut a 1.4.5 with it?

RebeccaMahany:
@directionless yes I think that would be a good idea!

PR #1561, opened 2024-01-26
Set up tracing as early as possible to capture initial traces

RebeccaMahany:
Create an initial trace buffer (plus tracer provider, etc) as soon as possible in runLauncher, before knapsack/osquery client are available. Transfer buffer and attrs to trace exporter once available.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -56,6 +56,7 @@ import (
 	osqueryInstanceHistory "github.com/kolide/launcher/pkg/osquery/runtime/history"
 	"github.com/kolide/launcher/pkg/rungroup"
 	"github.com/kolide/launcher/pkg/service"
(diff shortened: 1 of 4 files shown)


cmd/launcher/launcher.go
```diff
@@ -73,6 +74,9 @@ const (
 // rungroups with the various options, and goes! If autoupdate is
 // enabled, the finalizers will trigger various restarts.
 func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSlogger *multislogger.MultiSlogger, opts *launcher.Options) error {
+	initialTraceBuffer := exporter.NewInitialTraceBuffer()
+	ctx, startupSpan := traces.StartSpan(ctx)
```
directionless, line 78: Can we name this span?

directionless (approved):

James-Pickett (approved):

PR #1563, opened 2024-01-29
Collect more timing info on launcher startup

RebeccaMahany:
Building on https://github.com/kolide/launcher/pull/1561 -- collect more timing info to be able to decrease launcher startup time.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -117,7 +117,7 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 			"err", err,
 		)
 	}
-	startupSpan.AddEvent("dns_lookup_complete")
+	startupSpan.AddEvent("dns_lookup_completed")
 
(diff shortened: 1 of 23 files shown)


cmd/launcher/launcher.go
```diff
@@ -116,6 +117,7 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 			"err", err,
 		)
 	}
+	startupSpan.AddEvent("dns_lookup_complete")
```
zackattack01, line 120: just double checking you don't want `completed` here instead to match the others

zackattack01 (approved):

cmd/launcher/launcher.go
RebeccaMahany: Ah whoops, thank you! Updated

zackattack01 (approved):

James-Pickett (approved):

PR #1566, opened 2024-01-30
run first checkpointer in go routine

James-Pickett:


James-Pickett:
someone else got to it first https://github.com/kolide/launcher/pull/1567

PR #1567, opened 2024-01-30
Speed up launcher startup

RebeccaMahany:
* `Once()` can take up to 5 seconds to run -- don't block on running it
* We can't use the Secure Enclave right now -- stop waiting for up to a second to access it

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -276,8 +276,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	}
 
 	// Add the log checkpoints to the rungroup, and run it once early, to try to get data into the logs.
+	// The checkpointer can take up to 5 seconds to run, so do this in the background.
 	checkpointer := checkups.NewCheckupLogger(logger, k)
-	checkpointer.Once(ctx)
+	go checkpointer.Once(ctx)
 	runGroup.Add("logcheckpoint", checkpointer.Run, checkpointer.Interrupt)
 
 	// Create a channel for signals
--- a/ee/agent/keys.go
+++ b/ee/agent/keys.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"crypto"
 	"fmt"
+	"runtime"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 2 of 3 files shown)


ee/agent/keys.go
```diff
@@ -39,18 +40,21 @@ func SetupKeys(logger log.Logger, store types.GetterSetterDeleter) error {
 		return fmt.Errorf("setting up local db keys: %w", err)
 	}
 
-	err = backoff.WaitFor(func() error {
-		hwKeys, err := setupHardwareKeys(logger, store)
+	// Secure Enclave is not currently supported, so don't spend startup time waiting for it to work -- see keys_darwin.go for more details.
+	if runtime.GOOS != "darwin" {
```
zackattack01, line 44: might be simpler as `if runtime.GOOS == "darwin" { return nil }`

cmd/launcher/launcher.go
```diff
@@ -276,8 +276,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	}
 
 	// Add the log checkpoints to the rungroup, and run it once early, to try to get data into the logs.
+	// The checkpointer can take up to 5 seconds to run, so do this in the background.
 	checkpointer := checkups.NewCheckupLogger(logger, k)
-	checkpointer.Once(ctx)
+	go checkpointer.Once(ctx)
```
zackattack01, line 281: nice catch! 🔥

zackattack01 (dismissed):

James-Pickett (dismissed):
nice

ee/agent/keys.go
RebeccaMahany: Ah yup that reads much better, thanks -- updated

cmd/launcher/launcher.go
directionless, line 281: The prior way was intentional -- it was intending to get some of the identifying information into the logs _before_ any early crashes might occur. It's fine to revisit, maybe we don't need it with all the things we've subsequently added.

Though I think this begs the question -- why does it take 5s to log, and should we add go routines downstack?

ee/agent/keys.go
```diff
@@ -39,6 +40,11 @@ func SetupKeys(logger log.Logger, store types.GetterSetterDeleter) error {
 		return fmt.Errorf("setting up local db keys: %w", err)
 	}
 
+	// Secure Enclave is not currently supported, so don't spend startup time waiting for it to work -- see keys_darwin.go for more details.
+	if runtime.GOOS == "darwin" {
+		return nil
+	}
```
directionless, line 46: @James-Pickett Is this okay with the other work you have in flight?

cmd/launcher/launcher.go
RebeccaMahany: As suspected -- it's the connectivity checkup: https://github.com/kolide/launcher/blob/main/ee/debug/checkups/connectivity.go

directionless: I wonder if we should just async that one check. Not sure

RebeccaMahany: That seemed like maybe overly complicated to coordinate an exclusion for that one checkup for the checkpoint only, but not doctor/flare

directionless: Thinking about it... Maybe this simple thing is good enough for now.

We probably _do_ want to keep this in the log checkpoint, it provides an interesting point-in-time view. 

I suspect that the theoretical correct state would be to make the timer log checkpoints use a different set of debuggers than the initial startup ones. And I'm not sure it's worth doing that this moment

RebeccaMahany: Agreed!

PR #1571, opened 2024-01-31
message menu item

James-Pickett:


ee/control/control.go
```diff
@@ -278,6 +278,14 @@ func (cs *ControlService) RegisterSubscriber(subsystem string, subscriber subscr
 	cs.subscribers[subsystem] = append(cs.subscribers[subsystem], subscriber)
 }
 
+func (cs *ControlService) Message(method string, params interface{}) error {
+	if cs.fetcher == nil {
+		return errors.New("fetcher is nil, cannot send message to server")
+	}
+
+	return cs.fetcher.Message(method, params)
```
directionless, line 286: The other methods use `GetX` Should this be `SendMessage`? (I'm not sure)

ee/control/control.go
```diff
@@ -278,6 +278,14 @@ func (cs *ControlService) RegisterSubscriber(subsystem string, subscriber subscr
 	cs.subscribers[subsystem] = append(cs.subscribers[subsystem], subscriber)
 }
 
+func (cs *ControlService) Message(method string, params interface{}) error {
+	if cs.fetcher == nil {
```
directionless, line 282: When will this be nil? If it's just in testing, the error seems fine.

But if it's a startup sequencing thing, should we async in here? (Downside is that it's a lot of complexity, and maybe the caller can use backoff if it matters)

ee/desktop/runner/server/server.go
```diff
@@ -151,3 +163,42 @@ func (ms *RunnerServer) isAuthTokenValid(authToken string) bool {
...
+
+func (ms *RunnerServer) message(w http.ResponseWriter, r *http.Request) {
+	if r.Body == nil {
+		// TODO: log
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	defer r.Body.Close()
+
+	data := make(map[string]interface{})
```
directionless, line 175: As this has an expected structure, why not use a more clear struct?

directionless: This is also something users can freely send data too. I think we should do a little sanitization before we send to k2. (K2 can handle whatever, but why make trouble). I think the only sanitization that's obvious is capping the outbound message size. Something like 1k?

directionless (commented):
Looks about right. 

One call out -- there's no real way we can auth from desktop to runner. I think that's inherent, and I'm not sure it's an issue. After all, the user can just use the menu. But short of moving everything to being in the user process, I don't think there's a fix

ee/desktop/user/menu/action.go
```diff
@@ -51,6 +52,12 @@ func (a *Action) UnmarshalJSON(data []byte) error {
 		a.Performer = openURL
 	case Flare:
 		a.Performer = actionFlare{}
+	case Message:
```
RebeccaMahany, line 55: I think `Message` is a little vague -- I could see it being confusing down the line. Maybe `MessageX` to indicate the recipient of the message -- like`MessageK2`, `MessageControl`, `MessageCloud`?

RebeccaMahany (approved):

RebeccaMahany (approved):

ee/desktop/user/menu/action.go
James-Pickett: updated to MessageControl

ee/desktop/runner/server/server.go
James-Pickett: now using struct

ee/control/control.go
James-Pickett: it will never be nli, removed nil check

ee/control/control.go
James-Pickett: update to SendMessage

RebeccaMahany (approved):

directionless (approved):

PR #1573, opened 2024-02-01
Use TUF metadata to determine version to download when packaging

RebeccaMahany:
Eventually, we will stop using the legacy Notary system, which means we will stop duplicating releases at `<binary>-<channel>.tar.gz` and expect clients to find the releases using the TUF metadata for that binary instead. This PR prepares for that by performing those steps when packaging.

This duplicates `root.json` (also located at `ee/tuf/assets...`) to avoid an import cycle. I didn't think this was a big enough deal to warrant pulling it out and moving it to a new shared location, but if others feel differently I'm happy to update.


pkg/packaging/fetch.go
```diff
@@ -14,7 +16,8 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/kolide/kit/fsutil"
 	"github.com/kolide/launcher/pkg/contexts/ctxlog"
-
+	"github.com/theupdateframework/go-tuf/client"
+	filejsonstore "github.com/theupdateframework/go-tuf/client/filejsonstore"
```
directionless, line 20:
```suggestion
 "github.com/theupdateframework/go-tuf/client/filejsonstore"
```

I think?

pkg/packaging/fetch.go
```diff
@@ -108,6 +114,81 @@ func FetchBinary(ctx context.Context, localCacheDir, name, binaryName, version s
...
+	}
+
+	// Set up our local store i.e. point to the directory in our filesystem
+	localStore, err := filejsonstore.NewFileJSONStore(tempDir)
+	if err != nil {
+		return "", fmt.Errorf("could not initialize local TUF store: %w", err)
+	}
+
+	// Set up our remote store i.e. tuf.kolide.com
+	remoteStore, err := client.HTTPRemoteStore("http://tuf.kolide.com", &client.HTTPRemoteOptions{
```
directionless, line 161:
```suggestion
	remoteStore, err := client.HTTPRemoteStore("https://tuf.kolide.com", &client.HTTPRemoteOptions{
```

pkg/packaging/fetch.go
```diff
@@ -108,6 +114,81 @@ func FetchBinary(ctx context.Context, localCacheDir, name, binaryName, version s
...
+
+	return dlTarPathFromVersion(name, version, platform, arch), nil
+}
+
+func dlTarPathFromVersion(name, version, platform, arch string) string {
+	return path.Join("kolide", name, platform, arch, fmt.Sprintf("%s-%s.tar.gz", name, version))
+}
+
+//go:embed assets/tuf/root.json
+var rootJson []byte
```
directionless, line 139: I think this is okay. Though we may want some kind of two step, so that we can maintain a newer TUF repo, and not have to fetch/validate everything.

directionless:
Looks good. tiny nits

pkg/packaging/fetch.go
RebeccaMahany: It conflicts with the import of `"github.com/theupdateframework/go-tuf/client"` because that's actually the name of the package, which is a little annoying:

```
pkg/packaging/fetch.go:20:2: "github.com/theupdateframework/go-tuf/client/filejsonstore" imported as client and not used
```

pkg/packaging/fetch.go
RebeccaMahany: That'd be nice to have for the ee/tuf code too -- it's on my list (eventually)

pkg/packaging/fetch.go
directionless: Ha. So they sloppily named the package _not_ what the folder is. oops!

directionless (approved):

pkg/packaging/fetch.go
RebeccaMahany: https://github.com/kolide/launcher/issues/1577

PR #1578, opened 2024-02-02
overwrite multislogger.Logger memory rather than reassigning pointer

James-Pickett:
Noticed that some logs in main func were not getting shipped. When adding a handler we can overwrite the the memory address of the multslogger.Logger to make handlers propagate to previously assigned Loggers.


ee/agent/knapsack/knapsack.go
```diff
@@ -51,7 +51,7 @@ func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB
 	}
 
 	if k.slogger != nil {
-		k.slogger.Logger = k.slogger.Logger.With("launcher_run_id", k.launcherRunId)
+		*k.slogger.Logger = *k.slogger.Logger.With("launcher_run_id", k.launcherRunId)
```
directionless, line 54: This spot feels like it breaks an abstraction. Somehow it feels wrong for knapsack to be reaching into slogger.Logger. But 🤷 not the end of the world

directionless (dismissed):
seems okay to try

ee/agent/knapsack/knapsack.go
James-Pickett: I agree, it's a weird spot. Knapsack needs access to the multislogger to be able to add handlers. We also want the launcher_run_id to be universally included in all logs ..... hmm maybe multislogger should just add it, then knapsack can stop reaching in there

James-Pickett: I moved this down to the multislogger, now knapsack is staying in it's lane and not reaching down into multislogger's guts

directionless (approved):
LGTM

RebeccaMahany (approved):

Issue #1577, opened 2024-02-02
Autoupdate improvements


RebeccaMahany:
After we release the new autoupdater (https://github.com/kolide/launcher/issues/954), there are some improvements I'd like to make and new features we can add --

- [ ] https://github.com/kolide/launcher/issues/1412
- [x] https://github.com/kolide/launcher/issues/1705
- [x] Report new version to K2 after update https://github.com/kolide/launcher/pull/1728
- [ ] Modernize `dl.kolide.com` hosting
- [x] An "update now" functionality tied to control server: https://github.com/kolide/launcher/pull/1579, https://github.com/kolide/k2/pull/9488
- [x] Remove old Notary updater (https://github.com/kolide/launcher/pull/1576, https://github.com/kolide/launcher/pull/1600, https://github.com/kolide/launcher/pull/1601, https://github.com/kolide/launcher/pull/1602, https://github.com/kolide/launcher/pull/1572)
- [x] Add ability to pin a particular update channel or binary version: https://github.com/kolide/launcher/pull/1628, https://github.com/kolide/launcher/pull/1629

PR #1579, opened 2024-02-05
Add update-now functionality to TUF autoupdater

RebeccaMahany:
Relates to:
* https://github.com/kolide/launcher/issues/1577
* https://github.com/kolide/k2/pull/9488


ee/tuf/autoupdate.go
```diff
@@ -233,6 +268,53 @@ func (ta *TufAutoupdater) Interrupt(_ error) {
...
+			"err", err,
+		)
+
+		return fmt.Errorf("could not check for update: %w", err)
+	} else {
+		ta.slogger.Log(context.TODO(), slog.LevelInfo,
+			"successfully checked for update per control server request",
+			"binaries_to_update", fmt.Sprintf("%+v", binariesToUpdate),
+		)
+	}
```
James-Pickett, line 313:
```suggestion
		return fmt.Errorf("could not check for update: %w", err)
	}
	
        ta.slogger.Log(context.TODO(), slog.LevelInfo,
	        "successfully checked for update per control server request",
	        "binaries_to_update", fmt.Sprintf("%+v", binariesToUpdate),\		
	}
```

ee/tuf/autoupdate.go
```diff
@@ -233,6 +268,53 @@ func (ta *TufAutoupdater) Interrupt(_ error) {
 	ta.interrupt <- struct{}{}
 }
 
+func (ta *TufAutoupdater) Do(data io.Reader) error {
```
James-Pickett, line 271: maybe leave a comment over Do that it's satisfying the actor interface from ActionQueue or make a separate struct for "update now"

ee/tuf/autoupdate.go
```diff
@@ -314,7 +399,7 @@ func (ta *TufAutoupdater) checkForUpdate() error {
 	// Check for and download any new releases that are available
 	updatesDownloaded := make(map[autoupdatableBinary]string)
 	updateErrors := make([]error, 0)
-	for _, binary := range binaries {
+	for _, binary := range binariesToCheck {
```
directionless, line 402: I think we want to be cautious here. Because some of these trigger restarts, we should think about what order we want to do these in, and if we need to separate them temporally.

ee/tuf/autoupdate.go
```diff
@@ -44,10 +46,41 @@ const (
 
 var binaries = []autoupdatableBinary{binaryLauncher, binaryOsqueryd}
```
directionless, line 47: Do you think it's worth an array here, vs a couple of booleans? My gut sense is that we have 2 binaries now, and I can maybe imagine a 3rd. But it's not an unbounded list. So if the code is cleaner with a couple booleans...

ee/tuf/autoupdate.go
```diff
@@ -44,10 +46,41 @@ const (
...
+	if err := json.Unmarshal(data, &binary); err != nil {
+		return err
+	}
+	switch binary {
+	case "launcher":
+		*a = binaryLauncher
+	case "osqueryd":
+		*a = binaryOsqueryd
+	default:
+		return fmt.Errorf("unknown binary %s", binary)
```
directionless, line 60: Will this create an upgrade issue? like if launcher v3 wants to upgrade `launcher`, `osquery`, and `foobar`, then all the old launchers will fail to unmarshal

cmd/launcher/launcher.go
```diff
@@ -343,6 +343,7 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 
 	// Create the control service and services that depend on it
 	var runner *desktopRunner.DesktopUsersProcessesRunner
+	var actionsQueue *actionqueue.ActionQueue
```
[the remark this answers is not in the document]
RebeccaMahany: So we have access to it below outside this block

ee/tuf/autoupdate.go
RebeccaMahany: This is already addressed -- the order here does not matter. After we've finished downloading the updates, we check first to see if launcher was updated, and do a full reload if so (since that will also reload osqueryd). Otherwise, we check to see if osqueryd was updated, and restart it if so.

ee/tuf/autoupdate.go
RebeccaMahany: This whole package is designed to accommodate a list of arbitrary binaries, so it's more consistent and cleaner to keep doing it this way.

cmd/launcher/launcher.go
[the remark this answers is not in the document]
directionless: Downside is that it means needing to export `actionqueue.ActionQueue`. But 🤷

ee/tuf/autoupdate.go
RebeccaMahany: Oh fair -- I will update this

ee/tuf/autoupdate.go
directionless: I guess I'm also avoiding suggesting that we go the protobuf route

ee/tuf/autoupdate.go
RebeccaMahany: 😅 thank you, I will update

ee/tuf/autoupdate.go
RebeccaMahany: I meant that `var binaries []autoupdatableBinary` is already used extensively throughout this package, so it would be more consistent for these changes to also use an array rather than a couple of booleans. I didn't realize you were discussing the control <=> launcher data format and not the one used internally in `ee/tuf`.

I think this option is a little more flexible than having the control server send down booleans -- I think if we ended up wanting a "lock to version X.Y.Z for N days" then this data format works well.

directionless: Oh. Yeah, I'm thinking only about the API layer

directionless: > lock to version X.Y.Z for N days

How? If we send down an array of things to update, we can't list a separate versions for them. 

Honestly, I'd might keep "check for updates now" and "use version X for N days" separate. K2 can always send two actions.

RebeccaMahany: > How? If we send down an array of things to update, we can't list a separate versions for them.

Something like --

```
{
  "binaries_to_update": [
    {
      "name": "launcher",
      "lock_version": "v1.6.0", // omitempty
      "lock_expires": 123456789 // omitempty
    }
  ]
}
```

> Honestly, I'd might keep "check for updates now" and "use version X for N days" separate.

`use version X for N days` requires a check for updates (to confirm if version X is already downloaded and to download it if not), so they fit together nicely.

James-Pickett (commented):
NICE

ee/tuf/autoupdate.go
```diff
@@ -233,6 +256,62 @@ func (ta *TufAutoupdater) Interrupt(_ error) {
...
+			"received update request in unexpected format from control server, discarding",
+			"err", err,
+		)
+		// We don't return an error because we don't want the actionqueue to retry this request
+		return nil
+	}
+
+	binariesToUpdate := make([]autoupdatableBinary, 0)
+	for _, b := range updateRequest.BinariesToUpdate {
+		if val, ok := autoupdatableBinaryMap[b.Name]; ok {
```
directionless, line 274: You could invert this, and use https://pkg.go.dev/slices#DeleteFunc to remove non matches from `updateRequest`

directionless (approved):
Woo! LGTM

ee/tuf/autoupdate.go
directionless: (Which I see is what you did)

ee/tuf/autoupdate.go
RebeccaMahany: I've got to build a new slice from `updateRequest` anyway, might as well leave the request unmodified

PR #1586, opened 2024-02-07
grab logs from var/logs/kolide-k2 on macos

James-Pickett:


RebeccaMahany (approved):

ee/debug/checkups/logs.go
```diff
@@ -46,6 +47,12 @@ func (c *Logs) Run(_ context.Context, fullFH io.Writer) error {
 	defer logZip.Close()
 
 	matches, _ := filepath.Glob(filepath.Join(c.k.RootDirectory(), "debug*"))
+
+	if runtime.GOOS == "darwin" {
+		stdMatches, _ := filepath.Glob("/var/log/kolide-k2/*")
```
directionless, line 52: I don't love this in `logs`. I kinda want it in another directory. But I can live with it.

James-Pickett: Well, I'll go ahead and merge this one and I can move it to another directory when I add stderr/out for other platforms. Should we just call the dir stderr?

directionless: If we make a separate flare for it, I'd call it `InitLogs` or something. Because it collects the logs from the init system

directionless: But 🤷 maybe the windows side isn't quite that simple.

PR #1587, opened 2024-02-08
Update the `kolide_spotlight` table with a longer timeout

directionless:
Relates to https://github.com/kolide/k2/issues/9493

To address an `mdfind` issue, this PR does a couple of things:
1. It ups the mdfind exec timeout. I'm a little wary here, but we see cases where 60s time out
2. It moves the `kolide_spotlight` table to using the tablehelpers.Exec. 
3. TODO: improve the exec helper's timing info


pkg/osquery/interactive/interactive_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"testing"
 
+	"github.com/go-kit/log"
```
RebeccaMahany, line 17:
```suggestion
	"github.com/go-kit/kit/log"
```

RebeccaMahany: (will fix the build error)

James-Pickett (dismissed):
LGTM!

pkg/osquery/interactive/interactive_test.go
directionless: :shakes fist at vcode:

RebeccaMahany (approved):

PR #1590, opened 2024-02-08
add to traces

directionless:


pkg/traces/traces.go
```diff
@@ -77,8 +77,12 @@ func startSpanWithExtractedAttributes(ctx context.Context, keyVals ...interface{
 
 // SetError records the error on the span and sets the span's status to error.
 func SetError(span trace.Span, err error) {
+	// These are some otel ways to record errors. But we're not sure where they come through in GCP traces
 	span.RecordError(err)
 	span.SetStatus(codes.Error, err.Error())
+
+	// Dump the error into a span attribute, because :shrug:
+	span.SetAttributes(attribute.String("error.message", err.Error()))
```
RebeccaMahany, line 85:
```suggestion
	span.SetAttributes(semconv.ExceptionMessage(err.Error()))
```

(uses the well-known key, see if that helps)

directionless: I mean, what is the well known key, and can we find it? 😆

RebeccaMahany (dismissed):
🤞

zackattack01 (approved):

James-Pickett (approved):

PR #1592, opened 2024-02-09
add journalctl launcher logs to flare

James-Pickett:


ee/debug/checkups/init_logs_linux.go
```diff
@@ -3,8 +3,24 @@ package checkups
...
 	"context"
+	"fmt"
+
+	"github.com/kolide/launcher/ee/allowedcmd"
 )
 
-func writeInitLogs(_ context.Context, _ *zip.Writer) error {
-	return nil
+func writeInitLogs(ctx context.Context, logZip *zip.Writer) error {
+	cmd, err := allowedcmd.Journalctl(ctx, "-u", "launcher.kolide-k2.service", "-o", "json")
```
James-Pickett, line 12: putting these in JSON formate, both for this and for windows, produce very verbose logs due to the extra properties it adds associated with each message ... do we want this?

RebeccaMahany: The extra info doesn't seem super useful (except for `_PID` maybe, but that should be available in a different checkup anyway) -- I think I'd lean toward not including it just to avoid the extra noise?

directionless: My inkling is not. I think of this as being pretty simple to a quick "tail the log file". But maybe I'll regret saying that

James-Pickett: droped the json conversation and just let it print the default on linux. We could add some output options to make it prettier or more customized, but those options might not be present on all distros

ee/debug/checkups/init_logs_windows.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func writeInitLogs(ctx context.Context, logZip *zip.Writer) error {
-	cmdStr := `Get-WinEvent -FilterHashtable @{LogName='Application'; ProviderName='launcher'} | ConvertTo-Json`
+	cmdStr := `Get-WinEvent -FilterHashtable @{LogName='Application'; ProviderName='launcher'} | ForEach-Object { $_.Message }`
```
James-Pickett, line 12: this works pretty well except for powershell adding a new line between each log, we could remove with some post processing of the output but ... 🤷

RebeccaMahany (approved):

ee/debug/checkups/init_logs_windows.go
directionless: If you want to monkey here, you could also add `Get-EventLog  -Newest 10 -LogName System -Source "Service Control Manager" -Message "*launcherkolidek2svc*" | select -ExpandProperty message`

Issue #1473, opened 2023-11-22
Support self-service signup


directionless:
This is a tracking issue for a TBD project

RebeccaMahany:
- [ ] Launcher must be able to run when there's no enrollment secret present, instead of shutting down
- [ ] It should be possible to retrieve the enrollment secret from a variety of sources (e.g. the registry on Windows)
- [ ] Enrollment status should be exposed through localserver

RebeccaMahany:
Mostly done and also tracked separately in Jira

PR #1596, opened 2024-02-09
Make the osquery extension and runner separate rungroups

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1473

Currently, we have an `actorQuerier` rungroup wrapper around both the extension and the runner, which act as a semi-conjoined single rungroup. They don't need to be coupled like this: the extension doesn't need the runner (since we officially disabled the initial runner), and the runner doesn't need the extension (it just needs a couple opts passed in to register the extension correctly when creating the osquery instance). My hope is that decoupling them will make it easier for us to handle the "unenrolled" state and give us better flexibility for how we want to do that. I also think it'll make rewriting our rungroups easier in the future, too.

This PR makes the following changes:

* Makes the runner and the extension both adhere to the rungroup actor interface
* Decouples the runner and the extension
* Removes support for `osquery` transport method (we don't really use it and haven't for a while)
* Removes unused initial runner
* Gives the runner access to the knapsack
* Gives the runner access to its own slogger instead of having to reach into the instance to use its logger

<details>

- [x] launcher can start up osquery successfully
- [x] launcher will restart osquery if the process is manually killed
- [x] can query `kolide_` tables provided by the extension
- [x] enrollment works
- [x] logs show up with component `osquery_runner` and `osquery_extension` as expected

</details>


James-Pickett (approved):
nice, this is much cleaner!

pkg/osquery/extension.go
```diff
@@ -160,65 +146,60 @@ func NewExtension(ctx context.Context, client service.KolideService, k types.Kna
...
 
-	identifier, err := IdentifierFromDB(configStore)
-	if err != nil {
-		return nil, fmt.Errorf("get host identifier from db when creating new extension: %w", err)
-	}
-
 	nodekey, err := NodeKey(configStore)
 	if err != nil {
-		k.Slogger().Log(context.TODO(), slog.LevelDebug,
+		slogger.Log(context.TODO(), slog.LevelDebug,
```
zackattack01, line 151: not blocking but should we be using the parent context for these? it looks like there is one available from func args

zackattack01 (dismissed):
looks great! 🔥

pkg/osquery/extension.go
RebeccaMahany: Ah whoops, thank you! I will update

zackattack01 (approved):

James-Pickett (approved):

RebeccaMahany:
Will merge this after we declare v1.5.3 stable

PR #1603, opened 2024-02-14
Remove autoloaded extensions

RebeccaMahany:
We can't expect to find autoloaded extensions based on the current executable anymore, because the current executable may live in the new updates directory outside of the install (bin) directory. We don't use this flag and we don't imagine that it has widespread use elsewhere either, so we're opting to remove it.

### DIFF
--- a/pkg/autoupdate/findnew.go
+++ b/pkg/autoupdate/findnew.go
@@ -251,7 +251,7 @@ func FindNewest(ctx context.Context, fullBinaryPath string, opts ...newestOption
 func getUpdateDir(fullBinaryPath string) string {
 	if strings.Contains(fullBinaryPath, ".app") {
 		binary := filepath.Base(fullBinaryPath)
-		return filepath.Join(FindBaseDir(fullBinaryPath), binary+updateDirSuffix)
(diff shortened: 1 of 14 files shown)


zackattack01 (approved):

James-Pickett (approved):
Nice

pkg/autoupdate/findnew.go
```diff
@@ -317,10 +317,10 @@ func getPossibleUpdates(ctx context.Context, updateDir, binaryName string) ([]st
 	return possibleUpdates, nil
 }
 
-// FindBaseDir takes a binary path, that may or may not include the
+// findBaseDir takes a binary path, that may or may not include the
```
directionless, line 320: Do we still need this, and anything else in the FindNewest?

directionless (approved):

pkg/autoupdate/findnew.go
RebeccaMahany: Yeah, we fall back to the legacy autoupdate library currently if there's nothing in the new library. We can definitely get rid of that at some point, but I wasn't in a rush to rip that part out as quickly.

PR #1607, opened 2024-02-15
Don't trigger TUF when running from build directory

directionless:
The old updater had the side effect of _not_ triggering the update check when running from a build dir. I found that quite convenient, as I'm often running launcher in a development cycle. This recreates that behavior with a somewhat klunky path check.

If y'all think it will have some unknown impact around CI or testing, speak up...

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -48,12 +48,19 @@ func main() {
 		Level: slog.LevelInfo,
 	}))
 
+	// If this is a development build directory, we want to skip the TUF lookups and just run
+	// the requested build. Don't care about errors here. This is developer experience shim
+	inBuildDir := false
+	if execPath, err := os.Executable(); err == nil {
+		inBuildDir = strings.Contains(execPath, filepath.Join("launcher", "build")) && !env.Bool("LAUNCHER_FORCE_UPDATE_IN_BUILD", false)
+	}
+
 	// If there's a newer version of launcher on disk, use it.
 	// Allow a caller to set `LAUNCHER_SKIP_UPDATES` as a way to
 	// skip exec'ing an update. This helps prevent launcher from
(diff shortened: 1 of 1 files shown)


RebeccaMahany:
I don't think there will be a negative impact to doing this, so if you want it, sure.

I am personally not in favor of adding this -- I don't think we need to add additional complexity to an already complicated system when there are already two other workarounds that meet this need during development (`LAUNCHER_SKIP_UPDATES` and `localdev_path`).

directionless:
I'm willing to hold while we figure out what we want, and how to cleanly handle it.

Mostly, I want to be able to do things like `make && ./build/launcher` and get the build I just made. And I see the other things are serving different purposes. `LAUNCHER_SKIP_UPDATES` was designed as an internal safety to prevent forkbombs, `localdev_path` is for doing local TUF development.

RebeccaMahany:
`localdev_path` isn't just for local TUF development but for all local development purposes -- e.g. I set `localdev_path /path/to/my/launcher/build/darwin.universal/Kolide.app/Contents/MacOS/launcher` in my `launcher.flags` file and unload/build/reload for all local development.

directionless:
> `localdev_path` isn't just for local TUF development but for all local development purposes

If we believe that, then we should make it part of the development environment. I'm not totally sure how we'd want to do that.

RebeccaMahany:
@directionless I think the discussion is getting long enough that we should move it to slack, or if you want to table it for our next team meeting -- I would like to know what "make it a part of the development environment" entails for you (plus I would also like to know if `localdev_path` has been adequate/not painful for James and Zack, or if they also want an alternative like this)

directionless:
We discussed this today, and I think concluded we should merge it.

RebeccaMahany (approved):

PR #1608, opened 2024-02-15
Allow launcher to run without enrollment secret

RebeccaMahany:
### Changes

* launcher runs, instead of immediately shutting down, if it doesn't have an enroll secret
* lay groundwork for platform-specific ways of reading the enrollment secret in the future and for exposing enrollment status to localserver
* attempt enrollment immediately, before rungroups even start

### Details

Before, the osquery extension rungroup could not be created if the enroll secret wasn't present, preventing launcher from starting up at all. Now, if the secret isn't present, launcher will still start, and will just return an empty config to osquery. Whenever the secret becomes available, it will perform enrollment.

I pulled out reading the secret into the knapsack, with the idea being that in the future we could a) expose enrollment status from localserver, and b) have per-platform implementations of this function that would allow us to e.g. read the key from the Windows registry.

This PR also adds an attempt at immediate enrollment in the background, before the rungroups even start. This should hopefully make first-time launcher startup faster.

### Testing notes

<details>

<summary>In here</summary>

Force reenrollment and remove access to enrollment secret:

```
sudo launchctl unload /Library/LaunchDaemons/com.kolide-k2.launcher.plist
sudo mv /etc/kolide-k2/secret /etc/kolide-k2/secret.bak
sudo rm -rf /var/kolide-k2/k2device-preprod.kolide.com
sudo mkdir /var/kolide-k2/k2device-preprod.kolide.com
sudo chmod -R 0755 /var/kolide-k2/*
sudo launchctl load /Library/LaunchDaemons/com.kolide-k2.launcher.plist
```

Confirm that launcher starts up and that the osquery process starts up.

Eventually, make the secret available:

```
sudo mv /etc/kolide-k2/secret.bak /etc/kolide-k2/secret
```

Observe launcher immediately complete enrollment.

</details>

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -402,7 +402,7 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 	)
 	span.AddEvent("starting_enrollment")
 
-	enrollSecret, err := e.readEnrollSecret(ctx)
+	enrollSecret, err := e.knapsack.ReadEnrollSecret()
 	if err != nil {
 		return "", true, fmt.Errorf("could not read enroll secret: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/osquery/extension.go
```diff
@@ -507,6 +529,10 @@ func (e *Extension) GenerateConfigs(ctx context.Context) (map[string]string, err
 		confBytes, _ = e.knapsack.ConfigStore().Get([]byte(configKey))
 
 		if len(confBytes) == 0 {
+			if !e.enrolled() {
+				// Not enrolled yet -- return an empty config
+				return map[string]string{"config": "{}"}, nil
+			}
```
RebeccaMahany, line 535: I'm not positive we even need this? If I omit it, we get:

```
{
"time":"2024-02-15T18:55:35.676293Z",
"level":"INFO",
"source":{
  "function":"github.com/kolide/launcher/ee/log/osquerylogs.(*OsqueryLogAdapter).Write",
  "file":"/Users/rebeccamahany-horton/Repos/launcher/ee/log/osquerylogs/log.go","line":83},
  "msg":"W0215 13:55:35.676178 -347473728 init.cpp:760] Error reading config: error getting config: loading config failed, no cached config: enrollment invalid, reenrollment errored: could not read enroll secret: could not read enroll secret path /etc/kolide-k2/secret: open /etc/kolide-k2/secret: no such file or directory",
  "launcher_run_id":"01HPQ1ZGS0R3Q0WAZ7FDB4N7FV",
  "component":"osquery",
  "osqlevel":"stderr",
  "caller":"init.cpp:760"
}
```

And behavior doesn't otherwise change -- as soon as I make the secret file available, launcher completes enrollment and queries start to run.

directionless (dismissed):
This is a nice, very narrow, change on the way.

I think there's stuff we should evolve on -- I think we should rename osquery.Extension to something like agent.Core. I think we should trigger this from runLauncher, not from extension. But I think this could merge and iterate.

James-Pickett (approved):
nice

PR #1612, opened 2024-02-16
[slogger] Replace logger with slogger in dataflatten/exec

RebeccaMahany:
Last step for removing logger from Kolide tables -- replaces logger with slogger in dataflatten/exec.

### DIFF
--- a/cmd/launcher/interactive.go
+++ b/cmd/launcher/interactive.go
@@ -80,7 +80,7 @@ func runInteractive(args []string) error {
 		flOsqueryFlags = append(flOsqueryFlags, fmt.Sprintf("tls_server_certs=%s", certs))
 	}
 
-	osqueryProc, extensionsServer, err := interactive.StartProcess(logger, rootDir, osquerydPath, flOsqueryFlags)
+	osqueryProc, extensionsServer, err := interactive.StartProcess(rootDir, osquerydPath, flOsqueryFlags)
 	if err != nil {
 		return fmt.Errorf("error starting osqueryd: %s", err)
 	}
--- a/cmd/launcher/interactive.go
+++ b/cmd/launcher/interactive.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"log/slog"
 	"os"
 	"strings"
 
@@ -14,6 +15,7 @@ import (
 	"github.com/kolide/launcher/ee/tuf"
 	"github.com/kolide/launcher/pkg/autoupdate"
 	"github.com/kolide/launcher/pkg/launcher"
+	"github.com/kolide/launcher/pkg/log/multislogger"
 	"github.com/kolide/launcher/pkg/osquery/interactive"
 )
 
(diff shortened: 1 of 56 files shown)


cmd/launcher/interactive.go
```diff
@@ -80,7 +80,7 @@ func runInteractive(args []string) error {
 		flOsqueryFlags = append(flOsqueryFlags, fmt.Sprintf("tls_server_certs=%s", certs))
 	}
 
-	osqueryProc, extensionsServer, err := interactive.StartProcess(logger, rootDir, osquerydPath, flOsqueryFlags)
+	osqueryProc, extensionsServer, err := interactive.StartProcess(rootDir, osquerydPath, flOsqueryFlags)
```
directionless, line 83: Should we be passing a slogger here? It feels weird to not have logs?

directionless (approved):

cmd/launcher/interactive.go
RebeccaMahany: Yeah I think I need to adjust this, just haven't been sure how to configure the slogger for command-line yet -- I'll get it sorted!

ee/dataflatten/examples/flatten/main.go
```diff
@@ -40,13 +40,14 @@ func main() {
 		checkError(fmt.Errorf("parsing flags: %w", err))
 	}
 
-	logger := logutil.NewCLILogger(*flDebug)
-
 	opts := []dataflatten.FlattenOpts{
-		dataflatten.WithLogger(logger),
+		dataflatten.WithSlogger(multislogger.New().Logger),
```
directionless, line 44: This is example code. It feels slightly weirf to pass `multislogger`. Can we pass a simple slogger instead?

ee/dataflatten/flatten.go
```diff
@@ -141,7 +142,7 @@ func Flatten(data interface{}, opts ...FlattenOpts) ([]Row, error) {
 	}
 
 	if !fl.debugLogging {
```
directionless, line 144: This boolean was kinda weird.

IIRC I added it (and the associated `WithDebugLogging`) because I didn't want the flatten code to log most of the time. I want logs when I'm debugging flattening, and otherwise they're kinda noisy and in the way. Not sure if there's a better pattern

directionless (dismissed):

ee/dataflatten/examples/flatten/main.go
RebeccaMahany: This is what we tend to copy-paste everywhere we need an unused slogger -- e.g. it's all over our unit tests. I'm not sure that this is wrong.

directionless: :shrug:

ee/dataflatten/flatten.go
directionless: If I were doing it from scratch, I
Might write the logs at a level below debug. Maybe. 

I think you could remove the conditional, and leave it as effectively unused. We can either rip out logs or relevel them in a follow up.

RebeccaMahany: Yup, that handles it!

ee/dataflatten/flatten.go
```diff
@@ -140,8 +147,8 @@ func Flatten(data interface{}, opts ...FlattenOpts) ([]Row, error) {
 		opt(fl)
 	}
 
-	if !fl.debugLogging {
-		fl.logger = level.NewFilter(fl.logger, level.AllowInfo())
+	if fl.debugLogging {
+		fl.logLevel = slog.LevelDebug
 	}
```
directionless, line 152: I suspect we could drop this, and simply log _everything_ at at `levelFlattenDebug` and let the caller apply the appropriate filter. But 🤷 maybe better to merge

directionless (approved):

ee/dataflatten/flatten.go
RebeccaMahany: I liked this because it works the same way it used to 🤷‍♀️

Issue #1537, opened 2024-01-09
Can't open browser from menu bar app or notification on Ubuntu 23 with Wayland, PopOS 22.04, potentially others


RebeccaMahany:
For Ubuntu 23 with Wayland: we rely on [Xwayland](https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner_linux.go#L262) to be running in order to grab the `DISPLAY`, which we need to be able to open URLs from the menu bar app and from notifications. I updated some software on Ubuntu 23.04 and now I don't have an `xwayland` process, just `gdm-wayland-session`, which does not have `DISPLAY` as an arg.

We've also had reports of the browser not opening on PopOS 22.04.

Overall, the way(s) that we're finding and setting `DISPLAY` is feeling increasingly brittle. We've had issues come up like this on the last several releases. This issue is for, at the very least, fixing the issue on Ubuntu 23.04 with wayland and PopOS, but ideally to find a less brittle way to set DISPLAY in general.

______________

## Workarounds

Consolidating some workarounds from the comments here:

### Ubuntu-specific workarounds

[Switching from wayland to x11 temporarily](https://askubuntu.com/a/961345) should let you open links from the menu bar app, and from notifications. After switching to x11, you will have to restart the launcher service via `sudo systemctl restart launcher.kolide-k2.service` first to get the menu bar app to show up.

### Registering your device when you can't access the registration URL from the menu bar app

Any route through auth will also force registration on an unregistered device, so attempting to log in to any Kolide-protected app should get you to the correct registration URL.

rolandoscott:
I can confirm this is also happening for me on Ubuntu 23.10.. the tray icon appears.. the menu option appear.. but nothing happens when I click on them.. 

In the meantime, is there any way to trigger the "Register Device" option somehow else.. either via terminal or is it a specific URL that I can just go to?

RebeccaMahany:
I know we discussed on https://github.com/kolide/launcher/issues/1543, just adding info on the workaround here for ease of access in case anyone else runs into it --

[Switching from wayland to x11 temporarily](https://askubuntu.com/a/961345) should let you open links from the menu bar app, and from notifications. After switching to x11, you will have to restart the launcher service via `sudo systemctl restart launcher.kolide-k2.service` first to get the menu bar app to show up.

tobybellwood:
[...]

PR #1616, opened 2024-02-22
Add env vars to Kolide processes in checkup

RebeccaMahany:
We care about these specifically for the desktop process on Linux (see: https://github.com/kolide/launcher/issues/1537), but it doesn't hurt to grab them for Windows, too.

### DIFF
--- a/ee/debug/checkups/processes.go
+++ b/ee/debug/checkups/processes.go
@@ -56,6 +56,13 @@ func (c *Processes) Run(ctx context.Context, fullWriter io.Writer) error {
 		exe, _ := p.Exe()
 		if strings.Contains(strings.ToLower(exe), "kolide") {
 			c.kolideCount += 1
+
+			// Grab ENV vars if available -- we ignore the error because we expect
+			// this function to return `ErrNotImplementedError` on darwin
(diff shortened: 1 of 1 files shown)


zackattack01 (approved):

ee/debug/checkups/processes.go
```diff
@@ -56,6 +56,13 @@ func (c *Processes) Run(ctx context.Context, fullWriter io.Writer) error {
 		exe, _ := p.Exe()
 		if strings.Contains(strings.ToLower(exe), "kolide") {
 			c.kolideCount += 1
+
+			// Grab ENV vars if available -- we ignore the error because we expect
+			// this function to return `ErrNotImplementedError` on darwin
+			if envVars, err := p.EnvironWithContext(ctx); err == nil {
+				pMap["env"] = strings.Join(envVars, " ")
```
directionless, line 63: I don't think you need the `strings.Join` here, the map should handle `any`. And you can probably be even simpler and just do `"env": naIfError(p.EnvironWithContext(ctx)),`

RebeccaMahany: Thanks -- addressed here: https://github.com/kolide/launcher/pull/1622

PR #1620, opened 2024-02-23
Do not spin up desktop process for nobody user

RebeccaMahany:
launcher appears to be attempting to spin up a desktop process for `nobody` (observed on Debian 12 so far) -- it shouldn't do that. This PR adds an exclusion for the `nobody` user.

### DIFF
--- a/ee/consoleuser/consoleuser_linux.go
+++ b/ee/consoleuser/consoleuser_linux.go
@@ -13,9 +13,10 @@ import (
 )
 
 type listSessionsResult []struct {
-	Session string `json:"session"`
-	UID     int    `json:"uid"`
-	Seat    string `json:"seat"`
+	Session  string `json:"session"`
+	UID      int    `json:"uid"`
+	Username string `json:"user"`
+	Seat     string `json:"seat"`
 }
 
 func CurrentUids(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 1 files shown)


zackattack01 (dismissed):

ee/consoleuser/consoleuser_linux.go
```diff
@@ -36,8 +36,9 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 
 	var uids []string
 	for _, s := range sessions {
-		// generally human users start at 1000 on linux
-		if s.UID < 1000 {
+		// generally human users start at 1000 on linux. 65534 is reserved for https://wiki.ubuntu.com/nobody,
+		// which we don't want to count as a current user.
+		if s.UID < 1000 || s.UID == 65534 {
```
directionless, line 41: If we have username here, I'd add a case for that. But only if we already have it. OTOH this is probably good enough for any modern system.

directionless (approved):

ee/consoleuser/consoleuser_linux.go
RebeccaMahany: I didn't add a check for username because it's not in the struct -- but it is available in the loginctl output, so I can add it to the struct and add the check here.

PR #1628, opened 2024-02-26
Control server can set update channel for autoupdater

RebeccaMahany:
Allows the control server to set the update channel via agent flags, and have the autoupdater apply that change immediately and on subsequent launcher restarts as well.

### DIFF
--- a/ee/tuf/autoupdate.go
+++ b/ee/tuf/autoupdate.go
@@ -16,7 +16,6 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
-	"slices"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


zackattack01 (dismissed):

ee/tuf/autoupdate.go
```diff
@@ -312,6 +319,36 @@ func (ta *TufAutoupdater) Do(data io.Reader) error {
 	return nil
 }
 
+// FlagsChanged satisfies the FlagsChangeObserver interface, allowing the autoupdater
+// to respond to changes to autoupdate-related settings.
+func (ta *TufAutoupdater) FlagsChanged(flagKeys ...keys.FlagKey) {
+	if !slices.Contains(flagKeys, keys.UpdateChannel) {
+		return
+	}
```
James-Pickett, line 327: would the behavior be any different without checking for update channel key?

RebeccaMahany: You're right, don't need this at all! Removed 🙂

James-Pickett (approved):

zackattack01 (approved):

directionless:
This turned out very simple. Nice.

PR #1629, opened 2024-02-27
Allow control server to pin version for osqueryd and launcher

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1577

Some users may want to pin a particular version of osqueryd or launcher. This PR enables that functionality via the TUF autoupdater.

If a pinned version is set, the autoupdater will select that version first. If it cannot find that version, then it will fall back to using the release version for the configured update channel.

While working on this PR, I noticed I had a couple related functions about parsing and comparing versions, so I've moved them into their own file and separated out their tests as well.

Also -- updated to handle control server actions during the initial delay in a clearer and more opinionated way:

* Flag changes should be noted and stored during the initial delay, but --
* Autoupdates should never occur during the initial delay


ee/tuf/version.go
```diff
@@ -0,0 +1,55 @@
...
+		return false, fmt.Errorf("could not parse launcher version %s: %w", rawVersion, err)
+	}
+
+	return versionParsed.GreaterThan(tufVersionMinimum) || versionParsed.Equal(tufVersionMinimum), nil
+}
+
+// versionFromTarget extracts the semantic version for an update from its filename.
+func versionFromTarget(binary autoupdatableBinary, targetFilename string) string {
+	// The target is in the form `launcher-0.13.6.tar.gz` -- trim the prefix and the file extension to return the version
+	prefixToTrim := fmt.Sprintf("%s-", binary)
```
directionless, line 52: Could also `SplitN`

directionless (commented):
I am still ambivalent about whether we need 3 flags, or 2 flags. But I think I can live with this.

James-Pickett:
LGTM!

ee/tuf/autoupdate.go
```diff
@@ -520,15 +568,57 @@ func (ta *TufAutoupdater) downloadUpdate(binary autoupdatableBinary, targets dat
...
+}
+
+// findTarget selects the appropriate target from `targets` for the given binary, using the pinned version (if set)
+// and otherwise selecting the correct release for the given channel.
+func findTarget(ctx context.Context, binary autoupdatableBinary, targets data.TargetFiles, pinnedVersion string, channel string, slogger *slog.Logger) (string, data.TargetFileMeta, error) {
+	ctx, span := traces.StartSpan(ctx)
+	defer span.End()
+
+	if pinnedVersion != "" {
+		target, targetMetadata, err := findTargetByVersion(context.Background(), binary, targets, pinnedVersion)
```
zackattack01, line 589: just curious for myself why the use of `context.Background()` over the parent `ctx` here?

zackattack01 (dismissed):
nice!

ee/tuf/autoupdate.go
RebeccaMahany: Because I copy-pasted from somewhere and neglected to change it 😅 

Updated, thanks!

zackattack01 (approved):

Issue #1312, opened 2023-08-30
Launcher is somewhat fragile around log buffering


directionless:
Osquery produces logs, and sends them to launcher, which immediately writes them to boltdb. Asynchronously, a process reads a chunk of them of them from boltdb and sends them to us. This process is fairly straightforward, but it contains a a theoretical bug. (Which has impacted us at least twice)

To wit, if the amount of time it takes to send a batch of logs, exceeds our servers max connection time (30s), the send will fail, and retry. When this happens, launcher will effectively use all the network bandwidth trying to resend that log batch. This is likely in places with exceptionally bad internet.

The code in question is at:
https://github.com/kolide/launcher/blob/a322303d9f9ddd13ce1a505acc2ae3161b38216a/pkg/osquery/extension.go#L620

I think, in an ideal world, we should have some dynamic adjustment of the `MaxBytesPerBatch` setting. I don't know if we should make it some exponential backoff sort of thing, or if we should track how long it takes to send a batch, and adjust based on that. Happy to hear suggestions. 

Along the way, maybe look at #705 and #1300

directionless:
Let's say #1630 fixed this. If we need to r-open we can.

PR #1630, opened 2024-02-29
add dynamic buffer size for log publication

zackattack01:
These changes are to address the log buffering fragility issue noted [here](https://github.com/kolide/launcher/issues/1312).

The adjustment logic took a few different forms based on testing - i also have a version of this that keeps a queue of latest results if anyone is interested in going down that route let me know (I ended up trimming this down to a more simple mechanism after seeing the complexity there).

I am more than happy to take suggestions on the adjustment logic - it's easy to make the backoff more dramatic if desired but I found it to be safest to keep the changes to small increments, if the network is really in a bad state the batch limit will be adjusted eventually.

### complications/lessons learned

- increasing the batch limit:
    - We publish logs in batches every minute by log type. This means that while some log types may become backlogged others can always publish successfully - causing the batch limit to re-adjust upwards too quickly. To get around this we only increase the limit if the previous batch limit was **both** hit and published successfully.
- decreasing the batch limit:
    - depending on how bad the network connection is and how it is terminated, any number of errors can pop out (not just network/transport errors, sometimes just failures to parse the response) . To get around this we keep track of when the batch was started, and only reduce the limit if an error occurred after a 20 second duration

### DIFF
--- a/pkg/osquery/log_publication_state.go
+++ b/pkg/osquery/log_publication_state.go
@@ -16,7 +16,9 @@ const (
 // logPublicationState holds stateful logic to influence the log batch publication size
 // depending on prior successes or failures. The primary intent here is to prevent repeatedly
 // consuming the entire network bandwidth available to devices that are unable to ship
-// the standard maxBytesPerBatch inside of the connection timeout (currently 30 seconds, enforced cloud side)
+// the standard maxBytesPerBatch inside of the connection timeout (currently 30 seconds, enforced cloud side).
+// Note that we always expect these batches to be sent sequentially, BeginBatch -> EndBatch
+// this would need rework (likely state locking and batch ID tracking) to support concurrent batch publications
 type logPublicationState struct {
 	// maxBytesPerBatch is passed in from the extension opts and respected
 	// as a fixed upper limit for batch size, regardless of publication success/failure rates
@@ -24,7 +26,7 @@ type logPublicationState struct {
 	// currentMaxBytesPerBatch represents the (stateful) upper limit being enforced
 	currentMaxBytesPerBatch int
 	// currentBatchBufferFilled is used to indicate when a batch's success can be used
-	// to increase the threshold (we should avoid increasing threshold for a successful partial batch of logs)
+	// to increase the threshold (we only want to increase the threshold after sending full, not partial, batches successfully)
 	currentBatchBufferFilled bool
 	currentBatchStartTime    time.Time
 }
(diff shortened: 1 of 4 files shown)


RebeccaMahany (approved):
Nice -- this was easy to read/follow.

pkg/osquery/log_publication_state.go
```diff
@@ -0,0 +1,100 @@
...
+// consuming the entire network bandwidth available to devices that are unable to ship
+// the standard maxBytesPerBatch inside of the connection timeout (currently 30 seconds, enforced cloud side)
+type logPublicationState struct {
+	// maxBytesPerBatch is passed in from the extension opts and respected
+	// as a fixed upper limit for batch size, regardless of publication success/failure rates
+	maxBytesPerBatch int
+	// currentMaxBytesPerBatch represents the (stateful) upper limit being enforced
+	currentMaxBytesPerBatch int
+	// currentBatchBufferFilled is used to indicate when a batch's success can be used
+	// to increase the threshold (we should avoid increasing threshold for a successful partial batch of logs)
```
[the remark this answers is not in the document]
zackattack01: yes! your interpretation is correct, I will update this that sounds clearer

pkg/osquery/log_publication_state.go
```diff
@@ -0,0 +1,105 @@
...
+	maxBytesPerBatch int
+	// currentMaxBytesPerBatch represents the (stateful) upper limit being enforced
+	currentMaxBytesPerBatch int
+	// currentBatchBufferFilled is used to indicate when a batch's success can be used
+	// to increase the threshold (we only want to increase the threshold after sending full, not partial, batches successfully)
+	currentBatchBufferFilled bool
+	currentBatchStartTime    time.Time
+}
+
+func NewLogPublicationState(maxBytesPerBatch int) *logPublicationState {
```
directionless, line 34: Does this need to be exported?

pkg/osquery/log_publication_state.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	newTargetThreshold := lps.currentMaxBytesPerBatch - batchIncrementAmount
+	lps.currentMaxBytesPerBatch = maxInt(newTargetThreshold, minBytesPerBatch)
+}
+
+func (lps *logPublicationState) increaseBatchThreshold() {
+	if lps.currentMaxBytesPerBatch >= lps.maxBytesPerBatch {
+		return
+	}
```
directionless, line 101: Nice. I _suspect_ we'll need to figure out how to grow beyond this -- some customers have it coded really small. But maybe this is fine to start, and we can make a real max value later

directionless (approved):
Looks pretty solid. Not sure if it's worth stashing the values for next startup.

James-Pickett (approved):
nice

PR #1634, opened 2024-03-01
Add `kolide_brew_upgradeable` table

Micah-Kolide:
Add a Kolide table on both Linux and macOS for `homebrew` outdated packages. My initial concerns of being unable to run as `sudo` were correct as `homebrew` panics if an attempt is made to run it as `sudo`. This is by design in `homebrew`.

```
$ sudo brew --help
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
```

```
osquery> SELECT * FROM kolide_brew_upgradeable WHERE uid = 501;
+---------------------------------+-------------------------------+-----------------+------------+-------+-----+
| fullkey                         | parent                        | key             | value      | query | uid |
+---------------------------------+-------------------------------+-----------------+------------+-------+-----+
| formulae/0/name                 | formulae/0                    | name            | cfitsio    | *     | 501 |
| formulae/0/installed_versions/0 | formulae/0/installed_versions | 0               | 4.3.1      | *     | 501 |
| formulae/0/current_version      | formulae/0                    | current_version | 4.4.0      | *     | 501 |
| formulae/0/pinned               | formulae/0                    | pinned          | false      | *     | 501 |
| formulae/1/name                 | formulae/1                    | name            | libnghttp2 | *     | 501 |
| formulae/1/installed_versions/0 | formulae/1/installed_versions | 0               | 1.59.0     | *     | 501 |
| formulae/1/current_version      | formulae/1                    | current_version | 1.60.0     | *     | 501 |
| formulae/1/pinned               | formulae/1                    | pinned          | false      | *     | 501 |
| formulae/2/name                 | formulae/2                    | name            | libpng     | *     | 501 |
| formulae/2/installed_versions/0 | formulae/2/installed_versions | 0               | 1.6.42     | *     | 501 |
| formulae/2/current_version      | formulae/2                    | current_version | 1.6.43     | *     | 501 |
| formulae/2/pinned               | formulae/2                    | pinned          | false      | *     | 501 |
| formulae/3/name                 | formulae/3                    | name            | openjpeg   | *     | 501 |
| formulae/3/installed_versions/0 | formulae/3/installed_versions | 0               | 2.5.0_1    | *     | 501 |
| formulae/3/current_version      | formulae/3                    | current_version | 2.5.1      | *     | 501 |
| formulae/3/pinned               | formulae/3                    | pinned          | false      | *     | 501 |
| formulae/4/name                 | formulae/4                    | name            | orc        | *     | 501 |
| formulae/4/installed_versions/0 | formulae/4/installed_versions | 0               | 0.4.37     | *     | 501 |
| formulae/4/current_version      | formulae/4                    | current_version | 0.4.38     | *     | 501 |
| formulae/4/pinned               | formulae/4                    | pinned          | false      | *     | 501 |
| formulae/5/name                 | formulae/5                    | name            | xz         | *     | 501 |
| formulae/5/installed_versions/0 | formulae/5/installed_versions | 0               | 5.4.5      | *     | 501 |
| formulae/5/current_version      | formulae/5                    | current_version | 5.6.0      | *     | 501 |
| formulae/5/pinned               | formulae/5                    | pinned          | false      | *     | 501 |
+---------------------------------+-------------------------------+-----------------+------------+-------+-----+
```


ee/tables/homebrew/upgradeable.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	if err := cmd.Wait(); err != nil {
+		return nil, fmt.Errorf("deallocation of brew outdated command as user %s: %w", uid, err)
+	}
+
+	return data, nil
+}
+
+func runAsUser(ctx context.Context, uid string, cmd *exec.Cmd) error {
```
RebeccaMahany, line 109: We use this in enough places now that we should probably just make this available/reusable somewhere. I don't think it's necessary to address in this PR, but wanted to flag in case you (or James/Zack/seph) have a good suggestion for where we should move this in the future.

RebeccaMahany (approved):
LGTM!

ee/tables/homebrew/upgradeable.go
Micah-Kolide: Agreed. I feel fried this week, so I too didn't want to address it in this PR, but I might take a look at that next week if I can.

ee/allowedcmd/cmd_darwin.go
```diff
@@ -20,6 +20,10 @@ func Bputil(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 	return validatedCommand(ctx, "/usr/bin/bputil", arg...)
 }
 
+func Brew(ctx context.Context, arg ...string) (*exec.Cmd, error) {
+	return validatedCommand(ctx, "/opt/homebrew/bin/brew", arg...)
```
directionless, line 24: This is only correct for m1 machines, not for the intel x86 ones

ee/tables/homebrew/upgradeable.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/kolide/launcher/ee/tables/dataflattentable"
+	"github.com/kolide/launcher/ee/tables/tablehelpers"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+const allowedCharacters = "0123456789"
+
+type Table struct {
+	slogger *slog.Logger
+	execCC  allowedcmd.AllowedCommand
```
directionless, line 28: You don't need the `execCC` thing. It should only get used when we want to override commands in the test suite. Otherwise, just skip it

ee/tables/homebrew/upgradeable.go
directionless: Agree it feels like we should put it into a helper. We have some exec helpers. Which reminds me, this isn't using them. So whether we expand the exec helper, or make another 🤷

PR #1636, opened 2024-03-01
Add `kolide_snap_upgradeable` table and new `data_table` exec parser

Micah-Kolide:
I've created a new string delimited method for processing csv-esque/data table output.

Initially I was aiming to utilize this new method for both `flatpak` and `snap`, but due to many issues that arose with `flatpak` _(they are commented in that PR)_ I've only stuck with `snap` for now.

```
$ snap refresh --list
Name    Version   Rev   Size  Publisher   Notes
core22  20240111  1122  77MB  canonical✓  base

osquery> SELECT * FROM kolide_snap_upgradeable;
+-------------+--------+-----------+-------------+-------+
| fullkey     | parent | key       | value       | query |
+-------------+--------+-----------+-------------+-------+
| 0/Version   | 0      | Version   | 20240111    | *     |
| 0/Rev       | 0      | Rev       | 1122        | *     |
| 0/Size      | 0      | Size      | 77MB        | *     |
| 0/Publisher | 0      | Publisher | canonical** | *     |
| 0/Notes     | 0      | Notes     | base        | *     |
| 0/Name      | 0      | Name      | core22      | *     |
+-------------+--------+-----------+-------------+-------+
```


ee/dataflatten/string_delimited.go
```diff
@@ -15,6 +15,75 @@ const (
...
+		// Skip first N lines due to provided headers or otherwise. This would likely be 1 or 0.
+		for skipN > 0 {
+			skipN--
+			// Early exit if the scanner skipped past all data.
+			if !scanner.Scan() {
+				return Flatten(results, opts...)
+			}
+		}
+
+		header_count := len(headers)
```
RebeccaMahany, line 39: `headerCount` for camel case, please

ee/dataflatten/string_delimited.go
```diff
@@ -15,6 +15,75 @@ const (
...
 )
 
+// StringDelimitedLineFunc returns a function that conforms to the interface expected by
+// dataflattentable.Table's execDataFunc property. Properties are grouped into a single
+// record per line of data by splitting the line into fields based on a separator.
+//
+// Headers are expected to either be provided, or set from the first line of data using the separator.
+// If no separator is provided, it's assumed that fields are separated by whitespace.
+// The first N lines of data can be skipped in case garbage is sent before the data.
+func StringDelimitedLineFunc(sep string, headers []string, skipN int) dataFunc {
```
RebeccaMahany, line 25: I think it'd be nice to have unit tests for these new functions -- maybe in a follow-up PR later?

RebeccaMahany (approved):
LGTM -- just one nitpick about casing

ee/dataflatten/string_delimited.go
Micah-Kolide: For sure. I'll add that to my list of future things.

RebeccaMahany (dismissed):

ee/dataflatten/string_delimited.go
directionless, line 25: Should this move to the exec parsers?

ee/tables/dataflattentable/tables.go
```diff
@@ -37,7 +38,10 @@ type Table struct {
 	cmdGen   allowedcmd.AllowedCommand
 	execArgs []string
 
-	keyValueSeparator string
+	keyValueSeparator  string
```
directionless, line 41: I'm a little unsure about adding these. I kinda feel like we should be moving the parsing to it's own thing. But I can't _quite_ explain it

directionless: I think the `dataflattentable.NewExecAndParseTable` pattern is more modular, since it's not just a bunch of options/struct variables on the table. 

I could also believe this is good enough, and we can iterate later. Maybe

ee/dataflatten/string_delimited.go
Micah-Kolide: I first wrote this method for `flatpak`, but since I wasn't able to utilize it I still thought it might be of value to parse csv like data. Also since `snap` outputs data in that format I added that to use it as a model for `StringDelimitedLineFunc`.

I can understand wanting this in exec parsers, but I also feel like this will be a repeatable use case for binary output parsing.

ee/tables/dataflattentable/tables.go
Micah-Kolide: Yeah I felt icky with adding 3 new elements, but I wasn't sure of a clean way handle the requirements of the `StringDelimitedLineFunc ` method without them, at least for now.

Micah-Kolide: Since I was adding this to the `string_delimited` package I tried to format it closely to the `KeyValue` methods.

ee/dataflatten/string_delimited.go
Micah-Kolide: This has now been moved to the exec parsers.

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (p parser) Parse(reader io.Reader) (any, error) {
+	return p.parseLines(reader)
+}
+
+// parseLines scans a reader line by line and splits it into fields based on a delimiter.
+// The line fields are paired with a header, which is defined by an input array, or the first line of data.
+// If no delimiter is provided, it's assumed that fields are separated by whitespace.
+// The first N lines of data can be skipped in case garbage is sent before the data.
```
directionless, line 31:
```suggestion
```

This behavior is elsewhere

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,92 @@
...
+			row[strings.TrimSpace(p.headers[c])] = strings.TrimSpace(fields[c])
+		}
+
+		results = append(results, row)
+	}
+
+	return results, nil
+}
+
+// Switch to the appropriate function to return the current line's fields.
```
directionless, line 76: godoc style comments should start with the function name

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,92 @@
...
+// skipLines - The number of initial lines of data to skip. By default no lines are skipped. This can be useful if consistent undesired output/garbage is printed before the data to parse.
+// headers - The set of headers. If left blank, the parser assumes the headers are in the first line of data and splits the line to set them.
+// delimiter - The splitting string. If left blank, the parser assumes the delimiter is whitespace and uses `strings.Fields()` split method.
+type parser struct {
+	skipLines uint
+	headers   []string
+	delimiter string
+}
+
+func Parser(skipLines uint, headers []string, delimiter string) parser {
```
directionless, line 20: If this grows more options, I'd push you to functional arguments, but this is probably fine here

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+)
+
+// data_table is a general parser for an input of data which conforms to columns and rows styled output.
+// Parser options
+// skipLines - The number of initial lines of data to skip. By default no lines are skipped. This can be useful if consistent undesired output/garbage is printed before the data to parse.
+// headers - The set of headers. If left blank, the parser assumes the headers are in the first line of data and splits the line to set them.
+// delimiter - The splitting string. If left blank, the parser assumes the delimiter is whitespace and uses `strings.Fields()` split method.
+type parser struct {
+	skipLines uint
```
directionless, line 15: What case uses skipLines? I can imagine some that want it, like that `top` test data. But I wonder if you'd see more reliability by defining either a comment character (some outputs use `#` for comment style lines), or a stanza deliminator (in that top example, there's an `\n\n` between the header and the content.

ee/tables/execparsers/data_table/parser.go
directionless, line 20: Prefer to name this `New` or `NewParser`

directionless (approved):
I think this looks pretty good. I like the idiom of `New` with options to create a parser. I think it gives us lots of rope. 

I'm not totally sold on all the options, and I suspect we'll eventually want functional arguments there. But you've got good test cases, so it's easy to refactor later.

Oh, please rename `Parser` to `NewParser` or `New` before merging.

ee/tables/execparsers/data_table/parser.go
Micah-Kolide: I didn't have a lot of use cases in mind, but I was actively aware that in some instances I would want to skip initial lines. For example, I could switch the [apt](https://github.com/kolide/launcher/blob/main/ee/tables/execparsers/apt/test-data/apt_upgradeable.txt) or [dnf](https://github.com/kolide/launcher/blob/main/ee/tables/execparsers/dnf/test-data/dnf_upgradeable.txt) upgradeable parsers to use this one, but I would want to skip the initial lines they both output.

I agree that either the comment character or stanza delimiter could be a better pattern for this going forward.

ee/tables/execparsers/data_table/parser.go
Micah-Kolide: Mmm, yeah I'm not sure why I didn't look at doing that initially

Micah-Kolide: I like how `data_table.NewParser()` reads, so I went with that.

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,120 @@
...
+
+		row := make(map[string]string)
+		fields := p.lineSplit(line, headerCount)
+		// It's possible we don't have the same number of fields to headers, so use
+		// min here to avoid a possible array out-of-bounds exception.
+		min := min(headerCount, len(fields))
+
+		// For each header, add the corresponding line field to the result row.
+		// Duplicate headers overwrite the set value. <-- Perhaps this should be different?
+		for c := 0; c < min; c++ {
```
directionless, line 94:
```suggestion
		// For each header, add the corresponding line field to the result row.
		// Duplicate headers overwrite the set value. <-- Perhaps this should be different?
		for c := 0; c < min(headerCount, len(fields)); c++ {
```

`i` is also the normal enumerator

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,120 @@
...
+		p.delimiter = delimiter
+	}
+}
+
+func NewParser(opts ...parserOpt) *parser {
+	p := &parser{
+		skipLines: 0,
+		headers:   nil,
+		delimiter: "",
+	}
```
directionless, line 45:
```suggestion
	p := &parser{}
```

Those are the defaults, you should be able to omit them.

directionless (approved):
I think this is okay. A couple of nits, but you could ignore them

ee/tables/execparsers/data_table/parser.go
Micah-Kolide: Yeah I'll switch to `i`, I sometimes just go with the first letter of purpose for the value and I put c for counters, but I can use the standard first loop enumerator.

ee/tables/execparsers/data_table/parser.go
Micah-Kolide: Sounds good, I'm not sure why I didn't see this initially.

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,116 @@
...
+	results := make([]map[string]string, 0)
+	scanner := bufio.NewScanner(reader)
+
+	// Skip first N lines due to provided headers or otherwise.
+	// This would likely only ever be 1 or 0, but we may want more.
+	for p.skipLines > 0 {
+		p.skipLines--
+		// Early exit if the scanner skipped past all data.
+		if !scanner.Scan() {
+			return results, nil // <-- Do we want to error here?
```
RebeccaMahany, line 66: Should this comment be removed/resolved?

ee/tables/execparsers/data_table/parser.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		row := make(map[string]string)
+		fields := p.lineSplit(line, headerCount)
+		// It's possible we don't have the same number of fields to headers, so use
+		// min here to avoid a possible array out-of-bounds exception.
+		min := min(headerCount, len(fields))
+
+		// For each header, add the corresponding line field to the result row.
+		// Duplicate headers overwrite the set value. <-- Perhaps this should be different?
```
RebeccaMahany, line 89: This seems like a fine way to handle it to me

RebeccaMahany (dismissed):

ee/tables/execparsers/data_table/parser.go
directionless: This would occur if the data is smaller than the `skipLines`? If we end up in the case, how would we know and debug it?

If we return nil here, we don't have much indication of the issue. If we error out, it should get logged by the caller, right? So my inclination is to return an error

ee/tables/execparsers/data_table/parser.go
directionless:
```suggestion
		// Duplicate headers overwrite the set value.
```

directionless (approved):

RebeccaMahany (approved):

PR #1637, opened 2024-03-04
Add status to /id response, and include enrollment status

RebeccaMahany:
Building on https://github.com/kolide/launcher/pull/1632 -- make enrollment status available via /id response

### DIFF
--- a/ee/localserver/request-id.go
+++ b/ee/localserver/request-id.go
@@ -12,18 +12,25 @@ import (
 	"github.com/kolide/launcher/pkg/traces"
 )
 
-type identifiers struct {
-	UUID           string
-	InstanceId     string
-	HardwareSerial string
-}
+type (
+	identifiers struct {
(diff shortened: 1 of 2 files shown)


James-Pickett (approved):
LGTM!

ee/localserver/request-id.go
```diff
@@ -12,18 +12,25 @@ import (
...
+		RequestId string
+		identifiers
+		Nonce     string
+		Timestamp time.Time
+		Status    status
+	}
+
+	status struct {
+		EnrollmentStatus string
+	}
```
directionless, line 32: This feels like an okay PoC, but my hunch is if this grows we'll want to centralize what our status is. Unclear if that's going to be `knapsack.Status`, `control.Status`, or `agent.Status` depends a bit on where the brain is.

directionless (approved):
Let's try

PR #1641, opened 2024-03-06
Add support to restrict localserver response handling to specific origins

directionless:
Relates To: https://github.com/kolide/k2/issues/9634


ee/localserver/krypto-ec-middleware.go
```diff
@@ -216,6 +218,41 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+				return
+			}
+
+			e.slogger.Log(r.Context(), slog.LevelDebug,
+				"origin matches allowlist",
+				"origin", r.Header.Get("Origin"),
+			)
+		} else {
+			e.slogger.Log(r.Context(), slog.LevelDebug,
+				"origin is allowed by default, no allowlist",
```
RebeccaMahany, line 251: Why do we want this?

ee/localserver/krypto-ec-middleware.go
```diff
@@ -216,6 +218,41 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 			}()
 		}
 
+		// Check if the origin is in the allowed list. See https://github.com/kolide/k2/issues/9634
+		if len(cmdReq.AllowedOrigins) > 0 {
```
RebeccaMahany, line 222: I think this needs to move above the previous block, unless I'm misunderstanding what we're trying to accomplish -- we kick off `e.sendCallback` on 217 before this check, so the callback will still execute even if the origin isn't allowed.

RebeccaMahany: I missed the `defer` above + the fact that we want to send callbacks in the error case too

ee/localserver/krypto-ec-middleware.go
```diff
@@ -216,6 +218,41 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+			for _, ao := range cmdReq.AllowedOrigins {
+				if strings.EqualFold(ao, r.Header.Get("Origin")) {
+					allowed = true
+					break
+				}
+			}
+
+			if !allowed {
+				span.SetAttributes(attribute.String("origin", r.Header.Get("Origin")))
+				span.SetStatus(codes.Error, "origin is not allowed")
```
RebeccaMahany, line 233:
```suggestion
				traces.SetError(span, fmt.Errorf("origin %s is not allowed", r.Header.Get("Origin")))
```

`SetError` will set the status on the span plus make some other attempts to capture info

ee/localserver/krypto-ec-middleware_test.go
```diff
@@ -209,6 +209,139 @@ func TestKryptoEcMiddleware(t *testing.T) {
...
+	var tests = []struct {
+		name           string
+		requestOrigin  string
+		allowedOrigins []string
+		logStr         string
+		expectedStatus int
+	}{
+		{
+			name:           "no allowed specified",
+			requestOrigin:  "auth.example.com",
```
RebeccaMahany, line 230: The origin will include the scheme, so I think either a) logic should be adjusted to handle URLs with and without schemes, or b) the test cases should use origins with schemes.

ee/localserver/request-id.go
```diff
@@ -78,6 +79,7 @@ func (ls *localServer) requestIdHandlerFunc(w http.ResponseWriter, r *http.Reque
 	response := requestIdsResponse{
 		Nonce:     ulid.New(),
 		Timestamp: time.Now(),
+		Origin:    r.Header.Get("Origin"),
```
RebeccaMahany, line 82: This will be empty because of how we construct this request in `krypto-ec-middleware` -- we don't bring along any of the wrapping request's headers. I was able to include the origin by updating the instantiation in `(e *kryptoEcMiddleware) Wrap` like this --

```
newReq := &http.Request{
	Method: http.MethodPost,
	URL: &url.URL{
		Scheme: r.URL.Scheme,
		Host:   r.Host,
		Path:   cmdReq.Path,
	},
	Header: make(http.Header),
}
newReq.Header.Set("Origin", r.Header.Get("Origin"))
```

James-Pickett (commented):
LGTM pending Becca's comments

ee/localserver/krypto-ec-middleware.go
directionless: Oh right! We have a helper. I updated the other parts of the file too

ee/localserver/krypto-ec-middleware_test.go
directionless: good call. I added schemes.

ee/localserver/request-id.go
directionless: nice

James-Pickett (approved):

RebeccaMahany (approved):

PR #1644, opened 2024-03-07
add support for collecitng multiple user keys on via secure enclave

James-Pickett:
* adds a launcher cmd for creating secure enclave key
* adds secureenclavesigner that implements crypto.Signer
* secureenclavesigner determines current console user and returns public secureenclave key for that user, execing cmd in user context

for now this just creates keys, follow on work will implement actual signing


ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,240 @@
...
+	if ses.pathToLauncherBinary == "" {
+		p, err := os.Executable()
+		if err != nil {
+			return nil, fmt.Errorf("getting path to launcher binary: %w", err)
+		}
+
+		ses.pathToLauncherBinary = p
+	}
+
+	return ses, nil
```
James-Pickett, line 66: wonder if we should do some kind of key verification here? we can at least query secure enclave to make sure it recognizes the users key.

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,240 @@
...
+		ses.slogger.Log(context.TODO(), slog.LevelError,
+			"getting first console user",
+			"err", err,
+		)
+
+		return nil
+	}
+
+	if v, ok := ses.uidPubKeyMap[c.Uid]; ok {
+		return v
```
James-Pickett, line 86: same here, should we verify that secure enclave has this key?

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,240 @@
...
+// keys it creates, this is needed since we may need to create a new key anytime
+// a new user logs in on macos
+type persister interface {
+	Persist([]byte) error
+}
+
+func New(slogger *slog.Logger, persister persister, opts ...opt) (*SecureEnclaveSigner, error) {
+	ses := &SecureEnclaveSigner{
+		uidPubKeyMap: make(map[string]*ecdsa.PublicKey),
+		persister:    persister,
```
James-Pickett, line 48: instead of expecting the calling code to unmarshall a json in to this struct, mabye we should just pull the data in new?

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: ehh .. not sure it's worth the extra execs

we don't save a key unless secure enclave creates it
once we store the key server side, if the key is lost some how launcher would need a new install if were strict on validating hardware

ee/agent/keys.go
```diff
@@ -40,11 +39,6 @@ func SetupKeys(slogger *slog.Logger, store types.GetterSetterDeleter) error {
 		return fmt.Errorf("setting up local db keys: %w", err)
 	}
 
-	// Secure Enclave is not currently supported, so don't spend startup time waiting for it to work -- see keys_darwin.go for more details.
```
James-Pickett, line 43: should we roll this out internally only first?

pkg/osquery/table/launcher_info.go
```diff
@@ -98,6 +100,17 @@ func generateLauncherInfoTable(store types.GetterSetter) table.GenerateFunc {
 			return results, nil
 		}
 
+		if runtime.GOOS == "darwin" && agent.HardwareKeys() != nil && agent.HardwareKeys().Public() != nil {
+			jsonBytes, err := json.Marshal(agent.HardwareKeys())
+			if err != nil {
+				return nil, fmt.Errorf("marshalling hardware keys: %w", err)
+			}
+			results[0]["hardware_key"] = string(jsonBytes)
```
James-Pickett, line 108: this produces a json array of objects with uid & keys, need to see how K2 handles this

ee/agent/keys.go
RebeccaMahany: Are there risks associated with rolling it out?

ee/agent/keys_darwin.go
```diff
@@ -5,45 +5,24 @@ package agent
...
-	*/
+	ses, err := secureenclavesigner.New(slogger, store)
+	if err != nil {
+		return nil, fmt.Errorf("creating secureenclave signer: %w", err)
+	}
+
+	// this is kind of weird, but we need to call public to ensure the key is generated
+	// it's done this way to do satisfying signer interface which doesn't return an error
+	if ses.Public() == nil {
+		return nil, errors.New("public key was not be created")
```
RebeccaMahany, line 24: nitpick -- typo 🙂 

```suggestion
		return nil, errors.New("public key was not created")
```

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,253 @@
...
+
+type secureEnclaveSigner struct {
+	uidPubKeyMap         map[string]*ecdsa.PublicKey
+	pathToLauncherBinary string
+	store                types.GetterSetterDeleter
+	slogger              *slog.Logger
+	mux                  *sync.Mutex
+}
+
+func New(slogger *slog.Logger, store types.GetterSetterDeleter, opts ...opt) (*secureEnclaveSigner, error) {
```
RebeccaMahany, line 40: I think it would be nice throw some traces in here:

```
ctx, span := traces.StartSpan(<parent ctx>)
defer span.End()
```

(if it's not too much of a pain to pass in a parent ctx too) -- I'm suggesting it because I think this happens on launcher startup, so it would be helpful to get this into our startup span so we can see timing details.

ee/agent/keys.go
James-Pickett: I remember we had the issue with some beta versions causing trouble, but that might have been because it was failing over and over again. I think the suggestion about adding traces to this area was a good one and will give more confidence.

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: Added some tracing, it was pretty easy to pass ctx

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+	data, err := store.Get([]byte(PublicEccDataKey))
+	if err != nil {
+		traces.SetError(span, fmt.Errorf("getting public ecc data from store: %w", err))
+		return nil, fmt.Errorf("getting public ecc data from store: %w", err)
+	}
+
+	if data != nil {
+		if err := json.Unmarshal(data, ses); err != nil {
+			traces.SetError(span, fmt.Errorf("unmarshaling secure enclave signer: %w", err))
+			ses.slogger.Log(context.TODO(), slog.LevelError,
```
RebeccaMahany, line 62: If you end up pushing changes again -- `ses.slogger.Log(ctx, slog.LevelError,` to link the available ctx? NBD if not, though! Ditto with line 91

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+	// get current console user key to make sure it's available
+	if _, err := ses.currentConsoleUserKey(ctx); err != nil {
+		traces.SetError(span, fmt.Errorf("getting current console user key: %w", err))
+		ses.slogger.Log(context.TODO(), slog.LevelError,
+			"getting current console user key",
+			"err", err,
+		)
+
+		// intentionally not returning error here, because this runs on start up
+		// and maybe the console user or secure enclave is not available yet
```
RebeccaMahany, line 97: The documentation is helpful, thank you

RebeccaMahany (dismissed):
🔥 let's try it!

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: good catch, using ctxs now!

RebeccaMahany (dismissed):

ee/localserver/krypto-ec-middleware.go
```diff
@@ -303,7 +304,8 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 		// it's possible the keys will be noop keys, then they will error or give nil when crypto.Signer funcs are called
 		// krypto library has a nil check for the object but not the funcs, so if are getting nil from the funcs, just
 		// pass nil to krypto
-		if e.hardwareSigner != nil && e.hardwareSigner.Public() != nil {
+		// hardware signing is not implemented for darwin
+		if runtime.GOOS != "darwin" && e.hardwareSigner != nil && e.hardwareSigner.Public() != nil {
```
directionless, line 308: Why change this?

ee/agent/keys_tpm.go
```diff
@@ -28,17 +31,24 @@ func setupHardwareKeys(slogger *slog.Logger, store types.GetterSetterDeleter) (k
 		priData, pubData, err = tpm.CreateKey()
 		if err != nil {
 			clearKeyData(slogger, store)
+			traces.SetError(span, fmt.Errorf("creating key: %w", err))
 			return nil, fmt.Errorf("creating key: %w", err)
 		}
 
+		span.AddEvent("new_key_created")
```
directionless, line 38: do we need this? It's somewhat implied by the function name. I guess we're still learning the tools

pkg/osquery/extension.go
```diff
@@ -92,6 +92,9 @@ type ExtensionOpts struct {
 	// RunDifferentialQueriesImmediately allows the client to execute a new query the first time it sees it,
 	// bypassing the scheduler.
 	RunDifferentialQueriesImmediately bool
+	// skipHardwareKeysSetup is a flag to indicate if we should skip setting up hardware keys.
+	// This is useful for testing environments where we don't have required hardware.
+	skipHardwareKeysSetup bool
```
directionless, line 97: I like that this is explicit, but it does require changing a lot of test filed. Do you think it makes more sense than switching off the environment?

ee/secureenclavesigner/secureenclavesigner_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	if os.Getenv("CI") != "" {
+		t.Skipf("\nskipping because %s env var was not empty, this is being run in a CI environment without access to secure enclave", testWrappedEnvVarKey)
+	}
+
+	// put the root dir somewhere else if you want to persist the signed macos app bundle
+	// should build this into make at some point
+	rootDir := "/tmp/secure_enclave_test"
+
+	// rootDir := t.TempDir()
```
directionless, line 45: This should probably be using `t.TempDir()` as the commented code implies

pkg/osquery/table/launcher_info.go
directionless: This is changes the meaning of the column from prior _and_ makes it different per platform. That's going to be hard to manage inside k2. Instead, I would recommend we move to using `hardware_keys` (and `hardware_keys_source`) and be consistent consistent across platforms.

directionless (commented):
Looks pretty good. Mostly nits and questions. I'm a bit ambivalent about the signature changes, but okay.

I think we do need to change how these are exposed to k2. The change in column meaning is hard to support.

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+		mux:          &sync.Mutex{},
+	}
+
+	data, err := store.Get([]byte(PublicEccDataKey))
+	if err != nil {
+		traces.SetError(span, fmt.Errorf("getting public ecc data from store: %w", err))
+		return nil, fmt.Errorf("getting public ecc data from store: %w", err)
+	}
+
+	if data != nil {
```
directionless, line 59: What happens if it's nil? That branch isn't populated. Maybe early return?

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+		if err := json.Unmarshal(data, ses); err != nil {
+			traces.SetError(span, fmt.Errorf("unmarshaling secure enclave signer: %w", err))
+			ses.slogger.Log(ctx, slog.LevelError,
+				"unable to unmarshal secure enclave signer, data may be corrupt, wiping",
+				"err", err,
+			)
+
+			if err := store.Delete([]byte(PublicEccDataKey)); err != nil {
```
directionless, line 67: I think it's probably correct to delete the key if it's corrupt, but it feels a little odd to do it here in the secureenclavesigner.

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+				return nil, fmt.Errorf("deleting corrupt public ecc data: %w", err)
+			}
+		}
+	}
+
+	for _, opt := range opts {
+		opt(ses)
+	}
+
+	if ses.pathToLauncherBinary == "" {
```
directionless, line 78: Feels a little odd to record this. Why not look at `os.Executable` each time?

ee/secureenclavesigner/secureenclavesigner_darwin.go
```diff
@@ -0,0 +1,295 @@
...
+func (ses *secureEnclaveSigner) createKey(ctx context.Context, u *user.User) (*ecdsa.PublicKey, error) {
+	ctx, span := traces.StartSpan(ctx)
+	defer span.End()
+
+	cmd, err := allowedcmd.Launchctl(
+		ctx,
+		"asuser",
+		u.Uid,
+		"sudo",
+		"--preserve-env",
```
directionless, line 219: Hrm. No idea what the risk is here. It's probably fine, but it's a bit of a yellow flag

ee/localserver/krypto-ec-middleware.go
James-Pickett: In the current state, the signer will not be nil, but will error when sign function is called. So it didn't make sense to me to just let it error.

ee/agent/keys_tpm.go
James-Pickett: I like seeing the events when I look at a span, but I can live without them.

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: I don't we would gain anything with an early return. With the intention of having key available as soon as possible, were making a call to `currentConsoleUserKey` that will get/create keys near bottom of func. We would have to add another line calling this if we early returned.

cmd/launcher/secure_enclave_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+const (
+	testWrappedEnvVarKey = "SECURE_ENCLAVE_TEST_WRAPPED"
+	macOsAppResourceDir  = "../../ee/secureenclavesigner/test_app_resources"
+)
+
+// TestSecureEnclaveTestRunner creates a MacOS app with the binary of this packages tests, then signs the app with entitlements and runs the tests.
+// This is done because in order to access secure enclave to run tests, we need MacOS entitlements.
+// #nosec G306 -- Need readable files
+func TestSecureEnclaveTestRunner(t *testing.T) {
```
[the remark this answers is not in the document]
directionless: I see the distinction, I'm unsure where the tests should live. OTOH, it's easy to move them later.

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: Yea it feels a little odd. I though it might be better than switching on errors in the calling function, deciding what to do, then calling new again. Also, since, without some big changes to the way `SetUpKeys` works, we have to let `secureEnclaveSIgner` handle the saving of new keys. So it kinda made sense to just let it handle the whole storage  layer.

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: This is for testing. If we look at os.Executable in a test, it will have a go test binary and will error out.

James-Pickett: I will leave a comment about this

James-Pickett:
follow up PR addressing post merge comments here https://github.com/kolide/launcher/pull/1647

ee/secureenclavesigner/test_app_resources/info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleExecutable</key>
+	<string>launcher_test</string>
+	<key>CFBundleIdentifier</key>
+	<string>com.kolide.agent</string>
```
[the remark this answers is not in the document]
James-Pickett: addressed in https://github.com/kolide/launcher/pull/1647

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: comment left in https://github.com/kolide/launcher/pull/1647

ee/secureenclavesigner/secureenclavesigner_darwin.go
James-Pickett: I believe we need this to retain the os variable `LAUNCHER_SKIP_UPDATES`, though we could move the cmd to the top of main and bypass the update logic completely.

pkg/osquery/extension.go
James-Pickett: Ehh, I waffled on this one. It felt worse to have logic in the production code to try and figure out if we were under test.

cmd/launcher/secure_enclave_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/kolide/krypto/pkg/echelper"
+	"github.com/kolide/krypto/pkg/secureenclave"
+	"github.com/kolide/launcher/ee/secureenclavesigner"
+)
+
+// runSecureEnclave performs either a create-key operation using the secure enclave.
+// It's available as a separate command because launcher runs as root by default and since it's
+// not in a user security context, it can't use the secure enclave directly. However, this command
+// can be run in the user context using launchctl.
+func runSecureEnclave(args []string) error {
```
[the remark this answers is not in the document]
James-Pickett: Early on we considered trying to do this, but decided not to since it was easy to spoof ppid. If we're feeling different now and we do go this route, we could even exec codesign and verify that the calling binary is signed as we expect.

ee/secureenclavesigner/secureenclavesigner_test.go
James-Pickett: addressed in https://github.com/kolide/launcher/pull/1647

pkg/osquery/table/launcher_info.go
James-Pickett: addressed in https://github.com/kolide/launcher/pull/1647

ee/secureenclavesigner/secureenclavesigner_darwin.go
directionless: Hrm. That's reasonable. But I think points to some lurking need to refactor. Maybe the quick path is to leave a comment, and call it good enough

ee/agent/keys.go
directionless: IIRC one version of macOS threw ugly errors with the secure enclave call, and caused launcher to crash

ee/secureenclavesigner/secureenclavesigner_darwin.go
directionless: There are a couple of ways to approach this.

If that's the only one we care about, maybe we could `--preserve-env= LAUNCHER_SKIP_UPDATES`.

I think we could also invoke sudo with `LAUNCHER_SKIP_UPDATES=true` in the command line. from the man page:
> Environment variables to be set for the command may also be passed as options to sudo in the form    VAR=value, for example LD_LIBRARY_PATH=/usr/local/pkg/lib.

PR #1647, opened 2024-03-11
secure enclave keys init follow up

James-Pickett:


pkg/osquery/table/launcher_info.go
```diff
@@ -105,16 +105,37 @@ func generateLauncherInfoTable(store types.GetterSetter) table.GenerateFunc {
 			if err != nil {
 				return nil, fmt.Errorf("marshalling hardware keys: %w", err)
 			}
-			results[0]["hardware_key"] = string(jsonBytes)
```
directionless, line 108: I think it's fine to drop this, but we'll need to update k2 to stop querying it

directionless (approved):

James-Pickett:
closed for https://github.com/kolide/launcher/pull/1658

PR #1648, opened 2024-03-12
Hardcode use_tuf_autoupdater in startup settings so that v1.5.3 can use it

RebeccaMahany:
v1.5.3 still needs this flag, so it should not be removed from the startup settings. Instead, hardcode it to always be set.

### DIFF
--- a/ee/agent/startupsettings/writer.go
+++ b/ee/agent/startupsettings/writer.go
@@ -61,6 +61,7 @@ func (s *startupSettingsWriter) setFlags() error {
 	for flag, getter := range s.storedFlags {
 		updatedFlags[flag.String()] = getter()
 	}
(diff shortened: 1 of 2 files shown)


James-Pickett (approved):

ee/agent/startupsettings/writer.go
```diff
@@ -61,6 +61,7 @@ func (s *startupSettingsWriter) setFlags() error {
 	for flag, getter := range s.storedFlags {
 		updatedFlags[flag.String()] = getter()
 	}
+	updatedFlags["use_tuf_autoupdater"] = "enabled" // Hardcode for backwards compatibility
```
directionless, line 64: Might be worth a comment about what version this is for. I can imagine wanting to drop it in the future.

PR #1653, opened 2024-03-14
Refactor runAsUser into tablehelper package

Micah-Kolide:
Cleaning up the runAsUser idiom I was using in a couple of places for the table generates.


ee/tables/tablehelpers/posix_run_cmd_as_user.go
```diff
@@ -0,0 +1,66 @@
+//go:build !windows
```
RebeccaMahany, line 1: I think the convention is to make the OS the suffix, not the prefix -- `run_cmd_as_user_posix.go`. (I couldn't quickly find great docs on this, but it's mentioned [here](https://www.digitalocean.com/community/tutorials/building-go-applications-for-different-operating-systems-and-architectures#using-goos-and-goarch-filename-suffixes) or you can see e.g. [here](https://pkg.go.dev/golang.org/x/sys@v0.18.0/unix#section-sourcefiles).)

Micah-Kolide: Sounds good. I questioned how I should've named it and thinking a bit more I agree with the suffix.

directionless: IIRC early go used the suffixes as implicit build tags, and that ended up being an antipattern. So now the convention is filename suffixes, and build tags.

directionless: (This is how the `_test.go` files don't get built into the prod binaries)

directionless (commented):
Hrm... this might be okay, but it _is_ a bit of a sharp edge.

tablehelpers.Exec enforces that it's called with a `timeout`, and ensures consistent logging. I'm not totally sure the signature is correct, but it does unify a lot of things. 

In contrast, this takes a `cmd` and runs it. Which, in some ways, is a better idiom. But I'm not sure how we can enforce things like timeouts, and create consistent logging. 

I'd be inclined to make a functional argument on `tablehelpers.Exec` that takes `RunAs(uid)`

The more arduous path would be to rethink how we handle exec. I can imagine passing cmd around, and maybe there are other patterns.

Curious where y'all land

RebeccaMahany:
I don't feel strongly about it either way. I think enforcing the timeout/logging is a plus, but I'm not terribly worried about not enforcing either since we do a good job including that in the first place and/or requesting it in code review.

ee/tables/tablehelpers/exec.go
```diff
@@ -6,14 +6,57 @@ import (
...
+			return fmt.Errorf("converting uid %s to int: %w", runningUser.Uid, err)
+		}
+
+		runningUserGid, err := strconv.ParseUint(runningUser.Gid, 10, 32)
+		if err != nil {
+			return fmt.Errorf("converting gid %s to int: %w", runningUser.Gid, err)
+		}
+
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Credential: &syscall.Credential{
```
RebeccaMahany, line 50: I think you'll need an exec_posix.go and an exec_windows.go to be able to define `WithUid` separately

directionless (commented):
Looks pretty good. As Becca said -- drop the `WithUid` into a `_posix` flavor.

ee/tables/tablehelpers/exec.go
Micah-Kolide: Broke it out into a different go file

RebeccaMahany (approved):
🔥

PR #1662, opened 2024-03-27
dont start osq instance in stand by

James-Pickett:


pkg/osquery/runtime/runner.go
```diff
@@ -66,6 +66,13 @@ func New(k types.Knapsack, opts ...OsqueryInstanceOption) *Runner {
 }
 
 func (r *Runner) Run() error {
+	if r.knapsack.InModernStandby() {
+		r.slogger.Log(context.TODO(), slog.LevelInfo,
+			"modern standby detected, skipping osquery instance start",
+		)
+		return nil
```
directionless, line 73: If we return nil, what happens when it wakes up?

James-Pickett: I'm not certain this is the right way or if we should just do some kind of wait loop, it's difficult to test. Currently all the osq runtime tests don't work on windows, so I'm trying to update those.

directionless:
I don't think this makes sense.

PR #1664, opened 2024-03-28
add menu update change detection logs

zackattack01:
This add a `menuItemCache` to the desktop runner to detect when changes to the top level items occur for the menu bar data.

Within the menu template that is sent down, there are several timestamps and links that change regularly but are not meaningful for our change detection purposes. The simplest approach I found to filtering this noise was to catalog the top level items by label, and note any changes there. This will provide us with a record of any major changes (pending registrations, device status change, etc) without being too noisy. This logic is fairly isolated and easy to update if anyone has better ideas for detection or if there's more we'd want to track

### DIFF
--- a/ee/desktop/runner/menu_cache.go
+++ b/ee/desktop/runner/menu_cache.go
@@ -1,0 +1,74 @@
+package runner
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+)
+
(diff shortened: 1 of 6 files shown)


James-Pickett (approved):
NICE!

RebeccaMahany (approved):
🔥

directionless (approved):
LGTM

PR #1667, opened 2024-04-02
Small improvements to logging and flares

directionless:
Some tweaks as I debug things

- Gather service manager logs via 2 patterns. I think the new one provides better data, but I'm not so sure I want to remove the old one
- Don't log the full results during `PublishLogs`. They are too big and cause our log files to rotate way too fast. 
- Use json for things
- set a merge strategy on go.sum to reduce conflicts

### DIFF
--- a/ee/debug/checkups/services_windows.go
+++ b/ee/debug/checkups/services_windows.go
@@ -239,7 +239,7 @@ func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
 		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
 	}
 
-	filterExpression := fmt.Sprintf(`'@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}'`, kolideSvcName)
+	filterExpression := fmt.Sprintf(`@{LogName='System'; ProviderName='Service Control Manager'; Data='%s'}`, kolideSvcName)
 
 	cmdArgs := []string{
 		"Get-WinEvent",
--- a/ee/debug/checkups/services_windows.go
+++ b/ee/debug/checkups/services_windows.go
@@ -239,7 +239,7 @@ func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
 		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
 	}
 
-	filterExpression := fmt.Sprintf(`@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}`, kolideSvcName)
+	filterExpression := fmt.Sprintf(`'@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}'`, kolideSvcName)
 
 	cmdArgs := []string{
 		"Get-WinEvent",
--- a/ee/debug/checkups/services_windows.go
+++ b/ee/debug/checkups/services_windows.go
@@ -239,11 +239,14 @@ func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
 		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
 	}
 
+	filterExpression := fmt.Sprintf(`@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}`, kolideSvcName)
+
 	cmdArgs := []string{
 		"Get-WinEvent",
-		`-FilterHashtable @{LogName='System'; ProviderName='Service Control Manager'}`,
+		"-MaxEvents 100",
+		"-FilterHashtable", filterExpression,
 		"|",
-		"ForEach-Object { $_.Message }",
+		"ConvertTo-Json",
 	}
 
 	cmd, err := allowedcmd.Powershell(ctx, cmdArgs...)
@@ -268,7 +271,7 @@ func gatherServiceManagerEventLogs(ctx context.Context, z *zip.Writer) error {
 
 	cmdletArgs := []string{
 		"Get-EventLog",
-		"-Newest", "50",
+		"-Newest", "100",
 		"-LogName", "System",
 		"-Source", "\"Service Control Manager\"",
 		"-Message", fmt.Sprintf("*%s*", kolideSvcName),
--- a/ee/debug/checkups/services_windows.go
+++ b/ee/debug/checkups/services_windows.go
@@ -254,7 +254,8 @@ func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
 		return fmt.Errorf("creating powershell command: %w", err)
 	}
 	hideWindow(cmd)
-	cmd.Stdout = out // write directly to zip
+	cmd.Stdout = out
+	cmd.Stderr = out
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("running Get-WinEvent: error %w", err)
 	}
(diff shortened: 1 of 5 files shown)


zackattack01 (dismissed):

ee/debug/checkups/services_windows.go
```diff
@@ -228,10 +232,38 @@ func gatherServices(z *zip.Writer, serviceManager *mgr.Mgr) error {
...
+	if err != nil {
+		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
+	}
+
+	cmdArgs := []string{
+		"Get-WinEvent",
+		`-FilterHashtable @{LogName='System'; ProviderName='Service Control Manager'}`,
+		"|",
+		"ForEach-Object { $_.Message }",
+	}
```
RebeccaMahany, line 247: The other option was filtering Service Control Manager events on ones that specifically had to do with the Kolide service. I think we might want that here too -- it would keep us from having to sift through irrelevant events, and from collecting data we don't need. What do you think? (Is it possible to filter like that with `Get-WinEvent`?)

pkg/service/publish_results.go
```diff
@@ -184,6 +184,11 @@ func (s *grpcServer) PublishResults(ctx context.Context, req *pb.ResultCollectio
 func (mw logmw) PublishResults(ctx context.Context, nodeKey string, results []distributed.Result) (message, errcode string, reauth bool, err error) {
 	defer func(begin time.Time) {
 		resJSON, _ := json.Marshal(results)
+		resTruncated := string(resJSON[:200])
```
RebeccaMahany, line 187: Thanks for adding this -- I've been running into this issue as well

RebeccaMahany (approved):

James-Pickett (approved):

ee/debug/checkups/services_windows.go
RebeccaMahany: I do, yes! Happy to run some tests in powershell if you'd like

directionless: If you can, that'd be great. I don't have a handy windows host :|

RebeccaMahany: Ok, I took a look!

An unrelated item I noticed is that this won't include date/timestamp, which I figure we may want for correlation with launcher logs later. We could do the same pipe to `Format-Table -Wrap -AutoSize` that the command below does instead of filtering on the `Message` here. That gives us date, timestamp, event code, error level, and message -- not an overwhelming number of columns.

As far as I can tell from [Get-WinEvent docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.diagnostics/get-winevent?view=powershell-7.4), we can't filter on Message the same way we do for `Get-EventLog`. If we want a similar implementation, then this works for me:

```
Get-WinEvent  -FilterHashtable @{LogName='System'; ProviderName='Service Control Manager'} | 
    Where {$_.Message.Contains('LauncherKolideK2Svc')} | 
    Format-Table -Wrap -AutoSize
```

Let me know if you want me to try out anything else.

directionless: I found how to search in the message body, and tossed it into json. This seems to work on a windows test machine I have. Does it work on yours?

RebeccaMahany: The filtering works -- I needed to put single quotes around the service name, though. `@{LogName='System'; ProviderName='Service Control Manager'; Data='LauncherKolideK2Svc'}`

RebeccaMahany: (The `ConvertTo-Json` is nice!)

ee/debug/checkups/services_windows.go
```diff
@@ -228,15 +232,46 @@ func gatherServices(z *zip.Writer, serviceManager *mgr.Mgr) error {
...
 }
 
+// gatherServiceManagerEvents uses Get-WinEvent to fetch the service manager logs. This might be newer than Get-EventLog
+func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
+	out, err := z.Create("eventlog-Get-WinEvent.json")
+	if err != nil {
+		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
+	}
+
+	filterExpression := fmt.Sprintf(`@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}`, kolideSvcName)
```
directionless, line 242:
```suggestion
	filterExpression := fmt.Sprintf(`'@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}'`, kolideSvcName)
```

ee/debug/checkups/services_windows.go
directionless: Okay. I think I just added single ticks

directionless: One caveat about `ConvertTo-Json` -- the timestamps are in a windows native format, it's ms since the epoch. Just something to know when translating them

ee/debug/checkups/services_windows.go
```diff
@@ -228,15 +232,46 @@ func gatherServices(z *zip.Writer, serviceManager *mgr.Mgr) error {
...
+
+	cmd, err := allowedcmd.Powershell(ctx, cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("creating powershell command: %w", err)
+	}
+	hideWindow(cmd)
+	cmd.Stdout = out // write directly to zip
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("running Get-WinEvent: error %w", err)
+	}
```
RebeccaMahany, line 260:
```suggestion
	cmd.Stdout = out // write directly to zip
	errBytes := bytes.Buffer{}
	cmd.Stderr = &errBytes
	if err := cmd.Run(); err != nil {
		return fmt.Errorf("running Get-WinEvent: error %w; stderr %s", err, errBytes.String())
	}
```

When testing locally I remembered that Windows stdout doesn't contain useful error info and we have to capture stderr if we want to get the actual error message.

(This suggestion isn't standalone because it also requires adding a "bytes" import to this file -- if it's easier I can add this here and to the `Get-EventLog` cmd below in a follow-up PR.)

directionless: As a potential counter proposal, what do you think about `cmd.Stderr = out` ?

ee/debug/checkups/services_windows.go
```diff
@@ -228,15 +232,46 @@ func gatherServices(z *zip.Writer, serviceManager *mgr.Mgr) error {
...
 }
 
+// gatherServiceManagerEvents uses Get-WinEvent to fetch the service manager logs. This might be newer than Get-EventLog
+func gatherServiceManagerEvents(ctx context.Context, z *zip.Writer) error {
+	out, err := z.Create("eventlog-Get-WinEvent.json")
+	if err != nil {
+		return fmt.Errorf("creating eventlog-Get-WinEvent.json: %w", err)
+	}
+
+	filterExpression := fmt.Sprintf(`'@{LogName='System'; ProviderName='Service Control Manager'; Data=%s}'`, kolideSvcName)
```
RebeccaMahany, line 242:
```suggestion
	filterExpression := fmt.Sprintf(`@{LogName='System'; ProviderName='Service Control Manager'; Data='%s'}`, kolideSvcName)
```

Sorry, this is what I meant!

ee/debug/checkups/services_windows.go
RebeccaMahany: Works for me also!

ee/debug/checkups/services_windows.go
directionless: :smacks forehead: of course.

ee/debug/checkups/services_windows.go
directionless: Did that. If we discover issues, we can always change.

RebeccaMahany (dismissed):

RebeccaMahany (approved):

zackattack01 (approved):

PR #1672, opened 2024-04-03
set up windows event logger first in main

James-Pickett:
Set's up system slogger to log to windows events if running as windows elevated user, first thing in main.

You can see the logs with powershell via

```ps1
Get-WinEvent -FilterHashtable @{LogName='Application'; ProviderName='launcher'} | Sort-Object -Property TimeCreated | ForEach-Object { $_.Message }
```

They will now contain the "started kolide launcher" msg


cmd/launcher/syslogger_windows.go
```diff
@@ -0,0 +1,40 @@
...
+	eventLogWriter, err := eventlog.NewWriter(serviceName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("creating eventlog writer: %w", err)
+	}
+
+	systemSlogger := multislogger.New(slog.NewJSONHandler(eventLogWriter, &slog.HandlerOptions{
+		Level: slog.LevelInfo,
+	}))
+
+	systemSlogger.Info("@@@@@ THIS IS A TEST @@@@@")
```
James-Pickett, line 37: remove this

cmd/launcher/main.go
```diff
@@ -27,27 +27,28 @@ import (
...
-	// create initial logger. As this is prior to options parsing,
-	// use the environment to determine verbosity.  It will be
-	// re-leveled during options parsing.
-	logger := logutil.NewServerLogger(env.Bool("LAUNCHER_DEBUG", false))
+	systemSlogger, logCloser, err := systemSlogger()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error creating system logger: %v\n", err)
+		os.Exit(1)
+	}
+	defer logCloser.Close()
```
James-Pickett, line 35: should we close this just before starting the windows svc? it have two windows event writers does not seem to be an issue 🤷

cmd/launcher/syslogger_windows.go
```diff
@@ -0,0 +1,38 @@
...
+		syslogger.Log(context.TODO(), slog.LevelDebug,
+			"launcher running on windows without elevated permissions, using default stderr instead of eventlog",
+		)
+
+		return syslogger, io.NopCloser(nil), nil
+	}
+
+	eventLogWriter, err := eventlog.NewWriter(serviceName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("creating eventlog writer: %w", err)
```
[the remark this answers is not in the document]
James-Pickett: agree, I've updated to just log and return default logger

cmd/launcher/main.go
```diff
@@ -265,3 +266,9 @@ func runVersion(args []string) error {
 	os.Exit(0)
 	return nil
 }
+
+func defaultSystemSlogger() *multislogger.MultiSlogger {
```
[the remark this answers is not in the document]
James-Pickett: sounds good, moved

cmd/launcher/main.go
James-Pickett: yea, it would be an annoying refactor, also weird passing around a closer

directionless: I'm sure we want the refactor. I'm less sure we want it _today_ (I'm not sure how hard it would be, it's just adding slogger as a parameter to the subcommands, right?)

James-Pickett: seems like we would also need to pass around the closer within main.go if we don't want to use defer .... I'm torn between wanting clean code and wanting to make sure we close the handler before we exit

James-Pickett: > I'm sure we want the refactor. I'm less sure we want it _today_ (I'm not sure how hard it would be, it's just adding slogger as a parameter to the subcommands, right?)

Yea pass slogger as a parameter and this event handler that needs to be closed, you could use an interface that has a close method on it.

James-Pickett: we could do something like

```go
func main() {
	if err := runMain(); err != nil {
		os.Exit(1)
	}
}

func runMain() error {
        // do stuff
        // return an err or nil
}
```

pkg/log/multislogger/syslogger_windows.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/kolide/launcher/pkg/log/eventlog"
+	"golang.org/x/sys/windows"
+)
+
+const serviceName = "launcher"
+
+func SystemSlogger() (*MultiSlogger, io.Closer, error) {
+	if !windows.GetCurrentProcessToken().IsElevated() {
+		syslogger := defaultSystemSlogger()
```
directionless, line 19: This is an interesting idea. Is the intent to try to capture people running in a terminal? Im game for trying this. I don't know if we have better rope to know what our caller is

cmd/launcher/main.go
directionless: Yeah, that's not bad. I was thinking of something much more complicated:
```
var exitDeferred []func

func osExit(ret int) {
  for _, fn := range exitDeferred {
    fn()
  }
  os.Exit(ret)
}

func main() {
  exitDeferred = append(exitDeferred, ...)
  osExit(0)
}

RebeccaMahany: I think both of these are good options 👍 since we already have 2 other defer funcs in here that we'd probably want to move around too, I'm fine w punting this refactor to a subsequent PR in order to merge this one

directionless (approved):
We concluded this was okay to merge, and we'd refactor/tidy in a followup

cmd/launcher/main.go
RebeccaMahany: Adding system multislogger to subcommands: https://github.com/kolide/launcher/pull/1674

PR #1674, opened 2024-04-04
Pass system multislogger through to all launcher subcommands

RebeccaMahany:
In some subcommands (notably the Windows service subcommand), we create a system multislogger. However, we also have already created one in the main.go entrypoint. This PR updates all subcommands to accept the system multislogger created in main.go so that we only create/manage one of them.

I also pulled out `CleanUpRenamedDebugLogs` -- it's been long enough and I don't think we need it anymore.

### DIFF
--- a/cmd/launcher/svc_windows.go
+++ b/cmd/launcher/svc_windows.go
@@ -32,14 +32,14 @@ const serviceName = "launcher"
 // runWindowsSvc starts launcher as a windows service. This will
 // probably not behave correctly if you start it from the command line.
 func runWindowsSvc(systemSlogger *multislogger.MultiSlogger, args []string) error {
-	systemSlogger.Logger.Log(context.TODO(), slog.LevelInfo,
+	systemSlogger.Log(context.TODO(), slog.LevelInfo,
 		"service start requested",
 		"version", version.Version().Version,
 	)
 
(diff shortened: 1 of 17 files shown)


cmd/launcher/svc_windows.go
```diff
@@ -33,46 +31,28 @@ const serviceName = "launcher"
...
-	}))
-
 	localSlogger := multislogger.New()
+	logger := log.NewNopLogger()
 
 	// Create a local logger. This logs to a known path, and aims to help diagnostics
 	if opts.RootDirectory != "" {
 		ll := locallogger.NewKitLogger(filepath.Join(opts.RootDirectory, "debug.json"))
-		logger = teelogger.New(logger, ll)
+		logger = ll
```
directionless, line 55: Does anything use this `logger`, or is it just stale?

directionless (approved):
I _think_ it's okay, it's a little hard to track everything, but worst case we'll end up tweaking which things log where

cmd/launcher/svc_windows.go
RebeccaMahany: It's mostly stale, but there are a couple places the logger is still in use. This change just amounts to only writing to debug.json when accessing `logger` instead of debug.json+event logs, so it should be a safe change.

PR #1675, opened 2024-04-09
Do not make additional control server fetch request during control server interval update

RebeccaMahany:
Because we are already inside a `Fetch` call, the `fetchMutex` is locked -- so, currently, whenever the control request interval updates we deadlock forever and launcher will not make a request to the control server until it is restarted.

This PR removes the extra `Fetch` call. We should still make the next `Fetch` call after the new interval.

### DIFF
--- a/ee/control/control_test.go
+++ b/ee/control/control_test.go
@@ -4,11 +4,17 @@ import (
 	"context"
 	"errors"
 	"io"
+	"strconv"
 	"testing"
 	"time"
 
+	"github.com/kolide/launcher/ee/agent/flags"
 	"github.com/kolide/launcher/ee/agent/flags/keys"
+	"github.com/kolide/launcher/ee/agent/knapsack"
+	"github.com/kolide/launcher/ee/agent/storage"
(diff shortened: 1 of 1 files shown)


ee/control/control.go
```diff
@@ -147,15 +147,6 @@ func (cs *ControlService) requestIntervalChanged(interval time.Duration) {
 		return
 	}
 
-	// perform a fetch now
-	if err := cs.Fetch(); err != nil {
```
James-Pickett, line 151: I think we might still want the immediate fetch if it doesn't come from control server (comes from menubar / local server). What using a trylock inside the fetch function and exiting if we can't get lock?

RebeccaMahany: Oh, thank you, that's a really good point. It looks like that will work -- I'll document and push up that change.

James-Pickett (approved):

zackattack01 (dismissed):
great find 🔥

zackattack01 (approved):

directionless (approved):
I think this is fine

PR #1680, opened 2024-04-11
Ensure file permissions are set appropriately when untarring archives during autoupdate

RebeccaMahany:
Replaces kolide/kit's `UntarBundle` with a new implementation that creates files with sanitized permissions, ensuring that group/public will never have write access to the autoupdate as it's being extracted.


ee/tuf/library_manager.go
```diff
@@ -249,6 +250,87 @@ func (ulm *updateLibraryManager) moveVerifiedUpdate(binary autoupdatableBinary,
...
+func sanitizeExtractPath(filePath string, destination string) error {
+	destpath := filepath.Join(destination, filePath)
+	if !strings.HasPrefix(destpath, filepath.Clean(destination)+string(os.PathSeparator)) {
+		return fmt.Errorf("illegal file path %s", filePath)
+	}
+	return nil
+}
+
+// Allow owner read/write/execute, and only read/execute for all others.
+const allowedPermissionBits fs.FileMode = fs.ModeType | 0755
```
directionless, line 309: nice

PR #1682, opened 2024-04-18
Fix command exec's `WithUid` (RunAsUser) when running as self

Micah-Kolide:
I've been trying to test my new check that uses the `kolide_brew_upgradeable` table, when I kept hitting an issue in Live Query where all devices returned no results. I could easily get results locally, so I wasn't sure what was going on.

After more testing locally, I saw the error `fork/exec /opt/homebrew/bin/brew: operation not permitted`, and after some digging online I came to this [post](https://walac.github.io/golang-patch/).

Basically the syscall `SYS_SETGROUPS` requires elevated permissions, so a non-root user attempting to set the groups causes an `EPERM` error. There is a fix inside the `Credential` structure: `NoSetGroups`, but I figured since we are already running as the user, we can just early exit instead of adding that flag.

Outside of this issue, I had thought that the queries in Live Query go through a sudoed instance of launcher, but perhaps I was mistaken on that?

### DIFF
--- a/ee/tables/tablehelpers/run_as_user_posix.go
+++ b/ee/tables/tablehelpers/run_as_user_posix.go
@@ -24,7 +24,10 @@ func WithUid(uid string) ExecOps {
 			return fmt.Errorf("looking up user with uid %s: %w", uid, err)
 		}
 
-		if currentUser.Uid != "0" && currentUser.Uid != runningUser.Uid {
+		// If the current user is the user to run as, then early return to avoid needing NoSetGroups.
+		if currentUser.Uid == runningUser.Uid {
+			return nil
+		} else if currentUser.Uid != "0" {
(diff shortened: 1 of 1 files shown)


directionless:
> Outside of this issue, I had thought that the queries in Live Query go through a sudoed instance of launcher, but perhaps I was mistaken on that?

Sorta, yes. Outside development, launcher runs with root permission. But dev instances are usually running with user credentials.

directionless (approved):
I think this is reasonable. We've done similar elsewhere.

zackattack01 (approved):

Micah-Kolide:
> > Outside of this issue, I had thought that the queries in Live Query go through a sudoed instance of launcher, but perhaps I was mistaken on that?
> 
> Sorta, yes. Outside development, launcher runs with root permission. But dev instances are usually running with user credentials.

I saw this issue in both my dev env and when I tried a real Live Query in prod.

Issue #1683, opened 2024-04-19
`launcher interactive` should include ATC tables


directionless:
Now that we have this cool `launcher interactive` mode, I'd like to start populating it with ATC tables. I'm not sure if other parts of the configuration should come along. Maybe? Maybe not, depends...

I'd suggest we use the sqlite store for this. It's meant to be publically readable.

PR #1685, opened 2024-04-22
auto load ATC config in interactive

James-Pickett:
automatically load automatic table configuration from osquery config when running launcher in interactive mode

closes https://github.com/kolide/launcher/issues/1683


ee/agent/startupsettings/writer.go
```diff
@@ -63,6 +65,18 @@ func (s *startupSettingsWriter) setFlags() error {
 	}
 	updatedFlags["use_tuf_autoupdater"] = "enabled" // Hardcode for backwards compatibility circa v1.5.3
 
+	// Set flags will only be called when a flag value changes. The osquery config that contains the atc config
+	// comes in via osquery extension. So a new config will not trigger a re-write.
+	atcConfig, err := s.extractAutoTableConstructionConfig()
```
James-Pickett, line 70: Is it worth figuring out the plumbing to have this fire when we get a new osquery config? This would likely mean that launcher interactive will not pick up the ATCs until the first launcher start up after getting osq config.

directionless: I'm not sure I understand... Is this only called on startup? Which means the sqlite conf file may lag weeks to months?

pkg/launcher/paths.go
```diff
@@ -56,7 +57,14 @@ func DefaultPath(path defaultPath) string {
 	// not windows
 	switch path {
 	case RootDirectory:
-		return "/var/kolide-k2/k2device.kolide.com/"
+		const defaultRootDir = "/var/kolide-k2/k2device.kolide.com"
```
directionless, line 60: I thought this would get read out of the config file, and this was just the default?

directionless (commented):
Looks pretty good. A couple questions

ee/agent/startupsettings/writer.go
James-Pickett: Yes, it's possible. I'm trying to find a clever way to plumb this through .... we could treat config like a flag in knapsack and use it's observer pattern or just find a way to have the extension call startupsettings.OpenWriter when the config updates.

pkg/launcher/paths.go
James-Pickett: with interactive we don't currently read the `launcher.flags` file, I suppose we could look for it on start up and if we don't find it, just use defaults

ee/agent/startupsettings/writer.go
directionless: mmm, that's an interesting thought. If we ever decide we need to restart osquery on a config change, would it leverage that?

pkg/launcher/paths.go
directionless: That's probably reasonable. I think we implemented something like in another spot. Maybe related to the new updater.

ee/agent/startupsettings/writer.go
James-Pickett: I doubt we would ever leverage that when launcher (daemon) is restarting osquery, since launcher (daemon) can just read the config out of bolt db.

directionless: Right -- that extension would read it from boltdb. But we still need something like that to trigger the restart, don't we?

pkg/launcher/paths.go
directionless: I guess if we go down this route, we'd need to add `root_directory ` and `config` to interactive's flags.

pkg/osquery/interactive/interactive.go
```diff
@@ -37,6 +44,35 @@ func StartProcess(slogger *slog.Logger, rootDir, osquerydPath string, osqueryFla
...
+		}
+	}
+
+	// start building list of osq plugins with the kolide tables
+	osqPlugins := table.PlatformTables(slogger, osquerydPath)
+
+	// if we were not provided a config path flag, try to add default config
+	if !haveConfigPathOsqFlag {
+		// check to see if we can actually get a config plugin
+		configPlugin, err := configPlugin()
```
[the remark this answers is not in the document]
James-Pickett: updated

ee/agent/startupsettings/writer.go
James-Pickett: I've updated this to have the extension just open the startup settings writer to trigger the write when it gets a new config. I see what you mean about restarting on a new config, as discussed, maybe an issue, but out of scope for this PR.

pkg/osquery/interactive/interactive.go
```diff
@@ -153,3 +188,20 @@ func waitForFile(path string, interval, timeout time.Duration) error {
 		}
 	}
 }
+
+func configPlugin() (*config.Plugin, error) {
+	r, err := startupsettings.OpenReader(context.TODO(), launcher.DefaultPath(launcher.RootDirectory))
```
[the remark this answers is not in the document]
James-Pickett: I've updated it to use launcher opts + knapsack to handle figuring out root dir.

pkg/launcher/paths.go
James-Pickett: I've updated it to use launcher opts + knapsack to handle figuring out root dir.

ee/agent/startupsettings/writer.go
directionless: Agree it's out of scope for this PR. Mostly mentioning it in case it chanses the direction you want to go in

cmd/launcher/interactive.go
```diff
@@ -19,22 +21,17 @@ import (
...
-
-	flagset.Usage = commandUsage(flagset, "launcher interactive")
-	if err := flagset.Parse(args); err != nil {
+	opts, err := launcher.ParseOptions("interactive", args)
+	if err != nil {
 		return err
 	}
 
+	if opts.RootDirectory == "" {
+		opts.RootDirectory = launcher.DefaultPath(launcher.RootDirectory)
```
RebeccaMahany, line 30: Curious what this is used for since we also have the temp root dir?

James-Pickett: the opts.RootDirectory is where to look for the kv.sqlite db to load the ATC config from, the temp root dir is for interactive to store temp files like socket and aegues lense files.

I added some comments to clarify the 2

pkg/osquery/extension.go
```diff
@@ -517,6 +518,18 @@ func (e *Extension) GenerateConfigs(ctx context.Context) (map[string]string, err
 	} else {
 		// Store good config
 		e.knapsack.ConfigStore().Set([]byte(configKey), []byte(config))
+
+		// open the start up settings writer just to trigger a write of the config,
+		// then we can immediately close it
```
RebeccaMahany, line 523: I think maybe it would feel cleaner/less easy to introduce bugs in the future if we had a specific function to call to perform the config write, rather than relying on a side effect of `OpenWriter`...maybe `setFlags` could be exposed as `SyncFlags` or something? What do you think?

RebeccaMahany (commented):
I'm sorry for reviewing this piecemeal -- I didn't quite have enough time to finish up reviewing interactive.go. I'll come back to it first thing tomorrow!

RebeccaMahany (approved):
Never mind, I ended up having time to finish reviewing! This looks good to me overall -- had one final question but it's not blocking.

pkg/osquery/extension.go
James-Pickett: how about renaming to `WriteSettings` and removing from the constructor?

RebeccaMahany: That works for me 🙂

James-Pickett: Done

RebeccaMahany (approved):

directionless (approved):
I didn't test this, but I think it's probably fine

PR #1692, opened 2024-04-25
Standardize how we log and handle panics

RebeccaMahany:
Requires https://github.com/kolide/launcher/pull/1693

If one goroutine panics, all of launcher will terminate. Before termination, the panic will travel up to the top-level function in the executing goroutine. If we want to recover or log panics, we must do that in this top-level function.

This PR introduces `gowrapper` to wrap our goroutines, to make it easier to standardize recovering from and logging panics. It will always log the panic, and allows for providing an `onPanic` function that executes if gowrapper recovers from a panic.

This PR adds `gowrapper` usage to our windows service and to all of our rungroups.

References:
* https://forum.golangbridge.org/t/why-does-go-terminate-the-whole-process-if-one-goroutine-paincs/27122

<details>
<summary>Example logs from panic inside a rungroup</summary>

```
{
  "time":"2024-04-30T14:17:34.976926Z",
  "level":"ERROR",
  "source":{
    "function":"github.com/kolide/launcher/ee/gowrapper.Go.func1.1",
    "file":"/Users/rebeccamahany-horton/Repos/launcher/ee/gowrapper/goroutine.go",
    "line":18
  },
  "msg":"panic occurred in goroutine",
  "launcher_run_id":"01HWQNTJJ3RET2810PM254GV7H",
  "component":"run_group",
  "err":"runtime error: index out of range [3] with length 0"
}
{
  "time":"2024-04-30T14:17:34.983632Z",
  "level":"ERROR",
  "source":{
    "function":"github.com/kolide/launcher/ee/gowrapper.Go.func1.1",
    "file":"/Users/rebeccamahany-horton/Repos/launcher/ee/gowrapper/goroutine.go",
    "line":23
  },
  "msg":"panic stack trace",
  "launcher_run_id":"01HWQNTJJ3RET2810PM254GV7H",
  "component":"run_group",
[...]

PR #1693, opened 2024-04-25
Wrap main and remove os.Exit calls so all deferred functions will execute

RebeccaMahany:
Relates to https://github.com/kolide/launcher/pull/1692

When we call os.Exit, deferred functions will not execute, which means e.g. the event log writer on windows won't get closed. For a more graceful shutdown, we wrap main as `runMain` and disallow `os.Exit` calls outside of the new `main`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,6 +37,8 @@ linters-settings:
         msg: use ee/allowedcmd functions instead
       - p: ^os\.Exit.*$
         msg: do not use os.Exit so that launcher can shut down gracefully
+      - p: ^logutil\.Fatal.*$
+        msg: do not use logutil.Fatal so that launcher can shut down gracefully
   sloglint:
     kv-only: true
     context-only: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,6 +35,8 @@ linters-settings:
     forbid:
       - p: ^exec\.Command.*$
         msg: use ee/allowedcmd functions instead
+      - p: ^os\.Exit.*$
+        msg: do not use os.Exit so that launcher can shut down gracefully
   sloglint:
     kv-only: true
     context-only: true
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -78,7 +78,7 @@ func runMain() error {
 	// handle that argument. Fall-back to running launcher
 	if len(os.Args) > 1 && !strings.HasPrefix(os.Args[1], `-`) {
 		if err := runSubcommands(systemSlogger); err != nil {
-			systemSlogger.Log(ctx, slog.LevelInfo,
+			systemSlogger.Log(ctx, slog.LevelError,
 				"running with positional args",
 				"err", err,
 			)
@@ -92,7 +92,10 @@ func runMain() error {
 		if launcher.IsInfoCmd(err) {
 			return nil
 		}
-		level.Info(logger).Log("err", err)
+		systemSlogger.Log(ctx, slog.LevelError,
+			"could not parse options",
+			"err", err,
+		)
 		return fmt.Errorf("parsing options: %w", err)
 	}
 
@@ -203,6 +206,10 @@ func runNewerLauncherIfAvailable(ctx context.Context, slogger *slog.Logger) erro
 		// Fall back to legacy autoupdate library
 		newerBinary, err = autoupdate.FindNewestSelf(ctx)
 		if err != nil {
+			slogger.Log(ctx, slog.LevelError,
+				"could not check out latest launcher from legacy autoupdate library",
+				"err", err,
+			)
 			return nil
 		}
 	}
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -27,17 +27,17 @@ import (
 )
 
 func main() {
-	if err := runMain(); err != nil {
-		os.Exit(1) //nolint:forbidigo // Our only allowed usages of os.Exit are in this function
-	}
-	os.Exit(0) //nolint:forbidigo // Our only allowed usages of os.Exit are in this function
+	os.Exit(runMain()) //nolint:forbidigo // Our only allowed usage of os.Exit is in this function
 }
 
-func runMain() error {
+// runMain runs launcher main -- selecting the appropriate subcommand and running that (if provided),
+// or running `runLauncher`. We wrap it so that all our deferred calls will execute before we call
+// `os.Exit` in `main()`.
+func runMain() int {
 	systemSlogger, logCloser, err := multislogger.SystemSlogger()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "error creating system logger: %v\n", err)
-		return fmt.Errorf("creating system logger: %w", err)
+		return 1
 	}
 	defer logCloser.Close()
 
@@ -70,7 +70,11 @@ func runMain() error {
 	// good way to pass it through the flags.
 	if !env.Bool("LAUNCHER_SKIP_UPDATES", false) && !inBuildDir {
 		if err := runNewerLauncherIfAvailable(ctx, systemSlogger.Logger); err != nil {
-			return fmt.Errorf("running newer version of launcher: %w", err)
+			systemSlogger.Log(ctx, slog.LevelInfo,
+				"could not run newer version of launcher",
+				"err", err,
+			)
+			return 1
 		}
 	}
 
@@ -82,21 +86,21 @@ func runMain() error {
 				"running with positional args",
 				"err", err,
 			)
-			return fmt.Errorf("running with positional args: %w", err)
+			return 1
 		}
-		return nil
+		return 0
 	}
 
 	opts, err := launcher.ParseOptions("", os.Args[1:])
 	if err != nil {
 		if launcher.IsInfoCmd(err) {
-			return nil
+			return 0
 		}
 		systemSlogger.Log(ctx, slog.LevelError,
 			"could not parse options",
 			"err", err,
 		)
-		return fmt.Errorf("parsing options: %w", err)
+		return 0
 	}
 
 	// recreate the logger with  the appropriate level.
@@ -142,14 +146,14 @@ func runMain() error {
 	if err := runLauncher(ctx, cancel, slogger, systemSlogger, opts); err != nil {
 		if !tuf.IsLauncherReloadNeededErr(err) {
 			level.Debug(logger).Log("msg", "run launcher", "stack", fmt.Sprintf("%+v", err))
-			return fmt.Errorf("running launcher: %w", err)
+			return 1
 		}
 		level.Debug(logger).Log("msg", "runLauncher exited to run newer version of launcher", "err", err.Error())
 		if err := runNewerLauncherIfAvailable(ctx, slogger.Logger); err != nil {
-			return fmt.Errorf("running newer version of launcher: %w", err)
+			return 1
 		}
 	}
-	return nil
+	return 0
 }
 
 func runSubcommands(systemMultiSlogger *multislogger.MultiSlogger) error {
@@ -190,7 +194,6 @@ func runSubcommands(systemMultiSlogger *multislogger.MultiSlogger) error {
 	}
 
 	return nil
-
 }
 
 // runNewerLauncherIfAvailable checks the autoupdate library for a newer version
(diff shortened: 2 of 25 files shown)


cmd/launcher/main.go
```diff
@@ -71,13 +78,16 @@ func main() {
 		if err := runSubcommands(systemSlogger); err != nil {
 			logutil.Fatal(logger, "err", fmt.Errorf("run with positional args: %w", err))
```
James-Pickett, line 79: should probably remove all the `logutil.Fatal()` as well since they call `os.Exit(1)`

cmd/launcher/svc.go
```diff
@@ -4,20 +4,15 @@
...
+	"errors"
 
 	"github.com/kolide/launcher/pkg/log/multislogger"
 )
 
 func runWindowsSvc(_ *multislogger.MultiSlogger, _ []string) error {
-	fmt.Println("This isn't windows")
-	os.Exit(1)
-	return nil
+	return errors.New("This isn't windows")
```
James-Pickett, line 13: nit: i prefer our usual all lower case, minimal punctuation error messages
```suggestion
	return errors.New("this is not windows")
```

cmd/launcher/svc.go
```diff
@@ -4,20 +4,15 @@
...
 }
 
 func runWindowsSvcForeground(_ *multislogger.MultiSlogger, _ []string) error {
-	fmt.Println("This isn't windows")
-	os.Exit(1)
-	return nil
+	return errors.New("This isn't windows")
```
James-Pickett, line 17: same as above, prefer lower case, minimal punctuation

cmd/launcher/main.go
RebeccaMahany: Oh yup absolutely, thank you!

James-Pickett (approved):

cmd/launcher/main.go
```diff
@@ -62,22 +69,31 @@ func main() {
 	// fork-bombing itself. This is an ENV, because there's no
 	// good way to pass it through the flags.
 	if !env.Bool("LAUNCHER_SKIP_UPDATES", false) && !inBuildDir {
-		runNewerLauncherIfAvailable(ctx, systemSlogger.Logger)
+		if err := runNewerLauncherIfAvailable(ctx, systemSlogger.Logger); err != nil {
+			return fmt.Errorf("running newer version of launcher: %w", err)
```
directionless, line 73: Is there a place we can log this?

cmd/launcher/main.go
```diff
@@ -62,22 +69,31 @@ func main() {
...
 	}
 
 	opts, err := launcher.ParseOptions("", os.Args[1:])
 	if err != nil {
+		if launcher.IsInfoCmd(err) {
+			return nil
+		}
 		level.Info(logger).Log("err", err)
-		os.Exit(1)
+		return fmt.Errorf("parsing options: %w", err)
```
directionless, line 96: Same!

directionless (commented):
I think it's the right shape, but I wonder what we can do with these errors. 

`main` doesn't have a logger, and stdout may not go anywhere. I wonder about a roundabout thing where `runMain` returns a slogger, and and error, and if the slogger isn't nil it can log the error... Or we can make sure all the error returns log first.

cmd/launcher/main.go
RebeccaMahany: Almost all the errors in `runNewerLauncherIfAvailable` are already logged, but I see one that's not -- I'll get that updated 🙂

directionless: Mostly it feels weird to return errors if we're just going to throw them away. So I think something is slightlt weird in the runLauncher idiom here

RebeccaMahany: I thought about returning an exit code int instead -- does that feel like a cleaner return?

RebeccaMahany: (I'd decided against it originally because returning an error felt more idiomatically Go, but 🤷‍♀️ )

directionless: Might be cleaner to return an exit code -- then it's more clear that the usual go pattern of returning an error won't work

RebeccaMahany: That makes sense, I'll update and see how it looks!

RebeccaMahany: Yup, it's way better, thank you for the suggestion!

.golangci.yml
```diff
@@ -35,6 +35,8 @@ linters-settings:
     forbid:
       - p: ^exec\.Command.*$
         msg: use ee/allowedcmd functions instead
+      - p: ^os\.Exit.*$
```
directionless, line 38: Nice. You could add panic as well

cmd/launcher/main.go
```diff
@@ -62,22 +69,38 @@ func main() {
...
+		return 0
 	}
 
 	opts, err := launcher.ParseOptions("", os.Args[1:])
 	if err != nil {
-		level.Info(logger).Log("err", err)
-		os.Exit(1)
+		if launcher.IsInfoCmd(err) {
+			return 0
+		}
```
directionless, line 98: I'd take it, but would be be cleaner if we handled the InfoCmd in the main loop, instead of `launcher.ParseOptions`?

cmd/launcher/svc_windows.go
```diff
@@ -231,6 +232,15 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 					"change_request", fmt.Sprintf("%+v", c),
 				)
 			}
+		case <-runLauncherResults:
```
directionless, line 235: _This_ is interesting. I wonder if it's a fix for the occasional timeout issues we see from service manager.

directionless (approved):
I didn't test this, but it looks okay

.golangci.yml
RebeccaMahany: Oh sure, let me take a look! I was also waffling on adding `logutil.Fatal` -- I only had to remove one and we don't have any others and I think with slogger we probably won't add more, so maybe not useful, but couldn't hurt

cmd/launcher/main.go
RebeccaMahany: I'm not 100% sure I understand the suggestion.

This handles `launcher --version` (or `launcher --dev_help`), and then above (on line ~84ish, in the call to `runSubcommands`) we handle `launcher version` separately. I don't know if there's a clean way to combine those two cases?

.golangci.yml
RebeccaMahany: We have a couple `panic`s, largely in dataflattentable, that are a little bit complicated to pull out. I am going to punt this to a follow-up PR, if that's ok!

cmd/launcher/main.go
directionless: I'm not deep in the code, so this could be offbase. But maybe either:
1. `--dev_help` and `--version` become boolean flags, get returned and then we detect them and do the thing and exit?
2. they are not flags, and parsed as args?

RebeccaMahany: I think 1 isn't really different what than what we currently do -- just a slightly different set of return args for the `launcher.ParseOptions` function. `--dev_help` and `--version` are already boolean flags, and if they're set then we return the special `InfoCmd` error and handle it here.

My sense is that the implementation for 2 would feel a little unexpected, but maybe no more so than the current implementation is. The difficulty is mostly "we want to treat these two special flags as subcommands", and it's probably weird either way, regardless of whether we do it in the subcommand parsing section or we do it in the flag parsing section.

cmd/launcher/svc_windows.go
RebeccaMahany: Unfortunately this is the part that is the trickiest and that I'm the least certain about 😞 

Basically, no matter which option I try, when we exit the service like this, the service manager won't restart it. I'm not sure why.

RebeccaMahany: Okay, I think I figured it out. Basically when launcher shuts down and we want service manager to restart it, we should NOT send `service stopped` to the status changes channel. Prior to this PR, we were sending that status update -- but now my best guess is that because we sent `service stopped` to the status changes channel and _then_ immediately called `os.Exit`, that change didn't actually have time to propagate through.

zackattack01: I think this change makes sense but per the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winsvc/ns-winsvc-service_failure_actions_flag) we should be able to do one or the other now and still get the service control manager restart (e.g. either set non-zero exit code _or_ don't set service stopped state). So just in the interest of making sure I understand here - if you set the service state to stopped, and then return `windows.ERROR_EXCEPTION_IN_SERVICE`, we don't see a restart?

zackattack01: I'd expect based on the behavior you're describing that the exit code is ending up as `ERROR_SUCCESS (0)`, _and_ the service is stopped from service manager perspective

RebeccaMahany: > So just in the interest of making sure I understand here - if you set the service state to stopped, and then return windows.ERROR_EXCEPTION_IN_SERVICE, we don't see a restart?

Correct!

> I'd expect based on the behavior you're describing that the exit code is ending up as ERROR_SUCCESS (0), and the service is stopped from service manager perspective

That's largely what I was seeing before making the change in https://github.com/kolide/launcher/pull/1693/commits/63c2f1a02e9fce5dc157c6817424b903c97ebf9c -- no matter what exit code I returned, if I set the service stopped state, the service would be stopped and remain stopped (never restarted), and querying via `Get-CimInstance` would show `ExitCode` 0 as well.

.golangci.yml
RebeccaMahany: PR here: https://github.com/kolide/launcher/pull/1697

Based on this branch, so the diff will look cleaner after this is merged and that is rebased.

cmd/launcher/svc_windows.go
zackattack01: Gotcha, well regardless of what is masking that exit code to service manager - not setting the stopped state seems like a safer route towards ensuring restart anyway

cmd/launcher/main.go
directionless: Oh. I see why this is unclear. In (1) I mean returning it as a bool (or enum) in launcher.Options, instead of handling the the run there. But that's a bit of a deeper refactor, though possible a cleaner outcome. 🤷 I'd take this

cmd/launcher/svc_windows.go
```diff
@@ -231,6 +227,13 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
...
+			w.systemSlogger.Log(ctx, slog.LevelInfo,
+				"shutting down service after runLauncher exited",
+			)
+			// We don't want to tell the service manager that we've stopped on purpose,
+			// so that the service manager will restart launcher correctly.
+			return false, uint32(windows.ERROR_EXCEPTION_IN_SERVICE)
```
directionless, line 236: Love the comment. What let you to chose `ERROR_SERVICE_SPECIFIC_ERROR`?

directionless (approved):
I think I'm game. I don't think I was that confident when I wrote service manager stuff, so this feels like a reasonable upgrade.

cmd/launcher/svc_windows.go
RebeccaMahany: I did `windows.ERROR_EXCEPTION_IN_SERVICE` because that's what the windows service code defaults to (https://github.com/golang/sys/blob/master/windows/svc/service.go#L265). I took a look thru the [options](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--1000-1299-) and also considered `ERROR_PROCESS_ABORTED` and `ERROR_SERVICE_SPECIFIC_ERROR`. I didn't do a service-specific error code because we don't see or handle that error code (even if I set `ssec` to true) so it didn't seem worth defining launcher-specific codes. I figured the precedent was good enough of a reason to go with `ERROR_EXCEPTION_IN_SERVICE`.

I'll update the comment to include this context!

directionless (approved):

cmd/launcher/main.go
RebeccaMahany: Do you think that refactor might be a good first issue? I think it'd be fairly simple to test appropriately, and the refactor would be limited in scope. I could write something up and file it if so.

directionless: That's a good call. Yes.

Issue #1321, opened 2023-08-30
Launcher should report errors from all exec tables


directionless:
We have a couple places we create tables by parsing the output from execs. Today, if there are errors, we log it, and `continue`. This isn't wrong, but it does make it hard to understand what the error is, and how to fix it. I think we should start including the errors. Probably as a new attribute -- `error`

I'm thinking of:
- https://github.com/kolide/launcher/blob/main/pkg/osquery/tables/dataflattentable/exec.go#L39
- https://github.com/kolide/launcher/blob/main/pkg/osquery/tables/dataflattentable/exec_and_parse.go#L59

directionless:
I think this really means pushing the dataflatten stuff to use `tablehelpers.Exec`

directionless:
I think this had been blocked on adding `WithUid` functionality to our exec helper, and that's present now. 🎉

directionless:
We should evaulate it in other places. Either exclude, or move to the helper, depending

RebeccaMahany:
I think most of these have moved over to tablehelpers.Run -- https://github.com/kolide/launcher/blob/main/pkg/osquery/table/mdm.go and maybe the windowsupdate might be all that's left.

PR #1696, opened 2024-04-26
tables use tablehelpers.run

James-Pickett:
relates to https://github.com/kolide/launcher/issues/1321

Adds helper functions to enforce timeout and simplify creating / execing commands. Follow on PR will add linting.

```go
func RunSimple(ctx context.Context, slogger *slog.Logger, timeoutSeconds int, cmd allowedcmd.AllowedCommand, args []string, opts ...ExecOps)

func Run(ctx context.Context, slogger *slog.Logger, timeoutSeconds int, execCmd allowedcmd.AllowedCommand, args []string, stdout io.Writer, stderr io.Writer, opts ...ExecOps) error
```


pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -39,44 +38,35 @@ func (t *touchIDSystemConfigTable) generate(ctx context.Context, queryContext ta
 	var results []map[string]string
 	var touchIDCompatible, secureEnclaveCPU, touchIDEnabled, touchIDUnlock string
 
-	// Read the security chip from system_profiler
-	var stdout bytes.Buffer
-	cmd, err := allowedcmd.SystemProfiler(ctx, "SPiBridgeDataType")
```
directionless, line 44: There's a `context.WithTimeout` we can probably get rid of

directionless (commented):
love it

RebeccaMahany (approved):
LGTM, just one small/optional question!

pkg/osquery/table/touchid_system_darwin.go
```diff
@@ -33,50 +31,38 @@ type touchIDSystemConfigTable struct {
...
-	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
-	defer cancel()
-
 	var results []map[string]string
 	var touchIDCompatible, secureEnclaveCPU, touchIDEnabled, touchIDUnlock string
 
-	// Read the security chip from system_profiler
-	var stdout bytes.Buffer
-	cmd, err := allowedcmd.SystemProfiler(ctx, "SPiBridgeDataType")
+	stdout, err := tablehelpers.Exec(ctx, t.slogger, 15, allowedcmd.SystemProfiler, []string{"SPiBridgeDataType"}, false)
```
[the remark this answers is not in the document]
James-Pickett: woops.... I just missed using the same timeout and chose an arbitrary number, I'll update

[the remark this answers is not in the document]
directionless: The old 10s was also arbitrary.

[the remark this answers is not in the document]
directionless: If we want a default, we should have a default and drop it from the signature. Add it to Options. But I'm not sure yet

ee/tables/dism_default_associations/dism_default_associations.go
```diff
@@ -1,7 +1,7 @@
 //go:build windows
 // +build windows
 
-package dsim_default_associations
+package dism_default_associations
```
James-Pickett, line 4: I corrected this spelling, then later remembered Becca having the same issue with it. Should I just put it back?

ee/tables/firmwarepasswd/firmwarepasswd.go
```diff
@@ -90,42 +90,29 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
-func (t *Table) runFirmwarepasswd(ctx context.Context, subcommand string, output *bytes.Buffer) error {
-	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
-	defer cancel()
-
-	cmd, err := allowedcmd.Firmwarepasswd(ctx, subcommand)
-	if err != nil {
-		return fmt.Errorf("creating command: %w", err)
-	}
-
+func (t *Table) runFirmwarepasswd(ctx context.Context, subcommand string) ([]byte, error) {
```
James-Pickett, line 93: this is not quite the same as it was, should probably leave it as close as possible for a large refactor

ee/tables/tablehelpers/exec.go
```diff
@@ -19,66 +20,80 @@ import (
...
+// It enforces a timeout, logs, traces, and returns the stdout as a byte slice.
+// It discards stderr.
 //
-// `possibleBins` can be either a list of command names, or a list of paths to commands.
-// Where reasonable, `possibleBins` should be command names only, so that we can perform
-// lookup against PATH.
-func Exec(ctx context.Context, slogger *slog.Logger, timeoutSeconds int, execCmd allowedcmd.AllowedCommand, args []string, includeStderr bool, opts ...ExecOps) ([]byte, error) {
+// This is not suitable for high performance work -- it allocates new buffers each time
+// Use Run() for more control over the stdout and stderr streams.
+func RunSimple(ctx context.Context, slogger *slog.Logger, timeoutSeconds int, cmd allowedcmd.AllowedCommand, args []string, opts ...ExecOps) ([]byte, error) {
```
James-Pickett, line 43: still not 100% sold on the functions .... wonder if we should make an option for `WithStderr(w io.Writer)`

RebeccaMahany (dismissed):

directionless (approved):
I think this looks good. I'm not sure how to test it, aside from checking a lot of tables.

ee/tables/tablehelpers/exec.go
directionless: TBH I'm not 100% sure either. But I think this is a great step, and we can refactor if it feels wrong.

directionless (approved):
Still seems reasonable. Merge and iterate?

PR #1697, opened 2024-04-29
Forbid use of panic inside launcher

RebeccaMahany:
image

### DIFF
--- a/pkg/osquery/table/corefoundation_helpers_darwin.go
+++ b/pkg/osquery/table/corefoundation_helpers_darwin.go
@@ -9,7 +9,6 @@ import (
 	"C"
 )
 import (
-	"fmt"
 	"unsafe"
 )
 
@@ -82,6 +81,6 @@ func goValueFromCFPlistRef(ref C.CFPropertyListRef) interface{} {
 	case C.CFStringGetTypeID():
 		return goString(C.CFStringRef(ref))
 	default:
-		panic(fmt.Sprintf("unknown CF type id %v", typeID))
+		return "Unknown"
 	}
 }
--- a/pkg/service/client_jsonrpc.go
+++ b/pkg/service/client_jsonrpc.go
@@ -24,12 +24,9 @@ import (
 func forceNoChunkedEncoding(ctx context.Context, r *http.Request) context.Context {
 	r.TransferEncoding = []string{"identity"}
 
-	// read the body, set the content legth, and leave a new ReadCloser in Body
+	// read the body, set the content length, and leave a new ReadCloser in Body
 	bodyBuf := &bytes.Buffer{}
-	bodyReadBytes, err := io.Copy(bodyBuf, r.Body)
-	if err != nil {
-		panic(err)
-	}
+	bodyReadBytes, _ := io.Copy(bodyBuf, r.Body)
 	r.Body.Close()
 	r.ContentLength = bodyReadBytes
 	r.Body = io.NopCloser(bodyBuf)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -39,6 +39,8 @@ linters-settings:
         msg: do not use os.Exit so that launcher can shut down gracefully
       - p: ^logutil\.Fatal.*$
         msg: do not use logutil.Fatal so that launcher can shut down gracefully
+      - p: ^panic.*$
+        msg: do not use panic so that launcher can shut down gracefully
   sloglint:
     kv-only: true
     context-only: true
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,6 +37,8 @@ linters-settings:
         msg: use ee/allowedcmd functions instead
       - p: ^os\.Exit.*$
         msg: do not use os.Exit so that launcher can shut down gracefully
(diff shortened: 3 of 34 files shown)


ee/dataflatten/jsonl.go
```diff
@@ -13,11 +14,21 @@ func JsonlFile(file string, opts ...FlattenOpts) ([]Row, error) {
...
+
+	rawdata, err := io.ReadAll(f)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read JSON: %w", err)
+	}
+
+	return flattenJsonl(rawdata, opts...)
+}
+
+func Jsonl(rawdata []byte, opts ...FlattenOpts) ([]Row, error) {
```
RebeccaMahany, line 26: Had to update this to make `Jsonl` signature compliant with `dataflatten.DataFunc` (`func(data []byte, opts ...FlattenOpts) ([]Row, error)`).

ee/dataflatten/jwt.go
```diff
@@ -31,9 +24,20 @@ func flattenJWT(path string, opts ...FlattenOpts) ([]Row, error) {
 		return nil, fmt.Errorf("unable to read JWT: %w", err)
 	}
 
+	return flattenJWT(tokenRaw, opts...)
+}
+
+func JWT(rawdata []byte, opts ...FlattenOpts) ([]Row, error) {
```
RebeccaMahany, line 30: Added this so that `dataflattentable.DataSourceType` `JWTType` was fully implemented with both a file and a data function -- it is unused.

pkg/osquery/table/corefoundation_helpers_darwin.go
```diff
@@ -82,6 +81,6 @@ func goValueFromCFPlistRef(ref C.CFPropertyListRef) interface{} {
 	case C.CFStringGetTypeID():
 		return goString(C.CFStringRef(ref))
 	default:
-		panic(fmt.Sprintf("unknown CF type id %v", typeID))
+		return "Unknown"
```
RebeccaMahany, line 84: This seemed safe enough -- this function and the one that calls it appear to be wholly unused.

pkg/service/client_jsonrpc.go
```diff
@@ -24,12 +24,9 @@ import (
...
 	r.TransferEncoding = []string{"identity"}
 
-	// read the body, set the content legth, and leave a new ReadCloser in Body
+	// read the body, set the content length, and leave a new ReadCloser in Body
 	bodyBuf := &bytes.Buffer{}
-	bodyReadBytes, err := io.Copy(bodyBuf, r.Body)
-	if err != nil {
-		panic(err)
-	}
+	bodyReadBytes, _ := io.Copy(bodyBuf, r.Body)
```
RebeccaMahany, line 29: I don't know if there's anything better we can do here. We don't have access to a logger and it's not trivial to get access.

James-Pickett: I agree ... what else can we do here 🤷

James-Pickett (approved):
LGTM!

pkg/service/client_jsonrpc.go
directionless: It's worth a comment about why we're discarding it. But yeah 🤷

pkg/osquery/table/corefoundation_helpers_darwin.go
directionless: `""` might be better? Sentinel values are hard. Or `__unknown__`

ee/dataflatten/jsonl.go
directionless: I feel sad that past me didn't use io.Reader

ee/dataflatten/jsonl.go
```diff
@@ -13,11 +14,21 @@ func JsonlFile(file string, opts ...FlattenOpts) ([]Row, error) {
 		return nil, err
 	}
 	defer f.Close()
-	return Jsonl(f, opts...)
+
+	rawdata, err := io.ReadAll(f)
```
directionless, line 18: `rawdata, err := os.ReadFile(file)` maybe?

directionless (approved):
I _think_ this is probably fine. I can't really track all the changes, but I think we have okay tests here. So if we can still use the tables, there won't be subtle errors. 

I think it does point to some tech debt in how all the flatten tables come together. There's some work to be done in refactoring. But not in this pr

pkg/osquery/table/corefoundation_helpers_darwin.go
RebeccaMahany: "Unknown" is what we use directly above the switch (line 74) so I thought that was probably appropriate enough

ee/dataflatten/jsonl.go
RebeccaMahany: Updated!

pkg/service/client_jsonrpc.go
RebeccaMahany: Added a comment

directionless (approved):

PR #1700, opened 2024-04-30
dont interact with desktop client while in standby

James-Pickett:


ee/desktop/runner/runner.go
```diff
@@ -531,6 +531,13 @@ func (r *DesktopUsersProcessesRunner) writeDefaultMenuTemplateFile() {
 }
 
 func (r *DesktopUsersProcessesRunner) runConsoleUserDesktop() error {
+	if r.knapsack.InModernStandby() {
```
James-Pickett, line 534: should we skip here or in the execute loop?

RebeccaMahany: Here makes sense! Looking here versus execute, I think when we're in modern standby we would still want to refresh menu data (https://github.com/kolide/launcher/blob/42f20a7303ca8d66ae445db673201019b4129edc/ee/desktop/runner/runner.go#L235) but maybe _not_ tell clients to pick up the new data (https://github.com/kolide/launcher/blob/42f20a7303ca8d66ae445db673201019b4129edc/ee/desktop/runner/runner.go#L439) -- so putting the `InModernStandby` check into these individual functions makes sense to me.

directionless: We should _not_ tell the clients in standby.

I forget how menu refresh works, if we skip it, will we miss things, or will we pick things up?

James-Pickett: we will still pick up the menu changes and can still have the most updated menu available when we come out of standby

the data to update the menu comes in via the control server update

zackattack01 (approved):

directionless (approved):
let's try!

PR #1679, opened 2024-04-10
[Do not merge] Remove the rest of pkg/autoupdate

RebeccaMahany:
Building on https://github.com/kolide/launcher/pull/1677 -- removes the rest of pkg/autoupdate, i.e. FindNewest. This means new versions of launcher will only look for updates in the new library location.

Holding off on this until v1.6.4 is released to stable.

PR #1701, opened 2024-05-01
Skip tests legacy autoupdate tests in CI

RebeccaMahany:
1. We are not modifying the legacy autoupdate code anymore, so we don't need the tests to validate against regressions
2. After the next launcher release, we are going to remove the legacy autoupdate code altogether (https://github.com/kolide/launcher/pull/1679)
3. These tests are some of our flakiest


James-Pickett (approved):
👏 🎉

zackattack01 (approved):

PR #1712, opened 2024-05-08
uninstall via extension / jsonrpc / grpc

James-Pickett:
* updates kolide svc functions to look for `disable_device` field in jsonRpc / gRPC responses and return a `ErrDeviceDisabled` if found
* updates extension to check for `ErrDeviceDisabled` when it receives and error and perform uninstall if found


pkg/service/publish_logs.go
```diff
@@ -102,6 +102,9 @@ func encodeGRPCLogCollection(_ context.Context, request interface{}) (interface{
 func decodeGRPCPublishLogsResponse(_ context.Context, grpcReq interface{}) (interface{}, error) {
 	req := grpcReq.(*pb.AgentApiResponse)
 	return publishLogsResponse{
+		jsonRpcResponse: jsonRpcResponse{
+			DisableDevice: req.DisableDevice,
+		},
 		Message:     req.Message,
```
directionless, line 108: Naive question -- why are the other things really obvious looking struct members, but this one is a nested thing?

James-Pickett: this is because the jsonRpc implementation is doing a standard json unmarshalling where as the grpc implementation is taking the auto-generated protobuf struct`pb.AgentApiResponse` and converting it the struct type `publishLogsResponse` via manually setting the fields

we could also probably just marshall the protobuf to json then unmarshall it into the `publishLogsResponse`

pkg/service/disable_device_test.go
```diff
@@ -0,0 +1,179 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// set up JSON-RPC test server
+			jsonRpcServerResult := `{"disable_device": false}`
```
James-Pickett, line 72: I think the service package became a little frankensteiny over the years. Mixing jsonRPC and gRPC implementations. I'm tempted to split them up and perhaps have a common `types` package. Not sure if it's worth the effort at this time since we're not really using the grpc portion.

pkg/service/publish_logs.go
directionless: Hrm. Caveat, I'm not sure what correct is...

My hunch is that we should do the same thing for jsonrpc and grpc, but I'm not sure if we should improve the conversion or just go the manual route. 

At some point, I think we should level up all the API surfaces via some kind of API specification, but that feels a bit out of scope for this.

pkg/service/disable_device_test.go
directionless: I think go-kit has a recommendation for how to do this. It's core design is all about splitting up transport implementation from protocol definition.

pkg/osquery/extension.go
```diff
@@ -449,13 +450,20 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
 	// If no cached node key, enroll for new node key
 	// note that we set invalid two ways. Via the return, _or_ via isNodeInvaliderr
 	keyString, invalid, err := e.serviceClient.RequestEnrollment(ctx, enrollSecret, identifier, enrollDetails)
-	if isNodeInvalidErr(err) {
+
+	switch {
+	case errors.Is(err, service.ErrDeviceDisabled{}):
```
directionless, line 455: hahahaha. This really points at how this method signature was limiting. This feels like a reasonable workaround, but definitely points to needing to refresh this code.

pkg/osquery/extension.go
```diff
@@ -449,13 +450,20 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+		uninstall.Uninstall(ctx, e.knapsack, true)
```
directionless, line 456: Should we return here? Falling through looks wrong

pkg/osquery/extension.go
```diff
@@ -554,10 +562,17 @@ var reenrollmentInvalidErr = errors.New("enrollment invalid, reenrollment invali
 // Helper to allow for a single attempt at re-enrollment
 func (e *Extension) generateConfigsWithReenroll(ctx context.Context, reenroll bool) (string, error) {
 	config, invalid, err := e.serviceClient.RequestConfig(ctx, e.NodeKey)
-	if isNodeInvalidErr(err) {
+	switch {
+	case errors.Is(err, service.ErrDeviceDisabled{}):
+		uninstall.Uninstall(ctx, e.knapsack, true)
```
directionless, line 567: same -- should we return?

pkg/osquery/extension.go
RebeccaMahany: +1 -- I think helpful to add a comment here stating that uninstall will stop launcher immediately so we don't need to do anything further? And maybe do the early return anyway, in case uninstall fails?

directionless: I'm not sure it needs much comment, but it feels like the node-invalid flow. It's a different action.

(I almost wonder if middleware is a cleaner approach, but I don't want that can of worms)

pkg/osquery/extension.go
James-Pickett: added return

pkg/osquery/extension.go
James-Pickett: added return

RebeccaMahany (approved):

zackattack01 (approved):

PR #1713, opened 2024-05-08
Add brew no auto update to brew execs

Micah-Kolide:
Multiple homebrew commands will invoke `brew update` before running the command that's called. This can cause unpredictable data being returned. Adding `HOMEBREW_NO_AUTO_UPDATE=1` to the exec will ensure this doesn't happen.

### DIFF
--- a/ee/allowedcmd/cmd_darwin.go
+++ b/ee/allowedcmd/cmd_darwin.go
@@ -21,7 +21,14 @@ func Bputil(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 }
 
 func Brew(ctx context.Context, arg ...string) (*exec.Cmd, error) {
-	return validatedCommand(ctx, "HOMEBREW_NO_AUTO_UPDATE=1 /opt/homebrew/bin/brew", arg...)
+	cmd, err := validatedCommand(ctx, "/opt/homebrew/bin/brew", arg...)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


ee/allowedcmd/cmd_darwin.go
```diff
@@ -21,7 +21,7 @@ func Bputil(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 }
 
 func Brew(ctx context.Context, arg ...string) (*exec.Cmd, error) {
-	return validatedCommand(ctx, "/opt/homebrew/bin/brew", arg...)
+	return validatedCommand(ctx, "HOMEBREW_NO_AUTO_UPDATE=1 /opt/homebrew/bin/brew", arg...)
```
directionless, line 24: I'm not sure this will behave correctly. We should, instead, work with the ENV directly.

directionless: That's the usual way. We could also use `WithAppendEnv` on the wrapper

Micah-Kolide:
Of course it was just simply setting BOTH pwd and home, not only one of them, which I would've had figured out two days ago if I just tested that as well instead of thinking that wasn't the issue since setting one of them didn't work...

I figure there wouldn't really be a case where we want to keep root's context of PWD and HOME over the running user's, so that's why I decided to throw it in the `run_as_user` exec option instead of making something specific for brew.

Micah-Kolide:
Brew should be fully fixed with this pr... hopefully.

PR #1715, opened 2024-05-13
Add package to query indexeddb database and parse results

RebeccaMahany:
This is a partially-working implementation. It allows for querying an indexeddb database, given the location, the database name, and the object store; it parses and returns all objects in that store. The parsing works for the store I care about at the moment, but does _not_ work in all cases. I have noted unknowns and limitations where I've encountered them in the code.

My intention is to merge and iterate.

<details>

<summary>How we query for items by object store</summary>

```mermaid
flowchart TD
    A[QueryIndexeddbObjectStore: example.indexeddb.leveldb, exampleDatabaseName, exampleObjectStoreName] --> B[Open example.indexeddb.leveldb using leveldb package]
    B --> C[Query database ID for exampleDatabaseName]
    C --> |Cannot query for object store ID by name| D[Query for object store ID 'i' and check its name]
    D --> |No match| E[i = i+1] --> D
    D --> |Object store name matches exampleObjectStoreName| F[Construct known key prefix including database ID and object store ID]
    F --> G[Iterate over all items in leveldb]
    G --> |Item's key does not match prefix| H[Do not process item] --> G
    G --> |Item's key matches prefix| I[Deserialize item into map] --> G
    G ---> |No more items| J[Return deserialized items map]
```

</details>

<details>
<summary>Unknowns and limitations</summary>

IndexedDB in general:
* Opening the database will fail if Chrome is also open, so we have to make a copy of the database.
* Sometimes data updates don't appear to get flushed to the database, or aren't available the way that I'm iterating through the keys -- I've seen both duplicate rows, and missing rows, across both macOS and Windows.
  * I don't think the issue is on the leveldb <=> indexeddb layer (i.e. what's implemented in this PR). When I iterate over all leveldb keys, I see that rows are duplicated and/or missing at _that_ layer.
  * I don't think the issue is that the extension is duplicating data to the database or failing to add it to the database, either, because when I view the database using Chrome's dev tools, the data is correct (i.e. not duplicated or missing).
  * My current theory is that the issue lies with goleveldb. I see, for example, that it sometimes thinks my database is corrupt when that doesn't seem to be the case.
  * My other theory is that the data is pending, per [this documentation](https://chromium.googlesource.com/chromium/src/+/master/content/browser/indexed_db/docs/README.md#persisted-vs-pending).

IndexedDB keys:

* I had to append `@1` to the database origin name that I parsed from the filepath. I don't know why.
* The documentation indicates that global keys begin with the prefix `<0, 0, 0>`, but I found they began with `<0, 0, 0, 0>`. Similarly, it seems to indicate that database metadata keys begin with the prefix `<database id, 0, 0>`, but I found they began with `<0, database id, 0, 0>` -- i.e. there's an additional `0` before all these keys. I don't know why.
* I only handle key paths of type string, and keys of type number -- e.g., a key path "id" that has a numerical value.

IndexedDB object deserialization:

* I don't know if there's anything else useful we should be extracting from the header; I'm just ignoring the rest of it besides the version. (The version seems like it could be useful for parsing decisions in the future.)
* I only handled the data types that I encountered in the given store we're interested in. Notably, I didn't add support for utf-8 strings, any other type of number besides int32 and uint32, or any type of array besides a sparse array of objects.
* Sometimes there's an `0x01` and/or an `0x03` at the end of an array's data, before the `tokenEndSparseArray`. I don't know why it's there. I've been discarding it.

</details>

<details>
<summary>Useful references</summary>

* [Coding scheme for IndexedDB keys](https://github.com/chromium/chromium/blob/main/content/browser/indexed_db/docs/leveldb_coding_scheme.md)
* [Implementation of coding scheme for IndexedDB keys](https://github.com/chromium/chromium/blob/master/content/browser/indexed_db/indexed_db_leveldb_coding.cc)
* [Deserializing indexeddb values](https://github.com/v8/v8/blob/master/src/objects/value-serializer.cc)
* [IndexedDB Overview for Database People](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/modules/indexeddb/docs/idb_overview.md)
* [IndexedDB Data Path](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/modules/indexeddb/docs/idb_data_path.md)
* [LevelDB Scopes: Special Transactions for IndexedDB](https://docs.google.com/document/d/16_igCI15Gfzb6UYqeuJTmJPrzEtawz6Y1tVOKNtYgiU/edit#heading=h.7nki9mck5t64)
* [IndexedDB specification](https://w3c.github.io/IndexedDB/)

</details>


ee/indexeddb/comparer.go
```diff
@@ -0,0 +1,33 @@
+package indexeddb
+
+import (
+	"github.com/kolide/goleveldb/leveldb/comparer"
+)
+
+// chromeComparer is a thin wrapper around the default comparer to allow us
+// to use it as `idb_cmp1`.
```
directionless, line 8: Should `idb_cmp1` be meaningful to a reviewer?

ee/indexeddb/indexeddb.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/kolide/goleveldb/leveldb"
+	"github.com/kolide/goleveldb/leveldb/opt"
+	"github.com/kolide/launcher/ee/agent"
+)
+
+// maxObjectStoresToCheck is the number of indices for object stores we will check
+// before declaring failure to find the given object store. We cannot look up
+// object stores by their names, only by their IDs -- so we have to iterate through
+// up to maxObjectStoresToCheck IDs to find the desired store. We assume there are
+// fewer than 100 object stores in a given database.
```
directionless, line 22: Do we know enough to make this a good guess? (We might need more data from the field)

ee/indexeddb/indexeddb.go
```diff
@@ -0,0 +1,184 @@
...
+func QueryIndexeddbObjectStore(dbLocation string, dbName string, objectStoreName string) ([]map[string]any, error) {
+	// If Chrome is open, we won't be able to open the db. So, copy it to a temporary location first.
+	tempDbCopyLocation, err := copyIndexeddb(dbLocation)
+	if err != nil {
+		if tempDbCopyLocation != "" {
+			_ = os.RemoveAll(tempDbCopyLocation)
+		}
+		return nil, fmt.Errorf("unable to copy db: %w", err)
+	}
+	defer os.RemoveAll(tempDbCopyLocation)
```
directionless, line 36: Does this need a check on `tempDbCopyLocation != ""`? (I wouldn't normally think so, but there's one on line31)

ee/indexeddb/comparer.go
```diff
@@ -0,0 +1,33 @@
...
+type chromeComparer struct {
+	defaultComparer comparer.Comparer
+}
+
+func newChromeComparer() *chromeComparer {
```
directionless, line 13: Is every instance of this the same / do we need to create new ones on usage? (Vs making a package variable one)

ee/indexeddb/indexeddb.go
```diff
@@ -0,0 +1,184 @@
...
+	databaseIdRaw, err := db.Get(databaseNameKey, nil)
+	if err != nil {
+		return nil, fmt.Errorf("querying for database id: %w", err)
+	}
+	databaseId, _ := binary.Uvarint(databaseIdRaw)
+
+	// We can't query for the object store ID by its name -- we have to query each ID to get its name,
+	// and check against that. Object store indices start at 1.
+	var objectStoreId uint64
+	for i := 1; i < maxObjectStoresToCheck; i++ {
```
directionless, line 68: I don't care either way, but since index is at one, shouldn't be be `<=` for the max number to to check. Vs `<` if it's the max index. But maybe I'm really saying that the plural "object stores" `maxObjectStoresToCheck` makes me think it's a count, not not a named index.

directionless: Should `i` be a uint64, not an int? (Would it let you drop the casts later)

directionless (approved):
Looks pretty good. I think all my comments are very nit

ee/indexeddb/comparer.go
RebeccaMahany: Ah nope -- it is indexeddb-specific. I will improve the documentation here!

ee/indexeddb/indexeddb.go
RebeccaMahany: The extension that we care about currently has 10 object stores, so 100 felt like a safe number for this particular usecase (plenty of wiggle room if this extension starts to add more object stores).

My guess is that it would be weird/unusual for an extension to have more than 100 object stores, but that could be incorrect.

ee/indexeddb/indexeddb.go
RebeccaMahany: We don't need an additional check here -- we only get here if err == nil, and if err == nil the indexeddb was copied correctly to `tempDbCopyLocation`, so the location won't be an empty string.

The check on line 31 is to handle partial copy failures -- e.g. we made a directory and copied over a couple files, but not all of them, so we can't proceed but would still want to clean up that partial copy.

ee/indexeddb/indexeddb.go
RebeccaMahany: Renamed the const to `maxNumberOfObjectStoresToCheck` -- I think that makes it clearer? Agreed on both counts re: `<=` and making `i` a uint64; updated!

ee/indexeddb/indexeddb.go
RebeccaMahany: Updated the comment to note that we may need to adjust this number upward in the future

zackattack01 (approved):
the comments and PR description docs really helped readability here, thank you!

PR #1720, opened 2024-05-20
Add associated domains to entitlements

RebeccaMahany:
TBD: update `embedded.provisionprofile`

### DIFF
--- a/tools/packaging/entitlements.plist
+++ b/tools/packaging/entitlements.plist
@@ -2,5 +2,9 @@
 <dict>
     <key>com.apple.application-identifier</key>
     <string>X98UFR7HA3.com.kolide.agent</string>
+    <key>com.apple.developer.associated-domains</key>
+    <array>
+        <string>applinks:auth.kolide.com</string>
+    </array>
 </dict>
 </plist>


tools/packaging/entitlements.plist
```diff
@@ -2,5 +2,9 @@
 <dict>
     <key>com.apple.application-identifier</key>
     <string>X98UFR7HA3.com.kolide.agent</string>
+    <key>com.apple.developer.associated-domains</key>
+    <array>
+        <string>applinks:auth.kolide.com</string>
```
directionless, line 7: Can you add `sso.kolide.com` as well?

And I wonder if we should be thinking about how to handle dev here. No good URL for it though

RebeccaMahany: In dev, if you have an ngrok environment you can add the appropriate URL to the array here and build a new app bundle to test against -- that's what I did yesterday to test. (I think we probably can't, or at least shouldn't, add `*.ngrok.io` wholesale here.) I know having to build an app bundle might be an annoying workflow for folks, but hopefully it won't come up too frequently?

directionless: We should not add ngrok here. 

But most of the developers cannot codesign launcher. (though maybe if it's a local build they don't need to?)

I'm content to approve this, and figure out an ngrok story later.

directionless (approved):
Happy to approve, though we may need to iterate if we want to take the target domain to something dedicated.

James-Pickett (approved):

PR #1721, opened 2024-05-21
package-builder: add `--bin_root_dir` flag

tstromberg:
This PR allows the Kolide Launcher to be deployed to Fedora SilverBlue, where /usr is read-only. To build for this platform, pass `--bin_root_dir=/opt` to `./build/package-builder`.

Without this flag, Kolide fails to install on Silverblue with:

```
sudo rpm-ostree install zxkx_thomas-stromberg-0ac07b_kolide-launcher.rpm                                                         
...
Resolving dependencies... done
Relabeling... done
Checking out packages... done
error: Checkout launcher-kolide-k2-1.6.2-1.x86_64: opendir(local): No such file or directory
```

It turns out that on Silverblue, `/usr/local` is a symlink to `../var/usrlocal/`, which makes `rpm-ostree` upset with life.

### DIFF
--- a/cmd/package-builder/package-builder.go
+++ b/cmd/package-builder/package-builder.go
@@ -64,6 +64,11 @@ func runMake(args []string) error {
 			env.String("PACKAGE_VERSION", ""),
 			"the resultant package version. If left blank, auto detection will be attempted",
 		)
+		flBinRootDir = flagset.String(
+			"bin_root_dir",
+			"/usr/local",
+			"the root directory path for the launcher on macOS and Linux",
+		)
 		flOsqueryVersion = flagset.String(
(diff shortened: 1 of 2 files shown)


directionless (approved):
This looks pretty narrow and reasonable, thank you @tstromberg 

It's worth noting that making sure autoupdate works may be another ball of wax. But this is clearly a step.

@RebeccaMahany any objections?

RebeccaMahany:
Seems fine to me!

Setting the `update_directory` flag to an appropriate location (if `/var/kolide-k2/<hostname>/updates` isn't suitable) should solve the autoupdate issue as well.

PR #1722, opened 2024-05-21
package-builder: add podman support via --container_tool

tstromberg:
This allows the launcher to be built on Fedora, which ships with a working podman environment that runs as a regular user (no background daemon required). In comparison, setting up Docker is a pain and gives the user root-level privilege. 

This PR does add the ":Z" flag to container volume mounts, so that the volume permissions allow access from the host filesystem, which the build process assumes. Without this flag, rootless podman encounters an error:

```
could not generate packages: making package: packaging, target linux-systemd-rpm: creating fpm package: /fpm/gems/fpm-1.11.0/lib/fpm/command.rb:444:in `read': Permission denied @ rb_sysopen - /pkgscripts/postinstall (Errno::EACCES)
```

The ":z" mount flag also works, but has a looser security model. This flag is supported by Docker and Podman. 

This PR also changes the container pull location from an ambiguous "kolide/fpm:latest" to the absolute path of ""docker.io/kolide/fpm:latest". This is necessary as Podman is not pre-programmed to pull from "docker.io" by default.


pkg/packagekit/package_fpm.go
```diff
@@ -134,16 +134,16 @@ func PackageFPM(ctx context.Context, w io.Writer, po *PackageOptions, fpmOpts ..
 		fpmCommand = append(fpmCommand, "--before-remove", filepath.Join("/pkgscripts", "prerm"))
 	}
 
-	dockerArgs := []string{
+	args := []string{
 		"run", "--rm",
-		"-v", fmt.Sprintf("%s:/pkgsrc", po.Root),
-		"-v", fmt.Sprintf("%s:/pkgscripts", po.Scripts),
-		"-v", fmt.Sprintf("%s:/out", outputPathDir),
+		"-v", fmt.Sprintf("%s:/pkgsrc:Z", po.Root),
```
directionless, line 139: Hrm... Docker has several warnings about this:

> If you use selinux you can add the z or Z options to modify the selinux label of the host file or directory being mounted into the container. This affects the file or directory on the host machine itself and can have consequences outside of the scope of Docker.
> 
> * The z option indicates that the bind mount content is shared among multiple containers.
> * The Z option indicates that the bind mount content is private and unshared.
>
> Use extreme caution with these options. Bind-mounting a system directory such as /home or /usr with the Z option renders your host machine inoperable and you may need to relabel the host machine files by hand.

I think I understand the why of this, but it makes me kinda unsure.

tstromberg: Interesting. My interpretation of the warning is: don't attach important system directories as Docker volumes as it may allow the Docker volume to manipulate the data.

If it makes you feel better, I can add the `:Z` only for podman users. The reason this works in Docker is that by default, Docker runs as root. podman normally executes as a regular user (rootless mode). Docker also has a rootless mode, but nobody uses it :)

cmd/package-builder/package-builder.go
```diff
@@ -54,6 +54,11 @@ func runMake(args []string) error {
 			false,
 			"enable debug logging",
 		)
+		flContainerTool = flagset.String(
+			"container_tool",
+			"docker",
```
RebeccaMahany, line 59: Just a quick note for the team that if we merge this, we'll need to update our internal tooling to provide this default value since we don't call `package-builder` on the command line.

tstromberg:
Updated the patches. There should be no behavioral changes now unless `--container_tool=podman` is passed in.

pkg/packagekit/package_fpm.go
tstromberg: Done!

directionless (approved):
Seems okay to me.

pkg/packaging/packaging.go
```diff
@@ -315,6 +316,7 @@ func (p *PackageOptions) Build(ctx context.Context, packageWriter io.Writer, tar
 		Title:                    p.Title,
 		Root:                     p.packageRoot,
 		Scripts:                  p.scriptRoot,
+		ContainerTool:            p.ContainerTool,
```
RebeccaMahany, line 319: Quick request, sorry for the late review -- could we add a default value here? e.g. above -- 

```
if p.Title == "" {
	p.Title = fmt.Sprintf("Launcher agent for %s", p.Identifier)
}

if p.ContainerTool == "" {
	p.ContainerTool = "docker"
}
```

That way when we have `cmd := exec.CommandContext(ctx, po.ContainerTool, append(args, fpmCommand...)...)` in `package_fpm` we can be sure that we are passing in "docker" rather than an empty string.

This is in reference to https://github.com/kolide/launcher/pull/1722/files#r1683280222 -- when we do packaging, we don't call package-builder via the command line, so we don't get the helpful default value from `package-builder.go` 🙂

RebeccaMahany (changes requested):

pkg/packaging/packaging.go
directionless: Looks like maintainers have access, so I added it!

RebeccaMahany (approved):

PR #1728, opened 2024-05-30
Add data to startup message

RebeccaMahany:
We want to be able to identify devices better when they send startup messages -- this PR adds status and identifying data.

### DIFF
--- a/ee/control/control.go
+++ b/ee/control/control.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/kolide/kit/version"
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):
nice

directionless (commented):
I could take this, but I wonder if we need a timeout on the ctx

RebeccaMahany:
@directionless I only passed the ctx in to tie the logs together -- it's not used for fetching data from the stores, so a timeout on the ctx wouldn't do anything

PR #1731, opened 2024-05-31
Register the desktop app path with launch services

RebeccaMahany:
Ordinarily, it is not required to explicitly register an application with Launch Services -- this is done automatically on system boot, on new user login, and whenever Finder becomes aware of a new application. However, because our application autoupdates itself in a unique way, I think sometimes Launch Services doesn't learn about the new update quickly enough. This PR adds a call to [LSRegisterURL](https://developer.apple.com/documentation/coreservices/1446350-lsregisterurl?language=objc) to manually perform this registration.

It is acceptable to have multiple app bundle paths registered for the same app in Launch Services. Per the documentation, Launch Services should choose the registered application with the latest version. This is pretty good for our use case, at least for now. However, this doesn't appear to work with development versions the way we expect, and wouldn't necessarily cover the use case where we want to roll back to an earlier version -- so we will probably, at some point, want to unregister older app bundles too.

Useful sections in the [Launch Services](https://developer.apple.com/library/archive/documentation/Carbon/Conceptual/LaunchServicesConcepts/LSCConcepts/LSCConcepts.html#//apple_ref/doc/uid/TP30000999-CH202-TP9) documentation:

* `Application Registration`
* `Preferred Application for a URL` 

This relates to previous universal link support work: https://github.com/kolide/launcher/pull/1727.

### DIFF
--- a/ee/desktop/user/universallink/handler.m
+++ b/ee/desktop/user/universallink/handler.m
@@ -1,0 +1,19 @@
+//go:build darwin
+// +build darwin
+
+#import <Foundation/Foundation.h>
+
+BOOL registerAppBundle(char *cAppBundlePath) {
+    NSString *appBundlePath = [NSString stringWithUTF8String:cAppBundlePath];
+    CFURLRef url = (CFURLRef)[NSURL fileURLWithPath:appBundlePath];
+
+    // Set inUpdate to true to ensure app gets updated
+    OSStatus status = LSRegisterURL((CFURLRef)url, YES);
+
+    if (status != noErr) {
+        NSLog(@"could not register app bundle: LSRegisterURL returned error: %jd", (intmax_t)status);
+        return NO;
+    }
+
+    return YES;
+}
--- a/ee/desktop/user/universallink/handler_darwin.go
+++ b/ee/desktop/user/universallink/handler_darwin.go
@@ -9,10 +9,12 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/kolide/launcher/ee/allowedcmd"
 	"github.com/kolide/launcher/ee/gowrapper"
 	"github.com/kolide/launcher/ee/localserver"
 )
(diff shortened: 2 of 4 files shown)


ee/desktop/user/universallink/handler_darwin.go
```diff
@@ -74,6 +93,32 @@ func (u *universalLinkHandler) Interrupt(_ error) {
...
 
+func register() error {
+	currentExecutable, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("getting current executable: %w", err)
+	}
+
+	// If we're running the originally-installed version of launcher, no need to register;
+	// Launch Services should already know about this application from the install process.
+	if strings.HasPrefix(currentExecutable, "/usr/local") {
```
zackattack01, line 104: any harm in registering anyway? comment is not blocking either way, but if there's no cost to registering I do wonder if there's any way this optimization could bite us down the road that we aren't considering

zackattack01 (approved):

James-Pickett (approved):

ee/desktop/user/universallink/handler.m
```diff
@@ -0,0 +1,19 @@
+//go:build darwin
+// +build darwin
+
+#import <Foundation/Foundation.h>
+
+BOOL registerAppBundle(char *cAppBundlePath) {
```
directionless, line 6: Do we need to free/release anything in the C side?

directionless (commented):
I like this more than the exec.

It will cause potentially duplicate registrations. Do you think there's an issue with that?

ee/desktop/user/universallink/handler.m
RebeccaMahany: The string is freed on 106 in handler_darwin.go -- is there anything else I'm missing?

RebeccaMahany:
@directionless Not worried about multiple registrations! If called for the exact same app bundle path, it will just update the existing registration.

ee/desktop/user/universallink/handler.m
directionless: I don't know enough to know if we need to free `appBundlePath` or `url` or `status`

RebeccaMahany: Ah, no, I don't believe we need to! We don't have ownership of those objects. [This](https://stackoverflow.com/a/8702692) has a clear overview.

directionless: excellent. Thank you!

PR #1732, opened 2024-05-31
update dataflatten tables to parse raw_data in query

James-Pickett:
allow for passing raw data into query for dataflatten tables

```
osquery> select * from kolide_json where raw_data = '[0]';
+---------+--------+-----+-------+-------+------+----------+
| fullkey | parent | key | value | query | path | raw_data |
+---------+--------+-----+-------+-------+------+----------+
| 0       |        | 0   | 0     | *     |      | [0]      |
+---------+--------+-----+-------+-------+------+----------+
osquery> select * from kolide_json where raw_data = '{"hi":"there"}';
+---------+--------+-----+-------+-------+------+----------------+
| fullkey | parent | key | value | query | path | raw_data       |
+---------+--------+-----+-------+-------+------+----------------+
| hi      |        | hi  | there | *     |      | {"hi":"there"} |
+---------+--------+-----+-------+-------+------+----------------+
```

when adding both constraints with an OR, the first one wins unless there is a specific query

```
osquery> select * from kolide_json where path = '/tmp/test.json' or raw_data = '[0]';
+---------+--------+-----+-------+-------+----------------+----------+
| fullkey | parent | key | value | query | path           | raw_data |
+---------+--------+-----+-------+-------+----------------+----------+
| hi      |        | hi  | there | *     | /tmp/test.json |          |
+---------+--------+-----+-------+-------+----------------+----------+
osquery> select * from kolide_json where query = 0 and path = '/tmp/test.json' or raw_data = '[0]';
+---------+--------+-----+-------+-------+------+----------+
| fullkey | parent | key | value | query | path | raw_data |
+---------+--------+-----+-------+-------+------+----------+
| 0       |        | 0   | 0     | *     |      | [0]      |
+---------+--------+-----+-------+-------+------+----------+
```

not sure if this is the expected behavior

### DIFF
--- a/ee/tables/dataflattentable/tables.go
+++ b/ee/tables/dataflattentable/tables.go
@@ -179,7 +179,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 			subresults, err := t.generateRawData(ctx, rawdata, dataQuery, append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
 			if err != nil {
 				t.slogger.Log(ctx, slog.LevelInfo,
-					"failed to get data for raw_data",
+					"failed to generate for raw_data",
 					"err", err,
 				)
 				continue
--- a/ee/tables/dataflattentable/tables.go
+++ b/ee/tables/dataflattentable/tables.go
@@ -140,8 +140,8 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	requestedPaths := tablehelpers.GetConstraints(queryContext, "path")
 	requestedRawDatas := tablehelpers.GetConstraints(queryContext, "raw_data")
 
-	if len(requestedPaths)+len(requestedRawDatas) == 0 {
-		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path or raw_data", t.tableName)
+	if len(requestedPaths) == 0 && len(requestedRawDatas) == 0 {
+		return results, fmt.Errorf("The %s table requires that you specify at least one of 'path' or 'raw_data'", t.tableName)
 	}
 
 	flattenOpts := []dataflatten.FlattenOpts{
--- a/ee/tables/dataflattentable/tables.go
+++ b/ee/tables/dataflattentable/tables.go
@@ -120,11 +120,12 @@ func AllTablePlugins(slogger *slog.Logger) []osquery.OsqueryPlugin {
 }
 
 func TablePlugin(slogger *slog.Logger, dataSourceType DataSourceType) osquery.OsqueryPlugin {
-	columns := Columns(table.TextColumn("path"))
+	columns := Columns(table.TextColumn("path"), table.TextColumn("raw_data"))
 
 	t := &Table{
-		tableName:       dataSourceType.TableName(),
-		flattenFileFunc: dataSourceType.FlattenFileFunc(""),
+		tableName:        dataSourceType.TableName(),
+		flattenFileFunc:  dataSourceType.FlattenFileFunc(""),
+		flattenBytesFunc: dataSourceType.FlattenBytesFunc(""),
 	}
 
 	t.slogger = slogger.With("table", t.tableName)
@@ -137,8 +138,15 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	var results []map[string]string
 
 	requestedPaths := tablehelpers.GetConstraints(queryContext, "path")
-	if len(requestedPaths) == 0 {
-		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path", t.tableName)
+	requestedRawDatas := tablehelpers.GetConstraints(queryContext, "raw_data")
+
+	if len(requestedPaths)+len(requestedRawDatas) == 0 {
+		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path or raw_data", t.tableName)
+	}
+
+	flattenOpts := []dataflatten.FlattenOpts{
+		dataflatten.WithSlogger(t.slogger),
+		dataflatten.WithNestedPlist(),
 	}
 
 	for _, requestedPath := range requestedPaths {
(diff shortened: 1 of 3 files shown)


ee/tables/dataflattentable/tables.go
```diff
@@ -165,16 +173,43 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
+	if err != nil {
+		t.slogger.Log(ctx, slog.LevelInfo,
+			"failure parsing raw data",
+			"err", err,
+		)
+		return nil, fmt.Errorf("parsing data: %w", err)
 	}
 
+	rowData := map[string]string{
+		"raw_data": rawdata,
```
James-Pickett, line 206: not sure if we should actually include this, can return a lot of data

I suppose you could always just select columns you want in query.

directionless: Sadly, we have to. If you don't, sqlite will filter out the rows. I'm not sure if we have another option today

ee/tables/dataflattentable/tables.go
```diff
@@ -137,8 +138,15 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	var results []map[string]string
 
 	requestedPaths := tablehelpers.GetConstraints(queryContext, "path")
-	if len(requestedPaths) == 0 {
-		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path", t.tableName)
+	requestedRawDatas := tablehelpers.GetConstraints(queryContext, "raw_data")
+
+	if len(requestedPaths)+len(requestedRawDatas) == 0 {
```
directionless, line 143: I'd probably `len(requestedPaths) == 0 && len(requestedRawDatas) == 0`, but okay...

ee/tables/dataflattentable/tables.go
```diff
@@ -137,8 +138,15 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
+		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path or raw_data", t.tableName)
```
directionless, line 144:
```suggestion
		return results, fmt.Errorf("The %s table requires that you specify a at least one of  `path` or `raw_data`", t.tableName)
```

directionless (commented):
I think probably good enough

ee/tables/dataflattentable/tables.go
directionless: Er, at least for the case where `raw_data` is provided. We should skip for file ones

ee/tables/dataflattentable/tables.go
```diff
@@ -165,16 +173,43 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 			}
 		}
 	}
+
+	for _, rawdata := range requestedRawDatas {
+		for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
+			subresults, err := t.generateRawData(ctx, rawdata, dataQuery, append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
+			if err != nil {
+				t.slogger.Log(ctx, slog.LevelInfo,
+					"failed to get data for raw_data",
```
directionless, line 182:
```suggestion
					"failed to generate for raw_data",
```

directionless (dismissed):
I think it's good

ee/tables/dataflattentable/tables.go
```diff
@@ -137,8 +138,15 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
-		return results, fmt.Errorf("The %s table requires that you specify a single constraint for path", t.tableName)
+	requestedRawDatas := tablehelpers.GetConstraints(queryContext, "raw_data")
+
+	if len(requestedPaths) == 0 && len(requestedRawDatas) == 0 {
+		return results, fmt.Errorf("The %s table requires that you specify at least one of 'path' or 'raw_data'", t.tableName)
+	}
+
+	flattenOpts := []dataflatten.FlattenOpts{
+		dataflatten.WithSlogger(t.slogger),
+		dataflatten.WithNestedPlist(),
```
RebeccaMahany, line 149: Why this plist-related option?

ee/tables/dataflattentable/tables.go
```diff
@@ -151,7 +159,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 
 		for _, filePath := range filePaths {
 			for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
-				subresults, err := t.generatePath(ctx, filePath, dataQuery)
+				subresults, err := t.generatePath(ctx, filePath, dataQuery, append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
```
RebeccaMahany, line 162: Why do we split the query on forward slash here and on 179?

RebeccaMahany (commented):
Just a couple questions to understand the changes better, but overall this looks good to me 🙂

ee/tables/dataflattentable/tables.go
James-Pickett: This code handles plist in addition to json, xml, ini, jwt. I just copied these options from the original [generatePath](https://github.com/kolide/launcher/blob/main/ee/tables/dataflattentable/tables.go#L174) function to consolidate them here for both `generatePath` and `generateRawData`

ee/tables/dataflattentable/tables.go
James-Pickett: I'm not quite sure, I just copied it over form the [original code](https://github.com/kolide/launcher/blob/main/ee/tables/dataflattentable/tables.go#L175), I'll look into it a little more.

James-Pickett: Hmm searching the code for `dataflatten.WithQuery(strings.Split(dataQuery, "/")` reveals that we do this in many (all?) dataflatten tables, we should probably just bake that directly into dataflatten.

RebeccaMahany (approved):

ee/tables/dataflattentable/tables.go
directionless: `dataflatten.WithNestedPlist`is an older attempt at handling this nesting. We should carry it through

zackattack01 (approved):

PR #1736, opened 2024-06-04
make sqlite migrations allow launcher downgrades

zackattack01:
Quick fix so that future schema migrations included in launcher releases can still work with launcher downgrades (at least to our next release version which includes these changes). I wanted to get something in as quickly as possible to support the incoming sqlite changes for the watchdog effort. 
There are some additional notes added to ee/agent/storage/sqlite/migrations.md for reference


RebeccaMahany (approved):
Nice find, nice fix

PR #1738, opened 2024-06-06
do not pre-create data files in wix for uninstall cleanup

zackattack01:
Certain data directory files are pre-created as empty files during wix packaging, to ensure that these are cleaned up on uninstall.

However some deployments forcibly re-install a new MSI outside of the autoupdate flow. This will trigger the uninstall first, which is now able to remove those files, leaving the device in an unenrolled state

### DIFF
--- a/pkg/packagekit/wix/wix.go
+++ b/pkg/packagekit/wix/wix.go
@@ -263,19 +263,6 @@ func (wo *wixTool) setupDataDir(ctx context.Context) error {
 		return fmt.Errorf("create base data dir error for wix harvest: %w", err)
 	}
 
-	// touch these known file names before harvest to ensure they're cleaned up on uninstall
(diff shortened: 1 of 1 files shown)


directionless:
I'm not sure I quite understand what this will do. Can you make a table of the various install/uninstall pairs? Or I don't quite know what?

directionless (approved):
Okay, I talked to Zack about this. 

Our current understanding is that a lot of MSI upgrade/downgrade/repair behaviors work by first triggering an msi uninstall. In those cases we do _not_ want to uninstall the database.

And while we do want to uninstall, on balance, it's worse than removing the DB on upgrades.

RebeccaMahany (approved):

James-Pickett (approved):

PR #1742, opened 2024-06-07
windows root directory fallback

zackattack01:
These changes are to allow a windows device to fallback to a historical (or current) root directory location. When an MSI with a different root directory is freshly installed over another, launcher cannot see the existing DB and triggers a new enrollment.
To avoid this, svc_windows will override the root_directory set by the MSI if a populated database exists in another known root directory location

### DIFF
--- a/cmd/launcher/svc_windows.go
+++ b/cmd/launcher/svc_windows.go
@@ -57,7 +57,7 @@ func runWindowsSvc(systemSlogger *multislogger.MultiSlogger, args []string) erro
 	// Create a local logger. This logs to a known path, and aims to help diagnostics
 	if opts.RootDirectory != "" {
 		// check for old root directories before creating DB in case we've stomped over with windows MSI install
-		updatedRootDirectory := determineRootDirectory(opts.RootDirectory, systemSlogger.Logger)
+		updatedRootDirectory := determineRootDirectory(opts, systemSlogger.Logger)
 		if updatedRootDirectory != opts.RootDirectory {
 			systemSlogger.Log(context.TODO(), slog.LevelInfo,
 				"old root directory contents detected, overriding opts.RootDirectory",
@@ -252,7 +252,8 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 
 // determineRootDirectory is used specifically for windows deployments to override the
 // configured root directory if another one containing a launcher DB already exists
-func determineRootDirectory(optsRootDirectory string, slogger *slog.Logger, opts *launcher.Options) string {
+func determineRootDirectory(opts *launcher.Options, slogger *slog.Logger) string {
+	optsRootDirectory := opts.RootDirectory
 	// don't mess with the path if this installation isn't pointing to a kolide server URL
 	if opts.ControlServerURL != "k2device.kolide.com" && opts.ControlServerURL != "k2device-preprod.kolide.com" {
 		return optsRootDirectory
--- a/cmd/launcher/svc_windows.go
+++ b/cmd/launcher/svc_windows.go
@@ -29,6 +29,11 @@ import (
 // TODO This should be inherited from some setting
 const serviceName = "launcher"
 
+var likelyRootDirPaths = []string{
+	"C:\\ProgramData\\Kolide\\Launcher-kolide-k2\\data",
+	"C:\\Program Files\\Kolide\\Launcher-kolide-k2\\data",
+}
+
 // runWindowsSvc starts launcher as a windows service. This will
 // probably not behave correctly if you start it from the command line.
 func runWindowsSvc(systemSlogger *multislogger.MultiSlogger, args []string) error {
@@ -247,14 +252,14 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 
 // determineRootDirectory is used specifically for windows deployments to override the
 // configured root directory if another one containing a launcher DB already exists
-func determineRootDirectory(optsRootDirectory string, slogger *slog.Logger) string {
-	optsDBLocation := filepath.Join(optsRootDirectory, "launcher.db")
-	// database already exists in configured root directory, keep that
-	dbExists, err := nonEmptyFileExists(optsDBLocation)
-	if err == nil && dbExists {
+func determineRootDirectory(optsRootDirectory string, slogger *slog.Logger, opts *launcher.Options) string {
+	// don't mess with the path if this installation isn't pointing to a kolide server URL
+	if opts.ControlServerURL != "k2device.kolide.com" && opts.ControlServerURL != "k2device-preprod.kolide.com" {
 		return optsRootDirectory
 	}
 
+	optsDBLocation := filepath.Join(optsRootDirectory, "launcher.db")
+	dbExists, err := nonEmptyFileExists(optsDBLocation)
 	// If we get an unknown error, back out from making any options changes. This is an
 	// unlikely path but doesn't feel right updating the rootDirectory without knowing what's going
 	// on here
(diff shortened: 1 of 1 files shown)


cmd/launcher/svc_windows.go
```diff
@@ -51,6 +51,18 @@ func runWindowsSvc(systemSlogger *multislogger.MultiSlogger, args []string) erro
 
 	// Create a local logger. This logs to a known path, and aims to help diagnostics
 	if opts.RootDirectory != "" {
+		// check for old root directories before creating DB in case we've stomped over with windows MSI install
+		updatedRootDirectory := determineRootDirectory(opts.RootDirectory, systemSlogger.Logger)
```
directionless, line 55: I'd use the local/debug slogger, not the system one

directionless: Though I guess this is happening _before_ we know where to write debug.json. Tricky

cmd/launcher/svc_windows.go
```diff
@@ -232,3 +244,74 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
...
+	// database already exists in configured root directory, keep that
+	dbExists, err := nonEmptyFileExists(optsDBLocation)
+	if err == nil && dbExists {
+		return optsRootDirectory
+	}
+
+	// If we get an unknown error, back out from making any options changes. This is an
+	// unlikely path but doesn't feel right updating the rootDirectory without knowing what's going
+	// on here
+	if err != nil {
```
directionless, line 261: I'd put this condition above, in the usual `err != nil` spot.

cmd/launcher/svc_windows.go
```diff
@@ -232,3 +244,74 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
...
+			"location", optsDBLocation,
+			"err", err,
+		)
+
+		return optsRootDirectory
+	}
+
+	// we know this is a fresh install with no launcher.db in the configured root directory,
+	// check likely locations and return updated rootDirectory if found
+	likelyRootDirPaths := []string{
```
directionless, line 273: package level maybe?

cmd/launcher/svc_windows.go
```diff
@@ -232,3 +244,74 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 		}
 	}
 }
+
+// determineRootDirectory is used specifically for windows deployments to override the
+// configured root directory if another one containing a launcher DB already exists
+func determineRootDirectory(optsRootDirectory string, slogger *slog.Logger) string {
+	optsDBLocation := filepath.Join(optsRootDirectory, "launcher.db")
+	// database already exists in configured root directory, keep that
+	dbExists, err := nonEmptyFileExists(optsDBLocation)
```
directionless, line 253: I think there's another case here.... If the host _isn't_ the prod URL, then we should probably take whatever specified rootdirectory is, because it's probably dev

directionless (commented):
Thinking out loud... What are the cases:
| db[s] on disk  | option | desired outcome |
| -------------- | ------ | ----------------- |
| old location     | old location  | old location |
| new location   | new location  | new location |
| new location   | old location  | new location |
| old location   | new location  | old location |
| anywhere   | developer  | developer specified |

cmd/launcher/svc_windows.go
zackattack01: yeah my first pass at this failed to notice that and the logs end up being split across RootDirs, I thought maybe safer to just do this until we get it set up

cmd/launcher/svc_windows.go
zackattack01: thank you, pushed up a change- let me know if that's what you're looking for

cmd/launcher/svc_windows.go
```diff
@@ -51,6 +56,18 @@ func runWindowsSvc(systemSlogger *multislogger.MultiSlogger, args []string) erro
 
 	// Create a local logger. This logs to a known path, and aims to help diagnostics
```
directionless, line 57: comment looks wrong

cmd/launcher/svc_windows.go
```diff
@@ -232,3 +249,75 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 		}
 	}
 }
+
+// determineRootDirectory is used specifically for windows deployments to override the
+// configured root directory if another one containing a launcher DB already exists
+func determineRootDirectory(opts *launcher.Options, slogger *slog.Logger) string {
```
directionless, line 255:
```suggestion
func determineRootDirectory(slogger *slog.Logger, opts *launcher.Option) string {
```

cmd/launcher/svc_windows.go
```diff
@@ -29,6 +29,11 @@ import (
 // TODO This should be inherited from some setting
 const serviceName = "launcher"
 
+var likelyRootDirPaths = []string{
+	"C:\\ProgramData\\Kolide\\Launcher-kolide-k2\\data",
+	"C:\\Program Files\\Kolide\\Launcher-kolide-k2\\data",
```
directionless, line 34: I _think_ this is the correct order? It means that if there's a new db path, we should prefer it. And I think that's likely to be true if someone got themselves into this situation.  (They'll have already done the pain of it, and going back to the old db will be worse)

directionless (commented):
This feels pretty good

cmd/launcher/svc_windows.go
zackattack01: that was my thinking as well. I think for most (upgrade) cases moving forward we'll get ProgramData via opts anyway and check that first before we even iterate these

directionless: Mostly talking through my assumptions here.

cmd/launcher/svc_windows.go
zackattack01: 👍 will do in a separate commit so I can update the caller too

cmd/launcher/svc_windows.go
zackattack01: oh yeah that got out of place, will update

James-Pickett (dismissed):

cmd/launcher/svc_windows.go
```diff
@@ -232,3 +249,75 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 		}
 	}
 }
+
+// determineRootDirectory is used specifically for windows deployments to override the
+// configured root directory if another one containing a launcher DB already exists
+func determineRootDirectory(slogger *slog.Logger, opts *launcher.Options) string {
+	optsRootDirectory := opts.RootDirectory
+	// don't mess with the path if this installation isn't pointing to a kolide server URL
+	if opts.KolideServerURL != "k2device.kolide.com" && opts.KolideServerURL != "k2device-preprod.kolide.com" {
```
RebeccaMahany, line 258: For my own understanding -- is the point of doing this over checking`opts.KolideHosted` to preserve dev/test environments?

RebeccaMahany (dismissed):
I think this makes sense!

cmd/launcher/svc_windows.go
zackattack01: yes, when I was looking at that logic it seemed like we set the KolideHosted for localhost (and other) cases so I just kept the check to this

directionless: I think we haven't really committed to `opts.KolideHosted` and I don't even know if it's wholly working. (but yeah, also need to carveout dev)

James-Pickett (approved):

directionless (approved):
I think this is okay to try. We'll need to figure out some really good testing

PR #1755, opened 2024-06-14
Take and use backups of the launcher database

RebeccaMahany:
- launcher takes periodic backups of its own database, beginning 10 mins after startup
- bbolt checkup includes stats about the backup database
- If its original database is unavailable but a backup is, launcher uses that backup db
- On remote uninstall, the backup database is deleted

### DIFF
--- a/ee/agent/storage/bbolt/backup.go
+++ b/ee/agent/storage/bbolt/backup.go
@@ -112,19 +112,20 @@ func (d *databaseBackupSaver) backupDb() error {
 func (d *databaseBackupSaver) rotate() error {
 	baseBackupPath := BackupLauncherDbLocation(d.knapsack.RootDirectory())
 
-	for i := numberOfOldBackupsToRetain; i > 0; i -= 1 {
+	// Delete the oldest backup, if it exists
+	oldestBackupPath := fmt.Sprintf("%s.%d", baseBackupPath, numberOfOldBackupsToRetain)
+	if _, err := os.Stat(oldestBackupPath); err == nil {
+		if err := os.Remove(oldestBackupPath); err != nil {
+			return fmt.Errorf("removing oldest backup %s during rotation: %w", oldestBackupPath, err)
+		}
+	}
+
+	// Rename all other backups
+	for i := numberOfOldBackupsToRetain - 1; i > 0; i -= 1 {
 		currentBackupPath := fmt.Sprintf("%s.%d", baseBackupPath, i)
 
 		// This backup doesn't exist yet -- skip it
 		if _, err := os.Stat(currentBackupPath); err != nil && os.IsNotExist(err) {
-			continue
-		}
-
-		// If is the oldest backup, delete it so we can rotate a new one into its place
-		if i == numberOfOldBackupsToRetain {
-			if err := os.Remove(currentBackupPath); err != nil {
-				return fmt.Errorf("removing oldest backup %s during rotation: %w", currentBackupPath, err)
-			}
 			continue
 		}
 
--- a/ee/agent/storage/bbolt/backup.go
+++ b/ee/agent/storage/bbolt/backup.go
@@ -14,45 +14,44 @@ import (
 )
 
 const (
-	snapshotInitialDelay = 10 * time.Minute
-	snapshotInterval     = 1 * time.Hour
+	backupInitialDelay = 10 * time.Minute
+	backupInterval     = 1 * time.Hour
 )
 
-// A photographer takes snapshots.
-// TODO RM - A better name.
-type photographer struct {
+// databaseBackupMaintainer periodically takes backups of launcher.db.
+type databaseBackupMaintainer struct {
 	knapsack    types.Knapsack
 	slogger     *slog.Logger
 	interrupt   chan struct{}
 	interrupted bool
 }
 
-func NewDatabasePhotographer(k types.Knapsack) *photographer {
-	return &photographer{
+func NewDatabaseBackupMaintainer(k types.Knapsack) *databaseBackupMaintainer {
+	return &databaseBackupMaintainer{
 		knapsack:  k,
-		slogger:   k.Slogger().With("component", "database_photographer"),
+		slogger:   k.Slogger().With("component", "database_backup_maintainer"),
 		interrupt: make(chan struct{}, 1),
 	}
 }
 
-func (p *photographer) Execute() error {
-	// Wait a little bit after startup before taking first snapshot, to allow for enrollment
+func (d *databaseBackupMaintainer) Execute() error {
+	// Wait a little bit after startup before taking first backup, to allow for enrollment
 	select {
-	case <-p.interrupt:
-		p.slogger.Log(context.TODO(), slog.LevelDebug,
+	case <-d.interrupt:
+		d.slogger.Log(context.TODO(), slog.LevelDebug,
 			"received external interrupt during initial delay, stopping",
 		)
 		return nil
-	case <-time.After(snapshotInitialDelay):
+	case <-time.After(backupInitialDelay):
 		break
 	}
 
-	// Take periodic snapshots
-	ticker := time.NewTicker(snapshotInterval)
+	// Take periodic backups
+	ticker := time.NewTicker(backupInterval)
 	defer ticker.Stop()
 	for {
-		if err := p.backupDb(); err != nil {
-			p.slogger.Log(context.TODO(), slog.LevelWarn,
+		if err := d.backupDb(); err != nil {
+			d.slogger.Log(context.TODO(), slog.LevelWarn,
 				"could not perform periodic database backup",
 				"err", err,
 			)
@@ -61,8 +60,8 @@ func (p *photographer) Execute() error {
 		select {
 		case <-ticker.C:
 			continue
-		case <-p.interrupt:
-			p.slogger.Log(context.TODO(), slog.LevelDebug,
+		case <-d.interrupt:
+			d.slogger.Log(context.TODO(), slog.LevelDebug,
 				"interrupt received, exiting execute loop",
 			)
 			return nil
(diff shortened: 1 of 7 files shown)


ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,173 @@
...
+	// Take backup -- it's fine to just overwrite previous backups
+	backupLocation := BackupLauncherDbLocation(d.knapsack.RootDirectory())
+	if err := d.knapsack.BboltDB().View(func(tx *bbolt.Tx) error {
+		return tx.CopyFile(backupLocation, 0600)
+	}); err != nil {
+		return fmt.Errorf("backing up database: %w", err)
+	}
+
+	// Confirm file exists and is nonempty
+	if exists, err := launcher.NonEmptyFileExists(backupLocation); !exists {
```
directionless, line 92: I think it's better to just copy `nonEmptyFileExists` into a helper here. Importing `launcher` for a help function feels wrong

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("backup succeeded, but nonempty file does not exist at %s", backupLocation)
+	} else if err != nil {
+		return fmt.Errorf("backup succeeded, but error checking if file was created at %s: %w", backupLocation, err)
+	}
+
+	// Log success
+	d.slogger.Log(context.TODO(), slog.LevelDebug,
```
directionless, line 99: I'd log this from the caller on `err=nil` or whatever.

directionless (commented):
early nits

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,173 @@
...
+	d.interrupted = true
+
+	d.interrupt <- struct{}{}
+}
+
+func (d *databaseBackupSaver) backupDb() error {
+	// Take backup -- it's fine to just overwrite previous backups
+	backupLocation := BackupLauncherDbLocation(d.knapsack.RootDirectory())
+	if err := d.knapsack.BboltDB().View(func(tx *bbolt.Tx) error {
+		return tx.CopyFile(backupLocation, 0600)
```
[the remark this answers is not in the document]
RebeccaMahany: Yeah, I really liked how simple this was!

ee/agent/storage/bbolt/backup.go
RebeccaMahany: Fair, I will update

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,173 @@
...
+func NewDatabaseBackupSaver(k types.Knapsack) *databaseBackupSaver {
+	return &databaseBackupSaver{
+		knapsack:  k,
+		slogger:   k.Slogger().With("component", "database_backup_saver"),
+		interrupt: make(chan struct{}, 1),
+	}
+}
+
+func (d *databaseBackupSaver) Execute() error {
+	// Wait a little bit after startup before taking first backup, to allow for enrollment
```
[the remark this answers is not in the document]
RebeccaMahany: I was going back and forth on that. We don't have a really easy mechanism to get notified about the nodekey getting set, the way that we do for when agent flags change -- so we'd have to write something custom here to periodically check different keys, compare their values with previous values (which we would also have to store here), and then take action based on the comparison. It felt like a lot of complexity for not a lot of advantage over just waiting. (Is it reasonable to worry about someone enrolling and then reinstalling within 10 minutes and expecting their previous installation to work?) But maybe there's an advantage I'm not thinking of?

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,173 @@
...
+		"backup_location", backupLocation,
+		"original_location", originalDbLocation,
+	)
+}
+
+func LauncherDbLocation(rootDir string) string {
+	return filepath.Join(rootDir, "launcher.db")
+}
+
+func BackupLauncherDbLocation(rootDir string) string {
```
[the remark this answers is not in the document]
RebeccaMahany: I wasn't super worried about that because of how the access remains relatively restricted in this implementation --

* On launcher startup, before any rungroups execute, launcher may rename `launcher.db.bak` to `launcher.db`; nothing else is accessing either db at this time
* After rungroups begin to execute, this actor is the only thing touching `launcher.db.bak` -- nothing else will read from it or write to it
* (Except for the flare checkup -- and data inconsistency probably isn't a huge deal there, and I'm not sure how much we even care about having that checkup anyway)
* (There's also remote uninstall, but that just wants to delete the file)

Is there something I'm missing?

I'm happy to keep N and rotate if we feel like that's a safer option, regardless.

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
RebeccaMahany: If we're worried about file size (especially if we end up retaining N versions), I could try out using the `dbcompact` that we already have: https://github.com/kolide/launcher/blob/main/ee/agent/dbcompact.go

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
RebeccaMahany: (If we do decide we want to check for keys instead of delaying -- just the node key in the config store? Anything else?)

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	d.interrupted = true
+
+	d.interrupt <- struct{}{}
+}
+
+func (d *databaseBackupSaver) backupDb() error {
+	// Take backup -- it's fine to just overwrite previous backups
+	backupLocation := BackupLauncherDbLocation(d.knapsack.RootDirectory())
+	if err := d.knapsack.BboltDB().View(func(tx *bbolt.Tx) error {
```
zackattack01, line 84: I'm not sure we need to do this now but this is adding another direct bbolt invocation and I know we were trying to move away from that - it gets a little weird because this is a full db copy but I think we could add something like a BackupSaver interface to types/keyvalue_store.go and just define this bit as e.g. `BackupFull` on `bboltKeyValueStore`. then call from `k.ConfigStore`, etc. mostly just thinking out loud, happy to think more and come back to this later

RebeccaMahany: Yeah, I keep struggling with this. It seems incorrect to have a full-db operation `BackupFull` attached to a partial-db object (store/bucket `bboltKeyValueStore`) -- but maybe that's unavoidable and the lesser evil if we really want to avoid the direct bbolt invocation?

zackattack01: yeah i have not been able to think of a clean solution there, maybe it's best to avoid complicating this for now anyway. there is no major benefit right now

zackattack01 (dismissed):

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
RebeccaMahany: I added rotation in https://github.com/kolide/launcher/pull/1755/commits/15183fa45690392898d559137063fc8e6565b9c3 -- lmk what you're thinking!

zackattack01 (dismissed):

ee/agent/storage/bbolt/backup.go
RebeccaMahany: (ended up moving this back due to the extra rotate/rename steps)

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
directionless: Yeah, that does sound messy. Maybe this is good enough for now.

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
directionless: If `tx.CopyFile` is clever, it might compact along the way

ee/agent/storage/bbolt/backup.go
directionless: Seems fine for now. I think one path might be to make `Backup(ctx, path)` part of the store interface, and then call it from knapsack or something

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// Confirm file exists and is nonempty
+	if exists, err := nonEmptyFileExists(tempBackupLocation); !exists {
+		return fmt.Errorf("backup succeeded, but nonempty file does not exist at %s", tempBackupLocation)
+	} else if err != nil {
+		return fmt.Errorf("backup succeeded, but error checking if file was created at %s: %w", tempBackupLocation, err)
+	}
+
+	// Perform rotation of older backups so we can move this backup to `backupLocation`
```
directionless, line 104: I suspect this is cleaner if we reorder -- rotate first, then backup to the expected location. But I could roll with this

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,234 @@
...
+			continue
+		}
+
+		// If is the oldest backup, delete it so we can rotate a new one into its place
+		if i == numberOfOldBackupsToRetain {
+			if err := os.Remove(currentBackupPath); err != nil {
+				return fmt.Errorf("removing oldest backup %s during rotation: %w", currentBackupPath, err)
+			}
+			continue
+		}
```
directionless, line 139: You can probably drop this logic, by removing the `i`th backup before the loop, and looping `i := numberOfOldBackupsToRetain - 1`.

ee/agent/storage/bbolt/backup.go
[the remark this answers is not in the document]
RebeccaMahany: It doesn't look like it does, but I could be missing something

directionless (dismissed):
I don't quite have the focus to really think through some of this. I think it's okay, and I'm nitpicking small things.

ee/agent/storage/bbolt/backup.go
RebeccaMahany: Oh yup it's much cleaner that way -- I will update

ee/agent/storage/bbolt/backup.go
RebeccaMahany: Updated! Hopefully it reads a little better now

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,225 @@
...
+	"path/filepath"
+	"time"
+
+	"github.com/kolide/launcher/ee/agent/types"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	backupInitialDelay         = 10 * time.Minute
+	backupInterval             = 1 * time.Hour
```
directionless, line 17: I think we can even lower this. 6 hours, maybe? The only things we really care about are keys and eventually registrations.

directionless (approved):

ee/agent/storage/bbolt/backup.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	d.interrupted = true
+
+	d.interrupt <- struct{}{}
+}
+
+func (d *databaseBackupSaver) backupDb() error {
+	// Perform rotation of older backups to prepare for newer backup
+	if err := d.rotate(); err != nil {
+		return fmt.Errorf("backup succeeded, but rotation did not: %w", err)
```
zackattack01, line 85: error message should probably be updated now

RebeccaMahany: whoops ty

zackattack01 (approved):

Issue #1758, opened 2024-06-21
`loginctl --output-json` does not output json, eventually leading to missing systray icon


roobre:
I've been chasing for a couple days why the kolide agent wasn't showing a systray icon for me and some other coworkers who use Arch Linux (btw).

In `/var/kolide-k2/k2device.kolide.com/debug.json` I was able to find this message being spammed:
```
{"time":"2024-06-21T15:43:31.114726268Z","level":"INFO","source":{"function":"github.com/kolide/launcher/ee/desktop/runner.(*DesktopUsersProcessesRunner).Execute","file":"/home/runner/work/launcher/launcher/ee/desktop/runner/runner.go","line":226},"msg":"running console user desktop","launcher_run_id":"REDACTED","component":"desktop_runner","err":"getting console users: loginctl list-sessions unmarshall json output: invalid character '1' after top-level value"}
```

Which seems to be caused by this line

https://github.com/kolide/launcher/blob/d80b842f943ee480ecd61ec929bf4dc5400c5a67/ee/consoleuser/consoleuser_linux.go#L23

I believe this is not working because `loginctl --output-json` does not output json on recent systemd versions:

```
roobre@Archiroo 󱃾 curry-admin@Curry
18:00:28 ~ $> loginctl list-sessions --output=json
SESSION  UID USER   SEAT  LEADER CLASS   TTY IDLE SINCE
      1 1000 roobre seat0 2774   greeter -   no   -
      2 1000 roobre -     3153   manager -   no   -

2 sessions listed.
```

The correct flag to get JSON output now seems to be `--json=short`:
```
roobre@Archiroo 󱃾 curry-admin@Curry
18:03:08 ~ $> loginctl list-sessions --json=short
[{"session":"1","uid":1000,"user":"roobre","seat":"seat0","leader":2774,"class":"greeter","tty":null,"idle":false,"since":null},{"session":"2","uid":1000,"user":"roobre","seat":null,"leader":3153,"class":"manager","tty":null,"idle":false,"since":null}]
```

I was able to work around this by replacing `/usr/bin/loginctl` with a shim script that translates `--output=json` to `--json=short`, but this is obviously not a very good workaround as it won't survive a systemd update. Placing the shim in `/usr/local/bin/` or similar is not possible as Kolide hardcodes this path:
https://github.com/kolide/launcher/blob/d80b842f943ee480ecd61ec929bf4dc5400c5a67/ee/allowedcmd/cmd_linux.go#L85

[...]

PR #1759, opened 2024-06-21
Try both `--output=json` and `--json=short` when running `loginctl list-sessions` output

RebeccaMahany:
The output flag for `loginctl list-sessions` has changed in more recent versions of systemd: https://www.freedesktop.org/software/systemd/man/latest/loginctl.html#Session%20Commands

On Ubuntu at least up to 24.04, systemd doesn't use this new `--json` flag -- I think at this point, the older flag will be much more widely supported in our userbase. So I have defaulted to trying with the older flag first, and falling back to the newer flag in case of error. Eventually we will want to swap the order that we try the flags in.

Confirmed `--json=short` output is compatible with what we previously expected (session, uid, user, seat all present):

```
[becca@arch-2 ~]$ loginctl --version
systemd 256 (256.1-1-arch-g34ba18b^)
+PAM +AUDIT -SELINUX -APPARMOR -IMA +SMACK +SECCOMP +GCRYPT +GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBCRYPTSETUP_PLUGINS +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK +XKBCOMMON +UTMP -SYSVINIT +LIBARCHIVE

[becca@arch-2 ~]$ loginctl list-sessions --no-legend --no-pager --json=short
[{"session":"3","uid":1002,"user":"becca","seat":null,"leader":9206,"class":"user","tty":"pts/0","idle":false,"since":null}]
```

Relates to https://github.com/kolide/launcher/issues/1758

<details>

<summary>Notes on setting up a test environment</summary>

I set up an arch VM in Google Cloud to test this -- here are the steps.

I used a base image from [this project](https://github.com/GoogleCloudPlatform/compute-archlinux-image-builder) and created a VM:

```
gcloud compute instances create <VM name> --image-project=arch-linux-gce --image-family=arch --project <project name>
```

I SSHed to the VM and got it updated via [these instructions](https://wiki.archlinux.org/title/Pacman/Package_signing#Upgrade_system_regularly)

```
pacman -Sy --needed archlinux-keyring && pacman -Su
```

Validated that it's adequately updated via `loginctl --version` and `loginctl list-sessions --no-legend --no-pager --json=short`.

I stopped the VM, enabled `display device`, and restarted the VM – I think this is what got me a `Remote=no Active=yes` session to be able to get a desktop process going.

I uploaded launcher binary from this PR's artifacts to VM; I uploaded osqueryd binary [from the releases](https://github.com/osquery/osquery/releases) to the VM; `chmod +x` both.

I made a root dir for testing, then ran launcher and let it start up:

```
sudo ./launcher --hostname <k2 hostname here> --root_directory <root dir here> --osqueryd_path <path-to-your-osqueryd> --transport jsonrpc --enroll_secret <secret here>
```

In another window, I confirmed I saw the launcher desktop process eventually -- `ps -eaf | grep launcher`. There were some errors (e.g. no tray is running, so we get errors about no tray running), but seeing the desktop process running should be enough to prove this fix.

</details>

### DIFF
--- a/ee/consoleuser/consoleuser_linux.go
+++ b/ee/consoleuser/consoleuser_linux.go
@@ -20,15 +20,9 @@ type listSessionsResult []struct {
 }
 
 func CurrentUids(ctx context.Context) ([]string, error) {
-	output, err := listSessions(ctx)
+	sessions, err := listSessions(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("listing sessions: %w", err)
-	}
-
-	// unmarshall json output into listSessionsResult
-	var sessions listSessionsResult
-	if err := json.Unmarshal(output, &sessions); err != nil {
-		return nil, fmt.Errorf("loginctl list-sessions unmarshall json output: %w", err)
 	}
 
 	var uids []string
@@ -70,7 +64,12 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 	return uids, nil
 }
 
-func listSessions(ctx context.Context) ([]byte, error) {
+// listSessions execs `loginctl list-sessions` in order to retrieve the current list of sessions.
+// Depending on the systemd version, we have to use different flags to output the results as JSON.
+// We may want to attempt parsing the output regardless in the future -- see launcher #1522.
+func listSessions(ctx context.Context) (listSessionsResult, error) {
+	var sessions listSessionsResult
+
 	// Try with `--output=json` first, to support the more widely-used older versions of systemd
 	legacyCmd, err := allowedcmd.Loginctl(ctx, "list-sessions", "--no-legend", "--no-pager", "--output=json")
 	if err != nil {
@@ -78,7 +77,11 @@ func listSessions(ctx context.Context) ([]byte, error) {
 	}
 	legacyOut, err := legacyCmd.Output()
 	if err == nil {
-		return legacyOut, nil
+		// Newer versions of systemd ignore `--output=json` rather than returning an error, so we also
+		// need to unmarshal the result to confirm we got expected output.
+		if err := json.Unmarshal(legacyOut, &sessions); err == nil {
+			return sessions, nil
+		}
 	}
 
 	cmd, err := allowedcmd.Loginctl(ctx, "list-sessions", "--no-legend", "--no-pager", "--json=short")
@@ -89,6 +92,9 @@ func listSessions(ctx context.Context) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("loginctl list-sessions: %w", err)
 	}
+	if err := json.Unmarshal(output, &sessions); err != nil {
+		return nil, fmt.Errorf("unmarshalling loginctl list-sessions output: %w", err)
+	}
 
-	return output, nil
+	return sessions, nil
 }
--- a/ee/consoleuser/consoleuser_linux.go
+++ b/ee/consoleuser/consoleuser_linux.go
@@ -20,13 +20,9 @@ type listSessionsResult []struct {
 }
 
 func CurrentUids(ctx context.Context) ([]string, error) {
-	cmd, err := allowedcmd.Loginctl(ctx, "list-sessions", "--no-legend", "--no-pager", "--output=json")
+	output, err := listSessions(ctx)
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):

ee/consoleuser/consoleuser_linux.go
```diff
@@ -73,3 +69,26 @@ func CurrentUids(ctx context.Context) ([]string, error) {
...
+func listSessions(ctx context.Context) ([]byte, error) {
+	// Try with `--output=json` first, to support the more widely-used older versions of systemd
+	legacyCmd, err := allowedcmd.Loginctl(ctx, "list-sessions", "--no-legend", "--no-pager", "--output=json")
+	if err != nil {
+		return nil, fmt.Errorf("creating loginctl command --no-legend --no-pager --output=json: %w", err)
+	}
+	legacyOut, err := legacyCmd.Output()
+	if err == nil {
+		return legacyOut, nil
+	}
```
roobre, line 82: Hey, OP from https://github.com/kolide/launcher/issues/1758 here. I'm not sure if this will work as intended. I believe the current behavior of `loginctl` is to accept `--output=json` but simply ignore it. In other words, the command does not fail when this (now unrecognized) flag is provided:

```
roobre@Archiroo 󱃾 curry-admin@Curry
14:30:50 ~ $> loginctl list-sessions --output=json && echo Succeeded
SESSION  UID USER   SEAT  LEADER CLASS   TTY IDLE SINCE
      1 1000 roobre seat0 2771   greeter -   no   -
      2 1000 roobre -     3152   manager -   no   -

2 sessions listed.
Succeeded
```

RebeccaMahany: Thanks! Still working on getting a test environment with a new enough version of systemd on it to be able to test these changes -- will update this PR and move it out of draft once that's complete 🙂

RebeccaMahany: Updated and tested

zackattack01 (approved):

ee/consoleuser/consoleuser_linux.go
```diff
@@ -73,3 +63,38 @@ func CurrentUids(ctx context.Context) ([]string, error) {
...
+func listSessions(ctx context.Context) (listSessionsResult, error) {
+	var sessions listSessionsResult
+
+	// Try with `--output=json` first, to support the more widely-used older versions of systemd
+	legacyCmd, err := allowedcmd.Loginctl(ctx, "list-sessions", "--no-legend", "--no-pager", "--output=json")
+	if err != nil {
+		return nil, fmt.Errorf("creating loginctl command --no-legend --no-pager --output=json: %w", err)
+	}
+	legacyOut, err := legacyCmd.Output()
+	if err == nil {
```
directionless, line 79: Do we need this `if`?

RebeccaMahany: No point in trying to unmarshal if the command failed, right -- or have I missed something?

directionless: Of the command failed, it returns on line 76

RebeccaMahany: That's the creation of the exec.Cmd struct -- we return on 76 if we fail to validate the command (e.g. if we can't find the `loginctl` binary). But we don't exec the command till line 78, so we're checking for exec failures here.

James-Pickett (approved):

ee/consoleuser/consoleuser_linux.go
directionless: Oh, I see it now. This is inverted logic -- no error early returns. You could probably combine, `if legacyOut, err := legacyCmd.Output();  err == nil {` but I don' think that changes anything real. 

Thank you for putting up with my fuzziness.

directionless (approved):

PR #1763, opened 2024-06-26
[Kolide ATC] Construct KATC tables and add support for Firefox extension data

RebeccaMahany:
Builds on https://github.com/kolide/launcher/pull/1761

This PR adds config parsing for KATC tables and constructing those tables.

As a proof-of-concept, it implements one table type and a couple data processing steps, with the end result of supporting querying Firefox extension data (sqlite file + data is compressed via snappy + data is serialized with StructuredClone).

I think will be easiest to review the changes in this order:

1. `pkg/osquery/table/table.go` -- this is the entrypoint to the `ee/katc` package, and touches the changes made in the previous PR
1. `ee/katc/config.go` -- this is the parsing and construction of the KATC tables
1. `ee/katc/table.go` -- this is the generation of results when a query is run against a KATC table
1. `ee/katc/sqlite.go` -- this is querying a specific backend (in this case, a sqlite database) on behalf of a KATC table (this function is called by `ee/katc/table.go`)
1. `ee/katc/snappy.go` -- this is a fairly simple row transform step (called by `ee/katc/table.go` on the sqlite results)
1. `ee/katc/deserialize_firefox.go` -- this is a significantly more complicated row transform step (called by `ee/katc/table.go` on the sqlite results that have already been snappy decompressed); ultimately it's deserializing a JS object into a Golang map

An example KATC config:

```json
{
    "my_example_table": {
        "source_type": "sqlite",
        "source": "/some/%/path/to/db.sqlite",
        "platform": "darwin",
        "columns": ["someKeyInsideCompressedSerializedJSONObject"],
        "query": "SELECT data FROM object_data JOIN object_store ON (object_data.object_store_id = object_store.id) WHERE object_store.name=\"some-test-table\";",
        "row_transform_steps": ["snappy", "structured_clone"]
    }
}
```

Work TBD in separate PRs:
* Backfill tests: https://github.com/kolide/launcher/pull/1766
* Transform camel case columns to snake case: https://github.com/kolide/launcher/pull/1765
* Add support for indexeddb/leveldb tables: begun in https://github.com/kolide/launcher/pull/1767


ee/katc/structured_clone.go
```diff
@@ -0,0 +1,257 @@
...
+	tagBackReferenceObject uint32 = 0xffff000d
+	tagDoNotUse1           uint32 = 0xffff000e
+	tagDoNotUse2           uint32 = 0xffff000f
+	tagTypedArrayObjectV2  uint32 = 0xffff0010
+	tagMapObject           uint32 = 0xffff0011
+	tagSetObject           uint32 = 0xffff0012
+	tagEndOfKeys           uint32 = 0xffff0013
+)
+
+// structuredCloneDeserialize deserializes a JS object that has been stored in IndexedDB
```
directionless, line 41: Why is it called `structuredCloneDeserialize` ?

ee/katc/config.go
```diff
@@ -0,0 +1,123 @@
...
+		d.name = structuredCloneDeserializeProcessingStep
+		d.processingFunc = structuredCloneDeserialize
+		return nil
+	default:
+		return fmt.Errorf("unknown data processing step %s", s)
+	}
+}
+
+type katcTableConfig struct {
+	Type                katcTableType        `json:"type"`
```
directionless, line 85: `type` implies something about the type of table, but I think this refers to the source? I'm also not sold on `Path` or `Columns`, they feel specific to `sqlite`.

I wonder if we can take a page from databases and treat this as a connect url string. `sqlite:path` or similar. Each source type would need to define it's own schema there.

ee/katc/config.go
```diff
@@ -0,0 +1,123 @@
...
+	Platform            string               `json:"platform"`
+	Columns             []string             `json:"columns"`
+	Path                string               `json:"path"`  // Path to file holding data (e.g. sqlite file) -- wildcards supported
+	Query               string               `json:"query"` // Query to run against `path`
+	DataProcessingSteps []dataProcessingStep `json:"data_processing_steps"`
```
directionless, line 90: love this one

ee/katc/sqlite.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, fmt.Errorf("querying %s: %w", sqliteDb, err)
+		}
+		results = append(results, resultsFromDb...)
+	}
+
+	return results, nil
+}
+
+// querySqliteDb queries the database at the given path, returning rows of results
+func querySqliteDb(ctx context.Context, path string, query string, columns []string, slogger *slog.Logger) ([]map[string][]byte, error) {
```
directionless, line 33: signature should be `ctx, slogger, ...`

ee/katc/sqlite.go
```diff
@@ -0,0 +1,84 @@
...
+	conn, err := sql.Open("sqlite", path)
```
directionless, line 34: We should open RO

ee/katc/sqlite.go
```diff
@@ -0,0 +1,84 @@
...
+	defer func() {
+		if err := rows.Close(); err != nil {
+			slogger.Log(ctx, slog.LevelWarn,
+				"closing rows after scanning results",
+				"err", err,
+			)
+		}
+	}()
+
+	results := make([]map[string][]byte, 0)
```
directionless, line 60: Does this drop `path` ?

ee/katc/sqlite.go
```diff
@@ -0,0 +1,84 @@
...
+	// Prepare destination for scan
+	rawResult := make([][]byte, len(columns))
+	scanDest := make([]any, len(columns))
+	for i := 0; i < len(columns); i += 1 {
+		scanDest[i] = &rawResult[i]
+	}
+
+	// Scan all rows
+	for rows.Next() {
+		if err := rows.Scan(scanDest...); err != nil {
```
directionless, line 71: I think this will error if `scanDest` has a different number of columns. I can imagine that breaking if we want `select *` and only want a couple columns.

directionless (commented):
I have not absorbed all of this. It looks neat. Early feedback

ee/katc/config.go
RebeccaMahany: Yeah, I think `source` is a better name than `type`! I'll rename.

I think `path` should always apply -- e.g. path to JSON file, or indexeddb file, or sqlite file. I'm still waffling a bunch on columns.

ee/katc/sqlite.go
RebeccaMahany: Yeah, it will definitely error in that case -- this requires that the `query` and `columns` in the config agree, which does feel brittle

ee/katc/structured_clone.go
RebeccaMahany: Named based on the mozilla references below -- could name based on some combination of `firefox` + `indexeddb values` instead

ee/katc/config.go
RebeccaMahany: Maybe `filepath` is better/clearer? I am also not opposed to doing a connect url string approach, that feels like it could be tidy 🙂

ee/katc/sqlite.go
RebeccaMahany: I don't think there's an open read-only option, unless I'm missing it

ee/katc/sqlite.go
RebeccaMahany: It does -- should we include it?

ee/katc/config.go
directionless: What if it was a network call?

directionless: I guess we could overload `path`. It's an arbitrary string parsed by source. :shrug: mostly it comes down to whether you'd like `source` and `path` or `source://path`

ee/katc/sqlite.go
directionless: Yes. A common ATC table is paths like `/Users/%/.config/blah.sql` so you need the path to differentiate between users.

ee/katc/sqlite.go
directionless: Is there a way to iterate a row by KV or something?

RebeccaMahany: I think calling `rows.Columns()` might get us what we want! I will try it out and see

ee/katc/sqlite.go
RebeccaMahany: Thank you for finding that -- I'll give it a shot!

directionless: It's also an example of the URL style :P

FritzX6:
Perhaps this is a silly suggestion, but do we want to consider the capability to copy/read/delete for DB's that are locked while open? Perhaps an optional boolean column for "read_from_copy" that puts the DB in /tmp/ or elsewhere?

ee/katc/structured_clone.go
directionless: I think in the context of indexdb it makes sense to use the mozilla names. But in the context of a DSL for KATC it's much less clear. Something along your handwave there feels better

ee/katc/sqlite.go
FritzX6: > Yeah, it will definitely error in that case -- this requires that the `query` and `columns` in the config agree, which does feel brittle

iirc this is the behavior of the existing osquery ATC parser, `columns` and `query` have to be in alignment or it won't return data.

directionless:
> Perhaps this is a silly suggestion, but do we want to consider the capability to copy/read/delete for DB's that are locked while open? Perhaps an optional boolean column eg. `read_from_copy` that puts the DB in `/tmp` or elsewhere?

We should explore that use case. There may be other sqlite options that let us open locked files. `immutable` specifically

ee/katc/sqlite.go
directionless: Doesn't mean we can't do better.

RebeccaMahany:
@FritzX6 not a silly suggestion! I had thought about keeping that in the implementation details instead of the config (e.g. for indexeddb we _always_ want to copy the db, we shouldn't allow configuration otherwise) -- maybe there's a stronger way to require that. I will look into the `immutable` suggestion from seph also.

ee/katc/config.go
RebeccaMahany: More thoughts --

I like the idea of replacing `path` with a `connection_string` or `dsn`. That does get a little weird with supporting wildcards -- we'd have to 1) parse the connection string to extract the filepath portion, then 2) glob based on that part, then 3) re-construct the connection string to use it. But we're doing the second two steps anyway, so might as well?

I _think_ I like keeping source in a separate field regardless of what we decide for ^, since it makes config unmarshalling simpler and more self-contained -- but I will think on it more and see how it feels after updating `path`!

directionless: > re-construct the connection string to use it.

Do we really need that? I think it'll depend a lot on the `type` in question. Like, imagine:
```
type: sqlite
dsn: file:///Users/%/...?options
```

If the dsn starts with file, URL parse it and hand it to glob. But I don't think we need to reconstruct, because I think the options are parsed by our katc type, not the underlying implementation. (practically speaking, maybe they're passed along, but I think that's going to be type specific)

directionless: I don't think we'd do this, but I can imagine a curl style thing. `https://host/path?validate=false` or something, that validate might signal to the type thing to disable tls validation, but it's not going to be part of the remote URL.

ee/katc/sqlite.go
RebeccaMahany: `rows.Columns` did the trick, no potential for scan errors now

ee/katc/sqlite.go
RebeccaMahany: Path is now included

ee/katc/config.go
RebeccaMahany: Renamed `type` to `source` -- `path` to `dsn` to come later once I sort through some of the other parts!

ee/katc/structured_clone.go
RebeccaMahany: Mostly, the difficulty in naming here for me is that this function and the corresponding one in `ee/indexeddb/values.go` both implement [StructuredDeserialize](https://html.spec.whatwg.org/multipage/structured-data.html#structureddeserialize), which is a great, well-known name. But the functions are not interchangeable -- this one only works for Firefox data, and the `ee/indexeddb` one only works for Chrome data. Maybe we just want something like `structuredDeserializeFirefox` here, since that would be more immediately understandable?

ee/katc/structured_clone.go
```diff
@@ -0,0 +1,275 @@
...
+	tagStringObject  uint32 = 0xffff000b
+	tagEndOfKeys     uint32 = 0xffff0013
+)
+
+// structuredCloneDeserialize deserializes a JS object that has been stored by Firefox
+// in IndexedDB sqlite-backed databases.
+// References:
+// * https://stackoverflow.com/a/59923297
+// * https://searchfox.org/mozilla-central/source/js/src/vm/StructuredClone.cpp (see especially JSStructuredCloneReader::read)
+func structuredCloneDeserialize(ctx context.Context, slogger *slog.Logger, row map[string][]byte) (map[string][]byte, error) {
```
RebeccaMahany, line 37: To keep this PR from expanding too much further, I will write unit tests for this file in a follow-up PR. A decent portion of this file is already exercised via `Test_generate_SqliteBackedIndexedDB`.

ee/katc/config.go
RebeccaMahany: Okay, I've ended up with `source_type` (for `sqlite`, `leveldb`, etc) and `source` (filepath with wildcard to db, network call, etc). I think this is a lot clearer in the code too, e.g. with the `sourceData` struct.

ee/katc/structured_clone.go
RebeccaMahany: Went with `deserializeFirefox`

ee/katc/table.go
```diff
@@ -0,0 +1,186 @@
...
+		for column, data := range row {
+			if _, expectedColumn := k.columnLookup[column]; !expectedColumn {
+				// Silently discard the column+data
+				continue
+			}
+
+			filteredRow[column] = data
+
+			// No need to check the rest of the row
+			if !includeRow {
```
zackattack01, line 101: how does `includeRow` become false?

RebeccaMahany: Oh, haha -- it's not actually needed anymore. This was leftover from when I was doing additional row filtering here. Updated to remove, thank you!

ee/katc/table.go
```diff
@@ -0,0 +1,178 @@
...
+	if sourceConstraints == nil {
+		return true, nil
+	}
+
+	validSource := true
+	for _, sourceConstraint := range sourceConstraints.Constraints {
+		switch sourceConstraint.Operator {
+		case table.OperatorEquals:
+			if source != sourceConstraint.Expression {
+				validSource = false
```
James-Pickett, line 126: why not just `return false, nil` in these spots instead of having the `validSource` variable?

RebeccaMahany: Nice catch, thank you! No good reason

zackattack01 (approved):
🔥

ee/katc/config.go
```diff
@@ -0,0 +1,120 @@
...
+	rows []map[string][]byte
+}
+
+const (
+	sqliteSourceType = "sqlite"
+)
+
+func (kst *katcSourceType) UnmarshalJSON(data []byte) error {
+	var s string
+	err := json.Unmarshal(data, &s)
```
James-Pickett, line 36: What is the purpose of doing a json.Unmarshall here? If I understand correctly, were just passing a single word string as []bytes? Is it just to remove the quotes? That's okay I suppose, could you please leave a comment?

RebeccaMahany: Yes, we are just processing a single-word string here. This means we use the official implementation for unmarshalling JSON strings, which felt a little more correct than casting to bytes. It's also how I've seen it done in examples and in the docs -- e.g. see the `Example (CustomMarshalJSON)` at the top [here](https://pkg.go.dev/encoding/json#pkg-overview).

PR #1764, opened 2024-06-27
refactor power_event_watcher for configurable subscriber behavior

zackattack01:
To support the windows watchdog efforts, we will be re-using the `power_event_watcher` code from the watchdog service. Currently the power_event_watcher component takes a knapsack reference and directly manipulates the `InModernStandby` value as a result of the observed events. These are also persisted to bbolt, making this a little tricky for sharing with a separate service as is.

We can't simply read the values from bbolt within watchdog because if launcher proper stops functioning (and requires watchdog intervention) we shouldn't trust the latest stored value.

We also likely won't want two versions of the same logs always being added from both watchdog and launcher, and I didn't want to just duplicate all of this code. This was the cleanest approach I could come up with, and should allow watchdog to simply define a new `powerEventSubscriber` (and avoid knapsack/stores setup). happy to hear any alternative suggestions!

### DIFF
--- a/ee/powereventwatcher/power_event_watcher_windows.go
+++ b/ee/powereventwatcher/power_event_watcher_windows.go
@@ -47,9 +47,9 @@ type (
 		OnStartup() error
 	}
 
-	// knapsackSubscriber implements the powerEventSubscriber interface and
+	// knapsackSleepStateUpdater implements the powerEventSubscriber interface and
 	// updates the knapsack.InModernStandby state based on the power events observed
-	knapsackSubscriber struct {
+	knapsackSleepStateUpdater struct {
 		knapsack types.Knapsack
 		slogger  *slog.Logger
 	}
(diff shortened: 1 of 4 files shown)


ee/powereventwatcher/power_event_watcher_windows.go
```diff
@@ -28,14 +28,31 @@ type (
...
+		// OnPowerEvent will be called for the provided subscriber whenever any watched event is observed
+		OnPowerEvent(eventID int) error
+		// OnStartup will be called when the powerEventWatcher is initially set up, allowing subscribers
+		// to perform any setup behavior (e.g. cache clearing, state resetting)
+		OnStartup() error
+	}
+
+	// knapsackSubscriber implements the powerEventSubscriber interface and
+	// updates the knapsack.InModernStandby state based on the power events observed
+	knapsackSubscriber struct {
```
RebeccaMahany, line 52: This might just be me, but I was confused because `knapsackSubscriber` sounds like something subscribing to the knapsack and not (effectively) the knapsack subscribing to something else. What about something like `knapsackUpdater` (or something along those lines but more specific, like `knapsackSleepStateUpdater`)?

RebeccaMahany (commented):
I like this approach! Just had a nit about naming

ee/powereventwatcher/power_event_watcher_windows.go
zackattack01: oh yeah, that makes more sense thank you. will do `knapsackSleepStateUpdater`

RebeccaMahany (approved):

PR #1772, opened 2024-07-08
[KATC] Update config schema, including overlays

RebeccaMahany:
This PR makes the following config updates:

* Rename `source` column to `path` to be more consistent with our e.g. dataflatten tables, which also return a `path` column
* Rename `source` to `source_paths` for clarity; additionally, make it an array rather than a single string to allow for matching against multiple patterns
* Rename `query` to `source_query` to disambiguate from the `query` column we use elsewhere
* Rename `platform` to `filters` to clarify that we may have other types of filters in the future
* Update config structure to include a list of overlays inside the table definition, which will allow us to define the table configuration across multiple platforms; move `filters` inside the overlays

This is an alternative implementation to https://github.com/kolide/launcher/pull/1770.


ee/katc/config.go
```diff
@@ -98,36 +100,43 @@ func (r *rowTransformStep) UnmarshalJSON(data []byte) error {
...
+		Columns           []string                 `json:"columns"`
+		SourceType        katcSourceType           `json:"source_type"`
+		SourcePaths       []string                 `json:"source_paths"` // Describes how to connect to source (e.g. path to db) -- % and _ wildcards supported
+		SourceQuery       string                   `json:"source_query"` // Query to run against each source path
+		RowTransformSteps []rowTransformStep       `json:"row_transform_steps"`
+		Overlays          []katcTableConfigOverlay `json:"overlays"`
+	}
+
+	katcTableConfigOverlay struct {
+		Filters           map[string]string   `json:"filters"` // determines if this overlay is applicable to this launcher installation
```
RebeccaMahany, line 116: May end up going with `Filters []string` pending discussion tomorrow

ee/katc/config.go
```diff
@@ -16,8 +15,11 @@ import (
 // identifier parsed from the JSON KATC config, and the `dataFunc` is the function
 // that performs the query against the source.
 type katcSourceType struct {
-	name     string
-	dataFunc func(ctx context.Context, slogger *slog.Logger, sourcePattern string, query string, sourceConstraints *table.ConstraintList) ([]sourceData, error)
+	name string
+	// `pathConstraints` comes from the live query or scheduled query, and constrains `sourcePaths` to a particular value.
```
directionless, line 19: I'm still confused about what `pathConstraints` achieve that globs in `sourcePaths` do not. 

I might offer that if `sourcePaths` is a list of possible globs, the only thing it _cannot_ do is have negative paths. Like if you had `/Users/%/path/to/file` and wanted to omit `root`, or `/Users/%/.config/netrc%` and wanted to omit `netrc~`. So something as a negative glob can make sense. But I don't understand as a further narrowing. 

Also don't know that the juice is worth the squeeze yet.

ee/katc/config.go
```diff
@@ -98,36 +100,43 @@ func (r *rowTransformStep) UnmarshalJSON(data []byte) error {
...
+type (
+	// katcTableConfig is the configuration for a specific KATC table. The control server
+	// sends down these configurations.
+	katcTableConfig struct {
+		Columns           []string                 `json:"columns"`
+		SourceType        katcSourceType           `json:"source_type"`
+		SourcePaths       []string                 `json:"source_paths"` // Describes how to connect to source (e.g. path to db) -- % and _ wildcards supported
+		SourceQuery       string                   `json:"source_query"` // Query to run against each source path
+		RowTransformSteps []rowTransformStep       `json:"row_transform_steps"`
+		Overlays          []katcTableConfigOverlay `json:"overlays"`
```
directionless, line 112: I suspect this should be `[]katcTableConfig`

ee/katc/config.go
directionless: Becca got me in a huddle and said this is `queryConstraints`, from the query -- it's an optimization. We discussed some renaming to make that clear.

ee/katc/config.go
RebeccaMahany: Discussed in huddle, adding extra context here for others reviewing --

I had originally settled on keeping the structs `katcTableConfig` and `katcTableConfigOverlay` separate because I didn't think it made sense to allow `SourceType`, `SourcePaths`, `SourceQuery`, and `RowTransformSteps` to be nil at the top level (the `katcTableConfig` level), but I felt that it was necessary to allow them to be nil at the overlay level (the `katcTableConfigOverlay` struct).

I am considering:

```
type (
	// katcTableConfig is the configuration for a specific KATC table. The control server
	// sends down these configurations.
	katcTableConfig struct {
		Columns []string `json:"columns"`
		katcTableDefinition
		Overlays []katcTableConfigOverlay `json:"overlays"`
	}

	katcTableConfigOverlay struct {
		Filters map[string]string `json:"filters"` // determines if this overlay is applicable to this launcher installation
		katcTableDefinition
	}

	katcTableDefinition struct {
		SourceType        *katcSourceType     `json:"source_type,omitempty"`
		SourcePaths       *[]string           `json:"source_paths,omitempty"` // Describes how to connect to source (e.g. path to db) -- % and _ wildcards supported
		SourceQuery       *string             `json:"source_query,omitempty"` // Query to run against each source path
		RowTransformSteps *[]rowTransformStep `json:"row_transform_steps,omitempty"`
	}
)
```

However, I'm still not sure I like allowing `SourceType` to be nil at the top level. I am going to mull it over some more and come back tomorrow! But wanted to leave my thoughts in case folks wanted to chime in.

RebeccaMahany: Adding -- I did not want to combine the structs completely because I think it is incorrect to allow an overlay to define `Columns`.

directionless: > However, I'm still not sure I like allowing SourceType to be nil at the top level. I am going to mull it over some more and come back tomorrow! But wanted to leave my thoughts in case folks wanted to chime in.

Why not?

> I did not want to combine the structs completely because I think it is incorrect to allow an overlay to define Columns.

Why not?

RebeccaMahany: re SourceType:

Feels messy? I should just try implementing it and see if it actually feels annoying to work around or not.

re columns:

I think the columns that a table returns should be consistent.

ee/control/consumers/keyvalueconsumer/config_consumer.go
```diff
@@ -0,0 +1,47 @@
...
+
+	return c
+}
+
+func (c *ConfigConsumer) Update(data io.Reader) error {
+	if c == nil {
+		return errors.New("key value consumer is nil")
+	}
+
+	var kvPairs map[string]any
```
RebeccaMahany, line 29: Discovered when testing with @jbeker that the existing KV consumer in this package does not work for this config because Golang won't unmarshal the config as a `map[string]string` since it's a more complex object, which makes sense.

This is a quick fix, but maybe we want something cleverer?

ee/katc/config.go
directionless: I would expect us to need to vary source by platform _somewhere_. That's exactly the kind of nonsense I expect. We see this commonly in things like registry values vs plist files, but it's possible that it will make more sense to handle _that_ skew in SQL and not here.

directionless: columns I could go either way for. I agree it feels a little weird to vary. Kinda touches your question about dataflatten. I think we should do whatever for columns, and we can update it if we find use cases.

ee/katc/table.go
```diff
@@ -39,21 +43,60 @@ func newKatcTable(tableName string, cfg katcTableConfig, slogger *slog.Logger) (
...
+		}
+
+		break
+	}
+
+	// Add extra fields to slogger
+	k.slogger = slogger.With(
+		"table_name", tableName,
+		"table_type", cfg.SourceType,
+		"table_source_paths", cfg.SourcePaths,
```
directionless, line 82:
```suggestion
		"table_type", cfg.SourceType,
```
I'd probably leave this out -- I think we're likely to path large arrays of them in. There's value in know what path actually matches on disk. but probably not the full paths array

ee/katc/table.go
```diff
@@ -103,30 +146,29 @@ func (k *katcTable) generate(ctx context.Context, queryContext table.QueryContex
 	return filteredResults, nil
 }
 
-// getSourceConstraint retrieves any constraints against the `source` column
-func getSourceConstraint(queryContext table.QueryContext) *table.ConstraintList {
-	sourceConstraint, sourceConstraintExists := queryContext.Constraints[sourceColumnName]
-	if sourceConstraintExists {
-		return &sourceConstraint
+// getPathConstraint retrieves any constraints against the `path` column
+func getPathConstraint(queryContext table.QueryContext) *table.ConstraintList {
```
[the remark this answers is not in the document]
directionless: Oh, I see you're handling LIKE and GLOB as well. I'm not sure osquery passes those, but that's nice completeness.

directionless (approved):

ee/katc/indexeddb_leveldb.go
```diff
@@ -15,39 +15,44 @@ import (
...
+		// Extract database and table from query
+		dbName, objectStoreName, err := extractQueryTargets(query)
 		if err != nil {
-			return nil, fmt.Errorf("querying %s: %w", db, err)
+			return nil, fmt.Errorf("getting db and object store names: %w", err)
+		}
+
+		// Query databases
+		for _, db := range leveldbs {
+			// Check to make sure `db` adheres to pathConstraintsFromQuery
```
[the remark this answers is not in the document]
RebeccaMahany: Applied these changes! (I updated both of these to say "osquery sqlite filtering" because I was confused by the original wording -- hopefully that conveys the meaning intended?)

ee/control/consumers/keyvalueconsumer/config_consumer.go
RebeccaMahany: Correct, this handles the error that Jeremy and I were seeing: `failed to decode key-value json: json: cannot unmarshal object into Go value of type string`.

I wasn't sure if it would be cleaner for K2 to update or not.

ee/katc/config.go
RebeccaMahany: Ok, I'll go with the option I posted above, which will let source type be nil at the top level!

James-Pickett (approved):
nice

zackattack01 (approved):

ee/control/consumers/keyvalueconsumer/config_consumer.go
directionless: It's going to be pretty trivial on the k2 side. And I think it's pretty trivial here. 

So I think we get to decide what feels like a cleaner protocol. 

Having everything as a string _feels_ clean, but having the KATC definitions marshalled into strings feels kinda weird. 🤷

RebeccaMahany: I guess I like it this way a little better, then -- it feels tidy that k2 doesn't need to know about the marshalling and unmarshalling to a string; launcher handles that itself for ease of storage in bucket/startup_settings.

directionless: It's step towards a more structured interface and API between k2 and launcher. (vs what we have now, which feels like it requires a very complete knowledge of all the parts)

directionless (approved):
Let's try!

PR #1780, opened 2024-07-16
Don't store hashes for unknown subsystems

RebeccaMahany:
We've run into an issue a couple of times now:

1. We create a new subsystem on the server side, and write launcher code to handle this subsystem
2. We release the new subsystem first
3. Older versions of launcher store the hashes for the new subsystem's data, but they do not process that data because they don't know how to yet
4. launcher autoupdates to a new version, but does not know to fetch the data anew because the hashes are cached

The solution this PR proposes is to skip step 3: i.e., if a subsystem is unknown to launcher, do not store the hash of its data. That way, when launcher autoupdates and learns about this new subsystem, it will fetch that subsystem's data fresh for the first time.

I audited to check that I hadn't left off any subsystems. The only one this PR does not add is `desktop_notifier`, which is a legacy subsystem that has since moved under `actions`.

### DIFF
--- a/ee/control/control.go
+++ b/ee/control/control.go
@@ -35,6 +35,7 @@ type ControlService struct {
 	lastFetched          map[string]string
 	consumers            map[string]consumer
 	subscribers          map[string][]subscriber
+	knownSubsystems      map[string]struct{}
 }
 
 // consumer is an interface for something that consumes control server data updates. The
@@ -67,6 +68,7 @@ func New(k types.Knapsack, fetcher dataProvider, opts ...Option) *ControlService
 		lastFetched:          make(map[string]string),
 		consumers:            make(map[string]consumer),
 		subscribers:          make(map[string][]subscriber),
+		knownSubsystems:      make(map[string]struct{}),
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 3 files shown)


zackattack01 (approved):

ee/control/control.go
```diff
@@ -35,6 +35,7 @@ type ControlService struct {
 	lastFetched          map[string]string
 	consumers            map[string]consumer
 	subscribers          map[string][]subscriber
+	knownSubsystems      map[string]struct{}
```
directionless, line 38: Would `len(subscribers[...]) > 0` have been the same?

RebeccaMahany: I don't think so, though I might be misinterpreting your suggestion -- there can be a valid subsystem that has a consumer but no subscribers.

directionless: `len` of both?

directionless: I'm saying that we probably don't need `knownSubsystems`, we should be able to derive it from `consumers` and `subscribers`. And that this is infrequent enough there's no performance win in caching it. 

It's also pretty harmless, but I started by wanting to quibble about names, and then I realized we didn't need it at all.

RebeccaMahany: I'll update in a new PR

PR #1781, opened 2024-07-16
[KATC] Set immutable option when opening sqlite db

RebeccaMahany:
Fixes the following:

> error generating table: fetching data: querying /path/to/some/db.sqlite: running query: database is locked (5) (SQLITE_BUSY)

### DIFF
--- a/ee/katc/sqlite.go
+++ b/ee/katc/sqlite.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"io"
 	"log/slog"
+	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/kolide/launcher/ee/agent"
 	"github.com/osquery/osquery-go/plugin/table"
 	_ "modernc.org/sqlite"
 )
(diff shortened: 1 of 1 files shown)


ee/katc/sqlite.go
```diff
@@ -118,3 +132,34 @@ func querySqliteDb(ctx context.Context, slogger *slog.Logger, path string, query
 
 	return results, nil
 }
+
+// copySqliteDb makes a temporary directory and copies the given db into it.
+func copySqliteDb(path string) (string, error) {
+	dbCopyDir, err := agent.MkdirTemp("sqlite-temp")
```
directionless, line 138: If we have to, we have to... Did we explore the sqlite options that might alleviate this?

RebeccaMahany: `mode=ro` wasn't making a difference -- let me try setting `immutable` to see!

RebeccaMahany: It works! Thank you, that's much tidier

zackattack01 (approved):

ee/katc/sqlite.go
directionless: The docs say `database is opened read-only and all locking and change detection is disabled` so I was hopeful

PR #1782, opened 2024-07-17
Check for known subsystem against consumers and subscribers

RebeccaMahany:
Updates https://github.com/kolide/launcher/pull/1780

Addresses comment thread: https://github.com/kolide/launcher/pull/1780#discussion_r1680233506

### DIFF
--- a/ee/control/control.go
+++ b/ee/control/control.go
@@ -367,13 +367,14 @@ func (cs *ControlService) fetchAndUpdate(subsystem, hash string) error {
 // knownSubsystem checks our registered consumers and subscribers to see if the given
 // subsystem is one that has been registered with the control service.
 func (cs *ControlService) knownSubsystem(subsystem string) bool {
-	if _, subsystemFromConsumer := cs.consumers[subsystem]; !subsystemFromConsumer {
-		if _, subsystemFromSubscriber := cs.subscribers[subsystem]; !subsystemFromSubscriber {
-			return false
-		}
+	if _, subsystemFromConsumer := cs.consumers[subsystem]; subsystemFromConsumer {
+		return true
+	}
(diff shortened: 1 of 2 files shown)


ee/control/control.go
```diff
@@ -366,20 +364,30 @@ func (cs *ControlService) fetchAndUpdate(subsystem, hash string) error {
...
 }
 
+// knownSubsystem checks our registered consumers and subscribers to see if the given
+// subsystem is one that has been registered with the control service.
+func (cs *ControlService) knownSubsystem(subsystem string) bool {
+	if _, subsystemFromConsumer := cs.consumers[subsystem]; !subsystemFromConsumer {
+		if _, subsystemFromSubscriber := cs.subscribers[subsystem]; !subsystemFromSubscriber {
+			return false
+		}
+	}
```
directionless, line 374:
```suggestion
	if _, subsystemFromConsumer := cs.consumers[subsystem]; subsystemFromConsumer {
	  return true
	}
	if _, subsystemFromSubscriber := cs.subscribers[subsystem]; subsystemFromSubscriber {
			return true
	}
	
	return false
```

I think? But it seems fine either way

directionless (approved):

ee/control/control.go
RebeccaMahany: Oh yup, reads better without the additional nesting -- updating now

ee/control/control_test.go
```diff
@@ -410,6 +410,71 @@ func TestControlServicePersistLastFetched(t *testing.T) {
 	}
 }
 
+func Test_knownSubsystem(t *testing.T) {
```
RebeccaMahany, line 413: This is also tested within the context of `Fetch` in `TestControlServiceFetch_IgnoresUnknownSubsystems`

directionless (approved):
I think cleaner. Yay

PR #1783, opened 2024-07-17
Add new values to `kolide_remotectl`

Micah-Kolide:
With MacOS 15 there are some new values returned from `remotectl dumpstate`. This broke the parser, but fortunately it's a small change to fix. The new `EncryptedRemoteXPCPopulatedOIDs` is the same format as `Features`, so I could use that parse method. However, `Identity` was also added with a slightly different data format which I added a new method for.

I've added my test data for this change, but it shouldn't change how it worked prior to this PR.

### DIFF
--- a/ee/tables/execparsers/remotectl/parse.go
+++ b/ee/tables/execparsers/remotectl/parse.go
@@ -117,6 +117,15 @@ func (p *parser) parseDevice() (map[string]interface{}, error) {
 			continue
 		}
 
+		if strings.HasPrefix(strings.TrimSpace(p.lastReadLine), "Identity:") {
+			identity, err := p.parseKeyValList()
+			if err != nil {
+				return nil, err
+			}
+
+			deviceResults["Identity"] = identity
+		}
+
 		if p.isDeviceDelimiter() {
 			return deviceResults, nil
 		}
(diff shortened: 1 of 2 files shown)


Micah-Kolide:
```
osquery> SELECT * FROM kolide_remotectl WHERE parent IN ('Local device/Properties/EncryptedRemoteXPCPopulatedOIDs', 'Local device/Identity');
+-----------------------------------------------------------+---------------------------------------------------------+-------------------+----------------------------------------------+-------+
| fullkey                                                   | parent                                                  | key               | value                                        | query |
+-----------------------------------------------------------+---------------------------------------------------------+-------------------+----------------------------------------------+-------+
| Local device/Identity/Public key SHA256                   | Local device/Identity                                   | Public key SHA256 | /3829hunfe927fh972eht972efhjeo82hfo2ehgo2ru= | *     |
| Local device/Properties/EncryptedRemoteXPCPopulatedOIDs/0 | Local device/Properties/EncryptedRemoteXPCPopulatedOIDs | 0                 | 1.2.111.000000.100.6.22                      | *     |
| Local device/Properties/EncryptedRemoteXPCPopulatedOIDs/1 | Local device/Properties/EncryptedRemoteXPCPopulatedOIDs | 1                 | 1.2.111.000000.100.6.23                      | *     |
+-----------------------------------------------------------+---------------------------------------------------------+-------------------+----------------------------------------------+-------+
```

directionless (approved):

RebeccaMahany (approved):

PR #1785, opened 2024-07-17
[KATC] Do not allow indexeddb row data to be overwritten

RebeccaMahany:
Fixes an issue where we failed to deserialize data because it had an unexpected format. The underlying cause appears to be that we were continuously overwriting data in the results by setting data equal to `iter.Value()`. I am not totally sure why this issue only appeared on three out of ~25+ test devices, but I can confirm that this fixes the issue because my test device was luckily one of the ones affected. 😅 

I also added a little more info to error messages to improve troubleshooting ability.


zackattack01 (approved):
nice find!

Issue #1776, opened 2024-07-12
Something unclear about MSI re-installs and a missing osqueryd


directionless:
This is a placeholder for discussion happening in https://kolide.slack.com/archives/CGFJY1SP2/p1720646284831119

PR #1788, opened 2024-07-18
download osqueryd when missing on startup

zackattack01:
resolves https://github.com/kolide/launcher/issues/1776, where we discovered that installing an MSI with an older version of osqueryd over a newer version leaves the launcher installation with _no_ osqueryd.exe on windows.

There are several potential approaches here (e.g. via WiX configuration changes, path versioning etc.) but we already have all of the logic we need to download osqueryd (via autoupdates) so lets do that.

There are also several different ways to integrate/repurpose the autoupdate logic for this and I'd love to hear any alternative suggestions- this was the cleanest approach I could find that didn't require excessive refactoring of that code.

Tested by building a custom 1.9.3 MSI with these changes + an older osqueryd binary. Installing removes the osqueryd.exe on disk (as expected), but the download is triggered immediately and launcher keeps going happily once that process finishes

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -540,6 +540,33 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 		if actionsQueue != nil {
 			actionsQueue.RegisterActor(tuf.AutoupdateSubsystemName, tufAutoupdater)
 		}
+
+		// in some cases, (e.g. rolling back a windows installation to a previous osquery version) it is possible that
+		// the installer leaves us in a situation where there is no osqueryd on disk.
+		// we can detect this and attempt to download the correct version into the TUF update library to run from that.
+		// This must be done as a blocking operation before the rungroups start, because the osquery runner will fail to
+		// launch and trigger a restart immediately
+		currentOsquerydBinaryPath := k.LatestOsquerydPath(ctx)
+		if _, err = os.Stat(currentOsquerydBinaryPath); os.IsNotExist(err) {
+			slogger.Log(ctx, slog.LevelInfo,
+				"detected missing osqueryd executable, will attempt to download",
+			)
+
+			// simulate control server request for immediate update, noting to bypass the initial delay window
+			actionReader := strings.NewReader(`{
+				"bypass_initial_delay": true,
+				"binaries_to_update": [
+					{ "name": "osqueryd" }
+				]
+			}`)
+
+			if err = tufAutoupdater.Do(actionReader); err != nil {
+				slogger.Log(ctx, slog.LevelError,
+					"failure triggering immediate osquery update",
+					"err", err,
+				)
+			}
+		}
 	}
 
 	startupSpan.End()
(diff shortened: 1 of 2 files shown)


ee/tuf/autoupdate.go
```diff
@@ -62,7 +62,8 @@ const AutoupdateSubsystemName = "autoupdate"
 
 type (
 	controlServerAutoupdateRequest struct {
-		BinariesToUpdate []binaryToUpdate `json:"binaries_to_update"`
+		BinariesToUpdate   []binaryToUpdate `json:"binaries_to_update"`
+		BypassInitialDelay bool             `json:"bypass_initial_delay,omitempty"`
```
RebeccaMahany, line 66: I like this

RebeccaMahany (approved):
Cool! This is a really tidy solution.

ee/tuf/autoupdate.go
```diff
@@ -298,6 +289,16 @@ func (ta *TufAutoupdater) Do(data io.Reader) error {
...
+	if time.Now().Before(ta.initialDelayEnd) && !updateRequest.BypassInitialDelay {
+		ta.slogger.Log(context.TODO(), slog.LevelWarn,
+			"received update request during initial delay, discarding",
+			"initial_delay_end", ta.initialDelayEnd.UTC().Format(time.RFC3339),
+		)
+		// We don't return an error because there's no need for the actionqueue to retry this request --
+		// we're going to perform an autoupdate check as soon as we exit the delay anyway.
+		return nil
+	}
+
```
RebeccaMahany, line 301: Oh, will we see an issue with the osquery restart func being called before we've actually started an osquery instance?

zackattack01: this is triggered inline (blocking) before any of the rungroups are started, that was the only way I could find to avoid the osq runner erroring out and restarting everything. so I think we're safe there, if that's what you mean?

zackattack01: justification was there's really no reason to avoid hanging up the main thread or bypassing initial delay when nothing will work anyway. and `launchOsqueryInstance` does a fresh path lookup at runtime which in testing handled this all perfectly already

RebeccaMahany: Yeah, was asking if it was safe to call the restart func before the rungroup started! Took a quick look -- do we need a `r.instance != nil` check in the restart func? https://github.com/kolide/launcher/blob/main/pkg/osquery/runtime/runner.go#L207

RebeccaMahany: Thank you!

zackattack01: 🤦 sorry I see what you meant, yeah it seems to be okay! Thank you!

cmd/launcher/launcher.go
```diff
@@ -540,6 +540,33 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
...
+
+			// simulate control server request for immediate update, noting to bypass the initial delay window
+			actionReader := strings.NewReader(`{
+				"bypass_initial_delay": true,
+				"binaries_to_update": [
+					{ "name": "osqueryd" }
+				]
+			}`)
+
+			if err = tufAutoupdater.Do(actionReader); err != nil {
```
directionless, line 563: This _is_ very elegant. I might suggest exposing `tufAutoupdater.DoNow` or something, because hardcoding the wireprotocol here will totally come back to bite us. But this is very tidy.

PR #1790, opened 2024-07-22
Remove unused parameter

RebeccaMahany:
Cleaning up an unused parameter in the sqlite package.


directionless (approved):
I mean, okay. But it feels wrong. 😆 

Migrations should have timeouts, and that should be linked back to a context...

RebeccaMahany:
@directionless -- I agree, but unfortunately the migration package doesn't provide `WithContext` versions of these functions 😞

PR #1792, opened 2024-07-23
Log table_type correctly in KATC logs

RebeccaMahany:
Fixes `table_type` being set to `{}` in logs.

### DIFF
--- a/ee/katc/table.go
+++ b/ee/katc/table.go
@@ -86,9 +86,13 @@ func newKatcTable(tableName string, cfg katcTableConfig, slogger *slog.Logger) (
 	}
 
 	// Add extra fields to slogger
+	var tableType string
+	if cfg.SourceType != nil {
+		tableType = cfg.SourceType.name
+	}
 	k.slogger = slogger.With(
 		"table_name", tableName,
-		"table_type", cfg.SourceType,
+		"table_type", tableType,
 	)
 
 	return &k, columns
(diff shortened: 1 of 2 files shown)


ee/katc/table.go
```diff
@@ -86,9 +86,13 @@ func newKatcTable(tableName string, cfg katcTableConfig, slogger *slog.Logger) (
 	}
 
 	// Add extra fields to slogger
+	var tableType string
+	if cfg.SourceType != nil {
+		tableType = cfg.SourceType.name
+	}
```
directionless, line 92: I'll approve, but... should `katcSourceType` gain a `Name()` or `String()`?

directionless (approved):
Would this be

ee/katc/table.go
RebeccaMahany: So, I implemented `func (kst *katcSourceType) String() string` hoping slogger would just make it work automagically, and it didn't. It doesn't hurt to have regardless, though. I'll update.

RebeccaMahany:
> Would this be

@directionless did some of this comment get truncated?

zackattack01 (approved):

James-Pickett (approved):

directionless:
> > Would this be
> 
> @directionless did some of this comment get truncated?

Didn't delete it all. Was replaced by the inline comment thread

PR #1794, opened 2024-07-24
[IndexedDB] Generate test data; support dense arrays

RebeccaMahany:
The improvements in this PR should make it significantly easier to test a variety of data shapes without acquiring actual IndexedDBs. This PR allows for generating a test database, and adds tests that use the same.

Along the way, I added support for dense arrays.


ee/indexeddb/values.go
```diff
@@ -296,6 +302,69 @@ func deserializeSparseArray(ctx context.Context, slogger *slog.Logger, srcReader
 	return arrBytes, nil
 }
 
+// deserializeDenseArray deserializes the next dense array from the srcReader.
+// Dense arrays are arrays of items that are NOT paired with indices, as in spare arrays.
```
zackattack01, line 306:
```suggestion
// Dense arrays are arrays of items that are NOT paired with indices, as in sparse arrays.
```

RebeccaMahany: Whoops, thank you!

zackattack01 (approved):
great idea! and test emoji choices

PR #1796, opened 2024-07-24
build windows msi for arm and amd

James-Pickett:
This PR updates package builder to add an arm64 binaries and service config to the MSI. This is done putting both of 2 services / binaries in to the MSI with mutually exclusive conditions based on processor architecture. 

To accomplish we are using heat to harvest files set up like ...

```
└───pkg_root
    └───Launcher-kolide-k2
        ├───bin
        │   ├───amd64
        │   │       launcher.exe
        │   │       osqueryd.exe
        │   │
        │   └───arm64
        │           launcher.exe
        │           osqueryd.exe
        │
        └───conf
                ...
```

then post processing the generated xml to remove the amd64 and arm64 directory tags so that all the exe files are in the same directory in the xml, then add the mutually exclusive tags to the binary elements

Additionally, the PR adds the windows service recovery settings to go code because the wix extension we were using for that is not compatible with this approach since it's primary key cannot be changed form the service name and would cause unique constraint error (now there are 2 in the MSI). The wix extension for recovery actions is still applied to the amd install.

relates to https://github.com/kolide/launcher/issues/1149


pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +224,81 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
 	lines := strings.Split(string(heatContent), "\n")
 	for _, line := range lines {
+
+		if currentArchSpecificBinDir != none && strings.Contains(line, "</Directory>") {
+			// were in a arch specific bin dir that we want to remove, don't write closing tag
+			currentArchSpecificBinDir = none
+			continue
+		}
+
+		if strings.Contains(line, "Directory") {
```
James-Pickett, line 247:
```suggestion
		if strings.Contains(line, "<Directory>") {
```

pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +224,81 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
+			if strings.Contains(line, string(amd64)) {
+				// were in a arch specific bin dir that we want to remove so when we hit closing tag, we'll skip it
```
James-Pickett, line 249:
```suggestion
				// were in a arch specific bin dir that we want to remove, skip opening tag and set current arch specific bin dir so we'll skip closing tag as well
```

pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +224,81 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
+				currentArchSpecificBinDir = amd64
+				continue
+			}
+
+			if strings.Contains(line, string(arm64)) {
+				// were in a arch specific bin dir that we want to remove so when we hit closing tag, we'll skip it
```
James-Pickett, line 255:
```suggestion
				// were in a arch specific bin dir that we want to remove, skip opening tag and set current arch specific bin dir so we'll skip closing tag as well
```

pkg/packaging/detectLauncherVersion.go
```diff
@@ -19,7 +19,13 @@ func (p *PackageOptions) detectLauncherVersion(ctx context.Context) error {
 	logger := log.With(ctxlog.FromContext(ctx), "library", "detectLauncherVersion")
 	level.Debug(logger).Log("msg", "Attempting launcher autodetection")
 
-	launcherPath := p.launcherLocation(filepath.Join(p.packageRoot, p.binDir))
+	binDir := filepath.Join(p.packageRoot, p.binDir)
+	if p.target.Platform != Darwin {
```
James-Pickett, line 23: explain this with comment

pkg/packaging/packaging.go
```diff
@@ -32,37 +32,39 @@ const (
...
+	WixSkipCleanup     bool
+	DisableService     bool
 
 	// Normally we'd download the same version we bake into the
 	// autoupdate. But occasionally, it's handy to make a package
 	// with a different version.
-	LauncherDownloadVersionOverride string
-	OsqueryDownloadVersionOverride  string
+	LauncherDownloadVersionOverride    string
+	LauncherArmDownloadVersionOverride string
```
James-Pickett, line 66: if were going to add for launcher, should also add for osq?

cmd/package-builder/package-builder.go
```diff
@@ -79,6 +79,11 @@ func runMake(args []string) error {
 			env.String("LAUNCHER_VERSION", "stable"),
 			"What TUF channel to download launcher from. Supports filesystem paths",
 		)
+		flLauncherArmVersion = flagset.String(
+			"launcher_arm_version",
+			env.String("LAUNCHER_ARM_VERSION", "stable"),
+			"What TUF channel to download launcher from for ARM. Supports filesystem paths",
+		)
```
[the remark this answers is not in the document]
James-Pickett: I added this just so that I could specify a local arm build to package. I can't imagine a situation where we would want the arm and amd binaries to have different code.

This has me thinking though that maybe we should add in some safe guards to ensure the arm and amd versions are the same. If we some how managed to get into a situation where TUF updated one and not the other, we could end up with a package different arches are on different versions.

pkg/packaging/packaging.go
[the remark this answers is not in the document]
James-Pickett: removed

pkg/packagekit/wix/service.go
```diff
@@ -90,32 +89,14 @@ type ServiceControl struct {
 // This is used needed to set DelayedAutoStart
 type ServiceConfig struct {
 	// TODO: this should need a namespace, and yet. See https://github.com/golang/go/issues/36813
+	Id               string    `xml:",attr,omitempty"`
```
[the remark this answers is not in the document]
James-Pickett: yea, I left a comment about it

cmd/package-builder/package-builder.go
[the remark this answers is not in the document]
James-Pickett: I can't find a coherent way to verify the versions are the same. However, I set up the flag so that it's default is an empty string and it will just set that flag to what ever `launcher_version` is if no `launcher_arm_version` is provided.

cmd/launcher/svc_config_windows.go
```diff
@@ -184,3 +187,56 @@ func checkRestartActions(logger *slog.Logger) {
 
 	logger.Log(context.TODO(), slog.LevelInfo, "successfully set RecoveryActionsOnNonCrashFailures flag")
 }
+
+// setRecoveryActions sets the recovery actions for the launcher service.
+// previously defined via wix ServicConfig Element (Util Extension) https://wixtoolset.org/docs/v3/xsd/util/serviceconfig/
+func setRecoveryActions(logger *slog.Logger) {
```
directionless, line 193: Consider passing in ctx

cmd/package-builder/package-builder.go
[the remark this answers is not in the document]
directionless: I'm not sure there's a good way to validate a launcher version, without invoking it. Maaaybe we can string parse it, but I'm not sure it's worth it. (And given what Zack recently learned about file versions and MSIs, let's not go there...)

Maybe this is an argument for being _less_ clever. Command line either takes `launcher_version=` and downloads from TUF (we should have enough metadata there to validate that `stable` is the same) Or `use_local_binary=...` and it ignores TUF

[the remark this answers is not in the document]
directionless: I guess that's kinda what would happen. So maybe I'm over thinking it. Maybe we validate the versions from tuf, and skip that validation on local paths.

directionless (commented):
The cross platform stuff is _really neat_. 

But I'm nervous about moving those restart things out.

directionless:
Hrm. I kinda want to look at the XML. I know I never baked in something to save it. How do you think that would be to add?

cmd/launcher/svc_config_windows.go
James-Pickett: passing ctx

James-Pickett:
> Hrm. I kinda want to look at the XML. I know I never baked in something to save it. How do you think that would be to add?

There is a flag `wix_skip_cleanup` already there, I updated it so that it prints the paths to the temp files when clean up gets skipped

cmd/package-builder/package-builder.go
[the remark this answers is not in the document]
James-Pickett: invoking it could be troublesome since we'll have 2 binaries with to different arches, An arm64 machine will probably emulate the amd64 binary and run it, but the reverse doesn't work.

I think just having separate flags for paths is probably good with only a single option for `launcher_version` which is downloaded form tuf for both arches.

pkg/packaging/packaging.go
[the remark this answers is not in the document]
directionless: (I'm not sure we still need it)

directionless:
> > Hrm. I kinda want to look at the XML. I know I never baked in something to save it. How do you think that would be to add?
> 
> @directionless , there is a flag `wix_skip_cleanup` already there, I updated it so that it prints the paths to the temp files when clean up gets skipped

Can you attach one here?

James-Pickett:
> > > Hrm. I kinda want to look at the XML. I know I never baked in something to save it. How do you think that would be to add?
> > 
> > 
> > @directionless , there is a flag `wix_skip_cleanup` already there, I updated it so that it prints the paths to the temp files when clean up gets skipped
> 
> Can you attach one here?
@directionless 
```xml
<?xml version="1.0" encoding="utf-8"?>
<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
    <Fragment>
        <DirectoryRef Id="PROGDIR">
            <Directory Id="dir9E2FF676320D7BE586E7569535A94904" Name="Launcher-kolide-k2">
                <Directory Id="dir8782311652F143EF6DD238714EB2C5A3" Name="bin">
                        <Component Id="cmpFDADA56D9962ABBCAA5F3803F3A0E022" Guid="84DB7147-05C9-49ED-ACBA-E6CA9AA3C070">
                            <File Id="fil604CDF77A839EF3F45370B06E0D217F0" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\bin\amd64\launcher.exe" />
<Condition> %PROCESSOR_ARCHITECTURE="AMD64" </Condition>
                    <ServiceInstall Account="[SERVICEACCOUNT]" Arguments="svc -config &#34;C:\Program Files\Kolide\Launcher-kolide-k2\conf\launcher.flags&#34;" Description="The Kolide Launcher (kolide-k2)" ErrorControl="normal" Id="LauncherKolideK2Svc01J423A2NVBYJE6VWSZHZAYKK8" Name="LauncherKolideK2Svc" Start="auto" Type="ownProcess" Vital="yes">
                        <ServiceConfig xmlns="http://schemas.microsoft.com/wix/2006/wi" Id="LauncherKolideK2Svc01J423A2NVBYJE6VWSZHZAYKK8" DelayedAutoStart="no" OnInstall="yes" OnReinstall="yes"></ServiceConfig>
                        <ServiceDependency Id="Dnscache" Group="no"></ServiceDependency>
                    </ServiceInstall>
                    <ServiceControl Name="LauncherKolideK2Svc" Id="LauncherKolideK2Svc01J423A2NVBYJE6VWSZHZAYKK8" Remove="uninstall" Start="install" Stop="both" Wait="no"></ServiceControl>
                        </Component>
                        <Component Id="cmp2495C4D31241909F181B39921D2280C9" Guid="03B7DB5F-97AF-46B6-BBC1-C96B7ACEA8BB">
                            <File Id="fil526CC2960E9DEE23BDA92EE29CE5A6A4" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\bin\amd64\osqueryd.exe" />
<Condition> %PROCESSOR_ARCHITECTURE="AMD64" </Condition>
                        </Component>
                        <Component Id="cmpABE14524530AA1F1CB8D6939D826DC32" Guid="F7854DA5-4627-44A8-841B-56BB03F37AB8">
                            <File Id="fil4EFEFF758D7317539AD8F04A59622E4B" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\bin\arm64\launcher.exe" />
<Condition> %PROCESSOR_ARCHITECTURE="ARM64" </Condition>
                    <ServiceInstall Account="[SERVICEACCOUNT]" Arguments="svc -config &#34;C:\Program Files\Kolide\Launcher-kolide-k2\conf\launcher.flags&#34;" Description="The Kolide Launcher (kolide-k2)" ErrorControl="normal" Id="LauncherKolideK2Svc01J423A2NVESRWGYKH5D97DDP4" Name="LauncherKolideK2Svc" Start="auto" Type="ownProcess" Vital="yes">
                        <ServiceConfig xmlns="http://schemas.microsoft.com/wix/2006/wi" Id="LauncherKolideK2Svc01J423A2NVESRWGYKH5D97DDP4" DelayedAutoStart="no" OnInstall="yes" OnReinstall="yes"></ServiceConfig>
                        <ServiceDependency Id="Dnscache" Group="no"></ServiceDependency>
                    </ServiceInstall>
                    <ServiceControl Name="LauncherKolideK2Svc" Id="LauncherKolideK2Svc01J423A2NVESRWGYKH5D97DDP4" Remove="uninstall" Start="install" Stop="both" Wait="no"></ServiceControl>
                        </Component>
                        <Component Id="cmp61B74053A68DD7A35C953D86480F9D4C" Guid="BE7AF4FE-B34C-463A-AE16-A343162B7531">
                            <File Id="fil2A8BBBB8EBD5B626DC98686307B2A9C9" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\bin\arm64\osqueryd.exe" />
<Condition> %PROCESSOR_ARCHITECTURE="ARM64" </Condition>
                        </Component>
                </Directory>
                <Directory Id="dirC586BDDEA1A8C8B62AFD7017E646DD85" Name="conf">
                    <Component Id="cmp3B4F901A7E2AE41959321E267F5FA937" Guid="C2F68C91-D693-4F13-B943-6D108A2CE5D6">
                        <File Id="fil18DE5428F20D530D3039426F03C41AC7" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\conf\launcher.flags" />
                    </Component>
                    <Component Id="cmp9DD41EA0BAC4BB356CB0436EC8BAE15D" Guid="687C5B5A-4071-4EAA-88CA-F475CC9A58DE">
                        <File Id="filB41D4879C4A879139C72198A2D8D7862" KeyPath="yes" Source="$(var.SourceDir)\Launcher-kolide-k2\conf\secret" />
                    </Component>
                </Directory>
            </Directory>
        </DirectoryRef>
    </Fragment>
    <Fragment>
        <ComponentGroup Id="AppFiles">
            <ComponentRef Id="cmpFDADA56D9962ABBCAA5F3803F3A0E022" />
            <ComponentRef Id="cmp2495C4D31241909F181B39921D2280C9" />
            <ComponentRef Id="cmpABE14524530AA1F1CB8D6939D826DC32" />
            <ComponentRef Id="cmp61B74053A68DD7A35C953D86480F9D4C" />
            <ComponentRef Id="cmp3B4F901A7E2AE41959321E267F5FA937" />
            <ComponentRef Id="cmp9DD41EA0BAC4BB356CB0436EC8BAE15D" />
        </ComponentGroup>
    </Fragment>
</Wix>

```

cmd/launcher/svc_config_windows.go
```diff
@@ -184,3 +187,56 @@ func checkRestartActions(logger *slog.Logger) {
 
 	logger.Log(context.TODO(), slog.LevelInfo, "successfully set RecoveryActionsOnNonCrashFailures flag")
 }
+
+// setRecoveryActions sets the recovery actions for the launcher service.
+// previously defined via wix ServicConfig Element (Util Extension) https://wixtoolset.org/docs/v3/xsd/util/serviceconfig/
+func setRecoveryActions(_ context.Context, logger *slog.Logger) {
+	sman, err := mgr.Connect()
+	if err != nil {
+		logger.Log(context.TODO(), slog.LevelError,
```
RebeccaMahany, line 196: Might as well use the context passed in for these log messages? Or remove it from the args if we don't need it?

pkg/packagekit/wix/wix.go
```diff
@@ -161,6 +162,12 @@ func New(packageRoot string, identifier string, mainWxsContent []byte, wixOpts .
 // Cleanup removes temp directories. Meant to be called in a defer.
 func (wo *wixTool) Cleanup() {
 	if wo.skipCleanup {
+
+		fmt.Print("skipping cleanup of temp directories\n")
+		for _, d := range wo.cleanDirs {
+			fmt.Printf("skipping cleanup of %s\n", d)
+		}
```
RebeccaMahany, line 169: Are these print statements supposed to be here?

pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +230,93 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
 	lines := strings.Split(string(heatContent), "\n")
 	for _, line := range lines {
+
+		if currentArchSpecificBinDir != none && strings.Contains(line, "</Directory>") {
+			// were in a arch specific bin dir that we want to remove, don't write closing tag
+			currentArchSpecificBinDir = none
+			continue
+		}
+
+		// the directory tag will like like "<Directory Id="xxxx"...>"
```
RebeccaMahany, line 253: (nitpick -- typo 🙂 )

```suggestion
		// the directory tag will look like "<Directory Id="xxxx"...>"
```

pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +230,93 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
 		for _, service := range wo.services {
+
 			isMatch, err := service.Match(line)
 			if err != nil {
 				return fmt.Errorf("match error: %w", err)
 			}
+
 			if isMatch {
+				if currentArchSpecificBinDir == none {
+					return fmt.Errorf("service found, but not in a bin directory")
```
RebeccaMahany, line 283: fmt.Errorf => errors.New?

pkg/packagekit/wix/wix.go
```diff
@@ -223,19 +230,93 @@ func (wo *wixTool) addServices(ctx context.Context) error {
...
+			currentArchSpecificBinDir = none
+			continue
+		}
+
+		// the directory tag will like like "<Directory Id="xxxx"...>"
+		// so we just check for the first part of the string
+		if strings.Contains(line, "<Directory") {
+			if strings.Contains(line, string(amd64)) {
+				// were in a arch specific bin dir that we want to remove, skip opening tag
+				// and set current arch specific bin dir so we'll skip closing tag as well
```
RebeccaMahany, line 258: Somewhere (in an md file in this package or inside the docs folder) could you paste in some of your PR description? The description + the directory tree was helpful for me in understanding what was going on in this PR and I don't want to lose that documentation going forward 🙂

RebeccaMahany (commented):
LGTM, just a couple small comments!

pkg/packagekit/wix/service.go
```diff
@@ -90,6 +90,12 @@ type ServiceControl struct {
 // This is used needed to set DelayedAutoStart
 type ServiceConfig struct {
```
directionless, line 91: Just to say this out loud... I think even without the util extension, we're _still_ using wix to install the service. Which means that uninstall will _still_ remove it and we don't have to think harder?

cmd/launcher/svc_config_windows.go
James-Pickett: oops, now using context passed

pkg/packagekit/wix/wix.go
James-Pickett: yes, the only print if the `wix_skip_cleanup` flag is passed, left a comment explaining

pkg/packagekit/wix/wix.go
James-Pickett: fixed

pkg/packagekit/wix/wix.go
James-Pickett: done

pkg/packagekit/wix/wix.go
James-Pickett: added to the docs

pkg/packagekit/wix/service.go
James-Pickett: Yes, this is correct. The only real difference now is that the arm64 will not set recovery actions during install, they are set when windows service starts.

RebeccaMahany (approved):
🔥 amazing

cmd/launcher/svc_config_windows.go
```diff
@@ -184,3 +187,56 @@ func checkRestartActions(logger *slog.Logger) {
 
 	logger.Log(context.TODO(), slog.LevelInfo, "successfully set RecoveryActionsOnNonCrashFailures flag")
 }
+
+// setRecoveryActions sets the recovery actions for the launcher service.
+// previously defined via wix ServicConfig Element (Util Extension) https://wixtoolset.org/docs/v3/xsd/util/serviceconfig/
+func setRecoveryActions(ctx context.Context, logger *slog.Logger) {
+	sman, err := mgr.Connect()
```
directionless, line 194: I see this is what we're doing everywhere here, I wonder if we should consolidate and only open it once, instead of the rapid opening and closing. (If we think so, it's fine as a followup)

docs/package-builder.md
```diff
@@ -193,3 +193,27 @@ Note that the windows package will only install as `ALLUSERS`. You may
 need to use elevated privileges to install it. This will likely be
 confusing. `msiexec.exe` will either silently fail, or be
 inscrutable. But `start` will work.
+
+###### Windows Multi-Archecture MSI
+
+Package builder creates an MSI containing both amd64 & arm64 binaries. This is done putting both of binaries and their service installation in to the MSI with mutually exclusive conditions based on processor architecture.
```
directionless, line 199: Very clear.

pkg/packaging/packaging.go
```diff
@@ -58,12 +60,6 @@ type PackageOptions struct {
 	WixSkipCleanup    bool
 	DisableService    bool
 
-	// Normally we'd download the same version we bake into the
-	// autoupdate. But occasionally, it's handy to make a package
-	// with a different version.
-	LauncherDownloadVersionOverride string
-	OsqueryDownloadVersionOverride  string
```
directionless, line 65: It's okay to remove these, but I want to note that I have used them in the past. They might make less sense with the newer TUF updater

directionless (approved):
This is a big step forward.

And I love the conclusion of only removing the util stuff from the less common platform.

zackattack01 (approved):
🔥

PR #1798, opened 2024-07-25
pkg/osquery/runtime improvements, largely around improving test flakiness

RebeccaMahany:
I've noticed that the runtime tests have been particularly flaky on Windows lately and it's been holding up CI significantly because we often have to wait for the 10-min test timeout to hit before test failure. See e.g. https://github.com/kolide/launcher/actions/runs/10094889355/job/27918355827.

## Issues found

* `runner.Shutdown` was getting stuck; tests would panic after 10 minutes due to timeout. (This really holds up CI, plus the panic means we don't really get any useful information about what part of the test got stuck.)
* `waitHealthy` reported that the osquery instance was healthy before we'd actually finished launching the instance. I believe that this was what caused `runner.Shutdown` to get stuck.
* I saw in the logs that sometimes the PID file wasn't getting removed. Not really a big deal, but I fixed it anyway.

### Test improvements

* Tests fail faster: add a one-minute timeout to calling `runner.Shutdown`, so that we can quickly fail and get useful troubleshooting information when `Shutdown` is stuck -- instead of waiting 10 minutes to hit test timeout + panic, which gives us no useful information about the issue
* Confirm osquery instance is fully set up: instead of only validating that `runner.Healthy()` returns no error, also confirm that the launch function has finished all critical work, and add a little bit of a sleep buffer before proceeding
* Test failures include more useful troubleshooting information: include error message from calling `runner.Healthy` when that fails; include runner logs in failure messages

### Code improvements

* Retry PID file removal
* A log message was erroneously tagged as an errgroup exiting when it was not -- updated appropriately
* Mark instance as connected after all extension manager servers are set up, not just `kolide_grpc`

### DIFF
--- a/pkg/osquery/runtime/osqueryinstance.go
+++ b/pkg/osquery/runtime/osqueryinstance.go
@@ -596,8 +596,7 @@ func (o *OsqueryInstance) StartOsqueryExtensionManagerServer(name string, socket
 		<-o.doneCtx.Done()
 
 		o.slogger.Log(context.TODO(), slog.LevelDebug,
-			"exiting errgroup",
-			"errgroup", "starting extension shutdown",
+			"starting extension shutdown",
 			"extension_name", name,
 		)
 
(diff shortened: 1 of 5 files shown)


James-Pickett (approved):
Nice!

PR #1802, opened 2024-07-26
prevent watchdog installation for dev environments

zackattack01:
There were a few places already where we gate behavior on the KolideServerURL with the same checks. This adds a helper function to check a serverURL against our match criteria, and adds this check to the watchdog installation and log publication routines

### DIFF
--- a/ee/watchdog/controller_windows.go
+++ b/ee/watchdog/controller_windows.go
@@ -156,7 +156,8 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
 	// we don't alter watchdog installation (install or remove) if this is a non-prod deployment
 	if !launcher.IsKolideHostedServerURL(wc.knapsack.KolideServerURL()) {
 		wc.slogger.Log(ctx, slog.LevelDebug,
-			"skipping ServiceEnabledChanged for launcher watchdog",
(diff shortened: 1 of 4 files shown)


pkg/launcher/options.go
```diff
@@ -627,3 +627,9 @@ func commandUsage(fs *flag.FlagSet, short string) func() {
 		fmt.Fprintf(os.Stderr, "\n")
 	}
 }
+
+// IsKolideHostedServerURL is a convenience function to enable gating functionality for
+// developer (or other non-production) deployments
+func IsKolideHostedServerURL(serverURL string) bool {
```
RebeccaMahany, line 633: This is a nice improvement

RebeccaMahany (dismissed):

ee/watchdog/controller_windows.go
```diff
@@ -153,6 +158,17 @@ func (wc *WatchdogController) Interrupt(_ error) {
 
 func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
 	ctx := context.TODO()
+	// we don't alter watchdog installation (install or remove) if this is a non-prod deployment
```
directionless, line 161: We could probably also check if we're running as not-root.

zackattack01: oh good call, on it!

directionless (approved):

PR #1804, opened 2024-07-26
Log when osquery instance has completed setup; improve interactive tests

RebeccaMahany:
Continuing to look at flakiness in the Windows osquery runtime tests --

I've seen a decent number of test failures like [this one](https://github.com/kolide/launcher/actions/runs/10114504292/job/27973279477), where the instance is not marked as healthy in 30 seconds so the test fails. The logs help a little bit, but not as much as they could. This PR adds a couple debug-level logs to better document the osquery setup process, and adds a timestamp to the test failure message to better contextualize the runner logs.

Also updated interactive tests with similar changes as in previous PR: add timeouts so that we get test errors instead of test timeout panics; add logs to test failure messages.

### DIFF
--- a/pkg/osquery/runtime/runner.go
+++ b/pkg/osquery/runtime/runner.go
@@ -472,10 +472,6 @@ func (r *Runner) launchOsqueryInstance() error {
 	o.extensionManagerClient, err = o.StartOsqueryClient(paths)
 	if err != nil {
 		traces.SetError(span, fmt.Errorf("could not create an extension client: %w", err))
-		r.slogger.Log(ctx, slog.LevelError,
-			"error creating extension client",
-			"err", err,
-		)
 		return fmt.Errorf("could not create an extension client: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/osquery/runtime/runner.go
```diff
@@ -377,6 +377,9 @@ func (r *Runner) launchOsqueryInstance() error {
 	}
 
 	span.AddEvent("socket_created")
+	r.slogger.Log(ctx, slog.LevelDebug,
```
directionless, line 380: I wonder if it would be interesting to collect the spans from testing -- it looks like they already have this.

RebeccaMahany: Maybe! I imagine it could be interesting if we wanted to do benchmarking in the future too

James-Pickett (approved):

zackattack01 (approved):

PR #1806, opened 2024-07-26
optimize windows service startup

zackattack01:
See inline comments for additional technical context

We are seeing issues with windows devices timing out before launcher is able to start in certain environments. This typically coincides with significant resource contention/high activity during boot up. One bit of logic that stood out as possibly dragging things out is our registry and service manager interactions (to perform runtime service configuration). 

Moving these checks into the Execute function, after our Running status update should help to make startup faster, or at least more responsive to service control manager

### DIFF
--- a/cmd/launcher/svc_windows.go
+++ b/cmd/launcher/svc_windows.go
@@ -31,6 +31,14 @@ const serviceName = "launcher"
 
 // runWindowsSvc starts launcher as a windows service. This will
 // probably not behave correctly if you start it from the command line.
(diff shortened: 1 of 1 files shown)


RebeccaMahany (approved):

James-Pickett (approved):

directionless (approved):
Nice. As separate work, is it interesting to add spans or timing data to this critical section?

zackattack01:
regarding the spans - I will check that out next!

PR #1812, opened 2024-07-30
Add socketfilterfw parser table

Micah-Kolide:
With the release of MacOS 15, Apple has deprecated the use of a plist for application firewall settings. Instead they say to use `socketfilterfw` if you relied on the plist in past versions.

osquery does not have a pattern for executing and parsing data from binaries (yet), so I've added this table to fill in the gap if that table stays unsupported for MacOS 15.


Micah-Kolide:
Will be adding some tests next, but as for an example query:
```
osquery> SELECT * FROM kolide_socketfilterfw;
+------------------------+--------+----------------------+-----------+-------+
| fullkey                | parent | key                  | value     | query |
+------------------------+--------+----------------------+-----------+-------+
| 0/logging_option       | 0      | logging_option       | throttled | *     |
| 0/stealth_enabled      | 0      | stealth_enabled      | false     | *     |
| 0/allow_signed_enabled | 0      | allow_signed_enabled | true      | *     |
| 0/block_all_enabled    | 0      | block_all_enabled    | false     | *     |
| 0/global_state         | 0      | global_state         | true      | *     |
| 0/logging_enabled      | 0      | logging_enabled      | true      | *     |
+------------------------+--------+----------------------+-----------+-------+
```

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,84 @@
...
+
+// socketfilterfw returns lines for each `get` argument supplied.
+// The output data is in the same order as the supplied arguments.
+//
+// Each line describes a part of the feature and what state it's in.
+// These are not very well-formed, so I'm doing some regex magic to
+// know which option the current line is, and then sanitize the state.
+func socketfilterfwParse(reader io.Reader) (any, error) {
+	results := make([]map[string]string, 0)
+	row := make(map[string]string)
```
directionless, line 20: I'd recommend putting `row` into the `for scanner.Scan()` loop.

Outside the loop it optimizes reducing memory allocations, which can be a small win for high performance, but comes with the risk/overhead of needing to make sure you've cleared state each time. Since we're not high performance, it's simpler to put it into the loop and let the scoping handle it.

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,84 @@
...
+		case "downloaded":
+			key = "allow_downloaded_signed_enabled"
+		case "stealth":
+			key = "stealth_enabled"
+		case "log mode":
+			key = "logging_enabled"
+		case "log option":
+			key = "logging_option"
+		default:
+			continue
```
directionless, line 47: This is an interesting questions. Do you think this format is stable enough we should use the enum here, or should we capture everything and try to coerce it to kv pairs?

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		// Don't allow overwrites.
+		_, ok := row[key]
```
directionless, line 51: When would this occur?

RebeccaMahany (commented):
LGTM once seph's comments are addressed!

ee/tables/execparsers/socketfilterfw/parser.go
Micah-Kolide: I had placed this in for the malformed test data as the `state` string for global state is matched in the app states. Now that data shouldn't be ever received, so this may be overkill.

ee/tables/execparsers/socketfilterfw/parser.go
Micah-Kolide: Sounds good

Micah-Kolide: I updated how row is being handled since I added in apps, but I still have it first declared outside of the loop.

ee/tables/execparsers/socketfilterfw/parser.go
Micah-Kolide: I've made use of this space for the apps table.

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+
+			if len(row) > 0 {
+				results = append(results, row)
+				row = make(map[string]string)
+			}
+
+			continue
+		}
+
+		matches := make([]string, 0)
```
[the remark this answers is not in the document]
Micah-Kolide: Resolved

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+//
+// This supports parsing the list of apps and their allow state, or
+// each line describes a part of the feature and what state it's in.
+//
+// These are not very well-formed, so I'm doing some regex magic to
+// know which option the current line is, and then sanitize the state.
+func socketfilterfwParse(reader io.Reader) (any, error) {
+	results := make([]map[string]string, 0)
+	row := make(map[string]string)
+	parse_app_data := false
```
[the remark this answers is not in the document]
Micah-Kolide: Haha yeah of course I don't know why I randomly switched in the middle of making this. snake case always is stuck in my head 😄

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+		line := scanner.Text()
+
+		// When parsing the app list, the first line of output is a total
+		// count of apps. We can break on this line to start parsing apps.
+		if strings.Contains(line, "Total number of apps") {
+			parse_app_data = true
+
+			if len(row) > 0 {
+				results = append(results, row)
+				row = make(map[string]string)
```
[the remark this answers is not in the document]
Micah-Kolide: I probably didn't need to include it, but yeah I should've done it below as well if I was to keep this pattern.

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+
+			continue
+		}
+
+		var matches []string
+		if parse_app_data {
+			matches = appRegex.FindStringSubmatch(line)
```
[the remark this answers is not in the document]
Micah-Kolide: I lower cased the line parsing since I'm creating the key string manually, but I didn't want to change the case of app names, so I kept it off there.

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+var lineRegex = regexp.MustCompile("(state|block|built-in|downloaded|stealth|log mode|log option)(?:.*\\s)([0-9a-z]+)")
+
+// socketfilterfw returns lines for each `get` argument supplied.
+// The output data is in the same order as the supplied arguments.
+//
+// This supports parsing the list of apps and their allow state, or
+// each line describes a part of the feature and what state it's in.
+//
+// These are not very well-formed, so I'm doing some regex magic to
+// know which option the current line is, and then sanitize the state.
```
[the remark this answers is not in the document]
Micah-Kolide: Thank you :)

Now I just need to do the same for the rest of my PRs 😅

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,111 @@
...
+
+		row[key] = sanitizeState(matches[2])
+	}
+
+	if len(row) > 0 {
+		results = append(results, row)
+	}
+
+	return results, nil
+}
```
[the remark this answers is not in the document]
Micah-Kolide: I agree and went a bit back and forth on it until I landed here. I'm not opposed to splitting them out, and I think it does make it more readable and easier to track.

I can see a couple of differences I'll need to do from your example, but I'll see how I can re-work it to be closer to that pattern.

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		if parseAppData {
+			appRow := parseAppList(line)
+			if appRow != nil {
+				results = append(results, appRow)
+			}
+
+			continue
+		}
```
RebeccaMahany, line 44: I think you could combine like this -- it would let you get rid of `parseAppData` bool entirely.

```suggestion
		if strings.Contains(line, "Total number of apps") {
			if appRow := parseAppList(line); appRow != nil {
				results = append(results, appRow)
			}
			continue
		}
```

ee/tables/execparsers/socketfilterfw/parser.go
```diff
@@ -0,0 +1,125 @@
...
+
+		k, v := parseLine(line)
+		if k != "" {
+			row[k] = v
+		}
```
RebeccaMahany, line 49: Ahh, I understand better what `row` looks like now

RebeccaMahany (approved):
LGTM, had a final nitpick for conciseness

ee/tables/execparsers/socketfilterfw/parser.go
Micah-Kolide: I'm confused how this is supposed to work. Only the first line of text for the app list contains that string:
```
Total number of apps = 6
replicatord (state: 1)
Pop Helper.app (state: 4)
Google Chrome (state: 4)
rtadvd (state: 1)
com.docker.backend (state: 1)
sshd-keygen-wrapper (state: 1)
```

RebeccaMahany: Oh blah, I'm sorry, I see what I missed! Ignore this.

RebeccaMahany (approved):

ee/tables/execparsers/socketfilterfw/parser.go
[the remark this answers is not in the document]
directionless: Working across all these languages with different idioms always throws me

ee/tables/execparsers/socketfilterfw/parser.go
directionless: Oh, I see now. 

In the general case, you're returning a single row, with a lot of columns. 

I was thinking you could return this as EAV style data, and let data flatten handle it.

ee/tables/execparsers/socketfilterfw/parser.go
Micah-Kolide: No worries, I just wasn't sure if I was missing something

PR #1814, opened 2024-07-31
[Chrome IndexedDB] Fix header processing

RebeccaMahany:
I wasn't adequately accounting for the header structure before. This appears to be correct now -- needed to add a constraint that the header ENDS with `<version tag, version, object tag>`, instead of terminating processing the header as soon as we see `<object tag>` (because that may occur before the end of the header and should just be ignored).


zackattack01 (approved):

RebeccaMahany:
(I added and removed a bunch of labels to this PR while working on https://github.com/kolide/launcher/pull/1815 to test it, but I am done now and it is correctly labeled)

PR #1818, opened 2024-08-01
Add osqueryd file presence check to autoupdate running version check

zackattack01:
We noticed the on-demand osquery download was taking longer than expected and @RebeccaMahany found this was due to our built in retry logic for determining the current osquery version. Currently, we need to wait for that to time out before the download will continue.

We should still leave that logic in place, but can additionally check once for osqueryd file presence to bypass the lengthier (retried) version checks when the file does not exist. This speeds up the on-demand download process significantly

### DIFF
--- a/ee/tuf/autoupdate_test.go
+++ b/ee/tuf/autoupdate_test.go
@@ -75,6 +75,11 @@ func TestExecute_launcherUpdate(t *testing.T) {
 	testReleaseVersion := "1.2.3"
 	tufServerUrl, rootJson := tufci.InitRemoteTufServer(t, testReleaseVersion)
 	s := setupStorage(t)
+	// setup fake osqueryd binary to mock file existence for currentRunningVersion
+	fakeOsqBinaryPath := filepath.Join(testRootDir, "osqueryd")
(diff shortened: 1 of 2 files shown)


ee/tuf/autoupdate_test.go
```diff
@@ -75,6 +75,11 @@ func TestExecute_launcherUpdate(t *testing.T) {
 	testReleaseVersion := "1.2.3"
 	tufServerUrl, rootJson := tufci.InitRemoteTufServer(t, testReleaseVersion)
 	s := setupStorage(t)
+	// setup fake osqueryd binary to mock file existence for currentRunningVersion
+	fakeOsqBinaryPath := filepath.Join(testRootDir, "osqueryd")
```
RebeccaMahany, line 79: Think you will want this to be `osqueryd.exe` for running tests on Windows. I think there's a utility func for this, hang on...

RebeccaMahany: `executableLocation`?

zackattack01: 👍 will update, ty!

RebeccaMahany (approved):

PR #1824, opened 2024-08-07
add invoking user permissions metadata for flares

zackattack01:
it is sometimes confusing to me whether something failed due to a permissions issue with the flare invocation, this sticks an explicit check into the metadata for quick reference


RebeccaMahany (approved):
🔥

Issue #1791, opened 2024-07-23
launcher on Windows does not ship logs to cloud if `debug` flag is set


RebeccaMahany:
While doing release validation for https://github.com/kolide/launcher/releases/tag/v1.9.4, I noticed that my launcher installation on Windows does not ship its logs IFF `debug` is set in `launcher.flags`.

We should fix this issue, but I don't think it should hold up this release -- `debug` is not typically set in end user `launcher.flags` files. (It is likely that this issue pre-dates this release, as well.)

PR #1828, opened 2024-08-09
Do not echo debug logs to stderr on Windows

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1791

This appears to be what prevents logs from shipping when the `debug` flag is set. We don't receive logs send to stderr on Windows anyway, so we can remove this.


zackattack01 (approved):
nice find!

James-Pickett (approved):

PR #1830, opened 2024-08-12
[LevelDB IndexedDB] Return empty rows instead of error when db or object store doesn't exist

RebeccaMahany:
I think this is more consistent behavior -- we already ignore filepaths that don't exist. If the filepath to the indexeddb exists, but the indexeddb hasn't been initialized (db doesn't exist yet, object store doesn't exist yet), we should also return no data instead of an error.

### DIFF
--- a/ee/indexeddb/indexeddb.go
+++ b/ee/indexeddb/indexeddb.go
@@ -71,7 +71,7 @@ func QueryIndexeddbObjectStore(dbLocation string, dbName string, objectStoreName
 		if errors.Is(err, leveldberrors.ErrNotFound) {
 			return objs, nil
 		}
-		return nil, fmt.Errorf("querying for database id: %w", err) // TODO if leveldb: not found, return no rows?
+		return nil, fmt.Errorf("querying for database id: %w", err)
 	}
 	databaseId, _ := binary.Uvarint(databaseIdRaw)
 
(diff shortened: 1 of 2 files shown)


ee/indexeddb/indexeddb.go
```diff
@@ -64,7 +67,11 @@ func QueryIndexeddbObjectStore(dbLocation string, dbName string, objectStoreName
 	}
 	databaseIdRaw, err := db.Get(databaseNameKey, nil)
 	if err != nil {
-		return nil, fmt.Errorf("querying for database id: %w", err)
+		// If the database doesn't exist, return an empty list of objects
+		if errors.Is(err, leveldberrors.ErrNotFound) {
+			return objs, nil
+		}
+		return nil, fmt.Errorf("querying for database id: %w", err) // TODO if leveldb: not found, return no rows?
```
zackattack01, line 74: not sure i understand the TODO comment here but probably not important

zackattack01 (approved):

ee/indexeddb/indexeddb.go
RebeccaMahany: lol whoops, that was an implementation note I'd left for myself. removed, thank you!

zackattack01 (approved):

PR #1831, opened 2024-08-12
use runtime to populate enrollment values where possible

James-Pickett:
uses go runtime to populate enrollment details when possible

### DIFF
--- a/ee/uninstall/uninstall_windows.go
+++ b/ee/uninstall/uninstall_windows.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/kolide/launcher/ee/watchdog"
 	"golang.org/x/sys/windows/svc/mgr"
 )
 
@@ -28,6 +29,11 @@ func disableAutoStart(ctx context.Context) error {
 	cfg.StartType = mgr.StartManual
 	if err := launcherSvc.UpdateConfig(cfg); err != nil {
 		return fmt.Errorf("updating launcher service config: %w", err)
+	}
+
+	// attempt to remove watchdog service in case it is installed to prevent startups later on
+	if err := watchdog.RemoveService(svcMgr); err != nil {
+		return fmt.Errorf("removing watchdog service, error may be expected if not enabled: %w", err)
 	}
 
 	return nil
--- a/ee/watchdog/controller_windows.go
+++ b/ee/watchdog/controller_windows.go
@@ -202,7 +202,7 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
 	defer serviceManager.Disconnect()
 
 	if !enabled {
-		err := removeService(serviceManager, launcherWatchdogServiceName)
+		err := RemoveService(serviceManager)
 		if err != nil {
 			if err.Error() == serviceDoesNotExistError {
 				wc.slogger.Log(ctx, slog.LevelDebug, "watchdog service was not previously installed")
@@ -344,15 +344,19 @@ func (wc *WatchdogController) installService(serviceManager *mgr.Mgr) error {
 	return nil
 }
 
-// removeService utilizes the passed serviceManager to remove the existing service
-// after looking up the handle from serviceName
-func removeService(serviceManager *mgr.Mgr, serviceName string) error {
-	existingService, err := serviceManager.OpenService(serviceName)
+// RemoveService utilizes the passed serviceManager to remove any existing watchdog service if it exists
+func RemoveService(serviceManager *mgr.Mgr) error {
+	existingService, err := serviceManager.OpenService(launcherWatchdogServiceName)
 	if err != nil {
 		return err
 	}
 
 	defer existingService.Close()
+
+	// attempt to stop the service first, we don't care if this fails because we are going to
+	// remove the service next anyway (the removal happens faster if stopped first, but will
+	// happen eventually regardless)
+	existingService.Control(svc.Stop)
 
 	if err := backoff.WaitFor(func() error {
 		if err = existingService.Delete(); err != nil {


pkg/osquery/enrollment_details.go
```diff
@@ -31,13 +37,16 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
 		return details, errors.New("Skipping enrollment details")
 	}
 
+	details.OSPlatform = runtime.GOOS
+	details.OSPlatformLike = runtime.GOOS
```
directionless, line 41: Anything we can do here, we should be pulling out of the osquery query

James-Pickett: I just started with these 2 because I know they are consistent with osq. However, hostname is not consistent. On my mac, runtime pkg give a different value that osquery for hostname. Do you think we should just avoid inconsistent values all together? or grab what we can using go, then let osq overwrite the values.

directionless: I think these two are the most important. So if we don't have an answer for the others, even just having these is good.

macOS has 3 different hostnames. There might be a way to get the others

James-Pickett: I think we would have to exec to get the other hostnames. I suppose in theory we could replicate what osquery does on all platforms with execs, but I'm not sure if we want to go down that path. Or we can just do it for data that is critical to enrollment.

pkg/osquery/enrollment_details.go
```diff
@@ -31,22 +37,23 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
...
+		return details, fmt.Errorf("%w: %s is a directory", osqueryNotFoundErr{}, osquerydPath)
 	} else if err != nil {
-		return details, fmt.Errorf("statting %s: %w", osquerydPath, err)
+		return details, fmt.Errorf("%w: statting %s: %w", osqueryNotFoundErr{}, osquerydPath, err)
 	}
 
 	query := `
 	SELECT
 		osquery_info.version as osquery_version,
 		os_version.build as os_build,
```
directionless, line 55: Do you know if we can get the rest of `os_version` ?

James-Pickett: I think only if we exec for it. There is `unix.Sysctl("kern.osrelease")`, but it gives darwin kernel version as opposed to macos version.

James-Pickett: Or we could read `/System/Library/CoreServices/SystemVersion.plist` (and find the equivalent on other platforms)

directionless: Eh. Maybe we kick that forward

pkg/osquery/enrollment_details.go
```diff
@@ -31,22 +37,23 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
 		return details, errors.New("Skipping enrollment details")
 	}
 
+	details.OSPlatform = runtime.GOOS
+	details.OSPlatformLike = runtime.GOOS
+
 	// If the binary doesn't exist, bail out early.
 	if info, err := os.Stat(osquerydPath); os.IsNotExist(err) {
-		return details, fmt.Errorf("no binary at %s", osquerydPath)
+		return details, fmt.Errorf("%w: no binary at %s:", osqueryNotFoundErr{}, osquerydPath)
```
directionless, line 45: I don't think we should return content and errors. It feels like it breaks the usual go idioms. 

My inclination is to log this, and then use the best we have. 

Slightly longer term, I'd like to calculate all the enroll details when either launcher or osquery restart, and then store them in knapsack for subsequent usage. If there's some issue with osquery, I'd want to allow things to get the go only ones, and then fill in the osquery ones when possible

pkg/osquery/extension.go
```diff
@@ -418,35 +418,38 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+			)
+		}
+		return err
+	}, 30*time.Second, 5*time.Second); err != nil {
+		if os.Getenv("LAUNCHER_DEBUG_ENROLL_DETAILS_REQUIRED") == "true" {
+			return "", true, fmt.Errorf("query enrollment details: %w", err)
 		}
+
+		e.slogger.Log(ctx, slog.LevelError,
+			"failed to get enrollment details with osquery, using runtime values for enrollment",
```
RebeccaMahany, line 445: I'm confused by this log -- it makes it sound like we're about to fall back to setting `enrollDetails` another way, but then we don't do that. Are we supposed to be performing another operation here? Or does the log message mean "failed to get enrollment details via osquery/runtime"...?

pkg/osquery/enrollment_details.go
```diff
@@ -31,22 +37,23 @@ func getEnrollDetails(ctx context.Context, osquerydPath string) (service.Enrollm
 		return details, errors.New("Skipping enrollment details")
 	}
 
+	details.OSPlatform = runtime.GOOS
+	details.OSPlatformLike = runtime.GOOS
```
RebeccaMahany, line 41: Are these the only details we need to be able to avoid the "Unknown Linux Device" issue?

pkg/osquery/enrollment_details.go
RebeccaMahany: Could solve this by doing e.g. `return details, fmt.Errorf("%w: no binary", osqueryNotFoundErr{})`, and then in extension.go log `osqPath` at the same time that we log this error.

James-Pickett: > I don't think we should return content and errors. It feels like it breaks the usual go idioms.

I see what you mean, this kind of implies that `getEnrollDetails` will never return an error because we always have access to the runtime values?

pkg/osquery/enrollment_details.go
James-Pickett: I believe so, they are the only details we can "easily" retrieve. I think all the other details require execs or reading various files, basically doing what ever osq does to get the info ... maybe it's not a bad idea to move off of osq completely for enrollment. Definitely expands scope of this task though.

RebeccaMahany: I think it might be worth grabbing (via execs) whatever the bare minimum is that we need to avoid "Unknown Linux Device" -- I don't know which details those are, though. Maybe this is already sufficient 🤞

directionless: This should be it

Issue #1826, opened 2024-08-09
Retrieve enrollment details via runtime data where possible


RebeccaMahany:
We rely on execing osquery to fetch enrollment details; sometimes this exec fails, and we end up with an unknown linux device enrolled. We should instead try to fetch as much of this information as possible from other sources that are not as prone to failure during the launcher enrollment process. We hope this will make enrollment more reliable.

The details we fetch from osquery are [here](https://github.com/kolide/launcher/blob/main/pkg/osquery/enrollment_details.go#L43-L59).

PR #1833, opened 2024-08-13
get runtime enrollment details before osq details

James-Pickett:
splits enrollment details between funcs to get runtime (non-osq) enroll details and osq details

closes https://github.com/kolide/launcher/issues/1826


RebeccaMahany (approved):
LGTM! Not sure why govet only just started to complain about fetch.go

PR #1834, opened 2024-08-14
support preprod dual install

zackattack01:
to support multiple installations:
- wires build time identifier used into launcher.flags
- uses this from windows service configuration to determine the correct service name and launcher registry keys
  - minor refactoring to prevent repeatedly opening and closing service manager/service connections
- makes flare shipper's enroll secret collection fall back to knapsack's EnrollSecretPath if set, to ensure collecting the correct secret when run in situ
- adds multiple installation check to flares, to note any potential discrepancies that may arise from multiple installations in flare checkups


ee/debug/shipper/shipper.go
```diff
@@ -268,6 +268,7 @@ func enrollSecret(k types.Knapsack) string {
 		return k.EnrollSecret()
 	}
 
+	// TODO this will need to respect the identifier when determining the secret file location for dual-launcher installations
 	b, err := os.ReadFile(launcher.DefaultPath(launcher.SecretFile))
```
directionless, line 272: oooo.... This one we might need to fix. Can we shift it to using the secret in something configured instead of DefaultPath?

zackattack01: I think we can make this fallback to reading from `knapsack.EnrollSecretPath()` first and that will fix this when running in situ at least. fixing for CLI triggered flares looks tougher

James-Pickett (dismissed):
Nice

ee/debug/checkups/checkups.go
```diff
@@ -302,10 +305,62 @@ func RunFlare(ctx context.Context, k types.Knapsack, flareStream io.WriteCloser,
...
+
 	// we could defer this close, but we want to return any errors
 	return close()
 }
 
+// noteMultipleInstallations checks for whether the results of running flare for this installation may be complicated
+// by multiple installations. This is less of an issue when the flare is run in situ, but should be noted because
+// we may need to pay closer attention to the results. When run standalone without a config argument passed, it would be
+// possible for flare to default to reading the directories for the wrong installation
+func noteMultipleInstallations(z *zip.Writer) {
```
directionless, line 320: Feels weird not to push this through the normal checkup interface. But okay...

directionless (approved):
I think this is okay to try!

directionless (approved):
I think this is okay! Woo!

Issue #1717, opened 2024-05-15
Make windows desktop process responsive to repeated systray errors


zackattack01:
We often observe systray errors like `tray not ready` when devices are asleep - but these can be indicative of persistent issues when repeated after the device has woken up from sleeping. Specifically I saw this pattern on my own device after being in modern standby for a few hours:
<details>
<summary>logs</summary>
<p>

```
{"time":"2024-05-15T15:21:56.094156Z","level":"DEBUG","source":{"function":"github.com/kolide/launcher/ee/desktop/runner.(*DesktopUsersProcessesRunner).runConsoleUserDesktop","file":"C:/Users/zack-/code/launcher/ee/desktop/runner/runner.go","line":549},"msg":"modern standby detected, skipping desktop process spawning and health checks","launcher_run_id":"01HXXDRZPD8P7MREKCHWGM71GG","component":"desktop_runner"}
{"time":"2024-05-15T15:21:57.2525241Z","level":"DEBUG","source":{"function":"github.com/kolide/launcher/ee/powereventwatcher.(*powerEventWatcher).onPowerEvent","file":"C:/Users/zack-/code/launcher/ee/powereventwatcher/power_event_watcher_windows.go","line":206},"msg":"system is waking","launcher_run_id":"01HXXDRZPD8P7MREKCHWGM71GG","component":"power_event_watcher","event_id":507}
{"time":"2024-05-15T15:21:59.6302959Z","level":"DEBUG","source":{"function":"github.com/kolide/launcher/ee/desktop/runner.(*DesktopUsersProcessesRunner).desktopCommand.func1","file":"C:/Users/zack-/code/launcher/ee/desktop/runner/runner.go","line":870},"msg":"2024/05/15 11:21:59 systray error: unable to init instance: Unspecified error","launcher_run_id":"01HXXDRZPD8P7MREKCHWGM71GG","component":"desktop_runner","uid":"zackwindows\\zack-","subprocess":"desktop"}
[...]

PR #1835, opened 2024-08-14
Restart desktop process on systray errors

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1717.

### Explanation of underlying cause

If we fail to [init the instance](https://github.com/kolide/systray/blob/master/systray_windows.go#L892-L905), `initialized` will never get set, which means [isReady](https://github.com/kolide/systray/blob/master/systray_windows.go#L242-L245) will always return false. That means any time we attempt to perform an action on the systray in the future, like [setting an icon](https://github.com/kolide/systray/blob/master/systray_windows.go#L251), we'll always get the "tray not ready" errors.

### Evaluation of options for solving the issue

| Fix location | Has knowledge of systray errors?  | Has knowledge of sleep state? | Fix mechanism                                            |
|----------|-----------------------------------|-------------------------------|----------------------------------------------------------|
| systray  | Yes                               | No                            | Call `registerSystray()`                                 |
| desktop  | No                                | No                            | Shut down and restart the `menu`                         |
| runner   | Yes, via monitoring desktop logs  | Yes, via InModernStandby      | Restart desktop process, or ask it to restart the `menu` |

It would be cleanest to have systray manage its own errors. However, we do not want to take any action while we're in modern standby, and systray doesn't currently have knowledge of the `InModernStandby` flag. I don't think it's worthwhile to build in access to the flag.

I don't think it's practical for the desktop process to handle fixing the issue instead. It doesn't have knowledge of `InModernStandby` _and_ it doesn't have access to systray state: it can't see the systray logs, and systray doesn't return any errors when menu actions fail.

On the runner side, we already process desktop logs; we could monitor these for systray errors, and take action once we reach a threshold of "tray not ready" errors when not in modern standby. We could easily restart the entire desktop process, or we could build a new endpoint to ask the desktop process to shut down and restart the `menu` only.

### Implementation

The third option (handling in the runner, rather than in systray or desktop) is the approach I've taken.

I've also chosen the bigger and simpler hammer of restarting the entire desktop process, rather than implementing the ability to restart only the menu portion of the process.

### Testing

It's proving difficult to trigger this error condition, so I've feature-flagged these changes so that we can enable them only for internal test devices.


ee/desktop/runner/runner.go
```diff
@@ -350,6 +350,51 @@ func (r *DesktopUsersProcessesRunner) killDesktopProcesses(ctx context.Context)
...
+		"err", err,
+	)
+
+	if err := proc.Process.Kill(); err != nil {
+		return fmt.Errorf("could not kill desktop process for uid %s with pid %d: %w", uid, proc.Process.Pid, err)
+	}
+
+	// Successfully killed process
+	delete(r.uidProcs, uid)
+	return nil
```
zackattack01, line 395: worth adding an affirmative log here to match the `"killed user desktop process"` logging behavior above?

RebeccaMahany: Yes! Updated

zackattack01 (approved):
nice!

Issue #1513, opened 2023-12-14
Update flag parsing


RebeccaMahany:
We use [ff](https://pkg.go.dev/github.com/peterbourgon/ff), which uses [flag](https://pkg.go.dev/flag) for command-line arg parsing, right now. This hampers our ability to parse flags on launcher startup because `flag` cannot handle unknown command-line flags -- we have to wait until main has bootstrapped the most recent launcher that will know about all current flags.

There are limited other options that allow for handling unknown flags. [pflag](https://pkg.go.dev/github.com/spf13/pflag) and [go-flags](https://pkg.go.dev/github.com/jessevdk/go-flags) are the two I found, but neither seems to be maintained at the moment, and neither handles config files or environment variables the way that `ff` does.

We will likely want to move to a new solution that addresses all our needs, while remaining backward-compatible for older installations of launcher.

RebeccaMahany:
Not sure we want this

PR #1839, opened 2024-08-16
add WithIgnoreUndefined option to config parse

zackattack01:
Currently, if an unknown option is found in our launcher.flags the parser will stop parsing and return an error which is not checked/logged. This adds an option to the parser to prevent this behavior and continue parsing, discarding the unknown flag


directionless:
@RebeccaMahany Would this have eliminated the need for `github.com/spf13/pflag` ?

RebeccaMahany:
@directionless -- I don't think so. Unless the behavior has changed -- the last time I looked into it, `WithIgnoreUndefined` worked for flags supplied via config file, but would still break for us with undefined command-line args.

Separately, I am not sure if this PR will address https://github.com/kolide/launcher/issues/1513 -- older launcher installations without this option set will still run into this issue.

PR #1844, opened 2024-08-26
Do not override root directory to point to another tenant

RebeccaMahany:
With dual enrollment, we should not select another enrollment's root directory.


James-Pickett (dismissed):

directionless (approved):
Sure, let's try!

PR #1858, opened 2024-09-10
windows arm64 fixes, upgrade winio and thrift

James-Pickett:


ee/wmi/wmi.go
```diff
@@ -154,7 +155,12 @@ func Query(ctx context.Context, slogger *slog.Logger, className string, properti
 	defer serviceRaw.Clear()
 
 	service := serviceRaw.ToIDispatch()
-	defer service.Release()
+	if runtime.GOARCH != "arm64" {
+		// calling service.Release() and the serviceRaw.Clear() (or the reverse) causes
+		// a panic on arm64 (Exception 0xc0000005: Access Violation). The hunch is that on arm64
+		// one clears the memory of the other.
+		defer service.Release()
```
James-Pickett, line 162: I'm making the assumption here that the call to `serviceRaw.Clear()` that is defered above will release this as well because we get an access violation if both `serviceRaw.Clear()` and `service.Release()` get called (regardless of order)

directionless: Hrm. I wonder if something isn't doing what we think in `service := serviceRaw.ToIDispatch()`

All this stuff is deep in `unsafe` pointer land

directionless: But it seems like we should be able to make a pretty small test case and post it as an issue.

Though... https://github.com/go-ole/go-ole/issues/255

directionless (approved):

directionless (approved):

Issue #1852, opened 2024-09-09
Check recovery actions before editing them


RebeccaMahany:
On launcher service startup on Windows, we check the registry for settings related to running the launcher service, and edit those settings as needed, to ensure a consistent service definition across earlier and more recent launcher installations.

One of the settings we edit is recovery actions. Unlike with our other edits, we don't read the current state of those settings before setting them. We should update this function to read the current settings first, and only perform an edit if they aren't already set correctly, in order to reduce the number of times we edit the registry.

The relevant function lives [here](https://github.com/kolide/launcher/blob/main/cmd/launcher/svc_config_windows.go#L197).

PR #1859, opened 2024-09-11
Check windows service manager settings prior to setting them

mlwood-dev:
Fixes: #1852

Added a check for the Recovery actions prior to setting them


cmd/launcher/svc_config_windows.go
```diff
@@ -194,7 +194,18 @@ func checkRestartActions(logger *slog.Logger, service *mgr.Service) {
 
 // setRecoveryActions sets the recovery actions for the launcher service.
 // previously defined via wix ServicConfig Element (Util Extension) https://wixtoolset.org/docs/v3/xsd/util/serviceconfig/
-func setRecoveryActions(ctx context.Context, logger *slog.Logger, service *mgr.Service) {
+func checkRecoveryActions(ctx context.Context, logger *slog.Logger, service *mgr.Service) {
```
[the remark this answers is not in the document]
RebeccaMahany: Since we have `checkDelayedAutostart`, `checkDependOnService`, and `checkRestartActions`, calling this function `checkRecoveryActions` is more consistent. The `checkX` naming convention in this file is my fault 😅 -- I don't mind it, but I am not opposed to renaming the other three functions to `setX` or `checkAndSetX`.

If we keep the new `checkRecoveryActions` name, could you update the docblock to match?

cmd/launcher/svc_config_windows.go
```diff
@@ -213,10 +223,30 @@ func setRecoveryActions(ctx context.Context, logger *slog.Logger, service *mgr.S
...
+		if len(curRecoveryActions) == len(recoveryActions) {
+			for i := range curRecoveryActions {
+				if curRecoveryActions[i].Type != recoveryActions[i].Type && curRecoveryActions[i].Delay != recoveryActions[i].Delay {
+					return false
+				}
+			}
+			return true
+		}
+	}
+	return false
```
RebeccaMahany, line 251: We prefer to reduce nesting when possible for readability -- you can modify this to do some early checks and returns to accomplish this:

```suggestion
        if curRecoveryActions == nil || len(curRecoveryActions != len(recoveryActions) {
                return false
        }
	for i := range curRecoveryActions {
		if curRecoveryActions[i].Type != recoveryActions[i].Type && curRecoveryActions[i].Delay != recoveryActions[i].Delay {
			return false
		}
	}
	return true
```

mlwood-dev: That is much easier to read. Thanks for the suggestion!

cmd/launcher/svc_config_windows.go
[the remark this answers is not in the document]
mlwood-dev: Names are hard LOL 
I feel that `checkX` is easy to understand, and since the service methods specifically say `SetX` after that check validation it is easy to infer that since the check "failed" the setting is now set.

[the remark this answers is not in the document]
directionless: It's reasonable to follow convention here. As Becca says, please do update the comment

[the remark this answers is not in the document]
mlwood-dev: I got it updated in [this commit ](https://github.com/kolide/launcher/pull/1859/commits/7bbd8a86ceed6ac2506f78041d4672a28542d5af)

RebeccaMahany (dismissed):
LGTM, thank you for the contribution!

James-Pickett (approved):
thank you!

cmd/launcher/svc_config_windows.go
```diff
@@ -213,10 +224,28 @@ func setRecoveryActions(ctx context.Context, logger *slog.Logger, service *mgr.S
...
 	}
 }
+
+// recoveryActionsAreSet checks if the current recovery actions are set to the desired recovery actions
+func recoveryActionsAreSet(curRecoveryActions []mgr.RecoveryAction, recoveryActions []mgr.RecoveryAction) bool {
+	if curRecoveryActions == nil || len(curRecoveryActions) != len(recoveryActions) {
+		return false
+	}
+	for i := range curRecoveryActions {
+		if curRecoveryActions[i].Type != recoveryActions[i].Type && curRecoveryActions[i].Delay != recoveryActions[i].Delay {
```
directionless, line 246: Should this be an `||`

I might write it as two if blocks, because compound negative statements are confusing

cmd/launcher/svc_config_windows.go
```diff
@@ -213,10 +224,28 @@ func setRecoveryActions(ctx context.Context, logger *slog.Logger, service *mgr.S
...
 	if err := service.SetRecoveryActions(recoveryActions, 24*60*60); err != nil { // 24 hours
 		logger.Log(ctx, slog.LevelError,
 			"setting RecoveryActions",
 			"err", err,
 		)
 	}
 }
+
+// recoveryActionsAreSet checks if the current recovery actions are set to the desired recovery actions
+func recoveryActionsAreSet(curRecoveryActions []mgr.RecoveryAction, recoveryActions []mgr.RecoveryAction) bool {
```
directionless, line 241:
```suggestion
func recoveryActionsAreSet(curRecoveryActions, recoveryActions []mgr.RecoveryAction) bool {
```
(nit)

directionless (changes requested):

cmd/launcher/svc_config_windows.go
mlwood-dev: An OR would make more sense if the RecoveryActions differed in Type or Delay. I went with an AND here since they all have the same type and delay. Is that thought process flawed?

James-Pickett (approved):

RebeccaMahany (approved):

cmd/launcher/svc_config_windows.go
directionless: I think we'd want to overwrite if either the type or the delay didn't match. This is what the split statements do, and is readable.

I think if you combine them, it would be an `||`, otherwise the delay matching, but the type mismatching would _not_ return `false`. But I always get tangled in compound negation.

directionless (approved):
Thank you so much for the submission!

PR #1862, opened 2024-09-11
Fix `autoupdate_managed` table value for MacOS 15

Micah-Kolide:
In MacOS 15 Sequoia the SoftwareUpdate private framework has been updated and the `autoupdate_managed` value was no longer being detected for our `kolide_macos_software_update` table. After decompiling the framework it became clear what needed to change, but I was able to verify this change after we got a MDM profile on my laptop.

I've attached 4 photos of the decompiled framework (the first two are from MacOS 14 and the last two are from MacOS 15).
Screenshot 2024-09-11 at 3 39 52 PM
Screenshot 2024-09-11 at 4 04 10 PM
Screenshot 2024-09-11 at 3 41 48 PM
Screenshot 2024-09-11 at 4 04 23 PM
In these images we can see that the `isMacOSAutoUpdateManaged` and `isAutomaticallyCheckForUpdatesManaged` performed the exact same operation in the older version of the framework, but in the newest version it seems that `isAutomaticallyCheckForUpdatesManaged` has been deprecated.

Here is the result after updating to use `isMacOSAutoUpdateManaged` with a managed device:
Screenshot 2024-09-11 at 3 57 22 PM
Here is the result on my device without a profile to manage the autoupdates:
```
osquery> SELECT * FROM kolide_macos_software_update;
             autoupdate_managed = 0
             autoupdate_enabled = 1
                       download = 1
                    app_updates = 1
                     os_updates = 0
               critical_updates = 1
last_successful_check_timestamp = 1726030905
```

### DIFF
--- a/ee/tables/macos_software_update/SUSharedPrefs.h
+++ b/ee/tables/macos_software_update/SUSharedPrefs.h
@@ -62,6 +62,7 @@
 - (BOOL)bridgeOSUpdatesEnabled;
 - (BOOL)skipAPFSSnapshotting;
 - (BOOL)doesAllowBGStageWithoutInactivity;
+- (BOOL)isMacOSAutoUpdateManaged;
 - (BOOL)isAutomaticallyCheckForUpdatesManaged;
 - (BOOL)isAutomaticallyCheckForUpdatesEnabled;
 - (BOOL)adminDeferredInstallEnabled;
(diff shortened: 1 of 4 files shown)


directionless (approved):
I'm game to try this -- when it merges to nightly, let's make sure we test it on the nightly machines. (And maybe even push it to alpha)

PR #1863, opened 2024-09-12
james/remove wmi unneeded releases

James-Pickett:
I tested this by creating a small program that runs in a loop only clearing the `rawResult` and `rawService`. After 3K iterations, memory rose to around 9MB and remained stable. After 50K iterations memory was still at 9MB. This test was performed on both arm64 and amd64.

### DIFF
--- a/ee/wmi/wmi.go
+++ b/ee/wmi/wmi.go
@@ -153,8 +153,10 @@ func Query(ctx context.Context, slogger *slog.Logger, className string, properti
 	}
 	defer serviceRaw.Clear()
 
+	// the memory of result is released by `defer serviceRaw.Clear()` above,
+	// on windows arm64 machines, calling `service.Clear()` after `serviceRaw.Release()`
+	// would cause a panic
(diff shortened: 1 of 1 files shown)


ee/wmi/wmi.go
```diff
@@ -153,8 +153,10 @@ func Query(ctx context.Context, slogger *slog.Logger, className string, properti
 	}
 	defer serviceRaw.Clear()
 
+	// the memory of result is released by `defer serviceRaw.Clear()` above,
+	// on windows arm64 machines, calling `service.Clear()` after `serviceRaw.Release()`
+	// would cause a panic
```
directionless, line 158:
```suggestion
	// In testing, we find we do not need to `service.Release(). The memory of result is released
	// by `defer serviceRaw.Clear()` above, furthermore on windows arm64 machines, calling
	// `service.Clear()` after `serviceRaw.Release()` causes a panic
```

zackattack01 (approved):

directionless (approved):

PR #1865, opened 2024-09-13
Basic Windows Hello prompt

RebeccaMahany:
Hello steps:

1. Check to see if Hello is supported.
2. If supported, create credentials -- these are created for the user + application (in this case, launcher). This step is what prompts the Hello popup.
3. Once launcher sees that the user has accepted, it can then fetch the pubkey from those credentials.
4. launcher can then get an attestation against those credentials.

Notes:

* I pulled in winrt-go, which made dealing with async operations significantly easier.
* The Hello prompt works sort of weirdly for me -- it will recognize my face and let me accept the prompt, but then give me a new prompt with an error message, and repeat until I enter my PIN instead. I haven't investigated this yet at all.
* Should be able to refactor this code for reusability but I ran into some errors there that I haven't quite figured out yet.
* I think I should be able to move to using `UserConsentVerifier`, which will give us functionality closer to what we're looking for -- but this will be a bit more work, so I want to merge this in its current state and rewrite later, to avoid blocking James's overall presence detection work.


James-Pickett (approved):
NICE!

ee/presencedetection/presencedetection_windows.go
```diff
@@ -0,0 +1,413 @@
...
+	keyCredentialManager := (*KeyCredentialManager)(unsafe.Pointer(managerObj))
+
+	credentialNameHString, err := ole.NewHString(credentialName)
+	if err != nil {
+		return fmt.Errorf("creating credential name hstring: %w", err)
+	}
+	defer ole.DeleteHString(credentialNameHString)
+
+	var requestCreateAsyncOperation *foundation.IAsyncOperation
+	requestCreateReturn, _, _ := syscall.SyscallN(
```
zackattack01, line 214: is the err worth keeping/noting here? or is that just not as informative as `ole.NewError(requestCreateReturn)`

zackattack01 (approved):
same q for the other `SyscallN` calls but not blocking, looks great!

RebeccaMahany:
Closing in favor of https://github.com/kolide/launcher/pull/1890

PR #1867, opened 2024-09-13
macos presence detection

James-Pickett:
This PR adds a presence detection endpoint to the DesktopUserServer which will call the OS specific implementation of presence detection and return the result. The call to DesktopUserServer is made by middleware in the LocalServer. The LocalServer looks for values in cmdRequest (krypto) headers to determine when it needs to detect presence.

More docs here: https://github.com/kolide/monorepo/pull/240


ee/localserver/server.go
```diff
@@ -393,3 +402,21 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
```
James-Pickett, line 408: should we consider a control server flag here that only enforces this when set?

cmd/launcher/desktop.go
```diff
@@ -182,9 +184,10 @@ func runDesktop(_ *multislogger.MultiSlogger, args []string) error {
 		}
 	}()
 
+	// block until a send on showDesktopChan
+	<-showDesktopChan
```
RebeccaMahany, line 188: Ooh, this is tidy!

ee/desktop/user/server/server.go
```diff
@@ -152,6 +161,15 @@ func (s *UserServer) notificationHandler(w http.ResponseWriter, req *http.Reques
 	w.WriteHeader(http.StatusOK)
 }
 
+func (s *UserServer) showDesktop(w http.ResponseWriter, req *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	sync.OnceFunc(func() {
+		s.showDesktopChan <- struct{}{}
+	})()
```
RebeccaMahany, line 170: Since this function is instantiated every time there's a request to /show, does `sync.OnceFunc()` enforce only sending to the channel once? Or we need to instantiate it earlier, like --

```
userServer := &UserServer{
    showFunc: sync.OnceFunc(func() {
		s.showDesktopChan <- struct{}{}
    }),
    ....
}

...

func (s *UserServer) showDesktop(w http.ResponseWriter, req *http.Request) {
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)

	s.showFunc()
}
```

Or maybe I'm misreading the `sync` documentation?

ee/presencedetection/presencedetection_darwin.go
```diff
@@ -0,0 +1,105 @@
...
+    return authResult;
+}
+*/
+import "C"
+import (
+	"fmt"
+	"unsafe"
+)
+
+func Detect(reason string) (bool, error) {
```
RebeccaMahany, line 87: For Windows, I think we're going to want something like 

```suggestion
func Detect(credentialName string, reason string) (bool, string, error) {
```

where the string returned is the new credential name, if the credential didn't exist yet and had to be provisioned. I think we'd also want knapsack/store access in this package, so that we can store the credential name.

I can add both of those changes in a follow-up PR, but wanted to mention it in case it affected anything about what you're doing in this PR!

ee/presencedetection/presencedetection.go
```diff
@@ -0,0 +1,116 @@
...
+	Success bool   `json:"success"`
+	Error   string `json:"error,omitempty"`
+}
+
+type PresenceDetector struct {
+	lastDetectionUTC time.Time
+	mutext           sync.Mutex
+}
+
+func (pd *PresenceDetector) DetectForConsoleUser(reason string, detectionInterval time.Duration) (bool, error) {
```
RebeccaMahany, line 29: Is some of this going to move inside the desktop runner instead, now that desktop is always running?

RebeccaMahany (commented):
Nice!!

cmd/launcher/desktop.go
James-Pickett: I was torn between this route and just passing a flag when we spin up desktop. We are able to show the desktop on demand, but as far as I can tell, there is no way to hide the desktop on demand. When you tell systray to shutdown, it exits the program completely. So if we ever want to hide it, we have to kill the process and let a new one spin up hidden.

I chose this route because I think it's rare that we would turn off the desktop and I don't want to be starting / killing processes on a first install where a user is trying auth for the first time.

ee/desktop/user/server/server.go
James-Pickett: yea, I think I'm doing it wrong =)

ee/presencedetection/presencedetection.go
James-Pickett: Yea, will drop the whole exec stuff an move the call to DesktopRunner. Still trying to find some elegant way for LocalServer to ask DesktopRunner to to presence detection.

cmd/launcher/desktop.go
RebeccaMahany: >I think it's rare that we would turn off the desktop

I agree, I think that's a reasonable assumption -- this route makes sense to me

ee/desktop/runner/runner.go
```diff
@@ -88,6 +88,32 @@ func InstanceDesktopProcessRecords() map[string]processRecord {
 	return instance.uidProcs
 }
 
+func InstanceDetectPresence(reason string, interval time.Duration) (bool, error) {
```
James-Pickett, line 91: singleton instance lazy route to avoid lots of plumbing ... maybe okay, possible revisit and do plumbing as time permits

ee/desktop/runner/runner.go
```diff
@@ -499,11 +548,11 @@ func (r *DesktopUsersProcessesRunner) refreshMenu() {
 		return
 	}
 
-	// Tell any running desktop user processes that they should refresh the latest menu data
+	// Tell any running desktop user processes that they should show menu and refresh the latest menu data
```
James-Pickett, line 551: no longer correct

```suggestion
	// Tell any running desktop user processes that they should refresh the latest menu data
```

ee/desktop/runner/runner.go
```diff
@@ -669,13 +712,19 @@ func (r *DesktopUsersProcessesRunner) spawnForUser(ctx context.Context, uid stri
 	r.waitOnProcessAsync(uid, cmd.Process)
 
 	client := client.New(r.userServerAuthToken, socketPath)
-	if err := backoff.WaitFor(client.Ping, 10*time.Second, 1*time.Second); err != nil {
+
+	pingFunc := client.Ping
```
James-Pickett, line 716: comment explaining this

ee/localserver/server.go
```diff
@@ -56,6 +58,8 @@ type localServer struct {
 
 	serverKey   *rsa.PublicKey
 	serverEcKey *ecdsa.PublicKey
+
+	presenceDetector presencedetection.PresenceDetector
```
James-Pickett, line 62: dont need this any more

ee/presencedetection/presencedetection.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+type PresenceDetectionResponse struct {
+	Success bool   `json:"success"`
+	Error   string `json:"error,omitempty"`
+}
```
James-Pickett, line 12: I believe this is unused

ee/presencedetection/presencedetection_darwin.go
```diff
@@ -0,0 +1,105 @@
+//go:build darwin
+// +build darwin
+
+package presencedetection
+
+/*
+#cgo CFLAGS: -x objective-c -fmodules -fblocks
```
James-Pickett, line 7: consider putting this in it's own file, easer to read

James-Pickett, line 7: consider putting C code in own file, easier to read

ee/desktop/runner/runner.go
```diff
@@ -452,12 +470,35 @@ func (r *DesktopUsersProcessesRunner) Update(data io.Reader) error {
...
-			"runner processSpawningEnabled set by control server",
-			"process_spawning_enabled", r.processSpawningEnabled,
-		)
+	if !slices.Contains(flagKeys, keys.DesktopEnabled) {
+		return
+	}
+
+	r.slogger.Log(context.TODO(), slog.LevelDebug,
+		"desktop enabled set by control server",
+		"process_spawning_enabled", r.knapsack.DesktopEnabled(),
```
zackattack01, line 479: do you think worth renaming the `process_spawning_enabled` field here? It seems like we've largely moved away from that verbiage/pattern elsewhere so just wondering if this might be confusing down the road

ee/desktop/user/server/server.go
```diff
@@ -152,6 +166,63 @@ func (s *UserServer) notificationHandler(w http.ResponseWriter, req *http.Reques
...
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	s.showDesktopOnceFunc()
+}
+
+type DetectPresenceResponse struct {
+	Success bool   `json:"success"`
+	Error   string `json:"error,omitempty"`
+}
```
James-Pickett, line 179: I'm not convinced on this Response object. Do we really need a bool or can we just rely on error?

Maybe:
```suggestion
type DetectPresenceResponse struct {
	TimeSinceLastDetection string   `json:"time_since_last_detection"`
	Error   string `json:"error,omitempty"`
}
```

ee/localserver/server.go
```diff
@@ -119,11 +127,19 @@ func New(ctx context.Context, k types.Knapsack) (*localServer, error) {
...
 
 	srv := &http.Server{
-		Handler: otelhttp.NewHandler(ls.requestLoggingHandler(ls.preflightCorsHandler(ls.rateLimitHandler(mux))), "localserver", otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
-			return r.URL.Path
-		})),
+		Handler: otelhttp.NewHandler(
+			ls.requestLoggingHandler(
+				ls.preflightCorsHandler(
+					ls.rateLimitHandler(
+						ls.presenceDetectionHandler(mux),
```
directionless, line 138: Does this need to be on `ecAuthedMux`?

cmd/launcher/desktop.go
directionless: I think users will, but I think your reasoning is sound.

ee/desktop/user/client/client.go
```diff
@@ -55,6 +59,41 @@ func (c *client) Refresh() error {
...
 
+func (c *client) ShowDesktop() error {
+	return c.get("show")
+}
+
+func (c *client) DetectPresence(reason string, interval time.Duration) (time.Duration, error) {
+	encodedReason := url.QueryEscape(reason)
+	encodedInterval := url.QueryEscape(interval.String())
+
+	resp, requestErr := c.base.Get(fmt.Sprintf("http://unix/detect_presence?reason=%s&interval=%s", encodedReason, encodedInterval))
```
directionless, line 70: Do we need a timeout here, or is there already one on c.base?

ee/presencedetection/presencedetection.go
```diff
@@ -0,0 +1,51 @@
...
+	// Check if the last detection was within the detection interval
+	if pd.hadOneSuccessfulDetection && time.Since(pd.lastDetectionUTC) < detectionInterval {
+		return time.Since(pd.lastDetectionUTC), nil
+	}
+
+	success, err := pd.detectFunc(reason)
+
+	switch {
+	case err != nil && pd.hadOneSuccessfulDetection:
+		return time.Since(pd.lastDetectionUTC), fmt.Errorf("detecting presence: %w", err)
```
directionless, line 38: This is a weird case. Do we want it?

ee/presencedetection/presencedetection.go
```diff
@@ -0,0 +1,51 @@
...
+	"sync"
+	"time"
+)
+
+const DetectionFailedDurationValue = -1 * time.Second
+
+type PresenceDetector struct {
+	lastDetectionUTC          time.Time
+	mutext                    sync.Mutex
+	hadOneSuccessfulDetection bool
```
directionless, line 14: You probably don't need the bool, but it's reads cleaner than checking if `lastDetectionUTC` is 0

ee/presencedetection/presencedetection.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+)
+
+const DetectionFailedDurationValue = -1 * time.Second
+
+type PresenceDetector struct {
+	lastDetectionUTC          time.Time
```
directionless, line 12: if this is `time.Time` maybe drop the `UTC`?

ee/presencedetection/presencedetection_darwin_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const testPresenceEnvVar = "launcher_test_presence"
```
directionless, line 11: I'd generally make this uppercase.

directionless (commented):
love it

ee/localserver/krypto-ec-middleware.go
```diff
@@ -334,6 +344,14 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 
 		w.Header().Add(kolideKryptoHeaderKey, kolideKryptoEccHeader20230130Value)
 
+		for k, v := range bhr.Header() {
```
James-Pickett, line 347: These headers are outside the krypto box, is that okay?

James-Pickett, line 347: Do we want to blindly copy all headers or cherry pick?

ee/localserver/server.go
James-Pickett: Good Catch! I'm sure that saved me a lot of time. I had to update krypto wrapping to include headers ... which created some questions, please see https://github.com/kolide/launcher/pull/1867#discussion_r1778870790

ee/localserver/krypto-ec-middleware.go
directionless: It would be better if they were signed, but also not fatal. I thought if we put them into the CmdReq they'd be inside?

ee/desktop/user/client/client.go
James-Pickett: yea, there is a default timeout of 30s on c.base

directionless: Can you add a comment to that effect?

ee/localserver/krypto-ec-middleware.go
James-Pickett: The headers of the request are inside the CmdReq (in krypto box). We don't have a mechanism currently to add headers from the response to the krypto box. Currently we just grab the body from the response and put it in the krypto box. We could come up with some kind of base response that we imbed in each of the local server responses.

ee/presencedetection/presencedetection.go
James-Pickett: Yea I agree, but worth nothing that we are implicitly saying if a detection fails, we behave as if no detection in the past has ever been successful.

ee/presencedetection/presencedetection.go
James-Pickett: agree, even checking for zero is not necessary becaues we are saying if `time.Since(0) < interval` ... as long as the interval is not greater than the begging of time =)

James-Pickett: welp, after I wrote that out .... i thought, what if we added an interval that was a century ... and put it back

ee/presencedetection/presencedetection.go
James-Pickett: will drop it, but we still want UTC instead of local time I'm assuming

ee/desktop/user/client/client.go
James-Pickett: done

ee/localserver/krypto-ec-middleware.go
directionless: I'm confused. We have too many layers

I think it's important for the response from launcher to k2, to be in the box. Otherwise it's a bit too trivial to slap whatever you want in flight.

directionless: But I'm not sure I'm concerned about what's happening between localserver and the desktop.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -306,15 +316,44 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
...
+			e.slogger.Log(r.Context(), slog.LevelError,
+				"unable to unmarshal response",
+				"err", err,
+			)
+			responseMap = map[string]any{
+				"headers": bhr.Header(),
+
+				// the request body was not in json format, just pass it through as "msg"
+				// property of json
+				"msg": string(bhrBytes),
```
directionless, line 333:
```suggestion
				"body": string(bhrBytes),
```

_maybe_? I'm really not sure here. I don't think anything we have takes this code path, and I'm not sure what we should aim for.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -306,15 +316,44 @@ func (e *kryptoEcMiddleware) Wrap(next http.Handler) http.Handler {
 		bhr := &bufferedHttpResponse{}
 		next.ServeHTTP(bhr, newReq)
 
+		// add headers to the response map
```
directionless, line 319:
```suggestion
		// add headers to the response map
		// This assumes that the response to `bhr` was a json encoded blob. 
```

directionless (approved):
I'm game to try! wooo!

ee/localserver/krypto-ec-middleware.go
James-Pickett: done

ee/presencedetection/presencedetection.go
directionless: Generally it's better to keep things in UTC, yes. And when we emit to k2, we should use one of the standard formats, or just integer unix time

directionless (approved):

ee/presencedetection/presencedetection.go
James-Pickett: cool, we don't emit the time for this to K2, just the go style duration since last detection like `0s` or `1m30s`

ee/presencedetection/auth.m
```diff
@@ -0,0 +1,68 @@
...
+                if (policySuccess) {
+                    success = true; // Authentication successful
+                } else {
+                    success = false;
+                    errorCode = (int)[error code];
+                    errorMessage = [error localizedDescription];
+                    if (error.code == LAErrorUserFallback || error.code == LAErrorAuthenticationFailed) {
+                        // Prompting for password
+                        [myContext evaluatePolicy:LAPolicyDeviceOwnerAuthentication
+                            localizedReason:@"Please enter your password"
```
RebeccaMahany, line 32: I don't think it's worth doing in this PR -- but maybe we would want to have this text provided by K2 in the future in another header, so that if we ever do localization for text K2 can handle it instead of us?

RebeccaMahany (approved):
A couple very small comments -- feel free to ignore them or punt to another PR. LGTM!

ee/presencedetection/auth.m
James-Pickett: This value never actually appears if select "use password", but it's still required. Maybe this only appears if you don' t have bio / touch id enabled. Either way, I set it to use same reason as bio / touch id.

ee/localserver/server.go
```diff
@@ -393,3 +409,56 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
...
+
+		durationSinceLastDetection, err := ls.presenceDetector.DetectPresence(reason, detectionIntervalDuration)
+
+		ls.slogger.Log(r.Context(), slog.LevelError,
+			"presence_detection",
+			"reason", reason,
+			"interval", detectionIntervalDuration,
+			"duration_since_last_detection", durationSinceLastDetection,
+			"err", err,
+		)
```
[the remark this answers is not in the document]
James-Pickett: yes, good catch, I did both since I suppose if the user cancels the auth, it's not really a launcher error

ee/desktop/user/server/server.go
```diff
@@ -152,6 +166,70 @@ func (s *UserServer) notificationHandler(w http.ResponseWriter, req *http.Reques
...
+	// detect presence
+	durationSinceLastDetection, err := s.presenceDetector.DetectPresence(reason, interval)
+	response := DetectPresenceResponse{
+		DurationSinceLastDetection: durationSinceLastDetection.String(),
+	}
+
+	if err != nil {
+		response.Error = err.Error()
+
+		s.slogger.Log(context.TODO(), slog.LevelDebug,
```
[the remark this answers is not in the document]
James-Pickett: done!

PR #1869, opened 2024-09-17
Set slogger on rungroup after logshipping is set up, so that rungroup logs will be shipped

RebeccaMahany:
Despite the changes in https://github.com/kolide/launcher/pull/1822, I wasn't seeing rungroup shutdown logs being shipped. When looking through the shipped logs, I noticed that the only rungroup logs available were either from very old versions of launcher, or from the desktop process.

This PR updates when we set the slogger on the rungroup so that we set it after the shipping handler has been added.

### DIFF
--- a/cmd/launcher/desktop.go
+++ b/cmd/launcher/desktop.go
@@ -111,7 +111,8 @@ func runDesktop(_ *multislogger.MultiSlogger, args []string) error {
 		)
 	}
 
-	runGroup := rungroup.NewRunGroup(slogger)
+	runGroup := rungroup.NewRunGroup()
+	runGroup.SetSlogger(slogger)
 
 	// listen for signals
 	runGroup.Add("desktopSignalListener", func() error {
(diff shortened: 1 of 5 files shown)


directionless (approved):
Hrm. Not sure I love the changed signature. It feels error prone.

We could maybe move `NewRunGroup` to after the logshipper is setup? Though it would mean moving `runGroup.Add` till later too

zackattack01 (commented):
great find! this looks good to me I think you just have to fix up the watchdog_service_windows `NewRunGroup` call - those are only written to sqlite so I don't think order/timing should matter at all

RebeccaMahany:
@directionless 

> Hrm. Not sure I love the changed signature. It feels error prone.

> We could maybe move NewRunGroup to after the logshipper is setup? Though it would mean moving runGroup.Add till later too

I considered that, but felt like moving all the runGroup.Add calls later was more error prone 😅

zackattack01 (approved):

James-Pickett (approved):
nice

RebeccaMahany:
I wanted to understand why this was happening, and figured it out. When we set the new log shipping handler, we initialize a new slogger, so the previous reference to the slogger never gets that new handler. Because of this, I'm going to go ahead and update other uses of the old `slogger` in launcher main just to be safe.

> 	// we have to rebuild the handler everytime because the slogmulti package we're
	// using doesn't support adding handlers after the Fanout handler has been created

Issue #1864, opened 2024-09-13
`homebrew` table should not error when homebrew isn't installed


directionless:
I was perusing bugsnag, and I noticed we have a [lot](https://app.bugsnag.com/kolide/k2-device/errors/65368b57ff1d330008f571d2) of [errors](https://app.bugsnag.com/kolide/k2-device/errors/6661e0bb5f85da0008cd69a9) because the homebrew table can't find homebew. 
```
error generating table: failure allocating allowedcmd.Brew: homebrew not found
```

and 
```
error generating table: failure allocating allowedcmd.Brew: not found: /home/linuxbrew/.linuxbrew/bin/brew
```

This is coming from https://github.com/kolide/launcher/blob/766d939df65f7584e5fe8aedfc288d00efc610fc/ee/tables/homebrew/upgradeable.go#L45-L48

For _this_ table, I think we should identify that error, and return no results.

Digging in slightly, it seems like `validatedCommand` should have returned a different error, so I wonder if something is slightly amiss here.

In terms of solutions, it's probably more correct to plumb a custom error.

Dharma-09:
Can I work on this issue?

directionless:
Fixed By https://github.com/kolide/launcher/pull/1872

PR #1872, opened 2024-09-19
Custom error for homebrew table not found

Dharma-09:
Which issue(s) this PR fixes:
#1864 

Description:
In the given solution, an error is returned every time an issue occurs during the `validatedCommand` call, regardless of the cause. For instance, if the binary file does not exist, should we use a specific error to handle this scenario?

### DIFF
--- a/ee/allowedcmd/cmd.go
+++ b/ee/allowedcmd/cmd.go
@@ -8,11 +8,11 @@ package allowedcmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
-	"errors"
 )
 
 type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
--- a/ee/allowedcmd/cmd.go
+++ b/ee/allowedcmd/cmd.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"errors"
 )
 
 type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
@@ -19,6 +20,8 @@ type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
 func newCmd(ctx context.Context, fullPathToCmd string, arg ...string) *exec.Cmd {
 	return exec.CommandContext(ctx, fullPathToCmd, arg...) //nolint:forbidigo // This is our approved usage of exec.CommandContext
 }
+
+var ErrCommandNotFound = errors.New("command not found")
 
 func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*exec.Cmd, error) {
 	knownPath = filepath.Clean(knownPath)
@@ -31,7 +34,7 @@ func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*ex
 	// We expect to know the exact location for allowlisted commands on all
 	// OSes except for a few Linux distros.
 	if !allowSearchPath() {
-		return nil, fmt.Errorf("not found: %s", knownPath)
+		return nil, fmt.Errorf("%w: %s",ErrCommandNotFound, knownPath)
 	}
 
 	cmdName := filepath.Base(knownPath)
@@ -39,7 +42,7 @@ func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*ex
 		return newCmd(ctx, foundPath, arg...), nil
 	}
 
-	return nil, fmt.Errorf("%s not found at %s and could not be located elsewhere", cmdName, knownPath)
+	return nil, fmt.Errorf("%w: %s and could not be located elsewhere", ErrCommandNotFound, knownPath)
 }
 
 func allowSearchPath() bool {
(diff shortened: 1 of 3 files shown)


ee/allowedcmd/cmd_linux.go
```diff
@@ -13,10 +13,12 @@ func Apt(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 	return validatedCommand(ctx, "/usr/bin/apt", arg...)
 }
 
+var ErrHomebrewNotFound = errors.New("homebrew not found")
```
RebeccaMahany, line 16: I am wondering if we would instead want to move this to `cmd.go`, rename it something more generic like `ErrCommandNotFound` and return it from the `validatedCommand` function -- this new error is very useful here, and I imagine we might want to do similar in the future when other commands aren't found. 

(This will also have the bonus side effect of fixing the build errors, since we run `Brew` on both macOS and Linux so this new error needs to be available to darwin builds as well.)

Dharma-09: That's great idea, I'll update the PR with the changes.

ee/allowedcmd/cmd_linux.go
```diff
@@ -16,7 +16,7 @@ func Apt(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 func Brew(ctx context.Context, arg ...string) (*exec.Cmd, error) {
 	validatedCmd, err := validatedCommand(ctx, "/home/linuxbrew/.linuxbrew/bin/brew", arg...)
 	if err != nil {
-		return nil, err
+		return nil, nil
```
RebeccaMahany, line 19: I think this should still be `nil, err`, so we can return the wrapped `ErrCommandNotFound` to `upgradeable.go` to check

ee/allowedcmd/cmd.go
```diff
@@ -31,15 +34,15 @@ func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*ex
 	// We expect to know the exact location for allowlisted commands on all
 	// OSes except for a few Linux distros.
 	if !allowSearchPath() {
-		return nil, fmt.Errorf("not found: %s", knownPath)
+		return nil, fmt.Errorf("%w: %s",ErrCommandNotFound, knownPath)
```
RebeccaMahany, line 37: Tiny nitpick -- I think the spacing is off here --

```suggestion
		return nil, fmt.Errorf("%w: %s", ErrCommandNotFound, knownPath)
```

ee/allowedcmd/cmd.go
```diff
@@ -31,15 +34,15 @@ func validatedCommand(ctx context.Context, knownPath string, arg ...string) (*ex
...
 	}
 
 	cmdName := filepath.Base(knownPath)
 	if foundPath, err := exec.LookPath(cmdName); err == nil {
 		return newCmd(ctx, foundPath, arg...), nil
 	}
 
-	return nil, fmt.Errorf("%s not found at %s and could not be located elsewhere", cmdName, knownPath)
+	return nil, fmt.Errorf("%w: %s and could not be located elsewhere", ErrCommandNotFound, knownPath)
```
RebeccaMahany, line 45: Another nitpick -- I think the error message will read more clearly like this:

```suggestion
	return nil, fmt.Errorf("%w: not found at %s and could not be located elsewhere", ErrCommandNotFound, knownPath)
```

RebeccaMahany (commented):
Just a couple small comments, looks good to me overall! Will hold off on approval till @directionless or another team member gets a chance to look as well 🙂

ee/tables/homebrew/upgradeable.go
```diff
@@ -43,7 +44,12 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	// that user. To reduce duplicating the WithUid table helper, we can find the owner of the binary,
 	// and pass the said owner to the WIthUid method to handle setting the appropriate env vars.
 	cmd, err := allowedcmd.Brew(ctx)
+		
 	if err != nil {
+		if errors.Is(err, allowedcmd.ErrCommandNotFound) {
+			// No data, no error
+			return nil, err
```
RebeccaMahany, line 51: I think you updated the wrong one -- this should still be `return nil, nil` as you had it before 🙂

directionless (approved):
I think Becca covered all my comments. Thank you so much for you contribution!

ee/allowedcmd/cmd.go
```diff
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"errors"
```
RebeccaMahany, line 15: It looks like the lint job is complaining because this import is not added alphabetically -- same goes for other file as well

RebeccaMahany (commented):
LGTM once the lint passes!

ee/tables/homebrew/upgradeable.go
```diff
@@ -43,7 +44,12 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	// that user. To reduce duplicating the WithUid table helper, we can find the owner of the binary,
 	// and pass the said owner to the WIthUid method to handle setting the appropriate env vars.
 	cmd, err := allowedcmd.Brew(ctx)
+
```
RebeccaMahany, line 47: Apologies, missed this previously -- looks like the lint job is complaining about this line as well. (If you want to validate for yourself before CI, you can [install golangci-lint](https://golangci-lint.run/welcome/install/#local-installation) and then run `golangci-lint run`.)

RebeccaMahany (approved):
🔥

PR #1873, opened 2024-09-20
batch tmutil call args to prevent argument list too long errors

zackattack01:
We're seeing 18k errors logged in the last two days for `fork/exec /usr/bin/tmutil: argument list too long` - I am unable to reproduce this error on my machine but some brief research indicates that this is determined by the `ARG_MAX` set per machine.

We generate the filepaths for exclusion based on the output of `filepath.Glob` on our predefined patterns. This can also vary slightly per machine (mostly by the count of debug logs at the time this runs). To keep this simple I've approximated two equal (ish) batches of patterns, moved the exclusion logic into a helper, and run it twice. This should cut the args per call roughly in half and reduce the likelihood of hitting this error

### DIFF
--- a/ee/agent/timemachine/timemachine_darwin.go
+++ b/ee/agent/timemachine/timemachine_darwin.go
@@ -33,10 +33,17 @@ func AddExclusions(ctx context.Context, k types.Knapsack) {
 		launcher.pid
 	*/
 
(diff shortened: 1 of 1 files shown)


RebeccaMahany (approved):
Nice find!!

PR #1877, opened 2024-10-02
include arch and os in local server headers

James-Pickett:
In the event that presence detection is requested for a non-supported OS, return default failure duration `-1s`. Always include OS and Arch in headers.

### DIFF
--- a/ee/localserver/server.go
+++ b/ee/localserver/server.go
@@ -421,7 +421,7 @@ func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler 
 
 		// can test this by adding an unauthed endpoint to the mux and running, for example:
 		// curl -i -H "X-Kolide-Presence-Detection-Interval: 10s" -H "X-Kolide-Presence-Detection-Reason: my reason" localhost:12519/id
-		detectionIntervalStr := r.Header.Get(kolidePresenceDetectionInterval)
+		detectionIntervalStr := r.Header.Get(kolidePresenceDetectionIntervalHeaderKey)
 
 		// no presence detection requested
 		if detectionIntervalStr == "" {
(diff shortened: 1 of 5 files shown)


ee/localserver/server.go
```diff
@@ -415,13 +416,14 @@ func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler
 
 		// presence detection is only supported on macos currently
 		if runtime.GOOS != "darwin" {
+			w.Header().Add(kolideDurationSinceLastPresenceDetectionHeaderKey, presencedetection.DetectionFailedDurationValue.String())
```
directionless, line 419: Thinking about this, we'll need to figure out how to differentiate between "not supported" and "timeout" or "error"

directionless (dismissed):
Looks like forward progress

ee/localserver/server.go
James-Pickett: I adjusted it so we only return the presence response header when presence detection is actually requested.

That might give us enough data for some logic like

```
if failure && os == "linux"
  // we no it's not supported
```

but maybe we need something more concrete like a failure_reason header

directionless (approved):
I'm still pretty sure we're going to need to iterate. But I think it's okay to do it post merge.

directionless:
Let's see, I think we have the following conditions:
- presence detection supported, success
- presence detection supported, failure / timeout (Are there failures other than timeout? Maybe someone can click no)
- presence detection not supported
- old launcher so it's just 🤷 (eg: no headers at all)

It might additionally be interesting to track the reasons it failed. At least, if there are reasons we can track.

So we probably do need some kind of header for why it failed. And one reason can be `unsupported` (Or, if we don't like enums, we can have a boolean unsupported header and a string error header. But I think either is probably fine)

RebeccaMahany (approved):

PR #1878, opened 2024-10-07
added download info to flare

### LINKED ISSUE #1753: Add download directory information to flare
It is helpful for us when troubleshooting to know which Kolide installers a user has potentially downloaded.

Requirements:

- The download checkup should run only for flares (not for `launcher doctor` or for the log checkpoint -- see [here](https://github.com/kolide/launcher/blob/main/ee/debug/checkups/checkups.go#L88))
- The download checkup should write out a list of all Kolide installers found in the downloads directory; installers will have the filename or filename suffix `kolide-launcher.msi` on Windows, `kolide-launcher.pkg` on MacOS, and `kolide-launcher.rpm`, `kolide-launcher.deb`, or `kolide-launcher.pacman` on Linux
- If possible, the checkup should include launcher version information for those files, if it's feasible and reasonable to extract that information

Checkups live [here](https://github.com/kolide/launcher/tree/main/ee/debug/checkups) in the codebase.

cesarfda:
- The download checkup should run only for flares 

- The download checkup should write out a list of all Kolide installers found in the downloads directory; installers will have the filename or filename suffix kolide-launcher.msi on Windows, kolide-launcher.pkg on MacOS, and kolide-launcher.rpm, kolide-launcher.deb, or kolide-launcher.pacman on Linux

Local example:
image

Closes #1753

### DIFF
--- a/ee/debug/checkups/download_directory.go
+++ b/ee/debug/checkups/download_directory.go
@@ -10,51 +10,57 @@ import (
 	"strings"
 )
 
-type DownloadDirectory struct {
+type downloadDirectory struct {
 	status  Status
 	summary string
-	files   []string
+	files   []fileInfo
 }
 
-func (c *DownloadDirectory) Name() string {
-	return "Download directory contents"
+func (c *downloadDirectory) Name() string {
+	return "Download directory contents for all users"
 }
 
-func (c *DownloadDirectory) Run(_ context.Context, extraFH io.Writer) error {
-	downloadDir := getDownloadDir()
-	if downloadDir == "" {
-		return fmt.Errorf("no default download directory")
+func (c *downloadDirectory) Run(_ context.Context, extraFH io.Writer) error {
+	downloadDirs := getDownloadDirs()
+	if len(downloadDirs) == 0 {
+		c.status = Erroring
+		c.summary = "No download directories found"
+		return nil
 	}
 
-	err := filepath.Walk(downloadDir, func(path string, info os.FileInfo, err error) error {
+	for _, downloadDir := range downloadDirs {
+		pattern := filepath.Join(downloadDir, "kolide-launcher*")
+		matches, err := filepath.Glob(pattern)
+
 		if err != nil {
-			return err
+			fmt.Fprintf(extraFH, "Error listing files in directory (%s): %s\n", downloadDir, err)
+			continue
 		}
-		if !info.IsDir() && isKolideInstaller(info.Name()) {
-			c.files = append(c.files, path)
+
+		for _, match := range matches {
+			if info, err := os.Stat(match); err == nil {
+				c.files = append(c.files, fileInfo{
+					Name:    match,
+					ModTime: info.ModTime(),
+				})
+			}
 		}
-		return nil
-	})
+	}
 
 	switch {
-	case os.IsNotExist(err):
-		c.status = Warning
-		c.summary = fmt.Sprintf("download directory (%s) not present", downloadDir)
-	case err != nil:
-		c.status = Erroring
-		c.summary = fmt.Sprintf("error listing files in directory (%s): %s", downloadDir, err)
 	case len(c.files) == 0:
-		c.status = Warning
-		c.summary = fmt.Sprintf("no Kolide installers found in directory (%s)", downloadDir)
+		c.status = Informational
+		c.summary = "No Kolide installers found in any user's download directory"
 	default:
-		c.status = Passing
-		fileNames := make([]string, len(c.files))
+		c.status = Informational
+		fileInfos := make([]string, len(c.files))
 		for i, file := range c.files {
-			fileNames[i] = filepath.Base(file)
+			fileName := filepath.Base(file.Name)
+			modTime := file.ModTime.Format("Mon, 02 Jan 2006 15:04:05 MST")
+			fileInfos[i] = fmt.Sprintf("%s (Modified: %s)", fileName, modTime)
 		}
-		installerList := strings.Join(fileNames, ", ")
-		c.summary = fmt.Sprintf("Found Kolide installer(s) in directory (%s): %s", downloadDir, installerList)
-
+		installerList := strings.Join(fileInfos, ", ")
+		c.summary = fmt.Sprintf("Found Kolide installer(s) across user download directories: %s", installerList)
 	}
 
 	if len(c.files) > 0 {
@@ -67,51 +73,45 @@ func (c *DownloadDirectory) Run(_ context.Context, extraFH io.Writer) error {
 	return nil
 }
 
-func (c *DownloadDirectory) ExtraFileName() string {
-	return "kolide-installers"
+func (c *downloadDirectory) ExtraFileName() string {
+	return "kolide-installers-all-users.txt"
 }
 
-func (c *DownloadDirectory) Status() Status {
+func (c *downloadDirectory) Status() Status {
 	return c.status
 }
 
-func (c *DownloadDirectory) Summary() string {
+func (c *downloadDirectory) Summary() string {
 	return c.summary
 }
 
-func (c *DownloadDirectory) Data() any {
+func (c *downloadDirectory) Data() any {
 	return c.files
 }
 
-func getDownloadDir() string {
-	homeDir, err := os.UserHomeDir()
+func getDownloadDirs() []string {
+	var userDirs []string
+	var baseDir string
+
+	if runtime.GOOS == "windows" {
+		baseDir = "C:\\Users"
+	} else {
+		baseDir = "/Users"
+	}
+
+	entries, err := os.ReadDir(baseDir)
 	if err != nil {
-		return ""
+		return nil
 	}
 
-	switch runtime.GOOS {
-	case "darwin":
-		return filepath.Join(homeDir, "Downloads")
-	case "linux":
-		return filepath.Join(homeDir, "Downloads")
-	case "windows":
-		return filepath.Join(homeDir, "Downloads")
+	for _, entry := range entries {
+		if entry.IsDir() {
+			userDir := filepath.Join(baseDir, entry.Name(), "Downloads")
+			if _, err := os.Stat(userDir); err == nil {
+				userDirs = append(userDirs, userDir)
+			}
+		}
 	}
 
-	return ""
-}
-
-func isKolideInstaller(filename string) bool {
-	lowerFilename := strings.ToLower(filename)
-	switch runtime.GOOS {
-	case "darwin":
-		return strings.HasSuffix(lowerFilename, "kolide-launcher.pkg")
-	case "linux":
-		return strings.HasSuffix(lowerFilename, "kolide-launcher.rpm") ||
-			strings.HasSuffix(lowerFilename, "kolide-launcher.deb") ||
-			strings.HasSuffix(lowerFilename, "kolide-launcher.pacman")
-	case "windows":
-		return strings.HasSuffix(lowerFilename, "kolide-launcher.msi")
-	}
-	return false
+	return userDirs
 }
--- a/ee/debug/checkups/download_directory.go
+++ b/ee/debug/checkups/download_directory.go
@@ -95,8 +95,10 @@ func getDownloadDirs() []string {
 
 	if runtime.GOOS == "windows" {
 		baseDir = "C:\\Users"
-	} else {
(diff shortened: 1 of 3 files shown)


ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (c *DownloadDirectory) Name() string {
+	return "Download directory contents"
+}
+
+func (c *DownloadDirectory) Run(_ context.Context, extraFH io.Writer) error {
+	downloadDir := getDownloadDir()
+	if downloadDir == "" {
+		return fmt.Errorf("no default download directory")
```
RebeccaMahany, line 26: Nitpick -- `errors.New` instead of `fmt.Errorf` since there are no variables to interpolate

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return ""
+}
+
+func isKolideInstaller(filename string) bool {
+	lowerFilename := strings.ToLower(filename)
+	switch runtime.GOOS {
+	case "darwin":
+		return strings.HasSuffix(lowerFilename, "kolide-launcher.pkg")
```
RebeccaMahany, line 108: I think we'd expect to see files like `kolide-launcher(1).pkg` too. Maybe we want a regex match instead of a suffix check to be able to capture those? Or maybe `filepath.Glob` inside the download directory, instead of walking the download directory.

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+	if downloadDir == "" {
+		return fmt.Errorf("no default download directory")
+	}
+
+	err := filepath.Walk(downloadDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() && isKolideInstaller(info.Name()) {
+			c.files = append(c.files, path)
```
RebeccaMahany, line 34: Sorry, I completely missed documenting this on the original issue -- could we also include the time each file was created, so we know when the user performed the download? Can grab it here with `info.ModTime()`, I think.

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+		for _, file := range c.files {
+			fmt.Fprintln(extraFH, file)
+		}
+	}
+
+	return nil
+}
+
+func (c *DownloadDirectory) ExtraFileName() string {
+	return "kolide-installers"
```
RebeccaMahany, line 71: I think we usually set a file extension here -- .txt seems fine to me?

```suggestion
	return "kolide-installers.txt"
```

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+func (c *DownloadDirectory) Summary() string {
+	return c.summary
+}
+
+func (c *DownloadDirectory) Data() any {
+	return c.files
+}
+
+func getDownloadDir() string {
+	homeDir, err := os.UserHomeDir()
```
zackattack01, line 87: will this work correctly when running as root?

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+
+	switch {
+	case os.IsNotExist(err):
+		c.status = Warning
+		c.summary = fmt.Sprintf("download directory (%s) not present", downloadDir)
+	case err != nil:
+		c.status = Erroring
+		c.summary = fmt.Sprintf("error listing files in directory (%s): %s", downloadDir, err)
+	case len(c.files) == 0:
+		c.status = Warning
```
RebeccaMahany, line 47: I think this should be `Passing` -- `Warning` would imply a potential issue with how launcher is running.

ee/debug/checkups/download_directory.go
directionless: Ideally we do this for all the users.

RebeccaMahany (commented):
Couple small comments, but the shape of this looks good to me overall! 🔥

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+	return "Download directory contents"
+}
+
+func (c *DownloadDirectory) Run(_ context.Context, extraFH io.Writer) error {
+	downloadDir := getDownloadDir()
+	if downloadDir == "" {
+		return fmt.Errorf("no default download directory")
+	}
+
+	err := filepath.Walk(downloadDir, func(path string, info os.FileInfo, err error) error {
```
directionless, line 29: I think of `filepath.Walk` as being for recursion. Do you think we want to recurse through this, or examine only the toplevel? (I could go either way)

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
```
directionless, line 30: Prefer the single line -- `if err := ...; err != nil {`

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		if !info.IsDir() && isKolideInstaller(info.Name()) {
+			c.files = append(c.files, path)
+		}
+		return nil
+	})
+
+	switch {
+	case os.IsNotExist(err):
+		c.status = Warning
```
directionless, line 41: This is not a warning about the health of launcher -- it's `Erroring` probably.

ee/debug/checkups/download_directory.go
directionless, line 47: I would flag this as Informational.

ee/debug/checkups/download_directory.go
RebeccaMahany: I remembered we have an example elsewhere for grabbing home directories for all users -- https://github.com/kolide/launcher/blob/main/pkg/osquery/table/table_util.go#L23-L27 plus `findFileInUserDirs` below. Could probably do similar here!

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+		c.summary = fmt.Sprintf("download directory (%s) not present", downloadDir)
+	case err != nil:
+		c.status = Erroring
+		c.summary = fmt.Sprintf("error listing files in directory (%s): %s", downloadDir, err)
+	case len(c.files) == 0:
+		c.status = Warning
+		c.summary = fmt.Sprintf("no Kolide installers found in directory (%s)", downloadDir)
+	default:
+		c.status = Passing
```
directionless, line 50: Similar -- I would flag this as Informational.

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+type DownloadDirectory struct {
```
directionless, line 13: Does this need to be exported?

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+	return c.files
+}
+
+func getDownloadDir() string {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return ""
+	}
+
+	switch runtime.GOOS {
```
directionless, line 92: Why switch?

directionless (commented):
Nice beginning!

ee/debug/checkups/download_directory.go
cesarfda: I was expecting those to be different, I didn't have the chance to it on windows so I kept the switch in case I need to handle it differently for it. I'll remove it and handle it later if the windows download path doesn't work

ee/debug/checkups/download_directory.go
directionless: glob for `kolide*<suffix>` makes sense to me.

ee/debug/checkups/download_directory.go
cesarfda: Added support for all users

ee/debug/checkups/download_directory.go
cesarfda: updated

ee/debug/checkups/download_directory.go
cesarfda: added support for Mod time

ee/debug/checkups/download_directory.go
cesarfda: added .txt

ee/debug/checkups/download_directory.go
cesarfda: removed warnings

ee/debug/checkups/download_directory.go
cesarfda: used Glob for finding file match

ee/debug/checkups/download_directory.go
cesarfda: removed filepath.Walk

ee/debug/checkups/download_directory.go
cesarfda: Done

ee/debug/checkups/download_directory.go
cesarfda: Since it's only used in this package, I have changed it to not be exported

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func getDownloadDirs() []string {
+	var userDirs []string
+	var baseDir string
+
+	if runtime.GOOS == "windows" {
+		baseDir = "C:\\Users"
+	} else {
+		baseDir = "/Users"
```
RebeccaMahany, line 99: Sorry, missed on the last review -- at least on my Ubuntu install, the baseDir would be `/home` (e.g. my downloads directory is `/home/becca/Downloads`). I think we'll want to do `/Users` if runtime.GOOS is darwin, and `/home` otherwise.

cesarfda: thanks for catching that, I have addressed it

James-Pickett (dismissed):
Nice!

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,119 @@
...
+					ModTime: info.ModTime(),
+				})
+			}
+		}
+	}
+
+	switch {
+	case len(c.files) == 0:
+		c.status = Informational
+		c.summary = "No Kolide installers found in any user's download directory"
```
directionless, line 53: FWIW `status` and `summary` mostly matter for doctor commands, I basically always ignore them for flares

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,119 @@
...
+		baseDir = "/home"
+	}
+
+	entries, err := os.ReadDir(baseDir)
+	if err != nil {
+		return nil
+	}
+
+	for _, entry := range entries {
+		if entry.IsDir() {
```
directionless, line 110: Prefer early returns over deep nesting
```
if !entry.IsDir(){
  continue
}
```

directionless (dismissed):

zackattack01 (dismissed):
nice!!

ee/debug/checkups/download_directory.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type downloadDirectory struct {
+	status  Status
+	summary string
+	files   []fileInfo
+}
+
+func (c *downloadDirectory) Name() string {
+	return "Download directory contents for all users"
```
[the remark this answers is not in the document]
RebeccaMahany: Good point, I think `Installer Downloads` is a good idea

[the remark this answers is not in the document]
directionless: Or maybe `Kolide Downloads` ?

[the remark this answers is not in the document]
cesarfda: Agreed, will change this

ee/debug/checkups/download_directory.go
cesarfda: done

zackattack01 (approved):

RebeccaMahany (approved):
🔥

PR #1879, opened 2024-10-07
james/disable arm64 windows packaging

James-Pickett:
disable arm64 windows packaging


directionless (approved):
Do you want to flag this, so you can still have the dual packages for nababe?

James-Pickett:
ahh .. just realized the condition will stop it form installing (then emulating) on amd ... will reopen when ready

PR #1880, opened 2024-10-08
disable arm64 builds for non kolide

James-Pickett:
This disables arm64 builds in packaging to allow arm64 machines to emulate amd64. Internal builds will still have arm64.


James-Pickett:
We've stabilized arm64, so this is un-needed

PR #1881, opened 2024-10-08
use kolide go ole

James-Pickett:


go.sum
```diff
@@ -68,8 +68,6 @@ github.com/go-logr/logr v1.4.2/go.mod h1:9T104GzyrTigFIr8wt5mBrctHMim0Nb2HLGrmQ4
 github.com/go-logr/stdr v1.2.2 h1:hSWxHoqTgW2S2qGc0LTAI563KZ5YKYRhT3MFKZMbjag=
 github.com/go-logr/stdr v1.2.2/go.mod h1:mMo/vtBO5dYbehREoey6XUKy/eSumjCCveDpRre4VKE=
 github.com/go-ole/go-ole v1.2.6/go.mod h1:pprOEPIfldk/42T2oK7lQ4v4JSDwmV0As9GaiUsvbm0=
```
RebeccaMahany, line 70: Do we need to replace this reference also, or is it safe?

RebeccaMahany (approved):

go.sum
James-Pickett: I think it's okay ... reading about it apparently the go.sum file also keeps some kind of historical reference

directionless: `go.sum` does whatever it does, and I don't think we should be editing it. As long as `go.mod` is correct...

PR #1890, opened 2024-10-14
Perform presence detection on Windows using UserConsentVerifier interop

RebeccaMahany:
Supersedes https://github.com/kolide/launcher/pull/1865 with a more appropriate implementation for our usecase.

This PR uses [UserConsentVerifier.RequestVerificationAsync](https://learn.microsoft.com/en-us/uwp/api/windows.security.credentials.ui.userconsentverifier.requestverificationasync?view=winrt-26100) to perform presence detection. Because we are not a UWP app, we have to call this method by way of the interop interface --`IUserConsentVerifierInterop.RequestVerificationForWindowAsync`.

Calling `RequestVerificationForWindowAsync` requires a reference to a window, which we already have via systray, so I exposed that window in systray for use here.

I pulled in `winrt-go` because it makes it significantly easier to call async functions.


ee/presencedetection/presencedetection_windows.go
```diff
@@ -0,0 +1,305 @@
...
+	})
+	defer handler.Release()
+	requestVerificationAsyncOperation.SetCompleted(handler)
+
+	select {
+	case operationStatus := <-statusChan:
+		if operationStatus != foundation.AsyncStatusCompleted {
+			return fmt.Errorf("RequestVerificationForWindowAsync operation did not complete: status %d", operationStatus)
+		}
+	case <-time.After(1 * time.Minute):
```
directionless, line 167: We should set this duration as a constant, and make sure it's shared between the platforms. Similarly, the errors returned

ee/presencedetection/presencedetection_windows.go
```diff
@@ -0,0 +1,305 @@
...
+// See: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexw
+func createWindow() (syscall.Handle, error) {
+	instance, err := getInstance()
+	if err != nil {
+		return syscall.InvalidHandle, fmt.Errorf("getting instance: %w", err)
+	}
+
+	// TODO RM: need to close handles?
+
+	className := "launcher"
```
directionless, line 194: what's the className used for, do we need to worry about multiple launcher conflicts?

ee/presencedetection/presencedetection_windows.go
```diff
@@ -0,0 +1,305 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"syscall"
+	"time"
+	"unsafe"
+
+	ole "github.com/go-ole/go-ole"
```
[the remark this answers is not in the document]
RebeccaMahany: We've got an override in go.mod -- https://github.com/kolide/launcher/blob/main/go.mod#L129

ee/presencedetection/presencedetection_windows.go
RebeccaMahany: I believe it identifies the class that we have to register -- https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassexa. I don't know if it needs to be unique or not -- this section still requires a lot of cleanup 😅 and I am not sure how much of this will be in the final version.

ee/presencedetection/presencedetection_other.go
```diff
@@ -1,5 +1,5 @@
-//go:build !darwin
```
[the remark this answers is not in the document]
RebeccaMahany: Yes, missed that when rebasing! Will update

ee/presencedetection/presencedetection_windows.go
RebeccaMahany: I don't think className has to be unique -- systray uses className `SystrayClass` and we don't have any issues ther ewith multiple launcher conflicts

ee/presencedetection/presencedetection_windows.go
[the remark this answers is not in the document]
directionless: 🤦

ee/presencedetection/presencedetection_other.go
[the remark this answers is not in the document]
RebeccaMahany: Done!

ee/presencedetection/presencedetection_windows.go
RebeccaMahany: Added a shared constant for timeout -- still need to update to return the error in a friendlier way

RebeccaMahany: Added strings to capture error message more accurately, I imagine TBD on how we want to handle these

ee/presencedetection/presencedetection_windows.go
```diff
@@ -0,0 +1,178 @@
...
+	// Query for the interop interface, which we need to actually interact with this method
+	verifierObj, err := factory.QueryInterface(iUserConsentVerifierInteropGuid)
+	if err != nil {
+		return fmt.Errorf("getting UserConsentVerifier from factory: %w", err)
+	}
+	defer verifierObj.Release()
+	verifier := (*IUserConsentVerifierInterop)(unsafe.Pointer(verifierObj))
+
+	// Get the current window handle from systray
+	windowHwnd, err := systray.WindowHandle()
```
James-Pickett, line 103:
```suggestion
	windowHandler, err := systray.WindowHandle()
```
I think? ... just a nit

James-Pickett (dismissed):
Nice!

ee/presencedetection/presencedetection_windows.go
RebeccaMahany: `HWND` is the Windows type for a window handle -- `windowHwnd` (`window handle [to] window`) is redundant, though 😅 . I'll update!

James-Pickett (approved):

PR #1891, opened 2024-10-15
Patch `kolide_macos_software_update` for MacOS15

Micah-Kolide:
MacOS 15 has come with a lot of changes to the software update frameworks. With these changes some methods are still supported, some deprecated, and some are now static.

With this PR, new data is being added. The main thing of note is that we never tracked if "Install OS Updates" was managed, but this PR brings that data in with some other settings as well.

It appears that "Auto Check for Updates" is now static (readonly), so I believe it makes sense to always return true for it being enabled if the os version is on MacOS 15+. There's no way of validating if it's managed unless we figure out how to utilize `SUOSUClientProtocol`.

The new class does not have a property for if the "App store" auto updates or is managed, so that only relies on the old method.

I'd prefer to use `SUOSUClientProtocol`, as this class would give us all of the needed software update data, but attempts to initialize it (while returning correct data) ends up returning an error:
```
SUPreferenceManager: Failed to set object of class: __NSTaggedDate for key: LastSuccessfulDate with error: Error Domain=SUPreferenceManagerErrorDomain Code=1 "(null)"
SUPreferenceManager: Failed to set object of class: __NSCFConstantString for key: LastRecommendedMajorOSBundleIdentifier with error: Error Domain=SUPreferenceManagerErrorDomain Code=1 "(null)"
```

I would like to return to this in the future to try utilizing `SUOSUClientProtocol`, but for now this PR at least gets us the patches we need for MacOS 15.


Micah-Kolide:
In regards to how the check that uses this table is implemented. Since `autoupdate_managed` no longer has a way of being validated, we'll probably need to update it.

James-Pickett (approved):
LGTM, worked on my box =)

```
osquery> select * from kolide_macos_software_update;
                  autoupdate_managed = 1
                  autoupdate_enabled = 1
                    download_managed = 1
                            download = 1
                 app_updates_managed = 1
                         app_updates = 0
                  os_updates_managed = 1
                          os_updates = 1
config_data_critical_updates_managed = 0
                 config_data_updates = 1
                    critical_updates = 1
     last_successful_check_timestamp = 1729002002
```

Micah-Kolide:
Wanted to add my computers results as well. Both are on MacOS 15.1:
```
                  autoupdate_managed = 0
                  autoupdate_enabled = 1
                    download_managed = 1
                            download = 1
                 app_updates_managed = 0
                         app_updates = 1
                  os_updates_managed = 1
                          os_updates = 1
config_data_critical_updates_managed = 1
                 config_data_updates = 1
                    critical_updates = 1
     last_successful_check_timestamp = 1729016849
```
```
                  autoupdate_managed = 0
                  autoupdate_enabled = 1
                    download_managed = 0
                            download = 1
                 app_updates_managed = 0
                         app_updates = 1
                  os_updates_managed = 0
                          os_updates = 0
config_data_critical_updates_managed = 0
                 config_data_updates = 1
                    critical_updates = 1
     last_successful_check_timestamp = 1729018450
```

PR #1893, opened 2024-10-16
Log osquery version

### LINKED ISSUE #1840: Add osquery_version to all logs
We are finding that it would be useful to include the osquery version in logs, the same way that we do for `launcher_version`.

This issue has two parts.

First, we should make the osquery version available in the knapsack. Similar to the `InModernStandby` flag, this flag should not be set by the command-line options -- it should only be set by launcher itself. We should have the [osquery runner](https://github.com/kolide/launcher/blob/main/pkg/osquery/runtime/runner.go) check the version and set it, after starting up the osquery instance. This will allow us to set the version correctly a) on launcher startup and b) on osquery autoupdate, when the osquery instance is restarted. There is an example for performing a version check [here](https://github.com/kolide/launcher/blob/main/cmd/launcher/launcher.go#L596) during launcher startup.

Second, we should include this version, when available, when shipping logs. The relevant logshipping code is [here](https://github.com/kolide/launcher/blob/main/pkg/log/logshipper/logshipper.go#L190).

cesarfda:
resolves #1840 

- Adding osquery version check to Osquery runner 
- Making version available to knapsack and adding it to logshipper


ee/agent/flags/keys/keys.go
```diff
@@ -57,6 +57,7 @@ const (
 	LocalDevelopmentPath            FlagKey = "localdev_path"
 	LauncherWatchdogEnabled         FlagKey = "launcher_watchdog_enabled" // note that this will only impact windows deployments for now
 	SystrayRestartEnabled           FlagKey = "systray_restart_enabled"
+	OsqueryVersion                  FlagKey = "osquery_version"
```
RebeccaMahany, line 60: I think I would make this `running_osquery_version` or `current_osquery_version` a) for consistency with your `SetCurrrentRunningOsqueryVersion` method naming, and b) to indicate that this flag reflects but does not control the osquery version. (It's fine to leave the log field as `osquery_version` in logshipper.go -- I think it only matters here, where it's a control server flag.)

What do other folks think?

pkg/osquery/runtime/runner.go
```diff
@@ -656,3 +660,62 @@ func newRunner(opts ...OsqueryInstanceOption) *Runner {
...
+		return
+	}
+
+	versionCtx, versionCancel := context.WithTimeout(ctx, 30*time.Second)
+	defer versionCancel()
+
+	startTime := time.Now().UnixMilli()
+
+	osqErr := osq.RunVersion(versionCtx)
+	executionTimeMs := time.Now().UnixMilli() - startTime
```
RebeccaMahany, line 682: I think you can remove `startTime` and `executionTimeMs` -- we cared about those in `launcher.go => runOsqueryVersionCheck` when we were trying to troubleshoot a particular bug, but we don't really care about logging the execution time here 🙂

pkg/osquery/runtime/runner.go
```diff
@@ -656,3 +660,62 @@ func newRunner(opts ...OsqueryInstanceOption) *Runner {
 		opts:     opts,
 	}
 }
+
+func (r *Runner) runOsqueryVersionCheckAddToKnapsack(ctx context.Context, osquerydPath string) {
```
RebeccaMahany, line 664: Could we add a little bit of documentation here? `// runOsqueryVersionCheckAddToKnapsack checks the version of osquery at the given osquerydPath and saves it to the knapsack, so that it is available to add when logshipping` or similar?

pkg/osquery/runtime/runner.go
```diff
@@ -656,3 +660,62 @@ func newRunner(opts ...OsqueryInstanceOption) *Runner {
...
+	var output bytes.Buffer
+
+	osq, err := runsimple.NewOsqueryProcess(osquerydPath, runsimple.WithStdout(&output))
+	if err != nil {
+		r.slogger.Log(ctx, slog.LevelError,
+			"unable to create process",
```
RebeccaMahany, line 670:
```suggestion
			"unable to create runsimple process to check osquery version",
```

(adds a little more context to the log -- probably worth making a similar update in cmd/launcher/launcher.go, I see we have the same log there too?)

RebeccaMahany (commented):
LGTM -- just a couple nitpicks and a question about naming!

pkg/osquery/runtime/runner.go
cesarfda: addressed

pkg/osquery/runtime/runner.go
cesarfda: added documentation

pkg/osquery/runtime/runner.go
cesarfda: removed the execution time from runner

ee/agent/flags/keys/keys.go
cesarfda: Renamed to "running_osquery_version"

ee/agent/flags/flag_controller.go
```diff
@@ -420,6 +420,14 @@ func (fc *FlagController) OsqueryTlsDistributedWriteEndpoint() string {
 	return fc.cmdLineOpts.OsqueryTlsDistributedWriteEndpoint
 }
 
+func (fc *FlagController) OsqueryVersion() string {
+	return NewStringFlagValue(WithDefaultString("")).get(fc.getControlServerValue(keys.RunningOsqueryVersion))
+}
+
+func (fc *FlagController) SetCurrrentRunningOsqueryVersion(osqueryversion string) error {
```
zackattack01, line 427: nit - `SetCurrentRunningOsqueryVersion` i think fixing this will fix up some of the failing test mocks

cesarfda: fixed typo and added the mocks where it was missing

ee/agent/flags/flag_controller.go
```diff
@@ -420,6 +420,14 @@ func (fc *FlagController) OsqueryTlsDistributedWriteEndpoint() string {
 	return fc.cmdLineOpts.OsqueryTlsDistributedWriteEndpoint
 }
 
+func (fc *FlagController) OsqueryVersion() string {
```
directionless, line 423: This should probably match the `RunningOsqueryVersion` name

directionless (commented):
mmmm. I kinda see why we need the duplication between the cmd main routine, and the runner function. But I wonder if the runner one can be merged with the check-osquery-before-exec test

ee/agent/flags/flag_controller.go
cesarfda: removed all the logic from runner and updated the version from the LatestOsquerydPath in knapsack

RebeccaMahany (dismissed):
🔥

RebeccaMahany (approved):

PR #1900, opened 2024-10-21
darwin notifier listen after show desktop

James-Pickett:
In pervious iterations, I mis-diagnosed a launcher crash. I believed it was systray crashing do to being blocked on main thread. It was actually trying to to register then notifier on a blocked thread. 

This PR breaks notifier into `Execute` and `Listen` functions. `Listen` only has logic for macos to register listener app with app bundle.

Corresponding systray PR: https://github.com/kolide/systray/pull/10

Will update go.mod to point at Kolide repo after systray merges.


directionless (approved):

ee/desktop/user/notify/notify_darwin.go
```diff
@@ -46,6 +42,12 @@ func (m *macNotifier) Interrupt(err error) {
 	m.interrupt <- struct{}{}
 }
 
+func (m *macNotifier) Listen() {
```
RebeccaMahany, line 45: Maybe a quick comment here about what you found/why this is needed?

RebeccaMahany (dismissed):
🔥

zackattack01 (dismissed):

ee/desktop/user/notify/notify_darwin.go
James-Pickett: done

RebeccaMahany (approved):

RebeccaMahany (approved):

PR #1901, opened 2024-10-21
Set osquery version in packaging context

RebeccaMahany:
This change will allow us to record the osquery version when building packages. (We already do this for launcher version, using the same method.)


directionless (approved):

directionless:
It feels a little weird to do this as an exec -- since we've moved to TUF, I feel like we should know what the version should be. But okay

PR #1902, opened 2024-10-22
added metadata to server checkup

### LINKED ISSUE #1752: We should capture launcher metadata in flares
eg: `/var/kolide-k2/k2device.kolide.com/metadata.json` and the various platform specific paths.

cesarfda:
Resolves #1752 

Capturing metadata for server check up.

Preview:
image


RebeccaMahany (approved):
LGTM!

Issue #1827, opened 2024-08-09
launcher should be able to run between 0 and n osquery instances


RebeccaMahany:
We want launcher to be able to run with between 0 and n osquery instances:

1. When launcher can't start osquery, we don't want to restart launcher altogether. This is because, in the case where the current osquery instance is old and not compatible with the current database, it can be helpful for launcher to run for a while with 0 osquery instances so that launcher can reach the autoupdate cycle and self-update its osquery version.
2. In the short term, we want to be able to split up where Kolide queries run versus where tenant queries run, so that if a tenant creates a query that causes issues with osquery, the Kolide queries will still run. (This amounts to running 2 osquery instances.)
3. In the longer term, we want to support multitenant installations. This will mean 1 osquery instance for Kolide queries, and 1 osquery instance per tenant. I propose we refer to these consistently as "Kolide osquery instance" and "tenant osquery instances".

There are a couple steps we need to take to be able to do this. This issue tracks the foundational steps necessary to run 0 or 1 osquery instances.

RebeccaMahany:
Multitenancy is not an active project right now; closing related issues to clean up our backlog.

PR #1905, opened 2024-10-23
ADR for osquery extension, runner, and instance refactor

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1827.

(ER diagram docs are [here](https://mermaid.js.org/syntax/entityRelationshipDiagram.html).)


zackattack01 (approved):
this is really well written and the approach makes sense to me! being somewhat unfamiliar with these pieces I am curious if there were any alternative approaches you had considered? mostly just for my own learning but maybe worth adding if so

docs/architecture/2024-10-23_osquery_refactor.md
```diff
@@ -0,0 +1,79 @@
...
+    OSQUERYINSTANCE ||--|| EXTENSIONCLIENT : runs
+    EXTENSIONSERVER ||--|| EXTENSION : forwards
+    OSQUERYRUNNER ||--o{ STATS: updates
+```
+
+## Consequences
+
+The osquery instance will now be responsible for managing extension health -- if the extension exits, the instance will need to take corrective action. (This is potentially a better outcome than before, since now extension shutdown can be remediated without a full launcher restart.)
+
+This decision does preserve a shared JSONRPC client to talk to Kolide SaaS. If we determine this is not optimal, it would be fairly easy to update to a model where we have one JSONRPC client per extension.
```
zackattack01, line 75: any known reasons to share vs break out per extension? or we're just not aware of a reason to duplicate the client

James-Pickett (approved):
I think this would definitely be an improvement. More logical separation of concerns. Much better to let the osqueryrunner -> instance handle the extension so we don't have restart launcher if it dies.

RebeccaMahany:
@zackattack01 -- the main alternative was to leave the extension where it is, and then either 1) share the extension among all the instances, or 2) create multiple extensions. I added more details in this commit: https://github.com/kolide/launcher/pull/1905/commits/915b78b8f25ac773796b802c2cf42c5eb9fafbf6

docs/architecture/2024-10-23_osquery_refactor.md
RebeccaMahany: > or we're just not aware of a reason to duplicate the client

Basically this! I am imagining it's the sort of thing that I'll have a stronger opinion on once we get further through implementation. But either way, not hard to change after the fact.

James-Pickett (approved):

zackattack01 (approved):
looks great, thank you!

docs/architecture/2024-10-23_osquery_refactor.md
```diff
@@ -0,0 +1,92 @@
...
+    EXTENSION }o--|| JSONRPCCLIENT : uses
+    OSQUERYINSTANCE ||--|{ EXTENSIONSERVER : runs
+    OSQUERYINSTANCE ||--|| EXTENSIONCLIENT : runs
+    EXTENSIONSERVER ||--|| EXTENSION : forwards
+    OSQUERYRUNNER ||--o{ STATS: updates
+```
+
+## Consequences
+
+The osquery instance will now be responsible for managing extension health -- if the extension exits, the instance will need to take corrective action. (This is potentially a better outcome than before, since now extension shutdown can be remediated without a full launcher restart.)
```
directionless, line 86: > The osquery instance will now be responsible for managing extension health

Do you mean the go actor sometimes called an osquery instance, or do you mean the osquery process will launch extension under it?

docs/architecture/2024-10-23_osquery_refactor.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+In our current implementation, the extension is created by `runLauncher` and run via our rungroup, and passed into the osquery runner for use when creating the instances. To run multiple osquery instances using our _current_ implementation, we have two choices:
+
+1. We could share one osquery extension with all osquery instances
+2. We could create multiple extensions inside `runLauncher` and pass them into the runner to distribute one per osquery instance
+
+Ultimately, I rejected both of these options as unsuitable.
+
+We don't want to share the osquery extension among all osquery instances -- we will be unable to identify which osquery instance is talking to the extension, which makes this option untenable. Additionally, we may run into socket contention issues.
```
directionless, line 36: Wow. This is a tough call. I find myself able to argue either side. 

In favor of sharing: Fewer go actors; less process/CPU overhead; less places for windows sleep to get in the way

In favor of split: More clear ownership and separation.

My hunch is socket contention may not be an issue -- I think with the go server side we can be more clever. Maybe? At least it's testable. 

I think it's _also_ worth thinking about the jsonrpc extension and the tables extensions a bit differently. Even if we decide it doesn't matter.

directionless: The jsonrpc extension has some additional nuance. Do we want each osquery instance to talk to our SaaS via very distinct go processes, or do we want a single go processes that can redirect and distribute things. I highly suspect this will depends on how we handle auth flow.

I imagine we'll end up in a world where there's a set of root keys that identify launcher. And then we end up with some kind of registration data for each tenant. In that world, a shared extensions means a communication pattern where launcher root fetches everything, perhaps with each registration. In contrast, N extensions means each instance is using it's registration data to fetch.

It feels somewhat hard to predict. I think as long as we're prepared to rework it if we think we got it wrong, it doesn't matter which we try.

directionless (commented):
love the diagrams

docs/architecture/2024-10-23_osquery_refactor.md
RebeccaMahany: > In favor of sharing: Fewer go actors; less process/CPU overhead; less places for windows sleep to get in the way

I've updated to add this to the `Consequences` section as well.

zackattack01 (approved):

PR #1909, opened 2024-10-23
Remove and consolidate osquery instance options

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1827 -- getting some tidying out of the way before making more substantial changes.

This PR updates our osquery instance options:

1. Remove unused options
2. Remove options that are redundant with knapsack flags, opting to pull these values from the knapsack instead

While I was validating that the root directory will always be set by the time that we make it to the osquery runner, I noticed that in `runLauncher` if the root directory wasn't set in the launcher options and we fell back to making a temp directory, we wouldn't save that temp directory in the knapsack flags. I updated to save it in the launcher options, so that it would later be available via the knapsack. I imagine we don't run into this issue much/at all, but wanted to address it just in case.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -140,6 +140,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 			"using default system root directory",
 			"path", rootDirectory,
 		)
+		// Make sure we have record of this new root directory in the opts, so it will be set
+		// correctly in the knapsack later.
+		opts.RootDirectory = rootDirectory
 	}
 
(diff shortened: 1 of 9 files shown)


James-Pickett (approved):
LGTM! Always love negative lines of code. Do any of these removals make options in [launcher options](https://github.com/kolide/launcher/blob/main/pkg/launcher/options.go)?

directionless (commented):
I didn't test this, but I imagine it's fine. 

**However** I think we should hold slightly on merging -- I think we're going to want to be in a place to cut a release for Zack's flare stuff and maybe the systray tweaks.

James-Pickett (approved):

PR #1914, opened 2024-10-24
Added restrictions for datatypes for systemprofiler

### LINKED ISSUE #1876: `kolide_system_profiler` should not support an all option
Chatting with some coworkers, I learned that `SELECT * FROM kolide_system_profiler where datatype like "%" ` works. But it returns so much data it crashes `Terminal.app` and probably has a bunch of gnarly load. 

I think we should disable it. We should require a reasonable argument for `datatype`

https://github.com/kolide/launcher/blob/1ec19cddf4b2a6fa44a78ef1fccb5531c08a2677/ee/tables/systemprofiler/systemprofiler.go#L111

cesarfda:
Resolves #1876 

- Disallow passing % as datatype
- Do not allow more than 3 datatypes in a query (currently not supported, launcher receives one query for each datatype due to osquery and sqlite limitations)

### DIFF
--- a/ee/tables/systemprofiler/systemprofiler.go
+++ b/ee/tables/systemprofiler/systemprofiler.go
@@ -102,16 +102,20 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 
 	datatypeQ, ok := queryContext.Constraints["datatype"]
 	if !ok || len(datatypeQ.Constraints) == 0 {
-		return results, fmt.Errorf("The %s table requires that you specify a constraint for datatype", t.tableName)
+		return results, fmt.Errorf("the %s table requires that you specify a constraint for datatype", t.tableName)
+	}
+
+	// Check for maximum number of datatypes
+	if len(datatypeQ.Constraints) > 3 {
+		return results, fmt.Errorf("maximum of 3 datatypes allowed per query, got %d", len(datatypeQ.Constraints))
 	}
 
 	for _, datatypeConstraint := range datatypeQ.Constraints {
 		dt := datatypeConstraint.Expression
 
-		// If the constraint is the magic "%", it's eqivlent to an `all` style
+		// Block the % wildcard
 		if dt == "%" {
-			requestedDatatypes = []string{}
-			break
+			return results, fmt.Errorf("wildcard %% is not allowed as a datatype constraint")
 		}
 
 		requestedDatatypes = append(requestedDatatypes, dt)


ee/tables/systemprofiler/systemprofiler.go
```diff
@@ -102,16 +102,20 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
 
 	for _, datatypeConstraint := range datatypeQ.Constraints {
 		dt := datatypeConstraint.Expression
 
-		// If the constraint is the magic "%", it's eqivlent to an `all` style
+		// Block the % wildcard
 		if dt == "%" {
-			requestedDatatypes = []string{}
-			break
+			return results, fmt.Errorf("wildcard %% is not allowed as a datatype constraint")
```
directionless, line 118: small nit: errors.New and only a single `%`

Larger question: We often _don't_ return errors from tables. If we do, it can kill the whole query. Instead we often log the error, and `continue`. I'm a little agnostic here, but it's probably safer to `continue`

ee/tables/systemprofiler/systemprofiler.go
```diff
@@ -102,16 +102,20 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 
 	datatypeQ, ok := queryContext.Constraints["datatype"]
 	if !ok || len(datatypeQ.Constraints) == 0 {
-		return results, fmt.Errorf("The %s table requires that you specify a constraint for datatype", t.tableName)
+		return results, fmt.Errorf("the %s table requires that you specify a constraint for datatype", t.tableName)
+	}
+
+	// Check for maximum number of datatypes
+	if len(datatypeQ.Constraints) > 3 {
```
directionless, line 109: Might be cleaner to replace all the constraint checks with `tablehelpers.GetConstraints`

James-Pickett, line 109: would be nice to make the `3` a constant with a comment explaining why 3 is the max constraints

directionless: Not sure it needs a constant, but expanding the comment makes sense...

cesarfda: added use of `tablehelpers.GetConstraints`

ee/tables/systemprofiler/systemprofiler.go
cesarfda: I moved the allowed characters to be handled during the GetConstraints and kept the returning error only in the scenario where we have more datatypes than the limit

ee/tables/systemprofiler/systemprofiler.go
cesarfda: made it into const

RebeccaMahany (approved):
Nice!

James-Pickett (approved):
nice

PR #1918, opened 2024-10-25
Get launcher.flags from correct directory in checkup

RebeccaMahany:
In flare/doctor, we don't fetch the correct flag file if the identifier is not the default kolide-k2 -- this PR corrects that issue.

### DIFF
--- a/ee/debug/checkups/launcher_flags.go
+++ b/ee/debug/checkups/launcher_flags.go
@@ -76,7 +76,7 @@ func (lf *launcherFlags) Data() any {
 }
 
 func (lf *launcherFlags) flagsFilePath() string {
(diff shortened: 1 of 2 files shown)


ee/debug/checkups/launcher_flags.go
```diff
@@ -74,10 +76,14 @@ func (lf *launcherFlags) Data() any {
 }
 
 func (lf *launcherFlags) flagsFilePath() string {
+	identifier := "kolide-k2"
```
zackattack01, line 79: there is a `launcher.DefaultLauncherIdentifier` now if you want to use here

zackattack01 (dismissed):

zackattack01 (approved):

PR #1919, opened 2024-10-28
Make jsonrpc the default for transport

RebeccaMahany:
The expected value for the `transport` flag has de facto been `jsonrpc` for a significant amount of time now, but that hasn't been reflected in the default value in the options. This PR updates the default.


directionless (approved):
Sure!

PR #1923, opened 2024-10-29
added new store for uptime history and uptime tracking with knapsack

### LINKED ISSUE #1702: The `kolide_launcher_info` should expose runtime.
I often want to see what the launcher runtime is. We should add that to the `kolide_launcher_info` table

cesarfda:
Resolves #1702 

- Added a start time entry at launcher that is available at the knapsack
- Added UpTimeHistoryStore to store start_time
- Modified kolide_launcher_info table to take in the UpTimeHistoryStore, calculate the delta and save the value in a new Column called uptime.

Preview:
image

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -198,6 +198,11 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	flagController := flags.NewFlagController(slogger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db, multiSlogger, systemMultiSlogger)
 
+	// start counting uptime
+	processStartTime := time.Now()
+
+	k.UpTimeHistoryStore().Set([]byte("process_start_time"), []byte(processStartTime.Format(time.RFC3339)))
+
 	go runOsqueryVersionCheckAndAddToKnapsack(ctx, slogger, k, k.LatestOsquerydPath(ctx))
 	go timemachine.AddExclusions(ctx, k)
 
--- a/ee/agent/storage/bbolt/stores_bbolt.go
+++ b/ee/agent/storage/bbolt/stores_bbolt.go
@@ -33,6 +33,7 @@ func MakeStores(ctx context.Context, slogger *slog.Logger, db *bbolt.DB) (map[st
 		storage.ServerProvidedDataStore,
 		storage.TokenStore,
 		storage.ControlServerActionsStore,
+		storage.UpTimeHistoryStore,
 	}
 
 	for _, storeName := range storeNames {
--- a/ee/agent/storage/stores.go
+++ b/ee/agent/storage/stores.go
@@ -18,6 +18,7 @@ const (
 	ServerProvidedDataStore     Store = "server_provided_data"     // The store used for pushing values from server-backed tables.
 	TokenStore                  Store = "token_store"              // The store used for holding bearer auth tokens, e.g. the ones used to authenticate with the observability ingest server.
 	ControlServerActionsStore   Store = "action_store"             // The store used for storing actions sent by control server.
+	UpTimeHistoryStore          Store = "uptime_history"           // The store used for storing uptime history.
 )
 
 func (storeType Store) String() string {
--- a/pkg/osquery/table/launcher_info.go
+++ b/pkg/osquery/table/launcher_info.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/kolide/kit/version"
 	"github.com/kolide/launcher/ee/agent"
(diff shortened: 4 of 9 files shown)


cmd/launcher/launcher.go
```diff
@@ -198,6 +198,11 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	flagController := flags.NewFlagController(slogger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db, multiSlogger, systemMultiSlogger)
 
+	// start counting uptime
+	processStartTime := time.Now()
```
James-Pickett, line 202: We probably want this to be UTC

ee/agent/storage/stores.go
```diff
@@ -18,6 +18,7 @@ const (
 	ServerProvidedDataStore     Store = "server_provided_data"     // The store used for pushing values from server-backed tables.
 	TokenStore                  Store = "token_store"              // The store used for holding bearer auth tokens, e.g. the ones used to authenticate with the observability ingest server.
 	ControlServerActionsStore   Store = "action_store"             // The store used for storing actions sent by control server.
+	UpTimeHistoryStore          Store = "uptime_history"           // The store used for storing uptime history.
```
RebeccaMahany, line 21: Maybe we'd want something like `LauncherHistoryStore Store = "launcher_history"` -- I think that might be a little more extensible for the future, and would match `osquery_instance_history`. I don't feel strongly about it, though -- what do others think?

cmd/launcher/launcher.go
```diff
@@ -198,6 +198,11 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
...
+
+	k.UpTimeHistoryStore().Set([]byte("process_start_time"), []byte(processStartTime.Format(time.RFC3339)))
```
zackattack01, line 204: `HistoryStore` makes me think this would hold a series of process start times - but I feel like we would want to do that eventually anyway so probably not worth changing the name. might be nice to add a comment somewhere (maybe here or near knapsack definition) saying that we only currently track the running process start time, but would like to push these into a historical list in the future

RebeccaMahany (commented):
One question about naming -- LGTM!

pkg/osquery/table/launcher_info.go
```diff
@@ -60,13 +62,23 @@ func generateLauncherInfoTable(store types.GetterSetter) table.GenerateFunc {
...
 			// No logger here, so we can't easily log. Move on with blank values
 			publicKey = ""
 			fingerprint = ""
 		}
 
+		uptimeBytes, _ := upTimeHistoryStore.Get([]byte("process_start_time"))
+		uptime := ""
+
+		if uptimeBytes != nil {
+			startTime, _ := time.Parse(time.RFC3339, string(uptimeBytes))
```
zackattack01, line 76: should probably check the error here before processing `startTime`

pkg/osquery/table/launcher_info.go
```diff
@@ -60,13 +62,23 @@ func generateLauncherInfoTable(store types.GetterSetter) table.GenerateFunc {
...
 
-		publicKey, fingerprint, err := osquery.PublicRSAKeyFromDB(store)
+		publicKey, fingerprint, err := osquery.PublicRSAKeyFromDB(configStore)
 		if err != nil {
 			// No logger here, so we can't easily log. Move on with blank values
 			publicKey = ""
 			fingerprint = ""
 		}
 
+		uptimeBytes, _ := upTimeHistoryStore.Get([]byte("process_start_time"))
```
zackattack01, line 72: should probably check the error here before processing `uptimeBytes`

zackattack01 (commented):
looks great! just a few small comments

ee/agent/storage/stores.go
zackattack01: I like that more too, it makes sense to me to report as just `uptime` for the table column but internally here it is a little more confusing what this refers to

pkg/osquery/table/launcher_info.go
cesarfda: Done. I avoided returning the err because I don't want uptime blocking the rest of the info on the table, so I set a default value to use if the uptime err are not nil.

pkg/osquery/table/launcher_info.go
cesarfda: Done

ee/agent/storage/stores.go
cesarfda: That makes sense, I updated the name to LauncherHistoryStore

cmd/launcher/launcher.go
cesarfda: Updated to LauncherHistoryStore

cmd/launcher/launcher.go
cesarfda: Done

James-Pickett (approved):
LGTM!

zackattack01 (approved):
nice!

ee/agent/storage/bbolt/stores_bbolt.go
```diff
@@ -33,6 +33,7 @@ func MakeStores(ctx context.Context, slogger *slog.Logger, db *bbolt.DB) (map[st
 		storage.ServerProvidedDataStore,
 		storage.TokenStore,
 		storage.ControlServerActionsStore,
+		storage.LauncherHistoryStore,
```
directionless, line 36: Wow. Looking at the stores here, I'm not sure there was an obvious one. But creating a new one to hold a single value feels big. Done now though

PR #1927, opened 2024-10-30
Move osquery extension management into osquery instance

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1827

Per [ADR](https://github.com/kolide/launcher/blob/main/docs/architecture/2024-10-23_osquery_refactor.md), moves extension management into osquery instance.

This adds two new goroutines to the osquery instance's errgroup. To simplify a bit, this PR has also added the following:

* Standardize how we add goroutines to the errgroup, ensuring they are logged and that shutdown goroutines are always created in the same way
* Remove `kolide` extension, adding all tables to `kolide_grpc` instead


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -549,6 +488,116 @@ func (i *OsqueryInstance) launch() error {
...
+			"goroutine_name", goroutineName,
+		)
+
+		return goroutine()
+	})
+}
+
+// addShutdownGoroutineToErrgroup adds the given goroutine to the errgroup, ensuring that we log its start and exit.
+// The goroutine will not execute until the instance has received a signal to exit.
+func (i *OsqueryInstance) addShutdownGoroutineToErrgroup(ctx context.Context, goroutineName string, goroutine func() error) {
```
zackattack01, line 582: these are very nice 🔥

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -549,6 +488,116 @@ func (i *OsqueryInstance) launch() error {
...
+	gowrapper.Go(ctx, i.slogger, func() {
+		_, nodeInvalid, err := i.saasExtension.Enroll(ctx)
+		if nodeInvalid || err != nil {
+			i.slogger.Log(ctx, slog.LevelWarn,
+				"could not perform initial attempt at enrollment, will retry later",
+				"node_invalid", nodeInvalid,
+				"err", err,
+			)
+		}
+	}, func(r any) {})
```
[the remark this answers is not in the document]
zackattack01: oh ignore i forgot that happens automatically

James-Pickett (approved):

zackattack01 (approved):

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -375,19 +362,6 @@ func (i *OsqueryInstance) launch() error {
 		return i.doneCtx.Err()
 	})
 
-	// Here be dragons
-	//
-	// There are two thorny issues. First, we "invert" control of
-	// the osquery process. We don't really know when osquery will
-	// be running, so we need a bunch of retries on these connections
```
directionless, line 382: This part _might_ still be true?

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -375,19 +362,6 @@ func (i *OsqueryInstance) launch() error {
...
-	//
-	// Second, because launcher supplements the enroll
```
directionless, line 384: This part is slain.

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: This part didn't seem relevant to keep here anymore because we don't have the retries below this method -- the code was refactored (previously) to move those retries inside the `StartOsqueryClient` and `StartOsqueryExtensionManagerServer` methods.

directionless: Fair... the inverted control part is true, but maybe better elsewhere.

RebeccaMahany: I'll add it back as a comment in next PR!

PR #1929, opened 2024-10-31
Moved run ID management to knapsack

### LINKED ISSUE #1706: Make launcher run ID available in knapsack
Launcher has a "run ID" that is currently added to all log messages. This run ID is only available inside the multislogger itself, and is created during new multislogger instantiation. We'd like to be able to access this run ID elsewhere, e.g. to report the current run ID up to k2 during startup, so we want to move it into launcher's [knapsack](https://github.com/kolide/launcher/tree/main/ee/agent). The instantiation of run ID inside [multislogger.New](https://github.com/kolide/launcher/blob/main/pkg/log/multislogger/multislogger.go) also has the side effect that the same launcher run has multiple run IDs across different multisloggers -- for example, Windows event logs and debug.json logs will have different run IDs for the same run. Ideally, each time launcher runs, it should have a singular, unique run ID that it can access from the knapsack.

Notes for implementation:

* The run ID should be available in the knapsack so that launcher can e.g. report it up to k2 in the future
* The run ID should change if launcher reloads itself after an autoupdate (see [IsLauncherReloadNeededErr, hereabouts](https://github.com/kolide/launcher/blob/main/cmd/launcher/main.go#L148)), which means every call to `runLauncher` should get a new run ID
* The run ID should be applied to both the regular and system multislogger -- i.e., the run ID should be attached to all logs consistently, whether they are in the regular logs (debug.json) or the system logs (the Windows event logs on Windows, `/var/log/kolide-k2/launcher-stderr.log` on macOS, or journalctl logs on Linux)

cesarfda:
This pull request moves the run ID system to improve logging and tracking across the application. The changes include generating a unique run ID, setting it in the knapsack, and applying it to the logger and slogger for consistent tracking.

Resolves #1706 

### Run ID Implementation:

* [`cmd/launcher/launcher.go`](diffhunk://#diff-dd532c327d56a5d3623a61b7983e174e88bf93ba87f785ddfe9811a8029a9a63R197-R205): Added code to generate a new run ID and set it in the knapsack, logger, and slogger.
* [`ee/agent/knapsack/knapsack.go`](diffhunk://#diff-1fef7e482dbae76c26b95e73faa3353d09c1e04660f7156238bd991dd1ea004fR19-R21): Introduced a package-level `runID` variable and added `GetRunID` method to manage the run ID within the knapsack. [[1]](diffhunk://#diff-1fef7e482dbae76c26b95e73faa3353d09c1e04660f7156238bd991dd1ea004fR19-R21) [[2]](diffhunk://#diff-1fef7e482dbae76c26b95e73faa3353d09c1e04660f7156238bd991dd1ea004fR65-R76)

### Logger Updates:

* [`pkg/log/multislogger/multislogger.go`](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaL8): Removed the `launcherRunId` variable from the `MultiSlogger` struct and its initialization, and updated the `AddHandler` method to no longer include the `launcher_run_id`. [[1]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaL8) [[2]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaL38) [[3]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaL47) [[4]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaL72-R69)

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -195,9 +195,7 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	k := knapsack.New(stores, flagController, db, multiSlogger, systemMultiSlogger)
 
 	// Generate a new run ID
-	newRunID := ulid.New()
-	// Set the run ID in knapsack
-	k.SetRunID(newRunID)
+	newRunID := k.GetRunID()
 
 	// Apply the run ID to both logger and slogger
 	logger = log.With(logger, "run_id", newRunID)
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -194,6 +194,15 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	flagController := flags.NewFlagController(slogger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db, multiSlogger, systemMultiSlogger)
 
+	// Generate a new run ID
+	newRunID := ulid.New()
+	// Set the run ID in knapsack
+	k.SetRunID(newRunID)
+
+	// Apply the run ID to both logger and slogger
+	logger = log.With(logger, "run_id", newRunID)
+	slogger = slogger.With("run_id", newRunID)
+
 	// start counting uptime
 	processStartTime := time.Now().UTC()
 
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -16,6 +16,9 @@ import (
 	"go.etcd.io/bbolt"
 )
 
+// Package-level runID variable
+var runID string
+
 // type alias Flags, so that we can embed it inside knapsack, as `flags` and not `Flags`
 type flags types.Flags
 
@@ -57,6 +60,18 @@ func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB
 	}
 
 	return k
+}
+
+// SetRunID sets the run ID in the knapsack
+func (k *knapsack) SetRunID(id string) {
+	runID = id
+	k.slogger.With("run_id", id)
+	k.systemSlogger.With("run_id", id)
+}
+
+// GetRunID retrieves the run ID from the knapsack
+func (k *knapsack) GetRunID() string {
+	return runID
 }
 
 // Logging interface methods
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -9,6 +9,7 @@ import (
 
 	"log/slog"
 
+	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/agent/storage"
 	"github.com/kolide/launcher/ee/agent/types"
 	"github.com/kolide/launcher/ee/tuf"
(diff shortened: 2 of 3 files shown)


ee/agent/knapsack/knapsack.go
```diff
@@ -59,6 +62,18 @@ func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB
 	return k
 }
 
+// SetRunID sets the run ID in the knapsack
+func (k *knapsack) SetRunID(id string) {
+	runID = id
+	k.slogger.With("run_id", id)
+	k.systemSlogger.With("run_id", id)
```
RebeccaMahany, line 69: Do we need e.g. --

```golang
k.slogger = k.slogger.With("run_id", id)
k.systemSlogger = k.systemSlogger.With("run_id", id)
```

? I think `.With` creates and then modifies a clone, so it wouldn't update k.slogger.

cesarfda: Yeah, I did that in launcher but missed in Knapsack. I have pushed out another commit to fix this with the assignment.

ee/agent/knapsack/knapsack.go
```diff
@@ -59,6 +62,18 @@ func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB
 	return k
 }
 
+// SetRunID sets the run ID in the knapsack
+func (k *knapsack) SetRunID(id string) {
```
directionless, line 66: Do you need this method?

I'd probably do something like:
```
func (k *knapsack) GetRunID() string {
        if runID == "" {
            runID = ulid.New()
        }
	return runID
}
```

cesarfda: I think that wouldn't cover the scenario where we want every time runLauncher is called to have a newID, I assume we could run into a situation where the ID is persisted if runLauncher is called after an update. Let me know if that's not the case and I can change this

directionless: Chatted on slack -- I think this is worth testing. I'm not sure if the restart/exec would have access to the same memory

directionless: I think if we did need to create a specific create method, my inclination would be to make `k.NewRunID()` and call it on startup. I think passing in the string feels wrong.

cesarfda: I was able to test the scenario and after the update we do get a new ID when we generate the ID in the GET method. I refactored the code to not have a set.

RebeccaMahany (approved):

cmd/launcher/launcher.go
```diff
@@ -194,6 +194,13 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	flagController := flags.NewFlagController(slogger, stores[storage.AgentFlagsStore], fcOpts...)
 	k := knapsack.New(stores, flagController, db, multiSlogger, systemMultiSlogger)
 
+	// Generate a new run ID
+	newRunID := k.GetRunID()
```
directionless, line 198: I know this has merged, and it's fine, but I'd probably skip the variable here, and put `k.GetRunID()` into the two `With` functions

ee/agent/knapsack/knapsack.go
```diff
@@ -59,6 +63,16 @@ func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB
 	return k
 }
 
+// NewRunID sets the run ID in the knapsack
+func (k *knapsack) GetRunID() string {
+	if runID == "" {
+		runID = ulid.New()
+		k.slogger.Logger = k.slogger.Logger.With("run_id", runID)
+		k.systemSlogger.Logger = k.systemSlogger.Logger.With("run_id", runID)
```
directionless, line 71: Hrm. This here, and the part in `runLauncher` feels off somehow.

PR #1934, opened 2024-11-04
removed redundant slogger statement

cesarfda:
This pull request includes a minor update to the `cmd/launcher/launcher.go` file to clarify the handling of the `run_id` for logging purposes.

* [`cmd/launcher/launcher.go`](diffhunk://#diff-dd532c327d56a5d3623a61b7983e174e88bf93ba87f785ddfe9811a8029a9a63L200-L202): Updated comments to specify that the `run_id` is applied to the logger for use in the logshipper, and clarified that the `slogger` `run_id` is set in the knapsack `GetRunID()`.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -197,9 +197,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	// Generate a new run ID
 	newRunID := k.GetRunID()
 
-	// Apply the run ID to both logger and slogger
+	// Apply the run ID to logger which is later used in logshipper
+	// slogger run_id is set in the knapsack GetRunID()
 	logger = log.With(logger, "run_id", newRunID)
-	slogger = slogger.With("run_id", newRunID)
 
 	// start counting uptime
 	processStartTime := time.Now().UTC()


cmd/launcher/launcher.go
```diff
@@ -197,9 +197,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 	// Generate a new run ID
 	newRunID := k.GetRunID()
 
-	// Apply the run ID to both logger and slogger
+	// Apply the run ID to logger which is later used in logshipper
+	// slogger run_id is set in the knapsack GetRunID()
```
RebeccaMahany, line 201: I am not certain this is actually true -- I ran into a weird bug with this previously! (Bug fix: https://github.com/kolide/launcher/pull/1869.) I believe that updating the slogger in the knapsack does not actually update `slogger` here, so you do actually want to keep `slogger = slogger.With("run_id", newRunID)` below.

directionless, line 201: Approved, but I wonder if it's easier to just put `k.GetRunID()` here, and not create the `newRunID`

directionless (approved):

cmd/launcher/launcher.go
RebeccaMahany: Specific context in this comment: https://github.com/kolide/launcher/pull/1869#issuecomment-2356899896

cesarfda: That's fair, I had the assumption that the rest of the slogger references would come from knapsack but I wasn't able to check it through the whole code. I'll keep this in and close the PR

cesarfda: We still need the run_id for the other logger that is used in the logshipper since that doesn't use knapsack. Since Becca's comment that the setup should still be in launcher I'll leave things as is and close this PR.

Issue #1936, opened 2024-11-04
Refactor osquery runner to manage running multiple osquery instances (one Kolide, and a variable number of tenant osquery instances)


RebeccaMahany:
Preliminary refactor work described in ADR: https://github.com/kolide/launcher/blob/main/docs/architecture/2024-10-23_osquery_refactor.md.

Beyond ADR, this issue covers making it possible to run more than one osquery instance via the runner.

PR #1941, opened 2024-11-05
Support running multiple osquery instances

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1936

Updates the osquery runner to run multiple osquery instances. For now, we only run one -- the "default".


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -679,7 +679,7 @@ func calculateOsqueryPaths(rootDirectory string, runId string, opts osqueryOptio
 	// See: https://github.com/kolide/launcher/issues/1599
 	osqueryFilePaths := &osqueryFilePaths{
 		pidfilePath:           filepath.Join(rootDirectory, fmt.Sprintf("osquery-%s.pid", runId)),
-		databasePath:          filepath.Join(rootDirectory, "osquery.db"),
+		databasePath:          filepath.Join(rootDirectory, fmt.Sprintf("osquery-%s.db", runId)),
```
RebeccaMahany, line 682: TODO -- will move this to a separate PR, there is more work there that needs to be done

zackattack01 (approved):

PR #1942, opened 2024-11-06
Unique osquery database per instance

RebeccaMahany:
Similar change as https://github.com/kolide/launcher/pull/1935 -- we cannot share the database among multiple instances. Since these databases persist between launcher runs, I am introducing a `registration id` rather than using the ephemeral `run id`.

I also noticed that we don't clean up the osquery socket file on instance shutdown. Since the socket file path changes between runs now, this means we accumulate old socket files. I added a new goroutine to clean up the socket file on shutdown, similar to what we do for the pidfile.

Relates to https://github.com/kolide/launcher/issues/1936.


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -188,13 +189,14 @@ type osqueryOptions struct {
...
+func newInstance(registrationId string, knapsack types.Knapsack, serviceClient service.KolideService, opts ...OsqueryInstanceOption) *OsqueryInstance {
 	runId := ulid.New()
 	i := &OsqueryInstance{
-		knapsack:      knapsack,
-		slogger:       knapsack.Slogger().With("component", "osquery_instance", "instance_run_id", runId),
-		serviceClient: serviceClient,
-		runId:         runId,
+		registrationId: registrationId,
+		knapsack:       knapsack,
+		slogger:        knapsack.Slogger().With("component", "osquery_instance", "instance_run_id", runId),
```
zackattack01, line 197: do you think it's worth adding the `registrationId` to slogger here as well?

zackattack01 (approved):

James-Pickett (approved):

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: Sure, let me do that in the next PR (https://github.com/kolide/launcher/pull/1941)

RebeccaMahany: Done: https://github.com/kolide/launcher/pull/1941/commits/b3e2ee91c443c3c5623418b50df615b6e5a21f67

PR #1944, opened 2024-11-07
ActionQueue retry on failure

### LINKED ISSUE #1708: actionqueue should perform retries for failed actions without being prompted by control data changes
The [actionqueue](https://github.com/kolide/launcher/tree/main/ee/control/actionqueue) handles changes to the `actions` subsystem. When it successfully performs an action, it marks the action as performed and stores that information. Storing the performed actions allows the control server to safely send down the same action (with the same ID) an unlimited number of times -- the actionqueue will only perform that action if it wasn't able to perform it successfully previously, which implements retries for actions.

We think that this retry mechanism is not sufficient (especially for some of our more important actions like remote uninstall), since it relies on the control server changing the data in the `actions` subsystem. (E.g., if a remote uninstall action failed to process once, launcher would not retry until it received a new remote uninstall action, or a new notification, etc.)

We would like to update launcher's behavior so that the actionqueue will retry received-but-failed actions proactively, _without_ requiring a prompt from the control server.

Requirements for implementation:

* If any action fails, the actionqueue should retry it at a reasonable interval (once an hour?)
* The actionqueue should never retry actions that have already been completed successfully
* The actionqueue should never retry actions that are expired

cesarfda:
Resolves #1708 

We are implementing a retry pattern by returning an error from func (aq *ActionQueue) Update(data io.Reader) error if any action fails to process. Then the updated data for that subsystem shouldn't be stored, so Update would be called again in 1 minute by the control system, effectively handling the retry for us.

This pull request introduces error handling improvements and additional test cases for the `ActionQueue` and `ControlService` components. The main changes include adding error handling for action processing failures, updating test cases to handle these errors, and adding new test cases for retry logic in the `ControlService`.

Error handling improvements:

* [`ee/control/actionqueue/actionqueue.go`](diffhunk://#diff-cca5130d6e3079d2547d3a07ba5e38697d8c53b479fae5a3331e493f16bbdfc6R103-R104): Introduced `processError` to capture errors during action processing and return it at the end of the `Update` method. [[1]](diffhunk://#diff-cca5130d6e3079d2547d3a07ba5e38697d8c53b479fae5a3331e493f16bbdfc6R103-R104) [[2]](diffhunk://#diff-cca5130d6e3079d2547d3a07ba5e38697d8c53b479fae5a3331e493f16bbdfc6R133) [[3]](diffhunk://#diff-cca5130d6e3079d2547d3a07ba5e38697d8c53b479fae5a3331e493f16bbdfc6L139-R142)

Updates to existing tests:

* [`ee/control/actionqueue/actionqueue_test.go`](diffhunk://#diff-fdb0ac79b39f0f53a044a97ac956da7f0aacd9afc8bfc20d9882e25d8f2daa76L184): Modified the `TestActionQueue_HandlesDuplicatesWhenFirstActionCouldNotBeSent` test to check for errors on the first attempt and ensure success on the second attempt. [[1]](diffhunk://#diff-fdb0ac79b39f0f53a044a97ac956da7f0aacd9afc8bfc20d9882e25d8f2daa76L184) [[2]](diffhunk://#diff-fdb0ac79b39f0f53a044a97ac956da7f0aacd9afc8bfc20d9882e25d8f2daa76L197-R202)

New test cases:

* [`ee/control/control_test.go`](diffhunk://#diff-e617fc1085458f7b44b67e0310a766ecd3a50347f35400ecb88340499737ddc2R202-R286): Added `TestControlServiceUpdateErr` to verify that the `ControlService` handles update errors correctly and records the hash despite the error.
* [`ee/control/control_test.go`](diffhunk://#diff-e617fc1085458f7b44b67e0310a766ecd3a50347f35400ecb88340499737ddc2R202-R286): Added `TestControlServiceRetryAfterUpdateErr` to test the retry logic after an update error, ensuring that subsequent updates succeed and the final hash is recorded.

Mock updates:

* [`ee/control/control_test.go`](diffhunk://#diff-e617fc1085458f7b44b67e0310a766ecd3a50347f35400ecb88340499737ddc2R27): Modified the `mockConsumer` to support returning an error in the `Update` method. [[1]](diffhunk://#diff-e617fc1085458f7b44b67e0310a766ecd3a50347f35400ecb88340499737ddc2R27) [[2]](diffhunk://#diff-e617fc1085458f7b44b67e0310a766ecd3a50347f35400ecb88340499737ddc2L38-R39)

### DIFF
--- a/ee/control/actionqueue/actionqueue.go
+++ b/ee/control/actionqueue/actionqueue.go
@@ -130,7 +130,7 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 				"failed to do action with action, not marking action complete",
 				"err", err,
 			)
-			processError = fmt.Errorf("actor.Do failed: %w", err)
+			processError = fmt.Errorf("actor.Do, action type: %s, failed: %w", action.type, err)
 			continue
 		}
 
--- a/ee/control/actionqueue/actionqueue.go
+++ b/ee/control/actionqueue/actionqueue.go
@@ -100,6 +100,8 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 		return fmt.Errorf("failed to decode actions data: %w", err)
 	}
 
+	var processError error = nil
+
 	for _, rawAction := range rawActionsToProcess {
 		var action action
 		if err := json.Unmarshal(rawAction, &action); err != nil {
@@ -128,6 +130,7 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 				"failed to do action with action, not marking action complete",
 				"err", err,
 			)
+			processError = fmt.Errorf("actor.Do failed: %w", err)
 			continue
 		}
 
@@ -136,7 +139,7 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 		aq.storeActionRecord(action)
 	}
 
-	return nil
+	return processError
 }
 
 func (aq *ActionQueue) RegisterActor(actorType string, actorToRegister actor) {
--- a/ee/control/actionqueue/actionqueue.go
+++ b/ee/control/actionqueue/actionqueue.go
@@ -130,7 +130,7 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 				"failed to do action with action, not marking action complete",
 				"err", err,
 			)
-			processError = fmt.Errorf("actor.Do, action type: %s, failed: %w", action.type, err)
+			processError = fmt.Errorf("actor.Do, action type: %s, failed: %w", action.Type, err)
 			continue
 		}
 
--- a/ee/control/control_test.go
+++ b/ee/control/control_test.go
@@ -237,6 +237,53 @@ func TestControlServiceUpdateErr(t *testing.T) {
 	assert.Equal(t, "abc123", string(val))
 }
 
+func TestControlServiceRetryAfterUpdateErr(t *testing.T) {
+	t.Parallel()
+
+	errConsumer := &mockConsumer{
+		updateErr: errors.New("simulated update failure"),
+	}
+
+	mockKnapsack := typesMocks.NewKnapsack(t)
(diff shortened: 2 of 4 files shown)


ee/control/control_test.go
```diff
@@ -198,6 +199,91 @@ func TestControlServiceUpdate(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Trigger fetch which should cause update error
+	err = cs.Fetch()
+	require.NoError(t, err)
+
+	// Verify error consumer was called
+	assert.Equal(t, 1, errConsumer.updates)
+
+	// Verify hash was still recorded despite error
```
RebeccaMahany, line 234: Hm, maybe I'm missing something -- is this the desired behavior? I thought we didn't want to store/record the hash, so that we don't have to wait for a change in configuration (i.e. a new hash) to perform the retry. (I think that'd look like changing the behavior around here: https://github.com/kolide/launcher/blob/main/ee/control/control.go#L338-L346.)

cesarfda: I wasn't fully understanding that part but I think I get it now, the hash not being saved would be what will retrigger the fetch and update. Let me adjust that

ee/control/actionqueue/actionqueue.go
```diff
@@ -128,6 +130,7 @@ func (aq *ActionQueue) Update(data io.Reader) error {
 				"failed to do action with action, not marking action complete",
 				"err", err,
 			)
+			processError = fmt.Errorf("actor.Do failed: %w", err)
```
James-Pickett, line 133: Maybe we should return the action type in the error and add to logs above? Might help with debugging in the future.

```suggestion
			processError = fmt.Errorf("actor.Do, action type: %s, failed: %w", action.type, err)
```

RebeccaMahany (approved):
🔥

PR #1948, opened 2024-11-08
Add remote restart consumer to handle remote restart actions

RebeccaMahany:


ee/control/consumers/remoterestartconsumer/errors.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"errors"
+)
+
+type RemoteRestartRequested struct {
+	msg string
+}
+
+func NewRemoteRestartRequested() RemoteRestartRequested {
```
directionless, line 11: Should this have `Error` in the name?

RebeccaMahany: I'll update!

cmd/launcher/main.go
```diff
@@ -153,11 +154,11 @@ func runMain() int {
...
 
 	if err := runLauncher(ctx, cancel, slogger, systemSlogger, opts); err != nil {
-		if !tuf.IsLauncherReloadNeededErr(err) {
+		if !tuf.IsLauncherReloadNeededErr(err) && !remoterestartconsumer.IsRemoteRestartRequestedErr(err) {
 			level.Debug(logger).Log("msg", "run launcher", "stack", fmt.Sprintf("%+v", err))
 			return 1
 		}
-		level.Debug(logger).Log("msg", "runLauncher exited to run newer version of launcher", "err", err.Error())
+		level.Debug(logger).Log("msg", "runLauncher exited to reload launcher", "err", err.Error())
 		if err := runNewerLauncherIfAvailable(ctx, slogger.Logger); err != nil {
```
RebeccaMahany, line 162: This works on macOS/Linux _unless_ there is nothing to be found in the autoupdate library, in which case launcher will exit and we would have to rely on the init system (launchd/systemd) to restart launcher. (On Windows, we always exit completely and rely on the Service Manager to restart launcher.)

This seems like an okay tradeoff to me, since we will be doing this action only rarely as a troubleshooting step, but I wanted to flag it in case others disagree.

ee/control/consumers/remoterestartconsumer/remoterestartconsumer.go
```diff
@@ -0,0 +1,125 @@
...
+	if restartAction.RunID == "" {
+		r.slogger.Log(context.TODO(), slog.LevelInfo,
+			"received remote restart action with blank launcher run ID -- discarding",
+		)
+		return nil
+	}
+	if restartAction.RunID != r.knapsack.GetRunID() {
+		r.slogger.Log(context.TODO(), slog.LevelInfo,
+			"received remote restart action for incorrect (assuming past) launcher run ID -- discarding",
+			"run_id", restartAction.RunID,
```
zackattack01, line 70: should we namespace `run_id` for these as `action_run_id` or something? just wondering if this will get confusing or overwrite the `run_id` that we already include in logs

RebeccaMahany: Oh totally yup, let me update

zackattack01 (approved):
exciting! looks great

ee/control/consumers/remoterestartconsumer/errors.go
```diff
@@ -0,0 +1,35 @@
+package remoterestartconsumer
+
+import (
+	"errors"
+	"fmt"
+)
+
+// ErrRemoteRestartRequested is returned to the main launcher rungroup when
+// a remote restart has been requested.
+type ErrRemoteRestartRequested struct {
```
James-Pickett, line 10: nit, I think the typical convention is to have `Err` at the end of the struct name?

```suggestion
type RemoteRestartRequestedErr struct {
```

RebeccaMahany: Ah, I was confusing the naming convention for struct errors with the naming convention for sentinel errors like `os.ErrNotExist` and `http.ErrServerClosed`. Per your other comment I'll update to use a sentinel error.

James-Pickett: that's a good point, not sure what "correct" is =)

RebeccaMahany: The way you suggested is more concise and much easier to read!

ee/control/consumers/remoterestartconsumer/errors.go
```diff
@@ -0,0 +1,35 @@
...
+		msg:   "need to reload launcher: remote restart requested",
+		runId: runId,
+	}
+}
+
+func (e ErrRemoteRestartRequested) Error() string {
+	return fmt.Sprintf("%s (run_id %s)", e.msg, e.runId)
+}
+
+func (e ErrRemoteRestartRequested) Is(target error) bool {
```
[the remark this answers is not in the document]
RebeccaMahany: Removed in favor of `errors.Is`!

James-Pickett (approved):

ee/control/consumers/remoterestartconsumer/errors.go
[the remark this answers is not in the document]
directionless: If we believe this, is this something we can find a linter for?

PR #1951, opened 2024-11-11
upgrade watchdog service to scheduled task

zackattack01:
The current watchdog implementation installs a lightweight service to periodically check the launcher's service state, and restart if required. The main issue we've found with this in testing is that this service is subject to many of the same power management/sleep state transitions issues that impact the launcher service itself.

These changes alter this pattern to instead utilize a short-lived task that will perform the same state checking and restarts, driven instead by the windows task scheduler service. This is beneficial for a few reasons:
- it gives us the flexibility to operate both on a timer, and as an event-driven task. the task has both time based (every 30 minutes) and power state transition based (1 minute after wake events) triggers
- it offloads the event detection outside of launcher, which should be a more reliable methodology than depending on event subscriptions within a service that is also being heavily throttled immediately before or after these events take place

We still make use of much of the same log buffering, service restart, and controller + flag logic that was initially added for watchdog. This primarily includes the following changes:
- swapping out the remove/install service methods for remove/install task methods. the new logic utilizes go-ole/oleutil to interface with the windows scheduler subsystems and handle all registration and configuration logic
- removing the power event watcher and polling loops from the watchdog, which now just runs whichever checks and actions that we want to add, and exits
- adding `--install-task` and `--remove-task` options to the watchdog subcommand, this was something I had wanted for the original watchdog and it should allow us to easily install or remove on demand while testing
- ensuring that all service and task names used integrate the identifier set within our packaging logic - this was something that I missed for the service name in our original watchdog implementation, and it became very confusing for dual installs


pkg/launcher/pkg_utils_windows.go
```diff
@@ -25,3 +25,16 @@ func ServiceName(identifier string) string {
...
+// and returns the camelCased service name generated to match our packaging logic
+func TaskName(identifier, taskType string) string {
+	if strings.TrimSpace(identifier) == "" {
+		identifier = DefaultLauncherIdentifier
+	}
+
+	sanitizedIdentifier := nonAlphanumericRegex.ReplaceAllString(identifier, "_")                   // e.g. identifier=kolide-k2 becomes kolide_k2
+	sanitizedTaskType := nonAlphanumericRegex.ReplaceAllString(taskType, "_")                       // e.g. taskName=watchdog
+	sanitizedTaskName := fmt.Sprintf("launcher_%s_%s_task", sanitizedIdentifier, sanitizedTaskType) // wrapped as launcher_kolide_k2_watchdog_task
+	return snaker.SnakeToCamel(sanitizedTaskName)                                                   // will produce LauncherKolideK2WatchdogTask
```
RebeccaMahany, line 39: Docs are super helpful as always 💯

ee/watchdog/controller_windows.go
```diff
@@ -180,224 +191,426 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
...
+	wc.slogger.Log(ctx, slog.LevelInfo, "completed watchdog scheduled task installation")
+}
+
+func getExecutablePath(identifier string) (string, error) {
+	if strings.TrimSpace(identifier) == "" {
+		identifier = launcher.DefaultLauncherIdentifier
 	}
 
-	defer serviceManager.Disconnect()
+	binDirBase := fmt.Sprintf(`C:\Program Files\Kolide\Launcher-%s\bin`, identifier)
```
RebeccaMahany, line 225: Maybe we want to add this to `pkg/launcher/paths.go` so that we don't have two separate places where we define the bin directory?

RebeccaMahany: (Does `os.Executable()` work for this? Or not because of how we run launcher on Windows?)

ee/watchdog/controller_windows.go
```diff
@@ -180,224 +191,426 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
...
-	defer existingService.Close()
-	if err = wc.restartService(existingService); err != nil {
-		wc.slogger.Log(ctx, slog.LevelError,
-			"failure attempting to restart watchdog service",
-			"err", err,
-		)
+	regInfo := regInfoProp.ToIDispatch()
+	defer regInfo.Release()
+
+	if _, err = oleutil.PutProperty(regInfo, "Description", "Kolide agent waker"); err != nil {
```
RebeccaMahany, line 302: Instead of "waker" should we use "watchdog" for consistent terminology throughout codebase? Or maybe "restarter" for better precision?

ee/watchdog/controller_windows.go
```diff
@@ -180,224 +191,426 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
...
+		return fmt.Errorf("setting DisallowStartIfOnBatteries flag: %w", err)
+	}
+
+	// task will be continue even if the computer changes power source to battery
+	if _, err = oleutil.PutProperty(settings, "StopIfGoingOnBatteries", false); err != nil {
+		return fmt.Errorf("setting StopIfGoingOnBatteries flag: %w", err)
+	}
+
+	// see compatibility options here https://learn.microsoft.com/en-us/windows/win32/api/taskschd/ne-taskschd-task_compatibility
+	// 2=TASK_COMPATIBILITY_V2 - recommended unless you need to support Windows XP, Windows Server 2003, or Windows 2000
```
RebeccaMahany, line 359: Hopefully not 😅

ee/watchdog/watchdog_task_windows.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/kolide/launcher/pkg/log/multislogger"
+	"github.com/peterbourgon/ff/v3"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+// RunWatchdogTask is typically run as a check to determine the health of launcher and restart if required.
+// it is installed as an exec action via windows scheduled task. e.g. C:\path\to\launcher.exe watchdog -config <path>.
+// you can alternatively run this subcommand to install or remove the scheduled task via the --install-task or --remove-task flags
+func RunWatchdogTask(systemSlogger *multislogger.MultiSlogger, args []string) error {
```
zackattack01, line 24: note that this file was renamed and looks new but most of what has changed here is just the options parsing logic. the rest is a much simpler version (no power event watching, no polling) of what the service version did before

ee/watchdog/controller_windows.go
```diff
@@ -180,224 +191,426 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
...
+	if _, err = oleutil.PutProperty(execAction, "Path", `"`+installedExePath+`"`); err != nil {
+		return fmt.Errorf("setting action path: %w", err)
+	}
+
+	taskArgs := fmt.Sprintf(`watchdog -config "%s"`, configFilePath)
+	if _, err = oleutil.PutProperty(execAction, "Arguments", taskArgs); err != nil {
+		return fmt.Errorf("setting action arguments: %w", err)
+	}
+
+	// now register the task!
```
RebeccaMahany, line 498: Amazing documentation throughout, ty

ee/watchdog/controller_windows.go
```diff
@@ -180,224 +191,426 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
...
+// RemoveWatchdogTask will determine the task name based on the given identifier, and remove
+// the task from the scheduler service. This is exported for use by our remote uninstallation paths
+func RemoveWatchdogTask(identifier string) error {
+	if strings.TrimSpace(identifier) == "" {
+		identifier = launcher.DefaultLauncherIdentifier
+	}
+
+	taskName := launcher.TaskName(identifier, watchdogTaskType)
+	// init COM
+	ole.CoInitialize(0)
```
RebeccaMahany, line 525: Can you add a comment here and the other couple places that you call `CoInitialize` in this file that we discard the error returned by `CoInitialize` because it (harmlessly) returns `S_FALSE` if we call it more than once?

RebeccaMahany (commented):
LGTM overall, just a couple questions

ee/watchdog/controller_windows.go
zackattack01: oh yep! restarter sounds good to me, the task name should include watchdog already. thank you!

ee/watchdog/controller_windows.go
zackattack01: will do!

ee/watchdog/controller_windows.go
zackattack01: ah yeah it felt weird having this here, i'll look into somewhere better in paths, thanks! and yes the only issue with `os.Executable()` is that it could enter the scheduler database with an autoupdated exe path, and then have that path cleaned up after enough future autoupdates

pkg/launcher/paths.go
```diff
@@ -156,3 +157,32 @@ func nonEmptyFileExists(path string) (bool, error) {
...
+
+	var binDirBase string
+	var launcherExeName string
+
+	switch runtime.GOOS {
+	case "windows":
+		binDirBase = fmt.Sprintf(`C:\Program Files\Kolide\Launcher-%s\bin`, identifier)
+		launcherExeName = "launcher.exe"
+	default:
+		binDirBase = fmt.Sprintf(`/usr/local/%s/bin`, identifier)
```
RebeccaMahany, line 177: Ahhh can you throw a quick comment here or in the docblock that this won't work for NixOS, just so we don't forget if we try to use it for Linux in the future?

zackattack01: ahh yep! or do you think it make more sense to just add this to `pkg_utils_windows.go`? it is only in use for these windows tasks currently

zackattack01: added a comment but happy to move this over if you think that's more appropriate

RebeccaMahany (approved):
LGTM! 🔥

PR #1956, opened 2024-11-13
add current version to windows registry on startup

zackattack01:
Part 3: This builds on the changes [here](https://github.com/kolide/launcher/pull/1953) and adds the current running version to windows registry on startup if it is unset or needs to be updated.

See the full issue [here](https://github.com/kolide/k2/issues/10843) for additional context.

### DIFF
--- a/cmd/launcher/svc_config_windows.go
+++ b/cmd/launcher/svc_config_windows.go
@@ -308,6 +308,8 @@ func checkCurrentVersionMetadata(logger *slog.Logger, identifier string) {
 			"encountered error setting current version to registry",
 			"err", err,
 		)
+
+		return
 	}
 
 	logger.Log(context.TODO(), slog.LevelInfo,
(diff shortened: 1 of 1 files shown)


cmd/launcher/svc_config_windows.go
```diff
@@ -252,3 +260,59 @@ func recoveryActionsAreSet(curRecoveryActions, recoveryActions []mgr.RecoveryAct
...
+
+		return
+	}
+
+	// if we can't read, or the version is out of date, set the expected version as the new value
+	if err := launcherVersionKey.SetDWordValue("", uint32(expectedVersion)); err != nil {
+		logger.Log(context.TODO(), slog.LevelError,
+			"encountered error setting current version to registry",
+			"err", err,
+		)
```
James-Pickett, line 310: I think we would want to return here instead of logging the "success" message below?

zackattack01: yes, thank you!

RebeccaMahany (approved):

PR #1957, opened 2024-11-14
add windows installer-info from registry to flares

zackattack01:
We do not write out an installer-info.json file for windows - instead we write the relevant metadata to the registry during MSI installation.

This updates the package install checkup to collect that information (along with our newly added CurrentVersionNum and InstalledVersionNum keys) and write out an installer-info.json file with those values for our flares.

### DIFF
--- a/ee/debug/checkups/install_windows.go
+++ b/ee/debug/checkups/install_windows.go
@@ -1,0 +1,110 @@
+//go:build windows
+// +build windows
+
+package checkups
+
+import (
+	"archive/zip"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/kolide/launcher/pkg/launcher"
+	"golang.org/x/sys/windows/registry"
+)
+
+const (
+	installerInfoRegistryKeyFmt = `Software\Kolide\Launcher\%s\%s`
+	currentVersionKeyName       = `CurrentVersionNum`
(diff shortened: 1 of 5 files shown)


ee/debug/checkups/install_windows.go
```diff
@@ -0,0 +1,110 @@
...
+	return addStreamToZip(z, "installer-info.json", time.Now(), bytes.NewReader(infoJson))
+}
+
+// getDefaultRegistryStringValue queries for the default registry value set at the provided path
+// on the local machine. this is a best effort approach, grabbing whatever info we can. not all
+// devices will have all paths/values present, errors are ignored
+func getDefaultRegistryStringValue(path string) string {
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, path, registry.QUERY_VALUE)
+	if err != nil {
+		return ""
```
RebeccaMahany, line 81: Do we want to use `naIfError` here so we can capture the additional info in the errors too, in case they're useful? Same goes for below and in `getDefaultRegistryIntValue` too

zackattack01: oh yeah that makes sense to me! will do

zackattack01: i ended up just directly returning the errors as strings to avoid dealing with the `any` response everywhere, I think this should achieve the same. let me know what you think!

RebeccaMahany (dismissed):

James-Pickett (approved):

PR #1960, opened 2024-11-18
Set SystemDrive env var on osquery command

RebeccaMahany:
We need the `SystemDrive` environment variable to be set on the osquery command to ensure paths can be resolved appropriately.

### DIFF
--- a/pkg/osquery/runtime/osqueryinstance.go
+++ b/pkg/osquery/runtime/osqueryinstance.go
@@ -802,6 +802,14 @@ func (i *OsqueryInstance) createOsquerydCommand(osquerydBinary string, paths *os
 	// https://github.com/osquery/osquery/pull/6824
 	cmd.Env = append(cmd.Env, "SYSTEM_VERSION_COMPAT=0")
 
+	// On Windows, we want the `SystemDrive` environment variable to be set to ensure paths can be resolved appropriately.
+	// The cmd handles setting `SystemRoot` for us.
+	if runtime.GOOS == "windows" {
+		if systemDrive, found := os.LookupEnv("SystemDrive"); found {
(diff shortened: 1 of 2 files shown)


pkg/osquery/runtime/osqueryinstance_windows_test.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"strings"
+	"testing"
+
+	typesMocks "github.com/kolide/launcher/ee/agent/types/mocks"
+	"github.com/kolide/launcher/pkg/log/multislogger"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateOsqueryCommandEnvVars(t *testing.T) {
```
directionless, line 15: I'm not sure this test adds much, but I don't feel that strongly

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -802,6 +802,14 @@ func (i *OsqueryInstance) createOsquerydCommand(osquerydBinary string, paths *os
 	// https://github.com/osquery/osquery/pull/6824
 	cmd.Env = append(cmd.Env, "SYSTEM_VERSION_COMPAT=0")
 
+	// On Windows, we want the `SystemDrive` environment variable to be set to ensure paths can be resolved appropriately.
+	// The cmd handles setting `SystemRoot` for us.
+	if runtime.GOOS == "windows" {
+		if systemDrive, found := os.LookupEnv("SystemDrive"); found {
```
directionless, line 808: I think we probably need to think hard about the `!found` case. It might be better to set this to `""` than to leave it unset.

RebeccaMahany: @directionless I tested and confirmed setting `SystemDrive=` helps, so I made that change in this follow-up PR: https://github.com/kolide/launcher/pull/1961

PR #1961, opened 2024-11-19
Set all env vars on osquery process

RebeccaMahany:
This is a follow-up to https://github.com/kolide/launcher/pull/1960:

* We may want other env vars set, so this PR updates to grab all of the env vars from `cmd.Environ()` -- this is the default behavior for `cmd` when nothing is set in `cmd.Env`
* In the case that `SystemDrive` happens to be unset, we can prevent undesirable behavior by setting `SystemDrive=` in the env instead -- so this PR adds a check for that as well

### DIFF
--- a/pkg/osquery/runtime/osqueryinstance.go
+++ b/pkg/osquery/runtime/osqueryinstance.go
@@ -805,6 +805,19 @@ func (i *OsqueryInstance) createOsquerydCommand(osquerydBinary string, paths *os
 	// https://github.com/osquery/osquery/pull/6824
 	cmd.Env = append(cmd.Env, "SYSTEM_VERSION_COMPAT=0")
 
+	// On Windows, we need to ensure the `SystemDrive` environment variable is set to _something_,
+	// so if it isn't already set, we set it to an empty string.
+	systemDriveEnvVarFound := false
+	for _, e := range cmd.Env {
+		if strings.Contains(strings.ToLower(e), "systemdrive") {
+			systemDriveEnvVarFound = true
+			break
+		}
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -802,13 +805,18 @@ func (i *OsqueryInstance) createOsquerydCommand(osquerydBinary string, paths *os
...
 
-	// On Windows, we want the `SystemDrive` environment variable to be set to ensure paths can be resolved appropriately.
-	// The cmd handles setting `SystemRoot` for us.
-	if runtime.GOOS == "windows" {
-		if systemDrive, found := os.LookupEnv("SystemDrive"); found {
-			cmd.Env = append(cmd.Env, fmt.Sprintf("SystemDrive=%s", systemDrive))
+	// On Windows, we need to ensure the `SystemDrive` environment variable is set to _something_,
+	// so if it isn't already set, we set it to an empty string.
+	systemDriveEnvVarFound := false
+	for _, e := range cmd.Env {
```
directionless, line 811: Things like this make me grumpy it's an array and not a hash

directionless (approved):
I think this is okay, and just to talk through it...

My normal inclination is to set a _minimal_ ENV -- less likely to make a executated command behave unexpectedly. But, we're not running suid, and we only have the permissions we start with. So, this shouldn't negatively impact us --  if you can adjust launcher's env, you could just call osquery with that same adjusted ENV.

PR #1962, opened 2024-11-19
deny default write permissions from windows root directory

zackattack01:
this adds an additional check to the windows_svc_config startup routine to verify permissions for the configured root directory. I do not love the concept of adding an explicit denial instead of just removing the existing grant, but this has been the only way i've seen success so far, after trying several different ways to do this.

Once the changes run I am no longer able to create/write nested directories inside of the root directory without using an admin console. Subsequent runs see the entry exists and skip before making further changes.

Follow up PRs will attempt to set the proper directory permissions on package install

### DIFF
--- a/cmd/launcher/svc_config_windows.go
+++ b/cmd/launcher/svc_config_windows.go
@@ -382,11 +382,10 @@ func checkRootDirACLs(logger *slog.Logger, rootDirectory string) {
 	// if there is already a DENY entry set for user's group to avoid recreating every time
 	for i := 0; i < int(existingDACL.AceCount); i++ {
 		var ace *windows.ACCESS_ALLOWED_ACE
-		err = windows.GetAce(existingDACL, uint32(i), &ace)
-		if err != nil {
+		if aceErr := windows.GetAce(existingDACL, uint32(i), &ace); aceErr != nil {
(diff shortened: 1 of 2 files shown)


RebeccaMahany (dismissed):

James-Pickett (approved):

cmd/launcher/svc_config_windows.go
```diff
@@ -318,3 +327,127 @@ func checkCurrentVersionMetadata(logger *slog.Logger, identifier string) {
...
+		)
+
+		return
+	}
+
+	// first iterate the existing ACEs for the directory, we're checking to see
+	// if there is already a DENY entry set for user's group to avoid recreating every time
+	for i := 0; i < int(existingDACL.AceCount); i++ {
+		var ace *windows.ACCESS_ALLOWED_ACE
+		err = windows.GetAce(existingDACL, uint32(i), &ace)
```
directionless, line 385: reusing `err` in here always feels fixy. Can we use the common `if err := ...` form?

zackattack01: oh yep!

RebeccaMahany (approved):

James-Pickett (approved):

PR #1963, opened 2024-11-19
[KATC] Add support for doubles to indexeddb parsing

RebeccaMahany:
Our indexeddb parser currently doesn't handle doubles. This PR implements parsing doubles, and updates the test database to confirm the parsing does not return an error.


ee/indexeddb/values.go
```diff
@@ -26,6 +26,7 @@ const (
 	// numbers
 	tokenInt32  byte = 0x49 // I
 	tokenUint32 byte = 0x55 // U
+	tokenDouble byte = 0x4e // N
```
RebeccaMahany, line 29: The source for this value, as well as the troubleshooting steps I took to fix this parsing error, are documented here: https://github.com/kolide/launcher/tree/main/ee/katc/test_data#troubleshooting-parsing-errors

zackattack01 (approved):

PR #1966, opened 2024-11-21
Don't perform remote restart yet

RebeccaMahany:
Partially rolls back changes from https://github.com/kolide/launcher/pull/1948.

The tradeoff noted in the comment on the previous PR (https://github.com/kolide/launcher/pull/1948#discussion_r1837205965) was a bigger one than I realized on Linux, so I want to think more about how to handle the restart.

In the meantime, so we can get the release out, I'm updating to keep the consumer but not use it yet.


James-Pickett (approved):

zackattack01 (approved):

directionless:
I'd probably have left this in, and rolled it back in K2. Or popped up a warning

RebeccaMahany:
@directionless I preferred this approach because the result on Linux is that launcher shuts down and does not restart without manual user intervention -- so it seemed safest/most prudent to make remote restart not at all possible on the launcher side, for the time being

Issue #1892, opened 2024-10-16
Osquery checkup returns "failed to run" on Windows even when successful


RebeccaMahany:
On Windows, doctor output looks like this for the osquery checkup:

```
Osquery: failed to run: running launcher interactive: running C:\ProgramData\Kolide\Launcher-kolide-k2\data\updates\launcher\1.11.2\launcher.exe interactive: err exit status 1, output +-------+--------------------------------------+--------------------------------------+---------+------------------------------------------+--------------+------------+----------------+--------------+------------+---------+---------------+
| pid   | uuid                                 | instance_id                          | version | config_hash
                   | config_valid | extensions | build_platform | build_distro | start_time | watcher | platform_mask |
+-------+--------------------------------------+--------------------------------------+---------+------------------------------------------+--------------+------------+----------------+--------------+------------+---------+---------------+
| redacted | redacted | redacted | 5.13.1  | redacted | 1            | active     | windows        | 10           | 1729106122 | -1      | 2             |
+-------+--------------------------------------+--------------------------------------+---------+------------------------------------------+--------------+------------+----------------+--------------+------------+---------+---------------+
```

I think the interactive query is running successfully, but then the interactive process remains running, and eventually we hit the timeout?

We did `launcher interactive` over something like `runsimple` to test launcher <=> osquery communication, but if there's no other way to fix this, we should just switch to `runsimple`.

directionless:
Yeah -- my memory is that we did it this way to test 3 things. (Which is a little clever, but it's a bit of a e2e test)

1. Does interactive work?
2. Does launcher - osquery socket work?
3. Record the output

RebeccaMahany:
I was thinking that if we're having issues with 2), we'd see that reflected in the logs anyway -- but true re: interactive. I'll see if I can get it figured out!

directionless:
I assume this is weird fallout from that whole shift to windowsgui

RebeccaMahany:
We fixed the timeout issue, but I see a new one on v1.12.3 of launcher -- reopened.

PR #1968, opened 2024-11-26
Set config on interactive command in osquery checkup

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1892.

https://github.com/kolide/launcher/pull/1958 appeared to fix the above issue. However, during validation for v1.12.3, we saw a new iteration of the same error -- this time, though, it wasn't a timeout error.

When testing, I am finding that whether the osquery checkup fails depends on my `localdev_path` setting (which also explains why I did not see this issue when testing the previous PR).

| `localdev_path` | Checkup fails |
|---|---|
| Not set -- launcher path will be found at `<bin-dir>\launcher.exe`  | Yes |
| Not set -- launcher path will be found at`<updates-dir>\launcher\<nightly>\launcher.exe` | Yes |
| Path to my locally-built version of main | No |
| Path to `<bin-dir>\launcher.exe` | No |
| Path to `<updates-dir>\launcher\<nightly>\launcher.exe` | Yes |

I'm not quite sure how to explain this behavior yet, especially when I can't reproduce it in my test environment. (Plus -- for my secondary install on the same computer, as well as for both installs on @zackattack01's Windows machine, this checkup consistently passes now.)

In this PR, I'm trying setting the `--config` flag to see if that will change any of the behavior here. Regardless, it's useful to set the `--config` flag so e.g. the root directory can be picked up correctly by the interactive command.

### DIFF
--- a/ee/debug/checkups/osquery.go
+++ b/ee/debug/checkups/osquery.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/kolide/launcher/ee/agent/types"
+	"github.com/kolide/launcher/pkg/launcher"
 )
 
 type osqueryCheckup struct {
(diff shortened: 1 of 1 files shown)


ee/debug/checkups/osquery.go
```diff
@@ -66,10 +67,15 @@ func (o *osqueryCheckup) interactive(ctx context.Context) error {
 		return fmt.Errorf("getting current running executable: %w", err)
 	}
 
+	flagFilePath := launcher.DefaultPath(launcher.ConfigFile)
+	if o.k != nil && o.k.Identifier() != "" {
+		flagFilePath = strings.ReplaceAll(flagFilePath, launcher.DefaultLauncherIdentifier, o.k.Identifier())
```
zackattack01, line 72: 🔥

Issue #1938, opened 2024-11-05
Make the osquery instance status available in the knapsack


RebeccaMahany:
The osquery runner should be able to report on the status of each of the instances it manages. Status should probably be one of `unstarted|launching|healthy|unhealthy` or similar. Since we are moving toward running multiple instances, status must be stored per-instance -- use the `registration ID` to store. (Currently we only have one "default" registration ID.)

We should add the instances' statuses to the status blob returned by localserver.

PR #1976, opened 2024-12-04
Expose osquery instance status to knapsack

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1938

Adds a new function to expose the current status (not yet started, healthy, or unhealthy) of each instance. Adds this data to the status blob returned by the /id localserver endpoint, and to the osquery checkup.

This also makes the osquery runner available in the knapsack, which may be useful in the future.

### DIFF
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -39,9 +39,9 @@ type knapsack struct {
 
 	slogger, systemSlogger *multislogger.MultiSlogger
 
+	querier types.InstanceQuerier
+
 	// This struct is a work in progress, and will be iteratively added to as needs arise.
-	// Some potential future additions include:
-	// Querier
 }
 
 func New(stores map[storage.Store]types.KVStore, flags types.Flags, db *bbolt.DB, slogger, systemSlogger *multislogger.MultiSlogger) *knapsack {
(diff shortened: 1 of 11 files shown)


ee/agent/knapsack/knapsack.go
```diff
@@ -87,6 +87,18 @@ func (k *knapsack) AddSlogHandler(handler ...slog.Handler) {
 	k.systemSlogger.AddHandler(handler...)
 }
 
+// Osquery instance querier
+func (k *knapsack) SetInstanceQuerier(q types.InstanceQuerier) {
+	k.querier = q
+}
+
+func (k *knapsack) InstanceStatuses() map[string]types.InstanceStatus {
```
zackattack01, line 95: might be nice to add a comment in here noting what this is doing/how it gets the info (e.g. iterating registered instances and healthchecking inline). i personally just like being able to quickly tell if something coming from knapsack is cached/persisted somewhere vs real time checks

zackattack01 (approved):
nice!

ee/agent/knapsack/knapsack.go
RebeccaMahany: Yes, good call! Especially since I spent ages waffling on whether I wanted to throw a timeout in this. Will update!

zackattack01 (approved):

James-Pickett (approved):

PR #1977, opened 2024-12-04
hardware key runners

James-Pickett:
* adds secure enclave create key command to desktop user server, client, and root desktop runner
* adds a "runner" style struct to handle initializing keys for both tpm and secure enclave that is added to the rungroup in main launcher command, this allows for initializing (and retrying failures) of tpm and secure enclave keys
* signing with secure enclave will come in a follow on PR


ee/desktop/user/client/client.go
```diff
@@ -95,6 +98,35 @@ func (c *client) DetectPresence(reason string, interval time.Duration) (time.Dur
 	return durationSinceLastDetection, detectionErr
 }
 
+func (c *client) CreateSecureEnclaveKey() (*ecdsa.PublicKey, error) {
+	resp, err := c.base.Get("http://unix/secure_enclave_key")
```
James-Pickett, line 102: use `c.get("create_secure_enclave_key")`

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,281 @@
...
+					"err", err,
+				)
+			}
+		}
+	}
+
+	return ser, nil
+}
+
+// Public returns the public key of the current console user
```
James-Pickett, line 84: update comment

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,281 @@
...
+	defer retryTicker.Stop()
+
+	for {
+		if _, err := ser.currentConsoleUserKey(context.TODO()); err != nil {
+			ser.slogger.Log(context.TODO(), slog.LevelError,
+				"getting current console user key, will retry",
+				"err", err,
+			)
+
+			if currentRetryInterval < maxRetryInterval {
```
James-Pickett, line 98: lazy backoff ... is there a better way or is this good enough?

James-Pickett: actaully, I don't this is backing off, need to reset the ticker

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,215 @@
...
+	"io"
+	"log/slog"
+	"time"
+
+	"github.com/kolide/krypto/pkg/tpm"
+	"github.com/kolide/launcher/ee/agent/types"
+	"github.com/kolide/launcher/pkg/traces"
+)
+
+type tpmRunner struct {
```
James-Pickett, line 17: find a way to organize these types better, leave comments

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (d defaultTpmSignerCreator) New(private, public []byte) (crypto.Signer, error) {
+	return tpm.New(private, public)
+}
+
+type tpmRunnerOption func(*tpmRunner)
+
+func New(ctx context.Context, slogger *slog.Logger, store types.GetterSetterDeleter, opts ...tpmRunnerOption) (*tpmRunner, error) {
+	_, span := traces.StartSpan(ctx)
```
James-Pickett, line 44: dont think we need tracking in `New()`, but should be in execute / crypto opts.

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,215 @@
...
+
+		signer, err := tr.fetchCreateKeys(context.TODO())
+		if err != nil {
+			tr.slogger.Log(context.TODO(), slog.LevelError,
+				"creating tpm signer, will retry",
+				"err", err,
+			)
+
+			if currentRetryInterval < maxRetryInterval {
+				currentRetryInterval += time.Second
```
James-Pickett, line 83: need to actually reset the ticker

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,282 @@
...
+
+// Public returns the public key of the current console user
+// creating and peristing a new one if needed
+func (ser *secureEnclaveRunner) Execute() error {
+	currentRetryInterval, maxRetryInterval := 1*time.Second, 1*time.Minute
+	retryTicker := time.NewTicker(currentRetryInterval)
+	defer retryTicker.Stop()
+
+	for {
+		if _, err := ser.currentConsoleUserKey(context.TODO()); err != nil {
```
RebeccaMahany, line 92: If we don't currently have a logged-in user, `currentConsoleUserKey` will return nil -- so the ticker will stop, and we will never retry creating the key until launcher itself is restarted. Is that okay? It seems like in that case we might want to keep retrying, or maybe implement something where we only start retrying once we detect a logged-in user exists?

ee/secureenclaverunner/secureenclaverunner_test.go
```diff
@@ -0,0 +1,93 @@
+//go:build darwin
+// +build darwin
+
+package secureenclaverunner
```
RebeccaMahany, line 4: Would you add in a `Test_InterruptMultiple` test also? (You can basically just paste in https://github.com/kolide/launcher/blob/main/ee/control/consumers/remoterestartconsumer/remoterestartconsumer_test.go#L120.) I like to have a `Test_InterruptMultiple` everywhere we add a new rungroup to be certain we keep avoiding any regressions in the interrupt behavior.

ee/tpmrunner/tpmrunner_test.go
```diff
@@ -0,0 +1,80 @@
+package tpmrunner
```
RebeccaMahany, line 1: Same as for the secureenclaverunner -- would you paste in a `Test_InterruptMultiple` test?

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,241 @@
+package tpmrunner
```
RebeccaMahany, line 1: Would you add a quick diagram/writeup in the `docs/architecture` folder for this? Since this work traverses both root and user launcher, and the implementation differs between darwin/other OSes (plus the keys are pretty important) I think it's worth documenting how the different pieces fit together. Doesn't have to be in this PR!

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// This duplicates some of pkg/osquery/extension.go but that feels like the wrong place.
+// Really, we should have a simpler interface over a storage layer.
+const (
+	privateEccData = "privateEccData"
+	publicEccData  = "publicEccData"
+)
+
+// nolint:unused
```
RebeccaMahany, line 152: I think this nolint directive and the others below can be removed now?

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,241 @@
...
+type tpmRunnerOption func(*tpmRunner)
+
+func New(ctx context.Context, slogger *slog.Logger, store types.GetterSetterDeleter, opts ...tpmRunnerOption) (*tpmRunner, error) {
+	_, span := traces.StartSpan(ctx)
+	defer span.End()
+
+	_, _, err := fetchKeyData(store)
+	if err != nil {
+		return nil, err
+	}
```
RebeccaMahany, line 58: Why do we have to do this here? I'm curious because it doesn't seem to have any side effects. (Also, if we do have to do this but not _here_ necessarily, maybe we could move it to the top of `Execute` to not delay rungroup startup.)

ee/secureenclaverunner/secureenclaverunner.go
RebeccaMahany: I like this and think it's good enough! Since you use it here and in the tpmrunner (and since I can imagine it being useful in other areas too) I wonder if we might want to extract it into `pkg/backoff`?

RebeccaMahany (commented):
I really like how all of this fits together with the runners!

ee/secureenclaverunner/secureenclaverunner.go
RebeccaMahany: Oh, I just saw we do actually have a `Backoff` in that package -- I forgot we had anything in there besides `WaitFor` 😅 . Maybe an opportunity to use that/combine?

ee/tpmrunner/tpmrunner.go
directionless: I'm a little curious about this as well. I understand why we need to move the secure enclave signing. But I'm much less sure what's happening with the TPM

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,241 @@
...
+		return
+	}
+
+	tr.interrupted = true
+
+	// Tell the execute loop to stop checking, and exit
+	tr.interrupt <- struct{}{}
+}
+
+// Public returns the public key of the current console user
```
directionless, line 123: Is this actually connected to the console user?

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,282 @@
...
+		interrupt:           make(chan struct{}),
+	}
+
+	data, err := store.Get([]byte(publicEccDataKey))
+	if err != nil {
+		traces.SetError(span, fmt.Errorf("getting public ecc data from store: %w", err))
+		return nil, fmt.Errorf("getting public ecc data from store: %w", err)
+	}
+
+	if data != nil {
```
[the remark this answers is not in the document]
James-Pickett: done

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,282 @@
...
+	ser := &secureEnclaveRunner{
+		uidPubKeyMap:        make(map[string]*ecdsa.PublicKey),
+		store:               store,
+		secureEnclaveClient: secureEnclaveClient,
+		slogger:             slogger.With("component", "secureenclaverunner"),
+		mux:                 &sync.Mutex{},
+		interrupt:           make(chan struct{}),
+	}
+
+	data, err := store.Get([]byte(publicEccDataKey))
```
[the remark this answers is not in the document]
James-Pickett: Since this was just reading from bbolt and unmarshalling the json (without actual secure enclave op) I though it was reasonable to have it in new, but I suppose putting it in the top of execute would save a bit of start up time.

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,282 @@
...
+				"getting current console user key, will retry",
+				"err", err,
+			)
+
+			if currentRetryInterval < maxRetryInterval {
+				currentRetryInterval += time.Second
+				retryTicker.Reset(currentRetryInterval)
+			}
+		} else {
+			retryTicker.Stop()
```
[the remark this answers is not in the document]
James-Pickett: yes, I checked this in a go playground

ee/secureenclaverunner/secureenclaverunner.go
James-Pickett: good catch! I'll fix that

ee/tpmrunner/tpmrunner.go
James-Pickett: @RebeccaMahany , I will write up some docs for this PR

@directionless there are 2 reasons I went ahead and made the tpm a "runner" style like secure enclave.

1. We do see errors pop up with TPM operations that I can't reproduce. I'm hoping that these have something to do with the TPM not being ready or unavailable for some reason, taking this route we'll keep trying in the background hoping it becomes available.

2. It felt cleaner to have both hardware signature mechanisms operate in the same way. Understanding that no kind of hardware signing can be expected to be available immediately after start up.

ee/tpmrunner/tpmrunner.go
James-Pickett: done

ee/tpmrunner/tpmrunner.go
James-Pickett: It just verifies that the provided store is valid. I think your comment [here](https://github.com/kolide/launcher/pull/1977#discussion_r1873440394) applies to this as well. I think you're right, we should move these ops to execute, there is very little chance that the store will be invalid and if it ever was for some reason, I'm willing to delay that discovery to the execute loop.

ee/agent/keys_tpm.go
```diff
@@ -5,52 +5,20 @@ package agent
...
-			clearKeyData(slogger, store)
-			traces.SetError(span, fmt.Errorf("storing key: %w", err))
-			return nil, fmt.Errorf("storing key: %w", err)
-		}
-
-		span.AddEvent("new_key_stored")
-	}
-
-	k, err := tpm.New(priData, pubData)
+// SetHardwareKeysRunner creates a tpm runner and sets it as the agent hardware key as it also implements the keyInt/cyrpto.Signer interface.
```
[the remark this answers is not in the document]
James-Pickett: The runner also implements keyInt & crypto.Signer and gets set as the hardware key, maybe `CreateSetHardwareKeysRunner`? Or possible different functions for create and set?

ee/tpmrunner/tpmrunner.go
James-Pickett: no, that was just copy pasta, fixed

ee/secureenclaverunner/secureenclaverunner.go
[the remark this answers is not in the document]
James-Pickett: moved to top of execute

ee/tpmrunner/tpmrunner.go
James-Pickett: moved to top of execute

ee/secureenclaverunner/secureenclaverunner.go
James-Pickett: I didn't really like what was in there for this application, because there was no way to stop, so I created a custom ticker with "Multiplicative Backoff" (duration * ticks) ... maybe went a little overboard =)

ee/desktop/user/client/client.go
```diff
@@ -95,6 +98,35 @@ func (c *client) DetectPresence(reason string, interval time.Duration) (time.Dur
 	return durationSinceLastDetection, detectionErr
 }
 
+func (c *client) CreateSecureEnclaveKey() (*ecdsa.PublicKey, error) {
+	resp, err := c.base.Get("http://unix/create_secure_enclave_key")
```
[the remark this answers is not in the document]
James-Pickett: done

ee/tpmrunner/tpmrunner_test.go
James-Pickett: done, though I had to make the `interrupted` bools atmoic to deal with data race

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (ser *secureEnclaveRunner) currentConsoleUserKey(ctx context.Context) (*ecdsa.PublicKey, error) {
+	ctx, span := traces.StartSpan(ctx)
+	defer span.End()
+
+	ser.mux.Lock()
+	defer ser.mux.Unlock()
+
+	cu, err := firstConsoleUser(ctx)
+	if err != nil {
```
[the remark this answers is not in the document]
James-Pickett: I started to do this, but I don't see where we would need to switch on the error and behave differently. Do you have a specific case in mind?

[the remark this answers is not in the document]
RebeccaMahany: I suggested this while thinking about the other comment on handling the case where there's no console user -- https://github.com/kolide/launcher/pull/1977#discussion_r1873476377. I thought we might want special handling for the case where there's no console user (e.g. maybe back off to a longer interval faster? or something else?), so we'd need a specific error to be able to detect that state.

ee/secureenclaverunner/secureenclaverunner.go
RebeccaMahany: It looks like this is still unresolved, or am I missing the fix?

ee/tpmrunner/tpmrunner_test.go
RebeccaMahany: The data races with the rungroup actors are always such a pain in tests 😅

ee/secureenclaverunner/secureenclaverunner.go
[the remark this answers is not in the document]
James-Pickett: I have added this in, if we don't have a console users, we'll check every 15 seconds until we have one, then start regular execute loop.

ee/desktop/user/client/client.go
```diff
@@ -95,6 +98,35 @@ func (c *client) DetectPresence(reason string, interval time.Duration) (time.Dur
 	return durationSinceLastDetection, detectionErr
 }
 
+func (c *client) CreateSecureEnclaveKey() (*ecdsa.PublicKey, error) {
+	resp, err := c.base.Post("http://unix/secure_enclave_key", "application/json", http.NoBody)
```
RebeccaMahany, line 102: Oh neat, I didn't know about `http.NoBody`!

ee/desktop/user/client/client.go
```diff
@@ -95,6 +98,35 @@ func (c *client) DetectPresence(reason string, interval time.Duration) (time.Dur
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading response body: %w", err)
+	}
+
+	key, err := echelper.PublicB64DerToEcdsaKey(body)
+	if err != nil {
+		return nil, fmt.Errorf("converting key: %w", err)
+	}
```
RebeccaMahany, line 125: This is a nitpick, sorry I didn't notice it sooner --

I think it makes more sense from a separation of responsibilities standpoint to have this client function return the raw bytes of the body, and have this conversion happen inside the runner's `CreateSecureEnclaveKey` func instead -- since the runner implements an interface called `secureEnclaveClient`, it feels appropriate for the runner to know about the key format, but it doesn't seem necessary for the http client here to know about the key format.

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,305 @@
...
+)
+
+type secureEnclaveRunner struct {
+	uidPubKeyMap        map[string]*ecdsa.PublicKey
+	secureEnclaveClient secureEnclaveClient
+	store               types.GetterSetterDeleter
+	slogger             *slog.Logger
+	mux                 *sync.Mutex
+	interrupt           chan struct{}
+	interrupted         atomic.Bool
```
RebeccaMahany, line 39: Do you think it's worth updating the other rungroup actors to use atomic.Bool if they use the `interrupt`/`interrupted` pattern? Let me know if yes -- I can file an issue to tackle that in the future 🙂

ee/secureenclaverunner/secureenclaverunner.go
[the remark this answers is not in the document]
RebeccaMahany: I like how you updated this! Really easy to follow

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -0,0 +1,305 @@
...
+const (
+	publicEccDataKey = "publicEccData"
+)
+
+type secureEnclaveRunner struct {
+	uidPubKeyMap        map[string]*ecdsa.PublicKey
+	secureEnclaveClient secureEnclaveClient
+	store               types.GetterSetterDeleter
+	slogger             *slog.Logger
+	mux                 *sync.Mutex
```
RebeccaMahany, line 37: I think it would be helpful to rename `mux` to make it clearer what resource (I think `uidPubKeyMap`) is being locked, or what kind of concurrent action we're trying to prevent

ee/tpmrunner/tpmrunner.go
```diff
@@ -0,0 +1,249 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// clearKeyData is used to clear the keys as part of error handling around new keys. It is not intended to be called
+// regularly, and since the path that calls it is around DB errors, it has no error handling.
+// nolint:unused
```
RebeccaMahany, line 189: I think you can remove the nolint directive here

RebeccaMahany (commented):
Just a couple small nitpicks, LGTM! Sorry for the delay on re-review this time around

ee/secureenclaverunner/secureenclaverunner.go
James-Pickett: yea, I think so, feels more correct

ee/desktop/user/client/client.go
James-Pickett: done

ee/secureenclaverunner/secureenclaverunner.go
James-Pickett: done

ee/tpmrunner/tpmrunner.go
James-Pickett: done

ee/secureenclaverunner/secureenclaverunner.go
RebeccaMahany: Done! https://github.com/kolide/launcher/issues/2010

RebeccaMahany (approved):
Nice!!

RebeccaMahany (approved):

PR #1979, opened 2024-12-06
add watchdog trigger immediately following wake event

zackattack01:
The initial watchdog task triggers included 1 event based trigger to fire after a 1 minute delay from a wake event, and another 30 minute time based trigger. The event based trigger utilizes a delay to avoid running too early if resources are constrained immediately following a wake event, but in many cases we would want to try immediately anyway and avoid the delay.

This adds a third trigger, using the same event subscription but without the delay

### DIFF
--- a/ee/watchdog/controller_windows.go
+++ b/ee/watchdog/controller_windows.go
@@ -218,7 +218,8 @@ func (wc *WatchdogController) ServiceEnabledChanged(enabled bool) {
 // installWatchdogTask registers our watchdog subcommand as a scheduled task.
 // see inline comments for details on various settings, but here is a general overview:
 // Triggers:
-// - 1 minute after any wake event
+// - immediately following any wake event
+// - 1 minute after any wake event (to allow for retrying if resource constrained)
 // - every 30 minutes as a routine check
 // Action:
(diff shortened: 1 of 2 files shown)


ee/watchdog/controller_windows.go
```diff
@@ -410,6 +411,31 @@ func installWatchdogTask(identifier, configFilePath string) error {
 		return fmt.Errorf("setting event trigger delay: %w", err)
 	}
 
+	// add a second trigger using the same event subscription, but without a delay.
+	// this way we will fire immediately upon wake up, and again a minute later in case
+	// things were too busy during wake
+	noDelayTriggerResp, err := oleutil.CallMethod(triggers, "Create", uint(0)) // 0=TASK_TRIGGER_EVENT
+	if err != nil {
+		log.Fatalf("encountered error creating trigger: %s", err.Error())
```
RebeccaMahany, line 419: I don't think we want a `Fatalf` here -- `return fmt.Errorf`?

zackattack01: oop good catch! thank you

RebeccaMahany (approved):

Issue #1939, opened 2024-11-05
Osquery configuration and data must be stored per-organization


RebeccaMahany:
Store osquery data (node key, uuid, config, keys, ATC config, KATC config) in bbolt and in sqlite per-organization as well as globally. (We want to keep storing globally because this change must be backwards compatible.) To start, store everything under the current tenant. We may want to key off the node key, or a "registration ID" that doesn't yet exist in K2; we don't want to key off munemo in case we end up wanting multiple instances per single tenant.

PR #1980, opened 2024-12-06
Add basic registration tracking to knapsack

RebeccaMahany:
Some preliminary work for https://github.com/kolide/launcher/issues/1939 --

* Adds `RegistrationIds` function to knapsack so that anywhere with access to the knapsack can find the current registration IDs -- leaves list hardcoded to just the default for now
* Moves the default registration ID to more central location

### DIFF
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -93,8 +93,8 @@ func (k *knapsack) SetInstanceQuerier(q types.InstanceQuerier) {
 }
 
 // RegistrationTracker interface methods
-func (k *knapsack) RegistrationIds() []string {
-	return []string{types.DefaultRegistrationId}
+func (k *knapsack) RegistrationIDs() []string {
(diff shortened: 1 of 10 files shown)


ee/agent/knapsack/knapsack.go
```diff
@@ -92,6 +92,11 @@ func (k *knapsack) SetInstanceQuerier(q types.InstanceQuerier) {
 	k.querier = q
 }
 
+// RegistrationTracker interface methods
+func (k *knapsack) RegistrationIds() []string {
```
zackattack01, line 96: no strong preference here but do we want to use `RegistrationIDs` for consistency with e.g. `GetRunID`?

zackattack01 (approved):

ee/agent/knapsack/knapsack.go
RebeccaMahany: Done!

zackattack01 (approved):

PR #1982, opened 2024-12-06
Add registration ID to osquery history

RebeccaMahany:
Another smaller task for https://github.com/kolide/launcher/issues/1939 -- adds the registration ID to osquery instance history, and includes it appropriately when working with instance history.

Updates the `kolide_launcher_config` table to handle multiple registration IDs in preparation for registrations having different osquery configs.

### DIFF
--- a/pkg/osquery/runtime/history/history_test.go
+++ b/pkg/osquery/runtime/history/history_test.go
@@ -59,7 +59,7 @@ func TestNewInstance(t *testing.T) { // nolint:paralleltest
 
 			require.NoError(t, InitHistory(setupStorage(t, tt.initialInstances...)))
 
-			_, err := NewInstance(ulid.New())
+			_, err := NewInstance(ulid.New(), ulid.New())
 
 			assert.Equal(t, tt.wantNumInstances, len(currentHistory.instances), "expect history length to reflect new instance")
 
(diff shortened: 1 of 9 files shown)


pkg/osquery/table/launcher_info.go
```diff
@@ -57,7 +58,7 @@ func generateLauncherInfoTable(configStore types.GetterSetter, LauncherHistorySt
 			return nil, err
 		}
 
-		osqueryInstance, err := history.LatestInstance()
+		osqueryInstance, err := history.LatestInstanceByRegistrationID(types.DefaultRegistrationID)
```
RebeccaMahany, line 61: I don't think we want to return multiple rows for this table -- I'm not sure how K2 will handle that at the moment. So for now, I'm defaulting to using default ID (the same as we do in other places).

James-Pickett (approved):

pkg/osquery/runtime/history/history_test.go
```diff
@@ -176,6 +176,82 @@ func TestLatestInstance(t *testing.T) { // nolint:paralleltest
...
+
+			require.NoError(t, InitHistory(setupStorage(t, tt.initialInstances...)))
+
+			got, err := LatestInstanceByRegistrationID(tt.registrationID)
+
+			if tt.errString != "" {
+				assert.EqualError(t, err, tt.errString)
+			}
+
+			assert.Equal(t, tt.want, got)
```
zackattack01, line 250: should this be in an else block?

zackattack01: oh i see, works either way

zackattack01 (approved):

pkg/osquery/table/launcher_info.go
directionless: You can return whatever makes sense here, multiple rows are fine.

This isn't data k2 collects, it's here solely for our debugging.

directionless: Correcting myself... We _do_ query `kolide_launcher_info`, but I don't think we look at this column.

Issue #2004, opened 2024-12-17
Need to remediate when osquery cannot create database lock file


RebeccaMahany:
I saw this error recently in the automated tests:

> I1217 06:16:53.786639  2924 rocksdb.cpp:181] Rocksdb open failed (5:0) IO error: Failed to create lock file: some-root-dir\osquery.db/LOCK: The process cannot access the file because it is being used by another process.

At least at the time of the error, there was only one osquery process. It's not clear from the logs what happened, but I think the osquery process didn't get a chance to terminate completely and the lock file got left behind.

launcher cannot handle when the osquery process fails to start up fully with an error like this -- the only remediation currently available is manual.

We should a) improve the shutdown routine to make sure that launcher doesn't get into this situation, and b) update the osquery runner to detect this state and take corrective action.

Thoughts re: a) improve the shutdown routine to make sure that launcher doesn't get into this situation:

- Do we need a longer timeout for the shutdown routines? (This timeout is not enforced by the runner or instance -- it comes from the rungroup and/or from whatever is managing the launcher service.) I think timeout during shutdown is the most likely explanation for this issue. In this particular example, the issue occurred on Windows, so it may be something specific to the Windows service.
- Do we want to introduce some kind of ordering to the shutdown routines? I think it's possible we could be running into this issue because some of the shutdown routines happen simultaneously with the osquery process shutdown, and this disrupts clean shutdown.
- Do we need to change how we're shutting down osquery to shut it down more gracefully? I don't think there's much more we can do here, but maybe I'm missing something.

Thoughts re: b) update the osquery runner to detect this state and take corrective action:

- Is it a good idea for the osquery instance to remove the lock file (if one exists) before starting up a new osquery process? (I think we would prefer this to having the osquery instance remove the lock file on shutdown because the shutdown tasks run more or less simultaneously, so we could end up with e.g. the instance removing the lock file before the osquery process gets the chance to shut itself down cleanly.)
- The osquery instance and osquery runner both currently have no visibility into the osquery logs -- they're processed by the log adapter, which is entirely separate. However, here and previously (when trying to figure out how to handle launcher falling back to a old version of osquery that is incompatible with the database -- some details [here](https://github.com/kolide/launcher/issues/1937)) we've tentatively wanted the osquery instance to be able to respond to particular logs from osquery. Do we want to move log processing into the osquery instance? Or open up some line of communication between the instance and the log adapter?

RebeccaMahany:
[...]

Issue #1937, opened 2024-11-05
Retry launching osquery instance on failure


RebeccaMahany:
When the osquery runner cannot launch an osquery instance, we currently return an error, which will shut down launcher entirely.

Looking over the logs and past issues we've investigated, I see two primary errors: 1) `timeout waiting for osqueryd to create socket`, indicating the osquery process did not start up, and 2) `could not create an extension client` where the socket file does not exist or the connection is refused.

In both of these cases, restarting launcher is overkill, and even detrimental to solving the issue. In some cases, we can see these errors happen when the current osquery version is old and not compatible with the current database; restarting launcher in this case is actively harmful because it resets the autoupdate delay, preventing a newer osquery version from being downloaded.

So! We want to change the runner behavior to repeatedly retry starting osquery instances and _not_ exit from the runner.

1. If osquery instance launch fails, retry launching the instance -- potentially with backoff
2. If osquery instance launch fails, also consider triggering an autoupdate check for osquery
3. Runner should still be responsive to calls to `Shutdown`

PR #2006, opened 2024-12-17
Check for and log information about stale osquery database lock files

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/2004.

launcher is currently completely unable to recover if the osquery lock file is left behind. However, we want to be cautious about removing this file. For now, we detect whether there's a stale lock file, and collect + log information about it if so.

Also adds a test for launching an instance, plus a test for detecting lock file presence.


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -249,6 +249,12 @@ func (i *OsqueryInstance) Launch() error {
 		return fmt.Errorf("could not calculate osquery file paths: %w", err)
 	}
 
+	// Make sure no lock files exist that could prevent osquery from starting up
+	if err := i.cleanUpOldDatabaseLock(ctx, paths); err != nil {
```
James-Pickett, line 253: probably want to do this in a loop, windows can be so finicky about deleting files in use

James-Pickett (dismissed):

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: Oh yes, thank you! I'll move this to a backoff/retry

James-Pickett (dismissed):

zackattack01 (approved):

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -702,6 +708,45 @@ func calculateOsqueryPaths(rootDirectory string, registrationId string, runId st
...
+func (i *OsqueryInstance) cleanUpOldDatabaseLock(ctx context.Context, paths *osqueryFilePaths) error {
+	lockFilePath := filepath.Join(paths.databasePath, "LOCK")
+
+	lockFileInfo, err := os.Stat(lockFilePath)
+	if os.IsNotExist(err) {
+		// No lock file, nothing to do here
+		return nil
+	}
+
+	// We do a couple retries here -- Windows complains about removing files in use sometimes.
```
directionless, line 723: My hunch is that this is too aggressive -- if windows is blocking you from removing the file. the osquery probably has it open. I wonder if there's a way we can detect whether osquery is running, or what the state here really is. 

I think forcing the file out, with osquery running will lead to corrupt databases

directionless (changes requested):

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: I'd had another idea for handling this --

> The osquery instance and osquery runner both currently have no visibility into the osquery logs -- they're processed by the log adapter, which is entirely separate. However, here and previously (when trying to figure out how to handle launcher falling back to a old version of osquery that is incompatible with the database -- some details https://github.com/kolide/launcher/issues/1937) we've tentatively wanted the osquery instance to be able to respond to particular logs from osquery. Do we want to move log processing into the osquery instance? Or open up some line of communication between the instance and the log adapter?

We could do something like:

- detect via logs that the lock file is in use (like we do when the pidfile is in use)
- gather information about what's using the lock file (like we do when the pidfile is in use)
- as long as it's a dead process, remove the lock file

What do you think?

RebeccaMahany: Actually, I could probably update this PR to do the second and third bullet points as-is, without having to mess around w the log adapter 🤔

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -702,6 +712,79 @@
...
+		processStrs[i] = processStr(ctx, p)
+	}
+	infoToLog = append(infoToLog, "processes", processStrs)
+
+	return true, nil
+}
+
+func processStr(ctx context.Context, p *process.Process) string {
+	name := "unknown"
+	username := "unknown"
```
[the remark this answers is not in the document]
RebeccaMahany: These are not hardcoded credentials; this is a default display value that we log if we cannot determine the username that owns a given process. This value is only used for logging purposes.

James-Pickett (dismissed):

James-Pickett (dismissed):

pkg/osquery/runtime/runtime_helpers_windows.go
```diff
@@ -80,3 +82,39 @@ func isExitOk(err error) bool {
 	}
 	return false
 }
+
+func getProcessesHoldingFile(ctx context.Context, pathToFile string) ([]*process.Process, error) {
```
directionless, line 86: I think this is a fairly expensive call. Might be okay, given how infrequent it occurs. But we should know that it's not cheap

RebeccaMahany: It takes about 6 seconds in test, which is definitely not nothing. But we'll only hit this if the lock file exists -- and in that case, we'll never successfully start up osquery anyway, so the delay is a moot point. I figure we can remove this once we have more info from the logs to figure out how to detect/remediate.

zackattack01 (dismissed):

pkg/osquery/runtime/runtime_helpers.go
```diff
@@ -33,3 +40,41 @@ func platformArgs() []string {
 func isExitOk(_ error) bool {
 	return false
 }
+
+func getProcessesHoldingFile(ctx context.Context, pathToFile string) ([]*process.Process, error) {
+	cmd, err := allowedcmd.Lsof(ctx, "-t", pathToFile)
```
directionless, line 45: I wish we had a better option, but I concur we do not.

pkg/osquery/runtime/runtime_helpers_windows.go
directionless: Yeah...

directionless (approved):

PR #2007, opened 2024-12-17
Make osq runner responsive to registration updates

zackattack01:
- rework interfaces for runner change detection


pkg/osquery/runtime/runner.go
```diff
@@ -49,6 +53,31 @@ func New(k types.Knapsack, serviceClient service.KolideService, opts ...OsqueryI
 }
 
 func (r *Runner) Run() error {
+	for {
+		// if our instances ever exit unexpectedly, return immediately
```
RebeccaMahany, line 57: I think this comment comment isn't accurate? I think `runRegisteredInstances` only returns if a shutdown was requested. And it only returns an error if a) shutdown was requested and b) we were trying to restart one or more instances during that time and c) hadn't successfully restarted one of them yet.

Either way, why would we return the error here instead of checking `rerunRequired` first?

pkg/osquery/runtime/runner.go
```diff
@@ -333,3 +368,42 @@ func (r *Runner) InstanceStatuses() map[string]types.InstanceStatus {
...
+	}
+
+	r.slogger.Log(context.TODO(), slog.LevelDebug,
+		"detected changes to registrationIDs, will restart runner instances",
+		"previous_registration_ids", existingRegistrationIDs,
+		"new_registration_ids", newRegistrationIDs,
+	)
+
+	// we know there are changes, safe to update the internal registrationIDs now
+	r.registrationIds = newRegistrationIDs
```
RebeccaMahany, line 394: Do we need a lock on `r.registrationIds` to avoid a data race?

pkg/osquery/runtime/runner.go
zackattack01: that's a good point thank you. we'd probably want to rerun regardless if required, I will update that comment and get this fixed up!

pkg/osquery/runtime/runner.go
zackattack01: that seems like a good idea, will do!

pkg/osquery/runtime/runner.go
```diff
@@ -27,14 +28,16 @@ type settingsStoreWriter interface {
...
+	regIDLock       sync.Mutex                  // locks access to registrationIds
 	instances       map[string]*OsqueryInstance // maps registration ID to currently-running instance
 	instanceLock    sync.Mutex                  // locks access to `instances` to avoid e.g. restarting an instance that isn't running yet
 	slogger         *slog.Logger
 	knapsack        types.Knapsack
 	serviceClient   service.KolideService   // shared service client for communication between osquery instance and Kolide SaaS
 	settingsWriter  settingsStoreWriter     // writes to startup settings store
 	opts            []OsqueryInstanceOption // global options applying to all osquery instances
-	shutdown        chan struct{}
+	shutdown        chan struct{}           // buffered shutdown channel for to enable shutting down to restart or exit
```
RebeccaMahany, line 39: Nit -- think you have an extra word in here from rewriting this comment --

```suggestion
	shutdown        chan struct{}           // buffered shutdown channel to enable shutting down to restart or exit
```

pkg/osquery/runtime/runner.go
```diff
@@ -57,12 +61,56 @@ func New(k types.Knapsack, serviceClient service.KolideService, settingsWriter s
...
+// passed to mocked knapsack calls. when we AssertExpectations, the runner struct is traversed by the
+// Diff logic inside testify. This causes data races to be incorrectly reported for structs containing mutexes-
+// the second read is coming from testify.
+// see (one of) the issues here for additional context https://github.com/stretchr/testify/issues/1597
+// If we really needed to expose more here, we could acquire all locks and return fmt.Sprintf("%#v", r). but given
+// that we do not, it seems safer to avoid introducing any additional lock contention in case our Stringer call
+// is invoked someday in a production flow
+func (r *Runner) String() string {
+	return "runtime.Runner{}"
+}
```
RebeccaMahany, line 74: 😂 / 😭 thank you for figuring this out, fixing it, and documenting it

pkg/osquery/runtime/runner.go
```diff
@@ -367,3 +427,47 @@ func (r *Runner) InstanceStatuses() map[string]types.InstanceStatus {
...
+
+	if slices.Equal(newRegistrationIDs, existingRegistrationIDs) {
+		r.slogger.Log(context.TODO(), slog.LevelDebug,
+			"skipping runner restarts for updated registration IDs, no changes detected",
+		)
+
+		return nil
+	}
+
+	r.slogger.Log(context.TODO(), slog.LevelDebug,
```
RebeccaMahany, line 449: Nit -- I think reasonable to log this at the info level (so that we ship this log to the cloud)

pkg/osquery/runtime/runner.go
```diff
@@ -205,6 +253,13 @@ func (r *Runner) Query(query string) ([]map[string]string, error) {
 }
 
 func (r *Runner) Interrupt(_ error) {
+	if r.interrupted.Load() {
```
RebeccaMahany, line 256: Do you think it'd be useful to also do `r.rerunRequired.Store(false)` here? I'm thinking about the case where `UpdateRegistrationIDs` is called and then `Interrupt` is called while the shutdown/restart is ongoing. Could maybe add a test case for this as well?

pkg/osquery/runtime/runner.go
```diff
@@ -27,14 +28,16 @@ type settingsStoreWriter interface {
...
 	instances       map[string]*OsqueryInstance // maps registration ID to currently-running instance
 	instanceLock    sync.Mutex                  // locks access to `instances` to avoid e.g. restarting an instance that isn't running yet
 	slogger         *slog.Logger
 	knapsack        types.Knapsack
 	serviceClient   service.KolideService   // shared service client for communication between osquery instance and Kolide SaaS
 	settingsWriter  settingsStoreWriter     // writes to startup settings store
 	opts            []OsqueryInstanceOption // global options applying to all osquery instances
-	shutdown        chan struct{}
+	shutdown        chan struct{}           // buffered shutdown channel for to enable shutting down to restart or exit
+	rerunRequired   atomic.Bool
```
RebeccaMahany, line 40: Maybe add a quick comment here? `// if set, the runner will restart all instances after Shutdown is called, instead of exiting` or similar

RebeccaMahany (commented):
This makes sense to me! Really nice.

pkg/osquery/runtime/runner.go
zackattack01: oh good call, will do!

pkg/osquery/runtime/runner.go
zackattack01: ohh yeah that seems wise. will do both!

pkg/osquery/runtime/runner.go
```diff
@@ -27,14 +28,16 @@ type settingsStoreWriter interface {
 
 type Runner struct {
 	registrationIds []string                    // we expect to run one instance per registration ID
+	regIDLock       sync.Mutex                  // locks access to registrationIds
```
James-Pickett, line 31: I discovered that if we are locking for a single value, there is an `atmoic.Value` or `atmoic.Pointer` that can be used, like:

```go
	var registrationIds atomic.Value
	registrationIds.Store([]string{"a", "b", "c"})
	theIds := registrationIds.Load().([]string)
```
 
or

```go
	var registrationIds atomic.Pointer[[]string]
	registrationIds.Store(&[]string{"a", "b", "c"})
	theIds := registrationIds.Load()
```

the internets generally say that atmoics are faster ... but I doubt that it makes any noticeable difference for us, also pointer to an array feels weird and I'm not sure of the implications

feel free to not act on this, I don't know if it's any better

pkg/osquery/runtime/runner.go
zackattack01: ah yep! thank you

pkg/osquery/runtime/runner.go
```diff
@@ -218,14 +273,12 @@ func (r *Runner) Interrupt(_ error) {
...
 	defer span.End()
-
-	if r.interrupted.Load() {
-		// Already shut down, nothing else to do
-		return nil
+	// ensure one shutdown is sent for each instance to read
+	r.instanceLock.Lock()
+	for range r.instances {
+		r.shutdown <- struct{}{}
 	}
```
James-Pickett, line 280: what about giving each instance it's own shutdown channel to avoid the buffered channel?

```suggestion
	for instance := range r.instances {
	        close(instance.shutdown)
	}
```

zackattack01: oh yeah i'm gonna give this a try, ty! I'm actually seeing some interesting behavior while trying to write a test for Interrupt during restart per Becca's comment [here](https://github.com/kolide/launcher/pull/2007#discussion_r1932430973) and I wonder if this wouldn't clear things up a little. the behavior I'm seeing in the test is that the interrupt can end up ignored if timed correctly and I think It's related to the buffered channel use so far

PR #2008, opened 2024-12-18
drop reqs when detection locked

James-Pickett:
* try lock before presence detection, drop if can't get lock
* fix bug where desktop runner returned 0s on error
* drop call to ls.knapsack.InstanceStatuses() in `/id` endpoint that was causing request to be slow immediately after start up

### DIFF
--- a/ee/localserver/server.go
+++ b/ee/localserver/server.go
@@ -419,10 +419,6 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
 
 func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// ensure we only prompt for 1 presence detection at a time
-		ls.presenceDetectionMutex.Lock()
-		defer ls.presenceDetectionMutex.Unlock()
-
 		// can test this by adding an unauthed endpoint to the mux and running, for example:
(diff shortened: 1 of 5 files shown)


ee/localserver/server.go
```diff
@@ -461,6 +457,12 @@ func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler
 			)
 		}
 
+		if !ls.presenceDetectionMutex.TryLock() {
+			http.Error(w, "presence detection already in progress", http.StatusTooManyRequests)
+			return
+		}
+		defer ls.presenceDetectionMutex.Unlock()
```
James-Pickett, line 464: Should we not defer this and just unlock right after call to DetectPresence? I don't think it will unlock until request complete since the func is calling next instead of returning. But maybe it's better to let the whole request complete.

RebeccaMahany: I think the defer is fine!

RebeccaMahany (approved):

ee/localserver/server.go
```diff
@@ -461,6 +457,12 @@ func (ls *localServer) presenceDetectionHandler(next http.Handler) http.Handler
 			)
 		}
 
+		if !ls.presenceDetectionMutex.TryLock() {
+			http.Error(w, "presence detection already in progress", http.StatusTooManyRequests)
```
directionless, line 461: Should this be an error? Maybe?

James-Pickett: I think it's best to leave it as an error for because if we got here, the request is asking for presence detection which we cannot give at this time.

RebeccaMahany (approved):

PR #2017, opened 2024-12-30
Improvements for osqueryinstance's errgroup

RebeccaMahany:
- Moves errgroup to its own package for reusability
- Adds tests for errgroup
- Adds more and better logging so that we can track whether shutdown tasks are taking a long time
- Run repeated goroutine immediately after delay, rather than waiting for another interval
- Recover panicking goroutines


ee/errgroup/errgroup.go
```diff
@@ -0,0 +1,163 @@
...
+	"time"
+
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
+type LoggedErrgroup struct {
+	errgroup *errgroup.Group
+	cancel   context.CancelFunc
+	doneCtx  context.Context // nolint:containedctx
```
RebeccaMahany, line 16: This is the same nolint exception we had previously, just moved from osqueryinstance to here. In the future I think we might be able to get rid of this entirely -- but the errgroup is very sensitive when it comes to updating anything that touches `doneCtx` so I want to do it in a separate PR at a later time.

ee/errgroup/errgroup.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// StartGoroutine starts the given goroutine in the errgroup, ensuring that we log its start and exit.
+func (l *LoggedErrgroup) StartGoroutine(ctx context.Context, goroutineName string, goroutine func() error) {
+	l.errgroup.Go(func() (err error) {
+		slogger := l.slogger.With("goroutine_name", goroutineName)
+
+		// Catch any panicking goroutines and log them. We also want to make sure
+		// we return an error from this goroutine overall if it panics.
+		defer func() {
```
James-Pickett, line 43: do we want to add the same or similar defer func to catch panics to StartRepeatedGoroutine and AddShutdownGoroutine?

RebeccaMahany: They both call `StartGoroutine` so I think it should be covered -- I will add tests to confirm!

James-Pickett: ohh woops, duh, I see it now =) thx

James-Pickett (dismissed):

zackattack01 (dismissed):
🔥 🔥 🔥

ee/errgroup/errgroup.go
RebeccaMahany: It does work -- however, I noticed that setting the error after recovering from the panic is not what we actually want to do during the shutdown goroutine. I will get that fixed up and then push both tests. 🙂

RebeccaMahany: Updated!

James-Pickett (approved):

PR #2027, opened 2025-01-03
Revert "Switch from CoInitialize to CoInitializeEx"

RebeccaMahany:
Reverts kolide/launcher#1973

We think it's possible this is causing launcher to crash on first-time startup. Reverting to check that theory out.


directionless (approved):

directionless:
If you want to be fancy,. you could probably make this dynamic. But 🤷

James-Pickett (approved):

PR #2033, opened 2025-01-07
Clean up a couple places not yet using gowrapper

RebeccaMahany:
Found a couple places that didn't match our `go func` pattern, so golangci-lint didn't catch them. Cleaning them up for now without adding a new golangci-lint rule.

### DIFF
--- a/cmd/launcher/svc_windows.go
+++ b/cmd/launcher/svc_windows.go
@@ -175,7 +175,9 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
 
 	// Confirm that service configuration is up-to-date
-	go checkServiceConfiguration(w.slogger.Logger, w.opts)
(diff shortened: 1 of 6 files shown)


cmd/launcher/svc_windows.go
```diff
@@ -175,7 +175,9 @@ func (w *winSvc) Execute(args []string, r <-chan svc.ChangeRequest, changes chan
 	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
 
 	// Confirm that service configuration is up-to-date
-	go checkServiceConfiguration(w.slogger.Logger, w.opts)
+	gowrapper.Go(ctx, w.systemSlogger.Logger, func() {
```
zackattack01, line 178: oh great find!!

zackattack01 (approved):

cmd/launcher/svc_windows.go
RebeccaMahany: 🤞 🤞 trying not to get my hopes up lol

PR #2035, opened 2025-01-08
Always log stats for distributed queries

RebeccaMahany:
We would always like to log stats for distributed queries.

### DIFF
--- a/pkg/service/publish_results.go
+++ b/pkg/service/publish_results.go
@@ -223,12 +223,8 @@ func (mw logmw) PublishResults(ctx context.Context, nodeKey string, results []di
 				continue
 			}
 
-			// Log queries that took more than 5 seconds
-			if r.QueryStats.WallTimeMs < 5000 {
-				continue
-			}
-			mw.knapsack.Slogger().Log(ctx, slog.LevelWarn,
-				"noticed long-running query",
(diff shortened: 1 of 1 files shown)


pkg/service/publish_results.go
```diff
@@ -223,12 +223,8 @@ func (mw logmw) PublishResults(ctx context.Context, nodeKey string, results []di
 				continue
 			}
 
-			// Log queries that took more than 5 seconds
-			if r.QueryStats.WallTimeMs < 5000 {
-				continue
-			}
-			mw.knapsack.Slogger().Log(ctx, slog.LevelWarn,
-				"noticed long-running query",
```
zackattack01, line 231: do we still find the long running piece to be useful? just wondering if adding a `"long_running": r.QueryStats.WallTimeMs > 5000` would make spotting these a little easier for debugging

zackattack01 (dismissed):

James-Pickett (dismissed):

pkg/service/publish_results.go
RebeccaMahany: Oh yeah, good point, I was thinking about cloud log and not scanning a flare. I'll update!

RebeccaMahany: Done! Thank you

zackattack01 (approved):

PR #2038, opened 2025-01-09
More flaky test improvements

RebeccaMahany:
More attempts to get the osquery tests in a more reasonable state:

* If the instance doesn't become healthy by the timeout, check to see if the osquery process is in a running state or not
* Configure the device server service client mock to return a config and distributed queries, in case the error responses or missing configs was affecting the test osquery process
* Retry removing root dir during test cleanup

### DIFF
--- a/pkg/osquery/runtime/runtime_test.go
+++ b/pkg/osquery/runtime/runtime_test.go
@@ -743,7 +743,11 @@ func testRootDirectory(t *testing.T) string {
 	require.NoError(t, os.Mkdir(rootDir, 0700))
 
 	t.Cleanup(func() {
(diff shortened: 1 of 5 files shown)


pkg/osquery/runtime/runtime_test.go
```diff
@@ -236,9 +254,11 @@ func TestFlagsChanged(t *testing.T) {
 
 	startingInstance := runner.instances[types.DefaultRegistrationID]
 
+	// Now, WatchdogEnabled should return false
```
zackattack01, line 257:
```suggestion
	// Now, WatchdogEnabled should return true
```

zackattack01 (approved):

PR #2044, opened 2025-01-13
Shut down instances that fail to launch successfully

RebeccaMahany:
Previously, we would not shut down instances that only partially started up -- either because `Launch` returned an error, or because we received a shutdown request during instance startup. This PR corrects that issue. I am hoping it will help us with https://github.com/kolide/launcher/issues/2004.


James-Pickett (approved):
nice

zackattack01 (approved):

Issue #2050, opened 2025-01-16
Add timeout to osqueryinstance healthchecks


zackattack01:
Some logs today showed the following on a windows device running launcher v1.12.3:
- all osqueryinstance errgroup routines starting correctly
- eventually, osquery stops logging and localserver started logging the following errors:
  - `"msg":"updating id fields","component":"localserver","err":"id query failed: could not query the extension manager client: i/o timeout"`
- this state continued for days, without the healthcheck routine triggering an osquery restart (or logging any sort of failure)
- when a shutdown was finally issued, the healthcheck goroutine never returned

I think the most likely explanation here is that it got caught up somehow, and that we should enforce some sort of cancellation on our `Healthy()` calls. This way we will at least log the failures and restart to eventually recover from this state

PR #2051, opened 2025-01-17
add timeout to osqueryinstance Healthy calls

zackattack01:
closes https://github.com/kolide/launcher/issues/2050 (see issue for additional context)
I'm not exactly sure what the best timeout value would be, ten seconds seemed safe enough. the point here is only to prevent the healthchecking goroutine from becoming hung indefinitely

### DIFF
--- a/pkg/osquery/runtime/osqueryinstance.go
+++ b/pkg/osquery/runtime/osqueryinstance.go
@@ -162,7 +162,11 @@ func (i *OsqueryInstance) Healthy() error {
 	// Wait until we either receive an error or nil result from the healthcheck goroutine, or exceed our timeout threshold
 	select {
 	case maybeErr := <-resultsChan:
-		return maybeErr
+		if maybeErr != nil {
+			return fmt.Errorf("encountered error during healthcheck: %w", maybeErr)
+		}
+
+		return nil
 	case <-time.After(healtcheckTimeout):
 		return fmt.Errorf("osqueryinstance healthcheck exceeded timeout of %s", healtcheckTimeout.String())
 	}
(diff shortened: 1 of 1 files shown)


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -71,6 +71,9 @@ const (
 
 	// The maximum amount of time to wait for the osquery socket to be available -- overrides context deadline
 	maxSocketWaitTime = 30 * time.Second
+
+	// How long to wait for a single osqueryinstance healthcheck before forcibly returning error
+	healtcheckTimeout = 10 * time.Second
```
RebeccaMahany, line 76: Nit -- typo in `healthcheckTimeout` 🙂

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -129,31 +132,44 @@ func (i *OsqueryInstance) Healthy() error {
...
-				serverStatus.Message)
 
+		clientStatus, err := i.extensionManagerClient.Ping()
+		if err != nil {
+			resultsChan <- fmt.Errorf("could not ping osquery extension client: %w", err)
+			return
+		}
+		if clientStatus.Code != 0 {
+			resultsChan <- fmt.Errorf("ping extension client returned %d: %s", clientStatus.Code, clientStatus.Message)
+			return
```
RebeccaMahany, line 156: Do you think it'd be worth putting this in a separate `gowrapper.Go` call for increased speed/parallelization, or nah?

RebeccaMahany (commented):
Nice!!

pkg/osquery/runtime/osqueryinstance.go
zackattack01: so i got the impression (but totally could have misunderstood) that the Pings from `i.extensionManagerServers` should be extremely quick and hadn't really considered further optimization here since we'll have to wait for this anyway. i'm happy to break out if you think it would add value, it only seems slightly more complicated to me at first glance

RebeccaMahany: Seems fine to leave it as-is!

RebeccaMahany (approved):

James-Pickett (approved):

PR #2061, opened 2025-01-23
rough in of cmd tracing

James-Pickett:


ee/tables/pwpolicy/pwpolicy_test.go
```diff
@@ -71,8 +72,10 @@ func TestQueries(t *testing.T) {
 
 }
 
-func execFaker(filename string) func(context.Context, ...string) (*exec.Cmd, error) {
-	return func(ctx context.Context, _ ...string) (*exec.Cmd, error) {
-		return exec.CommandContext(ctx, "/bin/cat", filename), nil //nolint:forbidigo // Fine to use exec.CommandContext in test
+func execFaker(filename string) func(context.Context, ...string) (*allowedcmd.TracedCmd, error) {
```
James-Pickett, line 75: the method is the reason I had to export `allowedcmd.TracedCmd`

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,55 @@ import (
...
-type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
+type AllowedCommand func(ctx context.Context, arg ...string) (*TracedCmd, error)
+
+type TracedCmd struct {
+	*exec.Cmd
+}
+
+// Start overrides the Start method to add tracing before executing the command.
+func (t *TracedCmd) Start() error {
+	_, span := traces.StartSpan(context.Background(), "path", t.Cmd.Path)
```
James-Pickett, line 28: maybe?

```suggestion
	_, span := traces.StartSpan(context.Background(), "path", t.Cmd.Path, "args", t.Cmd.Args)
```

RebeccaMahany, line 28: Since `exec.Cmd.ctx` isn't exposed, I think you'd want to add the `ctx` into `TracedCommand` struct so that you can use the ctx here too and make sure that this span attaches to any existing parent spans

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,55 @@ import (
...
 	"path/filepath"
+
+	"github.com/kolide/launcher/pkg/traces"
 )
 
-type AllowedCommand func(ctx context.Context, arg ...string) (*exec.Cmd, error)
+type AllowedCommand func(ctx context.Context, arg ...string) (*TracedCmd, error)
+
+type TracedCmd struct {
+	*exec.Cmd
```
James-Pickett, line 23: one flaw here is that someone could just say `TracedCmd.Cmd.Run()` and bypass the func override, considered making it a private var, but there are a lot of places where we do need to access the embedded `*exec.Cmd` to add Env, Std in/out/err pipes, etc and it felt messy to try and implement all those funcs ... but make worth it?

ee/allowedcmd/cmd.go
RebeccaMahany: Maybe the other option would be to update these method signatures to e.g. `Start(ctx context.Context)`? I don't think the embedded ctx is bad here either, though, since it's just following what exec.Cmd also does

RebeccaMahany: I think would want to `fmt.Sprintf("%+v", t.Cmd.Args)` -- our trace frontend doesn't allow for array attributes unfortunately

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,55 @@ import (
...
+}
+
+// Start overrides the Start method to add tracing before executing the command.
+func (t *TracedCmd) Start() error {
```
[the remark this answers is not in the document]
James-Pickett: guess it doesn't hurt

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,59 @@ import (
...
+
+	return t.Cmd.Start() //nolint:forbidigo // This is our approved usage of t.Cmd.Start()
+}
+
+// Run overrides the Run method to add tracing before running the command.
+func (t *TracedCmd) Run() error {
+	_, span := traces.StartSpan(t.Ctx, "path", t.Cmd.Path, "args", fmt.Sprintf("%+v", t.Cmd.Args))
+	defer span.End()
+
+	return t.Cmd.Run() //nolint:forbidigo // This is our approved usage of t.Cmd.Start()
```
RebeccaMahany, line 40: Nitpick for the comment --

```suggestion
	return t.Cmd.Run() //nolint:forbidigo // This is our approved usage of t.Cmd.Run()
```

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,59 @@ import (
...
+}
+
+// Output overrides the Output method to add tracing before capturing output.
+func (t *TracedCmd) Output() ([]byte, error) {
+	_, span := traces.StartSpan(t.Ctx, "path", t.Cmd.Path, "args", fmt.Sprintf("%+v", t.Cmd.Args))
+	defer span.End()
+
+	return t.Cmd.Output() //nolint:forbidigo // This is our approved usage of t.Cmd.Start()
```
RebeccaMahany, line 48: (Ditto)

```suggestion
	return t.Cmd.Output() //nolint:forbidigo // This is our approved usage of t.Cmd.Output()
```

ee/allowedcmd/cmd.go
```diff
@@ -13,17 +13,59 @@ import (
...
+}
 
-func newCmd(ctx context.Context, fullPathToCmd string, arg ...string) *exec.Cmd {
-	return exec.CommandContext(ctx, fullPathToCmd, arg...) //nolint:forbidigo // This is our approved usage of exec.CommandContext
+// CombinedOutput overrides the CombinedOutput method to add tracing before capturing combined output.
+func (t *TracedCmd) CombinedOutput() ([]byte, error) {
+	_, span := traces.StartSpan(t.Ctx, "path", t.Cmd.Path, "args", fmt.Sprintf("%+v", t.Cmd.Args))
+	defer span.End()
+
+	return t.Cmd.CombinedOutput() //nolint:forbidigo // This is our approved usage of t.Cmd.Start()
```
RebeccaMahany, line 56: (Ditto)

```suggestion
	return t.Cmd.CombinedOutput() //nolint:forbidigo // This is our approved usage of t.Cmd.CombinedOutput()
```

ee/allowedcmd/cmd_darwin.go
```diff
@@ -6,148 +6,147 @@ package allowedcmd
...
-func Brew(ctx context.Context, arg ...string) (*exec.Cmd, error) {
+func Brew(ctx context.Context, arg ...string) (*TracedCmd, error) {
 	for _, p := range []string{"/opt/homebrew/bin/brew", "/usr/local/bin/brew"} {
 		validatedCmd, err := validatedCommand(ctx, p, arg...)
 		if err != nil {
 			continue
 		}
 
-		validatedCmd.Env = append(validatedCmd.Environ(), "HOMEBREW_NO_AUTO_UPDATE=1")
+		validatedCmd.Env = append(validatedCmd.Env, "HOMEBREW_NO_AUTO_UPDATE=1")
```
RebeccaMahany, line 30: I think this will change the behavior, resulting in a different set of vars in `validatedCmd.Env` than before. (If `cmd.Env` is nil when a command starts running, then the command will automatically add in `os.Environ`. If anything is set in `cmd.Env` prior to command running, it won't add in `os.Environ`.) So before, `validatedCmd.Env` contained both `os.Environ` plus `HOMEBREW_NO_AUTO_UPDATE=1`, and now it will only contain `HOMEBREW_NO_AUTO_UPDATE=1`. I don't know if that matters for this command or not, but maybe it would be safer to add an `Environ()` wrapper command to `TracedCmd` too?

RebeccaMahany (commented):
This is really nice! One small comment about `cmd.Environ`, otherwise LGTM!

ee/allowedcmd/cmd.go
James-Pickett: fixed all the comments

RebeccaMahany (approved):
🔥

zackattack01 (approved):

Issue #2057, opened 2025-01-22
Sometimes we can't write to the startup settings database


RebeccaMahany:
Sometimes we see errors that look like -- `writing settings: beginning transaction: sql: database is closed`.

We should look into these and fix them if possible.

PR #2067, opened 2025-01-27
Reuse startup settings writer for osquery extension instead of creating a new one for each new config

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2057

Previously, we were opening a new startup settings writer every time we needed to write a new config inside the osquery extension. Creating a new startup settings writer will _also_ register it as a change observer with the knapsack. However, the extension then closes the writer. This means the knapsack will try to notify this _closed_ change observer on subsequent `FlagsChanged` events, which will then result in `"msg":"writing startup settings after flag change","err":"writing settings: beginning transaction: sql: database is closed"` errors in the logs.

This PR updates the extension to use a preexisting startup settings writer (which it is not responsible for closing), which should fix the error.

The extension appears to be the only place in the codebase that does this, so I _think_ this PR should completely close the linked issue.


zackattack01 (approved):
nice find!

James-Pickett (approved):

PR #2068, opened 2025-01-27
Generate enrollment details on launcher startup

### LINKED ISSUE #1986: Include enrollment details in `/id` response
### Background

After launcher is first installed and starts up, it must enroll with K2. It generates "enrollment details" that identify the device -- for example, the device's OS and serial number. It includes these enrollment details in its enrollment request to K2's device server. K2 uses these details to create a `device` corresponding to this launcher installation.

In the future, we want enrollment to happen at a different point in time -- during device trust. (This will enable both secretless launcher installations and multiple enrollments.) In preparation for this work, we want to start including the enrollment details in launcher's localserver response to the `/id` endpoint.

The enrollment details should not change over time, so we should be able to generate them once on launcher startup, cache them, and then refer to the cached data when constructing the response to the `/id` endpoint.

### Requirements

- [ ] Generate the enrollment details on launcher startup, and cache them.
- [ ] Update the enrollment details after autoupdate happens.
- [ ] Since the enrollment details are cached, we don't need to generate them anew during enrollment -- update enrollment to use the cached enrollment details. (Nothing else needs to change about enrollment at this time, though -- launcher should still perform enrollment as expected.)
- [ ] Include enrollment details in every response to the `/id` endpoint. The K2 side to consume this data hasn't been implemented yet, so there isn't a hard requirement for how this data should be structured. It is probably reasonable for it to exist at the top level of the response (same as `RequestId` and `Status`), and for the data format to remain the same as it currently is defined in `service.EnrollmentDetails`.

### Relevant code links

- Generation of enrollment details: https://github.com/kolide/launcher/blob/main/pkg/osquery/enrollment_details.go
- `/id` endpoint: https://github.com/kolide/launcher/blob/main/ee/localserver/request-id.go

cesarfda:
This pull request introduces several changes to the `osquery` and `launcher` package to handle enrollment details.

- enrollment details are now getting triggered from runLauncher
- details are stored in the knapsack
- extension will poll until details are ready for enrollment
- if osquery details fails, extension will enroll with only runtime details to avoid unknown device type error
- osquery version gets refreshed anytime we try to get the details from knapsack, avoiding changes to AutoUpdater to cover the version update for the details.

 Closes #1986 

### Enrollment Details Handling:
* [`cmd/launcher/launcher.go`](diffhunk://#diff-dd532c327d56a5d3623a61b7983e174e88bf93ba87f785ddfe9811a8029a9a63R210-R211): Added a call to `osquery.CollectAndSetEnrollmentDetails` to collect and set enrollment details during the launcher run.
* [`ee/agent/knapsack/knapsack.go`](diffhunk://#diff-1fef7e482dbae76c26b95e73faa3353d09c1e04660f7156238bd991dd1ea004fR245-R278): Added methods `SetEnrollmentDetails` and `GetEnrollmentDetails` to manage enrollment details.
* [`ee/agent/types/enrollment.go`](diffhunk://#diff-1b0b19700142d01b37c6bf68a1609e885e7762ab005c35c1dc21155a36105a72R1-R32): Introduced `EnrollmentDetails` struct constants to encapsulate enrollment information.

### Interface and Mock Updates:
* [`ee/agent/types/knapsack.go`](diffhunk://#diff-2a75ee6ccb000780eb0256da925db3dd048db15d389e2d354ba3bd41c735d2beR23-R26): Updated `Knapsack` interface to include `GetEnrollmentDetails` and `SetEnrollmentDetails` methods.
* [`ee/agent/types/mocks/knapsack.go`](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1L40-R52): Updated mock functions to support new `KVStore` type and added mocks for `GetEnrollmentDetails` and `SetEnrollmentDetails`. [[1]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1L40-R52) [[2]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1R529-R556) [[3]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1R1488-R1505) [[4]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1L1987-R2065)

### Interface Updates:
* [`ee/agent/types/knapsack.go`](diffhunk://#diff-2a75ee6ccb000780eb0256da925db3dd048db15d389e2d354ba3bd41c735d2beR23-R26): Updated the `Knapsack` interface to include `GetEnrollmentDetails` and `SetEnrollmentDetails` methods.

### Miscellaneous:
* [`ee/agent/types/status.go`](diffhunk://#diff-89638e7ad81072661e4998fcfa7d6aef2e7270fd77a80d4c92502d256d74861aL1-L10): Removed redundant `EnrollmentStatus` type definitions.
* Updated mockery version in generated mock files. [[1]](diffhunk://#diff-c0d4e1b9c036b43e815c7ab1c26f8bb4f14023b0606127751fed7f04a2940e34L1-R1) [[2]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1L1-R1)


pkg/osquery/enrollment_details.go
```diff
@@ -145,3 +148,34 @@ func getOsqEnrollDetails(ctx context.Context, osquerydPath string, details *serv
...
+	k.SetEnrollmentDetails(details)
+
+	if err := backoff.WaitFor(func() error {
+		err := getOsqEnrollDetails(ctx, latestOsquerydPath, &details)
+		if err != nil {
+			span.AddEvent("failed to get enrollment details")
+		}
+		return err
+	}, collectTimeout, collectRetryInterval); err != nil {
+		span.AddEvent(fmt.Sprintf("enrollment details collection failed: %v", err))
```
RebeccaMahany, line 177: Should we return an error here?

pkg/osquery/enrollment_details_test.go
```diff
@@ -34,3 +38,40 @@ func Test_getEnrollDetails_executionError(t *testing.T) {
...
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "no osqueryd path")
+}
+
+func TestCollectAndSetEnrollmentDetails_Success(t *testing.T) {
+	t.Parallel()
+
+	k := typesmocks.NewKnapsack(t)
+
+	expectedOsquerydPath := "/usr/local/bin/osqueryd"
```
RebeccaMahany, line 60: Oh hm, this isn't the path I would expect for a test binary that we downloaded inside a GitHub runner, and I wouldn't expect it to work for a Windows runner. I think the test is only passing in CI because we don't return a failure currently from `CollectAndSetEnrollmentDetails` if we fail to query osquery.

We could download an osquery binary for this test with `packaging.FetchBinary`, like how `pkg/osquery/interactive/interactive_test.go` does it. I'm not sure whether it's worth it.

pkg/osquery/extension.go
```diff
@@ -426,38 +425,35 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
-			}
-			return err
-		}, 30*time.Second, 5*time.Second); err != nil {
-			if os.Getenv("LAUNCHER_DEBUG_ENROLL_DETAILS_REQUIRED") == "true" {
-				return "", true, fmt.Errorf("query osq enrollment details: %w", err)
-			}
+	var enrollDetails types.EnrollmentDetails
+	// Poll for complete enrollment details
+	pollTimeout := time.NewTimer(60 * time.Second)
+	defer pollTimeout.Stop()
```
RebeccaMahany, line 431: I don't think this is in use anymore

pkg/osquery/extension.go
```diff
@@ -426,38 +425,35 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
 
-			e.slogger.Log(ctx, slog.LevelError,
-				"failed to get osq enrollment details with retries, moving on",
-				"err", err,
-			)
-			traces.SetError(span, fmt.Errorf("query osq enrollment details: %w", err))
-		} else {
-			span.AddEvent("got_enrollment_details")
+	ticker := time.NewTicker(5 * time.Second)
+	defer ticker.Stop()
```
RebeccaMahany, line 434: Same -- I don't think this is in use since you switched back to using `backoff`

pkg/osquery/extension.go
```diff
@@ -426,38 +425,35 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+
+	// Replace the manual polling loop with backoff.WaitFor
+	err = backoff.WaitFor(func() error {
+		details, err := e.knapsack.GetEnrollmentDetails()
+		if err != nil {
+			traces.SetError(span, fmt.Errorf("error getting enrollment details: %w", err))
+			return err
+		}
+		if details.OSVersion == "" || details.Hostname == "" {
+			return fmt.Errorf("incomplete enrollment details")
```
RebeccaMahany, line 444: `fmt.Errorf` => `errors.New`

ee/agent/knapsack/knapsack.go
```diff
@@ -239,3 +242,42 @@ func (k *knapsack) CurrentEnrollmentStatus() (types.EnrollmentStatus, error) {
 
 	return types.Enrolled, nil
 }
+
+//Todo change where the enrollment details call is made
+// consider goroutine to unblock the startup of launcher
+// we can use the latest osquery version stored in the knapsack to add to the enrollment details when doing autoupdate
+
+func (k *knapsack) SetEnrollmentDetails(details types.EnrollmentDetails) error {
```
RebeccaMahany, line 250: We aren't ever returning an error here -- I think you could update this signature to not have a return

RebeccaMahany (commented):
Nice!

pkg/osquery/enrollment_details_test.go
cesarfda: I tried that in the commit history and downloading the binary worked for windows and linux but not for macos13 for some reason, it got really hairy for a unit test so I switched to just fake a osquery, but I'll do it with a mock so we can pass the err

pkg/osquery/enrollment_details.go
cesarfda: I removed the error from the signature because we just trigger the call from launcher and don't really evaluate the returned error. Let me know if you think we should keep the error return

pkg/osquery/enrollment_details.go
```diff
@@ -145,3 +148,31 @@ func getOsqEnrollDetails(ctx context.Context, osquerydPath string, details *serv
...
+			if err != nil {
+				span.AddEvent("failed to get enrollment details")
+			}
+			return err
+		}, collectTimeout, collectRetryInterval); err != nil {
+			span.AddEvent(fmt.Sprintf("enrollment details collection failed: %v", err))
+		}
+
+		k.SetEnrollmentDetails(details)
+	}
```
RebeccaMahany, line 177: Style nitpick -- we usually prefer to return early, i.e.:

```golang
if latestOsquerydPath == "" {
	return
}

if err := backoff.WaitFor(func() error {
...
```

pkg/osquery/enrollment_details.go
```diff
@@ -145,3 +148,31 @@ func getOsqEnrollDetails(ctx context.Context, osquerydPath string, details *serv
...
+
+	if latestOsquerydPath != "" {
+		if err := backoff.WaitFor(func() error {
+			err := getOsqEnrollDetails(ctx, latestOsquerydPath, &details)
+			if err != nil {
+				span.AddEvent("failed to get enrollment details")
+			}
+			return err
+		}, collectTimeout, collectRetryInterval); err != nil {
+			span.AddEvent(fmt.Sprintf("enrollment details collection failed: %v", err))
```
RebeccaMahany, line 173: Optional -- you can use `traces.SetError` to update the span with the error and error status -- `traces.SetError(span, fmt.Sprintf("enrollment details collection failed: %w", err))`

pkg/osquery/extension.go
```diff
@@ -426,38 +425,29 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+	var enrollDetails types.EnrollmentDetails
 
-			e.slogger.Log(ctx, slog.LevelError,
-				"failed to get osq enrollment details with retries, moving on",
-				"err", err,
-			)
-			traces.SetError(span, fmt.Errorf("query osq enrollment details: %w", err))
-		} else {
-			span.AddEvent("got_enrollment_details")
+	// Replace the manual polling loop with backoff.WaitFor
```
RebeccaMahany, line 430: I think this comment should be removed

pkg/osquery/enrollment_details_test.go
RebeccaMahany: Got it -- I would then either update the path (to something like `/some/fake/path/to/osqueryd`) or at least add a comment indicating we don't expect a real osquery binary at that path

RebeccaMahany (commented):
LGTM, just a couple small nitpicks!

RebeccaMahany (approved):
LGTM, nice work! (I don't know if @directionless wanted to review again before merge?)

ee/agent/knapsack/knapsack.go
```diff
@@ -239,3 +242,42 @@ func (k *knapsack) CurrentEnrollmentStatus() (types.EnrollmentStatus, error) {
...
+	if enrollmentDetails == nil {
+		newDetails := details
+		enrollmentDetails = &newDetails
```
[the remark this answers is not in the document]
James-Pickett: This actually let me down a bit of a rabbit hole to understand =) As long as the the types.EnrollmentDetails doesn't have points struct fields or arrays. You should need to do the extra assignments. See the example below.

```go
package main

import "fmt"

type EnrollmentDetails struct {
	OSVersion string `json:"os_version"`
}

var globalDetails *EnrollmentDetails

func main() {
	funcDetails := EnrollmentDetails{
		OSVersion: "1.0.0",
	}

	// Print the address of the pointer variable details
	fmt.Printf("address of details in main: %p\n", &funcDetails)

	// Call the function
	SetEnrollmentDetails(funcDetails)

	funcDetails.OSVersion = "2.0.0"
	if funcDetails.OSVersion == globalDetails.OSVersion {
		panic("global details should not change")
	}
}

func SetEnrollmentDetails(details EnrollmentDetails) {
	// Print the address of the pointer variable details in this function
	fmt.Printf("address of details in SetEnrollmentDetails: %p\n", &details)

	// Assign the address of the details to the globalEnrollmentDetails
	globalDetails = &details

	// Print the address stored in globalEnrollmentDetails
	fmt.Printf("globalEnrollmentDetails: %p\n", globalDetails)
}
```

prints:

```
address of details in main: 0x14000102020
address of details in SetEnrollmentDetails: 0x14000102030
globalEnrollmentDetails: 0x14000102030
```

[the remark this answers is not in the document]
James-Pickett: If think if we had pointers or arrays, the extra assignment wouldn't protect it. Either, I'm going to test this as well.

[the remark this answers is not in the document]
James-Pickett: Yea, if enrollment details and a field that was an array or a pointer, you would have to make an explicit deep copy of those fields. I don't think there is any way to protect against that other than something silly like converting it to JSON and back again.

ee/agent/knapsack/knapsack.go
```diff
@@ -20,6 +20,9 @@ import (
 // Package-level runID variable
 var runID string
 
+// Package-level enrollmentDetails variable
+var enrollmentDetails *types.EnrollmentDetails
```
James-Pickett, line 24: Should we make this an atomic value or protect with mutex since it can be accessed concurrently? I suppose the same would be true for runID.

James-Pickett (approved):

ee/agent/knapsack/knapsack.go
cesarfda: These are both only set in runLauncher and from them on they are basically read only

ee/agent/knapsack/knapsack.go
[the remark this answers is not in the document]
cesarfda: Removed, the extra assignment. Thanks for confirming the pointer behaviors

pkg/osquery/extension.go
```diff
@@ -426,38 +425,28 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
-			)
-			traces.SetError(span, fmt.Errorf("query osq enrollment details: %w", err))
-		} else {
-			span.AddEvent("got_enrollment_details")
+	err = backoff.WaitFor(func() error {
+		details := e.knapsack.GetEnrollmentDetails()
+		if err != nil {
+			traces.SetError(span, fmt.Errorf("error getting enrollment details: %w", err))
+			return err
+		}
```
James-Pickett, line 435: I think this is checking the err from `e.getHostIdentifier()` on 423? While it doesn't make any practical difference to the code, I think it's easier to read if we remove the double errcheck.

```suggestion
```

James-Pickett, line 435: I think this is checking the err from `e.getHostIdentifier()` on 423? While it doesn't make any practical difference to the code, I think it's easier to read if we remove the double errcheck.

```suggestion
```

James-Pickett (approved):

PR #2069, opened 2025-01-27
Force migration rather than halting launcher startup

RebeccaMahany:
This should prevent errors like this from completely halting launcher startup:

```
{
    "msg":"runLauncher exited",
    "err":"creating startup db: opening startup db in C:\\ProgramData\\Kolide\\Launcher-kolide-k2\\data: migrating the database: running migrations: Dirty database version 2. Fix and force version."
}
```

### DIFF
--- a/ee/agent/storage/sqlite/keyvalue_store_sqlite.go
+++ b/ee/agent/storage/sqlite/keyvalue_store_sqlite.go
@@ -187,7 +187,21 @@ func (s *sqliteStore) migrate() error {
 
 	// don't prevent DB access for a missing migration, this is the result of a downgrade after previously
 	// running a migration
-	if err := m.Up(); err != nil && !errors.Is(err, migrate.ErrNoChange) && !isMissingMigrationError(err) {
+	if err := m.Up(); err != nil {
+		// Not actually errors for us -- we're in a successful state
+		if errors.Is(err, migrate.ErrNoChange) || isMissingMigrationError(err) {
+			return nil
+		}
+
+		// If we need to force, do that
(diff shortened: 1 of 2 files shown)


ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -187,7 +187,21 @@ func (s *sqliteStore) migrate() error {
...
+		if errors.Is(err, migrate.ErrNoChange) || isMissingMigrationError(err) {
+			return nil
+		}
+
+		// If we need to force, do that
+		if errDirty, ok := err.(migrate.ErrDirty); ok {
+			if err := m.Force(errDirty.Version); err != nil {
+				return fmt.Errorf("forcing migration version %d: %w", errDirty.Version, err)
+			}
+			return nil
```
zackattack01, line 201: could we log that we intend to (or successfully did) the force call here regardless of error

zackattack01: oh no logger. hmmm

zackattack01: okay maybe not worth it, just seemed interesting to know if this ever causes issues

RebeccaMahany: Ah jk whoops -- we actually don't have logger access inside the store

zackattack01: yeah it seems messy this early too. that's fine with me this seems like a big improvement anyway

zackattack01 (approved):

PR #2070, opened 2025-01-28
Enable revive's defer rule in golangci-lint

RebeccaMahany:
Enables https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#defer in our ruleset, and fixes existing violations.

### DIFF
--- a/ee/desktop/user/notify/notify_linux.go
+++ b/ee/desktop/user/notify/notify_linux.go
@@ -10,7 +10,6 @@ import (
 	"log/slog"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/godbus/dbus/v5"
 	"github.com/kolide/launcher/ee/allowedcmd"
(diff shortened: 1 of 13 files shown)


ee/desktop/user/notify/notify_linux.go
```diff
@@ -101,9 +100,7 @@ func (d *dbusNotifier) Execute() error {
 			actionUri := signal.Body[1].(string)
 
 			for _, browserLauncher := range browserLaunchers {
-				ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
-				defer cancel()
-				cmd, err := browserLauncher(ctx, actionUri)
+				cmd, err := browserLauncher(context.TODO(), actionUri)
```
RebeccaMahany, line 103: Can't move a timeout with a `defer cancel` to a new function because then calling `cmd.Start` at the end of that function makes it fail immediately. I don't think the timeout is super meaningful here so I've removed it rather than adding a lint exception.

zackattack01 (approved):
nice!

James-Pickett (approved):
Nice, this looks like a pretty solid improvement.

PR #2073, opened 2025-01-28
Add tracing to kolide tables

RebeccaMahany:
Adds tracing basically wherever it is easy to do so, for all Kolide tables


James-Pickett (approved):
This is great ... I didn't go over it all with a fine tooth comb because it all just started looking the same to my brain after a while =)

PR #2077, opened 2025-01-30
Add wrapper to enforce timeout for extension tables

RebeccaMahany:
* Add a `tablewrapper` package to ensure that osquery does not ever have to wait on a call to `generate` for more than four minutes by default
* Update to use this package everywhere
* Enforce tablewrapper usage via lint rule

### DIFF
--- a/ee/tables/tablewrapper/tablewrapper.go
+++ b/ee/tables/tablewrapper/tablewrapper.go
@@ -13,55 +13,76 @@ import (
 
 const DefaultTableTimeout = 4 * time.Minute
 
+type wrappedTable struct {
+	slogger    *slog.Logger
+	name       string
+	gen        table.GenerateFunc
+	genTimeout time.Duration
+}
+
+type tablePluginOption func(*wrappedTable)
+
+// WithGenerateTimeout overrides the default table timeout of four minutes
+func WithGenerateTimeout(genTimeout time.Duration) tablePluginOption {
+	return func(w *wrappedTable) {
+		w.genTimeout = genTimeout
+	}
+}
+
 type generateResult struct {
 	rows []map[string]string
 	err  error
 }
 
-// NewTablePluginWithTimeout returns a table plugin using the default table timeout.
-// Most tables should use this function unless they have a specific need for a
-// custom timeout.
-func NewTablePluginWithTimeout(slogger *slog.Logger, name string, columns []table.ColumnDefinition, gen table.GenerateFunc) *table.Plugin {
-	return NewTablePluginWithCustomTimeout(slogger, name, columns, gen, DefaultTableTimeout)
-}
-
-// NewTablePluginWithCustomTimeout returns a table plugin that will attempt to execute a query
-// up until the given timeout, at which point it will instead return no rows and a timeout error.
-func NewTablePluginWithCustomTimeout(slogger *slog.Logger, name string, columns []table.ColumnDefinition, gen table.GenerateFunc, genTimeout time.Duration) *table.Plugin {
-	wrappedGen := func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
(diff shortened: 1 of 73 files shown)


ee/tables/tablewrapper/tablewrapper.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"log/slog"
+	"time"
+
+	"github.com/kolide/launcher/ee/gowrapper"
+	"github.com/kolide/launcher/pkg/traces"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+const DefaultTableTimeout = 2 * time.Minute
```
directionless, line 14:
```suggestion
const DefaultTableTimeout = 4 * time.Minute
```

I suspect some are slow, so let's start higher maybe?

ee/tables/tablewrapper/tablewrapper.go
```diff
@@ -0,0 +1,67 @@
...
+
+type generateResult struct {
+	rows []map[string]string
+	err  error
+}
+
+// NewTablePluginWithTimeout returns a table plugin using the default table timeout.
+func NewTablePluginWithTimeout(slogger *slog.Logger, name string, columns []table.ColumnDefinition, gen table.GenerateFunc) *table.Plugin {
```
directionless, line 22: This signature feels wrong. Feels like it should just be `New(...)` with a functional argument to override the default timeout

ee/tables/tablewrapper/tablewrapper.go
```diff
@@ -0,0 +1,67 @@
...
+func NewTablePluginWithCustomTimeout(slogger *slog.Logger, name string, columns []table.ColumnDefinition, gen table.GenerateFunc, genTimeout time.Duration) *table.Plugin {
+	wrappedGen := func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+		ctx, span := traces.StartSpan(ctx, "table_name", name)
+		defer span.End()
+
+		ctx, cancel := context.WithTimeout(ctx, genTimeout)
+		defer cancel()
+
+		// Kick off running the query
+		resultChan := make(chan *generateResult)
```
directionless, line 37: I love that this has the effect is interrupting no matter what's happening. But I'm also slightly leery of these piling up. Is there any way to clean this up? I assume not. So maybe we need to add something only spawn one generate (per table) at a time?

ee/tables/tablewrapper/tablewrapper.go
```diff
@@ -0,0 +1,67 @@
...
+				rows: rows,
+				err:  err,
+			}
+		})
+
+		// Wait for results up until the timeout
+		select {
+		case result := <-resultChan:
+			return result.rows, result.err
+		case <-ctx.Done():
```
directionless, line 51: I suspect we should be doing some more here. 

Annotate the span? Log directly? 

I don't know if we should also return some synthetic data. Probably not.

RebeccaMahany: The span will have the error added (https://github.com/osquery/osquery-go/blob/master/plugin/table/table.go#L75). I am also updating to add a log here -- that seems like it'd be useful to be able to search for.

James-Pickett (approved):
Nice

ee/tables/tablewrapper/tablewrapper.go
directionless: Annotate with the timeout value?

Yeah. log for searching is what I was thinking about.

cesarfda (dismissed):

ee/tables/tablewrapper/tablewrapper.go
RebeccaMahany: Discussed offline -- I will add some limits in a follow-up PR

James-Pickett (approved):

PR #2080, opened 2025-01-31
fix send buffer (log shipping) out of range panic

James-Pickett:
fixes panic: runtime error: index out of range [x] with length x that can occur in sendBuffer if buffer gets purged while waiting on network call to ship logs

### DIFF
--- a/pkg/sendbuffer/sendbuffer.go
+++ b/pkg/sendbuffer/sendbuffer.go
@@ -284,7 +284,7 @@ func (sb *SendBuffer) sendAndPurge() error {
 	// There is a small possibility that the buffer gets full while were in the middle of sending
 	// and gets deleted. However, we don't want to block writes while were waiting on a network call
 	// to send the logs. To live with this, we just have the deleteLogs func make sure that the buffer
-	// is not empty (ie: got purged while waiting on network call).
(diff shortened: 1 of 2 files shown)


RebeccaMahany (approved):

pkg/sendbuffer/sendbuffer.go
```diff
@@ -30,6 +30,9 @@ type SendBuffer struct {
 	sender                                      sender
 	sendTicker                                  *time.Ticker
 	isSending                                   bool
+
+	// logsJustPurged is used to prevent attempting to delete logs that were just purged
+	logsJustPurged bool
```
zackattack01, line 35: is this safe as bool and not atomic.Bool because we think it's adequately protected by the writeMutex? (it seems like it at first glance just wanted to double check)

zackattack01 (approved):

pkg/sendbuffer/sendbuffer.go
James-Pickett: yes, it's protected by the writeMutex... trying to think of a better way to do this, but I think this will suffice

RebeccaMahany (approved):

zackattack01 (approved):

PR #2085, opened 2025-02-03
Get timeout for table generate function from control server

RebeccaMahany:
Updates `tablewrapper` to pull the generate timeout value from the control server, so that if we need to adjust the timeout on the fly, we can.


James-Pickett (dismissed):
LGTM! I took a good look at the table wrapper, didn't go over everything else with fine tooth comb. Let me know if there is another part I should focus on.

RebeccaMahany:
@James-Pickett that's the important part, thank you!!

ee/agent/flags/flag_controller.go
```diff
@@ -703,3 +703,14 @@ func (fc *FlagController) Identifier() string {
 
 	return identifier
 }
+
+func (fc *FlagController) SetGenerateTimeout(interval time.Duration) error {
+	return fc.setControlServerValue(keys.GenerateTimeout, durationToBytes(interval))
+}
+func (fc *FlagController) GenerateTimeout() time.Duration {
```
directionless, line 710: I'd suggest "TableGenerateTimeout"

ee/agent/flags/keys/keys.go
```diff
@@ -58,6 +58,7 @@ const (
 	LauncherWatchdogEnabled         FlagKey = "launcher_watchdog_enabled" // note that this will only impact windows deployments for now
 	SystrayRestartEnabled           FlagKey = "systray_restart_enabled"
 	CurrentRunningOsqueryVersion    FlagKey = "osquery_version"
+	GenerateTimeout                 FlagKey = "generate_timeout"
```
directionless, line 61: same -- `table_generate_timeout`

ee/agent/flags/flag_controller.go
RebeccaMahany: Done! For both

zackattack01 (approved):

PR #2087, opened 2025-02-05
Enable rowserrcheck linter and fix existing violations

RebeccaMahany:
Makes sure we check and log `rows.Err()` after calling `rows.Close()`.

### DIFF
--- a/ee/agent/storage/sqlite/keyvalue_store_sqlite.go
+++ b/ee/agent/storage/sqlite/keyvalue_store_sqlite.go
@@ -6,6 +6,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	"log/slog"
 	"math"
 	"os"
 	"path"
@@ -45,6 +46,7 @@ func (s storeName) String() string {
 var migrations embed.FS
 
 type sqliteStore struct {
+	slogger       *slog.Logger
 	conn          *sql.DB
 	readOnly      bool
 	rootDirectory string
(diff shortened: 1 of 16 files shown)


ee/agent/storage/sqlite/keyvalue_store_sqlite.go
```diff
@@ -324,7 +327,20 @@ ON CONFLICT (name) DO UPDATE SET value=excluded.value;`
...
+		if err := rows.Close(); err != nil {
+			s.slogger.Log(context.TODO(), slog.LevelWarn,
+				"closing rows after scanning results",
+				"err", err,
+			)
+		}
+		if rows.Err() != nil {
+			s.slogger.Log(context.TODO(), slog.LevelWarn,
+				"encountered iteration error",
+				"err", err,
```
zackattack01, line 340: should this capture the `err` from `rows.Err()`?

ee/agent/storage/sqlite/logstore_sqlite.go
```diff
@@ -82,7 +84,20 @@ func (s *sqliteStore) ForEach(fn func(rowid, timestamp int64, v []byte) error) e
...
+		if err := rows.Close(); err != nil {
+			s.slogger.Log(context.TODO(), slog.LevelWarn,
+				"closing rows after scanning results",
+				"err", err,
+			)
+		}
+		if rows.Err() != nil {
+			s.slogger.Log(context.TODO(), slog.LevelWarn,
+				"encountered iteration error",
+				"err", err,
```
zackattack01, line 97: same note here, I think we need to capture `err`

ee/agent/storage/sqlite/keyvalue_store_sqlite.go
RebeccaMahany: Oops yes thank you!

zackattack01 (approved):

PR #2093, opened 2025-02-06
Enable noctx linter and fix existing violations

RebeccaMahany:
The `noctx` linter alerts on sending http requests without `context.Context`.


James-Pickett (approved):
nice

Issue #2086, opened 2025-02-05
`kolide_chrome_login_data_emails` table missing sqlite3 driver


RebeccaMahany:
Queries to this table frequently generate logs with error `connecting to sqlite db: sql: unknown driver "sqlite3" (forgotten import?)` (and msg `generating chrome keychain result`). I don't see this error for any of our other sqlite tables.

Looks like this might also only happen on Linux and Windows -- I'm not seeing any logs for it from macOS devices.

PR #2094, opened 2025-02-06
Ensure sqlite driver is imported for tables that query sqlite databases

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2086.

We were importing a sqlite driver inside `platform_tables_darwin.go`, which meant that calling `kolide_chrome_login_data_emails` worked only on Darwin. This PR updates to import the sqlite driver everywhere that we use it. I didn't see errors for any of the other tables updated in this PR, but updated them too for the sake of completeness.

We use the `modernc.org/sqlite` over `mattn/go-sqlite3` driver because it is supported on Windows. (We already use `modernc.org/sqlite` for our sqlite kv store.) This changes the driver name from `sqlite3` to `sqlite`, but [modernc.org/sqlite is indeed sqlite3](https://pkg.go.dev/modernc.org/sqlite).

I also added tests to confirm that the tables will fail without the added import -- example failure here: https://github.com/kolide/launcher/actions/runs/13187761763/job/36814141625?pr=2094. I did not add a test for `kolide_chrome_login_keychain` since that table is deprecated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/kolide/kit v0.0.0-20241126150023-fbf6f0f5bf6a
 	github.com/kolide/krypto v0.1.1-0.20241212211625-46a8d5cad1cc
 	github.com/mat/besticon v3.9.0+incompatible
-	github.com/mattn/go-sqlite3 v1.14.19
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):
Nice

zackattack01 (approved):

PR #2095, opened 2025-02-07
check munemo in local server

James-Pickett:
Add handler to local server to reject requests that don't match munemo from enroll secret.

This should stop the issue where presence detection window get "refreshed" because 2 launcher instances are trying to do presence detection in Kolide's dual launcher set up.

corresponding k2 pr https://github.com/kolide/k2/pull/11339


zackattack01:
The code seems great to me just a couple questions/comments:
- it seems like the intention server side (from the linked PR) is to send that munemo header for every organization, which means this enroll secret read + parse will get added for every request. maybe faster if we don't include that header at all by default for other orgs? since this is only supposed to fix the internal dual-install scenarios
- otherwise I wonder if we shouldn't cache the munemo somewhere? just seems kinda expensive to do this inside the request handling

James-Pickett:
> The code seems great to me just a couple questions/comments:
> 
> * ~it seems like the intention server side (from the linked PR) is to send that munemo header for every organization, which means this enroll secret read + parse will get added for every request. maybe faster if we don't include that header at all by default for other orgs? since this is only supposed to fix the internal dual-install scenarios.~ edit: never mind I see why you had to do it like this
> * otherwise I wonder if we shouldn't cache the munemo somewhere? just seems kinda expensive to do this inside the request handling

Yea, I'm cool with caching the munemo, probably in knapsack

ee/localserver/server.go
```diff
@@ -407,3 +410,93 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
...
+			"header", munemoHeaderValue,
+			"enroll_secret", localServerMunemo,
+		)
+
+		http.Error(w, "munemo in request does not match munemo in enroll secret", http.StatusUnauthorized)
+		return
+	})
+}
+
+// munemoCheckHandler returns an http error if the request's X-Kolide-Munemo header does not match the organization in the enroll secret
```
zackattack01, line 465: should probably update or remove this comment

ee/localserver/server.go
```diff
@@ -407,3 +410,93 @@ func (ls *localServer) rateLimitHandler(next http.Handler) http.Handler {
...
+			)
+
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		localServerMunemo, err := ls.getMunemoFromEnrollSecret()
+		if err != nil {
+			ls.slogger.Log(r.Context(), slog.LevelError,
+				"getting munemo from enroll secret, conitnuing",
```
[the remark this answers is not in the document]
James-Pickett: fixed

ee/localserver/server.go
James-Pickett: updated

zackattack01 (dismissed):

zackattack01 (approved):

PR #2101, opened 2025-02-10
remove unneeded rsa references, upgrade krypto

James-Pickett:
K2 & Krypto PRs

https://github.com/kolide/k2/pull/11371
https://github.com/kolide/krypto/pull/47


directionless (dismissed):
What about https://github.com/kolide/launcher/blob/df315fe57b4c58d981b40b465e6ac94ddb09d544/pkg/osquery/extension.go#L221

pkg/osquery/table/launcher_info.go
```diff
@@ -101,8 +94,8 @@ func generateLauncherInfoTable(configStore types.GetterSetter, LauncherHistorySt
 				"registration_id":     types.DefaultRegistrationID,
 				"identifier":          identifier,
 				"osquery_instance_id": osqueryInstance.InstanceId,
-				"fingerprint":         fingerprint,
-				"public_key":          publicKey,
+				"fingerprint":         "not_in_use",
+				"public_key":          "not_in_use",
```
James-Pickett, line 98: is this okay? We typically don't want to just remove columns from osq tables.

zackattack01 (dismissed):

pkg/osquery/table/launcher_info.go
directionless: Good call on not removing. I'd say NULL or empty string.

pkg/osquery/extension.go
```diff
@@ -59,12 +56,6 @@ const (
 	nodeKeyKey = "nodeKey"
 	// DB key for last retrieved config
 	configKey = "config"
-	// DB keys for the rsa keys
-	privateKeyKey = "privateKey"
-
-	// Old things to delete
-	xPublicKeyKey      = "publicKey"
-	xKeyFingerprintKey = "keyFingerprint"
```
directionless, line 67: Should we still be deleting these, as well as the `privateKey`?

pkg/osquery/extension.go
```diff
@@ -215,94 +206,6 @@ func (e *Extension) getHostIdentifier() (string, error) {
 	return IdentifierFromDB(e.knapsack.ConfigStore(), e.registrationId)
 }
 
-// SetupLauncherKeys configures the various keys used for communication.
-//
-// There are 3 keys:
-// 1. The RSA key. This is stored in the launcher DB, and was the first key used by krypto. We are deprecating it.
-// 2. The hardware keys -- these are in the secure enclave (TPM or Apple's thing) These are used to identify the device
-// 3. The launcher install key -- this is an ECC key that is sometimes used in conjunction with (2)
```
directionless, line 223: I think this comment is old -- we don't generate the ECC keys in this function. Thus it's removal is safe

directionless (commented):
Seems scary, but I think this is something `nightly` will test fine. As long as we can manage auth flows, I think we're good.

cesarfda (dismissed):

pkg/osquery/table/launcher_info.go
James-Pickett: updated to empty string

zackattack01 (approved):

PR #2106, opened 2025-02-18
make uninitialized osquery history informational for doctor output

zackattack01:
we may run doctor/flare in standalone mode, where osquery restart history access wouldn't be possible. to accommodate this without alarming flare consumers, we should just make this state informational (not erroring)


cesarfda (approved):

RebeccaMahany (approved):

RebeccaMahany:
Looks like https://github.com/kolide/launcher/pull/2107/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9 will fix the lint error 🙂

James-Pickett (approved):

PR #2107, opened 2025-02-18
add config file option to flare command for improved flexibility

### LINKED ISSUE #2058: Add `-config` flag for `flare` subcommand
We should be able to specify the location of the config file for the `flare` subcommand, the same way we can for the `doctor` subcommand, so that for installations with non-standard identifiers we pick up the correct config file, root directory, etc.

cesarfda:
This pull request includes updates to the `runFlare` function in the `cmd/launcher/flare.go` file to enhance the handling of configuration files and improve the flexibility of the flare command. Resolves #2058 

Enhancements to configuration file handling:

* Added a new flag `flConfigFilePath` to specify the path to a configuration file, with a default value set to `launcher.DefaultConfigFilePath`.
* Updated the `ff.Parse` function to include options for parsing the configuration file using `ff.WithConfigFileFlag` and `ff.WithConfigFileParser`.

Improvements to flare command flexibility:

* Modified the `ParseOptions` function call to pass the actual flagset arguments instead of an empty array, allowing for more accurate option parsing.
* Added the `config_path` to the log entry to include the path of the configuration file used during the flare creation process.


James-Pickett (dismissed):

cmd/launcher/flare.go
```diff
@@ -37,14 +37,21 @@ func runFlare(systemMultiSlogger *multislogger.MultiSlogger, args []string) erro
...
+		ff.WithConfigFileParser(ff.PlainParser),
+	}
+
+	if err := ff.Parse(flagset, args, ffOpts...); err != nil {
 		return fmt.Errorf("parsing flags: %w", err)
 	}
 
 	// were passing an empty array here just to get the default options
-	opts, err := launcher.ParseOptions("flareupload", make([]string, 0))
+	opts, err := launcher.ParseOptions("flare", flagset.Args())
```
zackattack01, line 54: if someone specifies e.g. `output_dir` as part of the flare command, will that be passed through here and error out when it is unknown to `launcher.ParseOptions`?

cesarfda: On local it looks like the output_dir still works. I only tested with relative paths, let me know if there's any specific edge case that could trigger that unknown behavior.

image

cesarfda: Done

zackattack01 (approved):

James-Pickett (approved):

PR #2111, opened 2025-02-19
make sure db reset logs encapsulate change values

zackattack01:
I noticed in a flare recently that we had logged a change to serial and hardware uuid values, but then immediately logged `"finished check to see if database should be reset..."` without any of those changed values being reflected.

This was just because the deferred log was done outside of an anonymous function, causing all values to be evaluated when the defer was processed instead of during its execution later. This wraps the call in a func to ensure those values reflect any updates at the time of logging


RebeccaMahany (approved):
💯 nice catch

PR #2116, opened 2025-02-20
verify secure enclave keys exist in secure enclave

James-Pickett:
Adds required plumbing and updates secure enclave runner to verify that the key exists in secure enclave when it loaded from db.

While testing I learned that you'll get a errKCItemNotFound = -25300 error if the secure enclave is unavailable for some reason, so it can't be relied on to determine if a key exists.

To test if the secure enclave is available, we can create a key, then delete it. Corresponding krypto PR to add ability to create delete key https://github.com/kolide/krypto/pull/51


ee/desktop/user/server/server_darwin.go
```diff
@@ -27,3 +29,55 @@ func (s *UserServer) createSecureEnclaveKey(w http.ResponseWriter, _ *http.Reque
...
+	signer, err := secureenclave.New(pubKey)
+	if err != nil {
+
+		// errKCItemNotFound = -25300
+		// means item was not found, any other error we assume is a different problem
+		// apple docs
+		// https://developer.apple.com/documentation/coreservices/1559994-anonymous/errkcitemnotfound/
+		// apple site where you can search for error codes, just enter error code
+		// https://developer.apple.com/bugreporter/
+		if strings.Contains(err.Error(), "-25300") {
```
James-Pickett, line 58: Dont really like the string contains .... but I don't think there is a way to pull "proper go errors" out of the apple C apis. At some point in the chain I think there would have to be a string or int compare and then creation of the go error.

ee/secureenclaverunner/secureenclaverunner.go
```diff
@@ -231,8 +287,8 @@ func (ser *secureEnclaveRunner) currentConsoleUserKey(ctx context.Context) (*ecd
 func (ser *secureEnclaveRunner) MarshalJSON() ([]byte, error) {
 	keyMap := make(map[string]string)
 
-	for uid, pubKey := range ser.uidPubKeyMap {
-		pubKeyBytes, err := x509.MarshalPKIXPublicKey(pubKey)
+	for uid, entry := range ser.uidPubKeyMap {
+		pubKeyBytes, err := x509.MarshalPKIXPublicKey(entry.pubKey)
```
RebeccaMahany, line 291: Maybe worth documenting here also (like you do in the unmarshal function) that we don't want to include `entry.verifiedInSecureEnclave` when marshalling because we don't want to save that info in the store, so that verified state won't get copied if the launcher db gets copied to a new machine?

ee/secureenclaverunner/secureenclaverunner_test.go
```diff
@@ -215,4 +247,116 @@ func Test_secureEnclaveRunner(t *testing.T) {
 		// and public not called
 		require.Len(t, ser.uidPubKeyMap, 0)
 	})
+
+	t.Run("creates new key when existing not found in secure enclave", func(t *testing.T) {
```
RebeccaMahany, line 251: Just curious -- I noticed these tests use `t.Run` instead of separate test functions. Is that a pattern you think we might want to adopt more widely across launcher? I like how specific the test names are without having to be like `func TestFunc_someSpecificHappyOrUnhappyPath(t *testing.T)`.

RebeccaMahany (dismissed):
Really nice, I like it!

ee/secureenclaverunner/secureenclaverunner_test.go
James-Pickett: I just kind of tried it out and I really like it for a few reasons:

1. The table driven tests can get really complicated as the the code it's testing grows and more edge cases are found
2. Using VS Code I can click run on a single test, for the table driven tests I have to comment out the ones I don't want to run
3. Feels easier to add or remove things from the tests, with the table driven tests, you change one field in the struct and your updating n tests

I still think table driven tests have their place and are a better choice for smaller scoped areas of code with simple inputs, outputs and few error / success paths.

ee/secureenclaverunner/secureenclaverunner.go
James-Pickett: 👍  added a comment

RebeccaMahany (dismissed):

RebeccaMahany (approved):

PR #2117, opened 2025-02-21
Check origin header against allowlist for /zta endpoint

RebeccaMahany:
We want to restrict access to localserver's `/zta` endpoint to allowlisted origins only.

### DIFF
--- a/ee/localserver/zta.go
+++ b/ee/localserver/zta.go
@@ -9,6 +9,11 @@ import (
 
 var (
 	localserverZtaInfoKey = []byte("localserver_zta_info")
+
+	// allowlistedZtaOriginsLookup contains the complete list of origins that are permitted to access the /zta endpoint.
+	allowlistedZtaOriginsLookup = map[string]struct{}{
+		"https://app.kolide.com": {}, // TODO RM -- placeholder, awaiting actual values
+	}
 )
 
 func (ls *localServer) requestZtaInfoHandler() http.Handler {
@@ -21,6 +26,13 @@ func (ls *localServer) requestZtaInfoHandlerFunc(w http.ResponseWriter, r *http.
 
 	if r.Method != http.MethodGet {
 		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	// Validate origin
+	requestOrigin := r.Header.Get("Origin")
+	if _, ok := allowlistedZtaOriginsLookup[requestOrigin]; !ok {
+		w.WriteHeader(http.StatusForbidden)
 		return
 	}
 
--- a/ee/localserver/zta_test.go
+++ b/ee/localserver/zta_test.go
@@ -40,6 +40,7 @@ func Test_requestZtaInfoHandler(t *testing.T) {
 
 	// Make a request to our handler
 	request := httptest.NewRequest(http.MethodGet, "/zta", nil)
+	request.Header.Set("origin", acceptableOrigin(t))
 	responseRecorder := httptest.NewRecorder()
 	ls.requestZtaInfoHandler().ServeHTTP(responseRecorder, request)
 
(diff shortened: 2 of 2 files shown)


ee/localserver/zta.go
```diff
@@ -9,6 +9,11 @@ import (
 
 var (
 	localserverZtaInfoKey = []byte("localserver_zta_info")
+
+	// allowlistedZtaOriginsLookup contains the complete list of origins that are permitted to access the /zta endpoint.
+	allowlistedZtaOriginsLookup = map[string]struct{}{
+		"https://app.kolide.com": {}, // TODO RM -- placeholder, awaiting actual values
```
RebeccaMahany, line 15: Leaving PR as a draft until I have these 🙂

ee/localserver/zta.go
```diff
@@ -24,6 +29,13 @@ func (ls *localServer) requestZtaInfoHandlerFunc(w http.ResponseWriter, r *http.
 		return
 	}
 
+	// Validate origin
+	requestOrigin := r.Header.Get("Origin")
+	if _, ok := allowlistedZtaOriginsLookup[requestOrigin]; !ok {
+		w.WriteHeader(http.StatusForbidden)
```
directionless, line 35: We should log something here

RebeccaMahany: Sure! I can sanitize+log the origin -- anything else?

RebeccaMahany: Added -- let me know if you'd like it updated!

zackattack01 (dismissed):

James-Pickett (approved):

ee/localserver/zta.go
```diff
@@ -3,12 +3,31 @@ package localserver
...
+	allowlistedZtaOriginsLookup = map[string]struct{}{
+		// Release extension
+		"chrome-extension://gejiddohjgogedgjnonbofjigllpkmbf":         {},
+		"chrome-extension://khgocmkkpikpnmmkgmdnfckapcdkgfaf":         {},
+		"chrome-extension://aeblfdkhhhdcdjpifhhbdiojplfjncoa":         {},
+		"chrome-extension://dppgmdbiimibapkepcbdbmkaabgiofem":         {},
+		"moz-extension://dfbae458-fb6f-4614-856e-094108a80852":        {},
+		"moz-extension://25fc87fa-4d31-4fee-b5c1-c32a7844c063":        {},
+		"moz-extension://d634138d-c276-4fc8-924b-40a0ea21d284":        {},
+		"safari-web-extension://7cba6ce0-bc55-43bd-b2a3-562b703ecbea": {},
```
complexspaces, line 24: As noted over Slack, we'll need to prefix match `safari-web-extension://` instead of hardcode an expected value since the UUID for an SWE is randomized each time that SWE is loaded.

RebeccaMahany: Addressed: https://github.com/kolide/launcher/pull/2117/commits/7eb33b1903ff9696f786965dd2129b627e0755a3

directionless: > prefix match safari-web-extension:// 

Doesn't this allow _all_ safari extensions, thus undermining the point?

James-Pickett (dismissed):

ee/localserver/zta.go
complexspaces: This is only going to be one defense in depth layer but more importantly: Every SWE is a native code bundle by definition (they are a special app bundle installed via the app store), so just as `Origin` matching isn't a strong defense against Chrome/Firefox extensions with NMH or native apps, it isn't a strong defense against Safari extensions. These Safari extensions can run HTTP requests from their Swift code and spoof their `Origin` if desired.

RebeccaMahany: @directionless -- yes, unfortunately it will allow all safari extensions. We're still preventing any other random site from accessing `/zta`, so this is an improvement, albeit incomplete. I think we've got to call this "good enough for now".

complexspaces (approved):

ee/localserver/zta.go
```diff
@@ -3,12 +3,32 @@ package localserver
...
+		"moz-extension://25fc87fa-4d31-4fee-b5c1-c32a7844c063": {},
+		"moz-extension://d634138d-c276-4fc8-924b-40a0ea21d284": {},
+		// Development and internal builds
+		"chrome-extension://hjlinigoblmkhjejkmbegnoaljkphmgo":  {},
+		"moz-extension://0a75d802-9aed-41e7-8daa-24c067386e82": {},
+		"chrome-extension://hiajhnnfoihkhlmfejoljaokdpgboiea":  {},
+		"chrome-extension://kioanpobaefjdloichnjebbdafiloboa":  {},
+	}
+
+	safariWebExtensionScheme = "safari-web-extension://"
```
zackattack01, line 31: nit- should this be const instead of var?

RebeccaMahany: Ahhh yes, thank you, whoops

RebeccaMahany: Updated!

zackattack01 (approved):

ee/localserver/zta.go
directionless: If @complexspaces is good, I'm good. I just want to note it.

PR #2120, opened 2025-02-24
Ensure enrollment details are set

RebeccaMahany:
In `SetEnrollmentDetails` we were setting the enrollment details to the old, rather than the updated, value.

Additionally, I gave us a longer timeout before failing to set enrollment details -- since it's happening in the background, I figure a longer timeout can only help. I also added a couple more logs to help us debug this issue in the future.

### DIFF
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -248,24 +248,13 @@ func (k *knapsack) CurrentEnrollmentStatus() (types.EnrollmentStatus, error) {
 	return types.Enrolled, nil
 }
 
-func (k *knapsack) SetEnrollmentDetails(details types.EnrollmentDetails) {
-	if enrollmentDetails == nil {
-		enrollmentDetails = &details
-		k.Slogger().Log(context.Background(), slog.LevelDebug,
-			"initializing enrollment details",
-			"details", fmt.Sprintf("%+v", enrollmentDetails),
-		)
-
-	}
-
-	current := *enrollmentDetails
-
+func (k *knapsack) SetEnrollmentDetails(newDetails types.EnrollmentDetails) {
 	k.Slogger().Log(context.Background(), slog.LevelDebug,
 		"updating enrollment details",
-		"old_details", fmt.Sprintf("%+v", current),
-		"new_details", fmt.Sprintf("%+v", details),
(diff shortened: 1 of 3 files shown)


ee/agent/knapsack/knapsack.go
```diff
@@ -262,10 +262,10 @@ func (k *knapsack) SetEnrollmentDetails(details types.EnrollmentDetails) {
 
 	k.Slogger().Log(context.Background(), slog.LevelDebug,
 		"updating enrollment details",
-		"old_details", fmt.Sprintf("%+v", enrollmentDetails),
-		"new_details", fmt.Sprintf("%+v", current),
+		"old_details", fmt.Sprintf("%+v", current),
```
zackattack01, line 265: nit- I think this entire function might be more readable/less error prone as something like 
```
func (k *knapsack) SetEnrollmentDetails(newDetails types.EnrollmentDetails) {
	k.Slogger().Log(context.Background(), slog.LevelDebug,
		"updating enrollment details",
		"old_details", fmt.Sprintf("%+v", enrollmentDetails),
		"new_details", fmt.Sprintf("%+v", newDetails),
	)

	enrollmentDetails = &newDetails
}
```

zackattack01: i also don't mind keeping the separate initializing log if that's more clear to people than old_details being nil, but it felt like most of the pointer swapping was not needed

zackattack01 (dismissed):

ee/agent/knapsack/knapsack.go
RebeccaMahany: Sure thing -- updated

RebeccaMahany: It felt clear enough to me in the log without the separate initializing log --

```
"msg":"updating enrollment details","old_details":"<nil>","new_details":"{OSVersion: OSBuildID: OSPlatform:darwin Hostname: HardwareVendor: HardwareModel: HardwareSerial: OsqueryVersion: LauncherHardwareKey: LauncherHardwareKeySource: LauncherLocalKey: LauncherVersion:1.15.2-31-g25c6278d-dirty OSName: OSPlatformLike:darwin GOOS:darwin GOARCH:arm64 HardwareUUID:}"
```

zackattack01 (approved):

Issue #2097, opened 2025-02-10
osquery instance should not restart when the KATC configuration changes


RebeccaMahany:
Currently, the osquery runner restarts all instances when `Ping` is called. (See: https://github.com/kolide/launcher/blob/main/pkg/osquery/runtime/runner.go#L286-L302.) On launcher first-time startup, launcher will receive an updated KATC configuration the first time it talks to the control server. Consequently, on launcher first-time startup, the osquery runner will start an instance, _then restart it almost immediately afterward_. This immediate restart is problematic because it delays the time before osquery is fully functional and running queries, and additionally because osquery startup is a little fragile on Windows and having to do it not once but _twice_ is not desirable.

This issue is for researching options that will allow us to avoid this restart.

Research notes:

* It doesn't look like we can de-register individual plugins, so I don't think we can de-register the KATC table plugins and then re-register new ones.
* Because the `kolide_grpc` extension is required (we indicate this with the `--extensions_require` flag when starting up osquery), I don't think we can safely de-register the entire extension and start up a new one in its place -- but it's probably worthwhile to validate this assumption.
* We cannot register plugins after the server has started, so we cannot add the new KATC tables when we receive them after osquery instance startup.
* Therefore -- we have to either wait for the control server data to become available, or ensure that the KATC configuration is available ahead of time or on demand.
* We could make the config available ahead of time by packaging launcher with the current KATC config. (This wouldn't help with organizations that continue to use older installers, though.)
* We could make the config available on demand via e.g. unauthed K2 API endpoint.

PR #2121, opened 2025-02-25
Add new KATC tables without restarting

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2097


RebeccaMahany:
Won't work -- plugins have to be registered before the extension manager server is started

PR #2126, opened 2025-02-26
Revert "Check origin header against allowlist for /zta endpoint"

RebeccaMahany:
Reverts kolide/launcher#2117


directionless (approved):

directionless (changes requested):
Change in plans. no revert any more.

PR #2127, opened 2025-02-26
Allow origin (in requests to /zta) to be missing or empty

RebeccaMahany:
We only want to disallow requests where the origin is set _and_ it isn't in the allowlist.


directionless (approved):
don't love the nesting, but good enough!

PR #2128, opened 2025-02-27
Move KATC tables to restartable extension manager server

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2097

Background from issue:

Currently, the osquery runner restarts all instances when Ping is called. On launcher first-time startup, launcher will receive an updated KATC configuration the first time it talks to the control server. Consequently, on launcher first-time startup, the osquery runner will start an instance, _then restart it almost immediately afterward_. This immediate restart is problematic because it delays the time before osquery is fully functional and running queries, and additionally because osquery startup is a little fragile on Windows and having to do it not once but _twice_ is not desirable.

Changes in this PR:

* Splits out KATC tables from other platform tables
* Adds a new extension manager server just for supplying the KATC tables
* On call to `Ping` (when KATC config changes), runner will attempt to reload the instance's KATC tables extension before restarting the instance
* Improve healthcheck to validate existence of extensions


pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -259,6 +278,66 @@ func (i *OsqueryInstance) Exited() <-chan struct{} {
...
+	i.emsLock.Unlock()
+
+	if err := i.startKatcExtensionManagerServer(ctx); err != nil {
+		return fmt.Errorf("starting katc server: %w", err)
+	}
+
+	return nil
+}
+
+func (i *OsqueryInstance) instanceStarted() bool {
```
zackattack01, line 313: maybe worth adding a comment here that we expect callers to manage emsLock while using

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -259,6 +278,66 @@ func (i *OsqueryInstance) Exited() <-chan struct{} {
...
+// without restarting the entire instance.
+func (i *OsqueryInstance) ReloadKatcExtension(ctx context.Context) error {
+	ctx, span := traces.StartSpan(ctx)
+	defer span.End()
+
+	if !i.instanceStarted() {
+		return errors.New("instance not started, cannot reload extension")
+	}
+
+	i.emsLock.Lock()
```
zackattack01, line 292: should this lock before calling instanceStarted()?

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: Ah I'll have this one lock too, seems simpler -- will update

pkg/osquery/runtime/osqueryinstance.go
RebeccaMahany: Moved lock into `instanceStarted`

James-Pickett (dismissed):
Nice

pkg/osquery/runtime/osqueryinstance.go
```diff
@@ -820,13 +916,14 @@ func (i *OsqueryInstance) StartOsqueryClient(paths *osqueryFilePaths) (*osquery.
 
 // startOsqueryExtensionManagerServer takes a set of plugins, creates
 // an osquery.NewExtensionManagerServer for them, and then starts it.
-func (i *OsqueryInstance) StartOsqueryExtensionManagerServer(name string, socketPath string, client *osquery.ExtensionManagerClient, plugins []osquery.OsqueryPlugin) error {
+// If allowRestart is set
```
zackattack01, line 919: maybe something like the errgroup will not return an exit error after successful calls to extensionManagerServer.Start(), to avoid canceling the group where restarts are expected

RebeccaMahany: Oh whoops, I don't know what happened to this comment here, thank you! Updated

zackattack01 (approved):
🔥

PR #2130, opened 2025-02-28
secure enclave more comments, logging

James-Pickett:
more comments, logging, fix unneeded header writes

### DIFF
--- a/ee/desktop/user/server/server_darwin.go
+++ b/ee/desktop/user/server/server_darwin.go
@@ -21,7 +21,7 @@ func (s *UserServer) createSecureEnclaveKey(w http.ResponseWriter, r *http.Reque
 			"secure enclave unavailable, could not create key",
 			"err", err,
 		)
-		http.Error(w, fmt.Errorf("secure enclave unavailable, could not creating key: %w", err).Error(), http.StatusServiceUnavailable)
+		http.Error(w, fmt.Errorf("secure enclave unavailable, could not create key: %w", err).Error(), http.StatusServiceUnavailable)
 		return
 	}
 
--- a/ee/desktop/user/server/server_darwin.go
+++ b/ee/desktop/user/server/server_darwin.go
@@ -6,6 +6,7 @@ package server
 import (
 	"crypto/ecdsa"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


ee/desktop/user/server/server_darwin.go
```diff
@@ -6,17 +6,22 @@ package server
...
-func (s *UserServer) createSecureEnclaveKey(w http.ResponseWriter, _ *http.Request) {
+func (s *UserServer) createSecureEnclaveKey(w http.ResponseWriter, r *http.Request) {
 	key, err := secureenclave.CreateKey()
 	if err != nil {
-		http.Error(w, fmt.Errorf("creating key: %w", err).Error(), http.StatusInternalServerError)
+		s.slogger.Log(r.Context(), slog.LevelDebug,
+			"secure enclave unavailable, could not create key",
+			"err", err,
+		)
+		http.Error(w, fmt.Errorf("secure enclave unavailable, could not creating key: %w", err).Error(), http.StatusServiceUnavailable)
```
zackattack01, line 24:
```suggestion
		http.Error(w, fmt.Errorf("secure enclave unavailable, could not create key: %w", err).Error(), http.StatusServiceUnavailable)
```

zackattack01: I guess I don't know what these errors typically look like- just verifying, is that something we should expose directly in the response?

James-Pickett: they are just apple error code statuses like `errKCItemNotFound -25300` , there shouldn't be anything sensitive

ee/desktop/user/server/server_darwin.go
```diff
@@ -68,12 +81,21 @@ func (s *UserServer) getSecureEnclaveKey(w http.ResponseWriter, r *http.Request)
...
 		}
 
+		s.slogger.Log(r.Context(), slog.LevelError,
+			"encountered unexpected error, cannot determine if key exists in secure enclave",
+			"err", err,
+		)
+
 		// encountered some other error, cannot confirm if key exists
-		http.Error(w, fmt.Errorf("encounter error, cannot determine if key exists in secure enclave: %w", err).Error(), http.StatusInternalServerError)
+		http.Error(w, fmt.Errorf("encounter unexpected error, cannot determine if key exists in secure enclave: %w", err).Error(), http.StatusInternalServerError)
```
zackattack01, line 98: same question as above, if these are errors we don't really control I kinda wonder if appropriate to put them directly in the response

zackattack01 (approved):

PR #2136, opened 2025-03-03
add system startup trigger to watchdog

zackattack01:
We're seeing some instances of VDIs failing to start launcher on system startup due to timeouts connecting to the service control manager. Watchdog should be well positioned to fix this, but most of the on-demand triggers are based off of wake events that these VDIs don't seem to transition through.

To account for this we can also add a system startup trigger with an execution delay of 2 minutes. See inline comments for the reasoning behind 2 minutes.


James-Pickett (approved):

ee/watchdog/controller_windows.go
```diff
@@ -483,6 +483,35 @@ func installWatchdogTask(identifier, configFilePath string) error {
...
+	createBootTriggerResp, err := oleutil.CallMethod(triggers, "Create", uint(8)) // 8=TASK_TRIGGER_BOOT
+	if err != nil {
+		return fmt.Errorf("encountered error creating event trigger: %w", err)
+	}
+
+	bootTrigger := createBootTriggerResp.ToIDispatch()
+	defer bootTrigger.Release()
+
+	if _, err = oleutil.PutProperty(bootTrigger, "ExecutionTimeLimit", "PT1M"); err != nil {
+		return errors.New("setting execution time limit property")
```
RebeccaMahany, line 496:
```suggestion
		return fmt.Errorf("setting execution time limit property: %w", err)
```

?

RebeccaMahany (commented):
LGTM!

ee/watchdog/controller_windows.go
zackattack01: good catch thank you!

RebeccaMahany (approved):

PR #2141, opened 2025-03-05
add secure enclave signing

James-Pickett:


ee/localserver/krypto-ec-middleware.go
```diff
@@ -617,9 +617,13 @@ func (e *kryptoEcMiddleware) kryptoResponse(box *challenge.OuterChallenge, data
...
 	// pass nil to krypto
-	// hardware signing is not implemented for darwin
+
+	// we add these kolide tags to the data so that we can identify the response as a kolide response
+	// and mitigate the agent signing arbitrary data
+	data = fmt.Appendf(nil, "kolide:%s:kolide", data)
+
 	var err error
-	if runtime.GOOS != "darwin" && agent.HardwareKeys() != nil && agent.HardwareKeys().Public() != nil {
+	if agent.HardwareKeys() != nil && agent.HardwareKeys().Public() != nil {
```
James-Pickett, line 626: if we get an error here signing with hardware keys ... should we just fall back to signing without?

ee/desktop/user/server/server_darwin.go
```diff
@@ -114,3 +91,97 @@ func (s *UserServer) getSecureEnclaveKey(w http.ResponseWriter, r *http.Request)
 
 	w.Write(keyBytes)
 }
+
+func (s *UserServer) signWithSecureEnclave(w http.ResponseWriter, r *http.Request) {
```
James-Pickett, line 95: TODOS
* verify krypto box from server
* add timestamp and nonce

PR #2142, opened 2025-03-05
Root.json generator

### LINKED ISSUE #1705: Embed the most recent version of root.json in our launcher binary
Currently, we embed the root.json file inside our launcher binary: https://github.com/kolide/launcher/blob/main/ee/tuf/assets/tuf/root.json

We want to change how we embed this file for a couple reasons:

1. The root.json metadata file has a one-year expiration. If the version we are shipping is expired, this can prevent launcher from being able to select the correct version of a binary until the TUF autoupdater exits its initial delay and performs an update.
1. We don't want to have to remember to update the root.json file every time we create a new version.
1. Shipping an outdated version of root.json means that the TUF autoupdater has to perform a couple extra HTTP calls to get the most recent version of root.json.
1. While we _could_ ship an outdated version of root.json and force the TUF autoupdater to immediately fetch the latest version based on that -- via a call to `metadataClient.UpdateRoots()` -- this would slow down launcher startup due to the extra HTTP calls, which we don't want.

Our build process should be updated to embed the most recent version of root.json in the launcher binary.

Notes for implementation:

* We build `launcher` using our `make` command, which lives [here](https://github.com/kolide/launcher/blob/main/cmd/make/make.go). `github-build` is a good place to start looking in the `Makefile` to understand the build process.
* We should hardcode the contents of the first version of the root JSON, which is located at https://tuf.kolide.com/repository/1.root.json -- we want to ensure that we are definitely starting with valid, untampered metadata
* The build process should use the go-tuf library's [UpdateRoots()](https://github.com/theupdateframework/go-tuf/blob/v0.5.2/client/client.go#L205) to update from the first version to the most recent version -- the go-tuf library provides `UpdateRoots()` as a safe way to update and validate root metadata
* [autoupdate.go's initMetadataClient](https://github.com/kolide/launcher/blob/main/ee/tuf/autoupdate.go) may be a good starting point for getting set up to call `UpdateRoots()`
* `pkg/packaging/fetch.go` also relies on a separately-hardcoded `root.json`. Since this code is for packaging launcher into installers (so we don't care as much about performance), and since it already calls `Update()` (which calls `UpdateRoots()`), you can just leave this as it is.

cesarfda:
This pull request introduces a new TUF (The Update Framework) metadata update process to the project. The changes include adding a new `go generate` directive, creating a new Go file to hold the directive, and implementing a generator to update the `root.json` file during the build process. Resolves #1705 

Key changes:

### TUF Metadata Update Process:

* [`Makefile`](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L229-R229): Added the `./ee/tuf/...` directory to the `go generate` command to include the new TUF metadata update process.
* [`ee/tuf/gen.go`](diffhunk://#diff-410f0a3691f92b36c48432b811d893f954d76b335ba4be0885b9cd1dbc6b5f35R1-R6): Created a new file to hold the `go:generate` directive for running the TUF metadata generator.
* [`ee/tuf/generator/generator.go`](diffhunk://#diff-5878f8bbde541fd3210354f1f6c2178edfb3837bcb3c18ad1c181ccbc864877cR1-R152): Implemented the TUF metadata generator, which includes initializing a TUF client, updating the metadata, and writing the updated `root.json` to specified directories.


ee/tuf/generator/generator.go
```diff
@@ -0,0 +1,153 @@
+package main
```
directionless, line 1: I think this can be (and maybe should be) `package generator`

ee/tuf/generator/generator.go
```diff
@@ -0,0 +1,153 @@
...
...6bb63337d1f2cec961dbe68ccafdc874"}},"bd9bbc72f9a1c7cd5998c94b5a31fedb997315f81758f90dbe393743ec984508":{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"aea7938b3b8f9cc0fc910899beb6a025487db87d6c2ea9a12fdf54c25817c814"}},"e434c83555ba93337e91c073095a164e4ae66f01a91cf5e8d4228f2382568d19":{"keytype":"ed25519","scheme":"ed25519","keyid_hash_algorithms":["sha256","sha512"],"keyval":{"public":"785323acb700ba835da550c1ac56300cf87890b4367c7cb1c14c1299e5e37075"}}},"roles":{"root":{"keyids":["bd9bbc72f9a1c7cd5998c94b5a31fedb997315f81758f90dbe393743ec984508"],"threshold":1},"snapshot":{"keyids":["bcd5f36e7a357e7fba38b6b7c531f7fcf509eb2da2cfcdfa94bce99ad18a7317"],"threshold":1},"targets":{"keyids":["e434c83555ba93337e91c073095a164e4ae66f01a91cf5e8d4228f2382568d19"],"threshold":1},"timestamp":{"keyids":["53a3eb57d9f5e755ec73e95429033b78423c9c01cef8c8d4e9911b6fe01198c8"],"threshold":1}},"consistent_snapshot":true},"signatures":[{"keyid":"bd9bbc72f9a1c7cd5998c94b5a31fedb997315f81758f90dbe393743ec984508","sig":"b0624815cf3b8749edc68c96ffca588ce532e9cdf8ac4db4748e260c186ecb3f975d555cf25fe5ddba418f2d3d2af4e5c2a2046b20619ddfd30f3390560f850d"}]}`)
```
directionless, line 22: Feels weird to have this, `ee/tuf/assets/tuf/root.json` is right there. Can we call something there?

ee/tuf/generator/generator.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("updating TUF metadata: %w", err)
+	}
+
+	// Get the latest root.json
+	rootPath := filepath.Join(tempDir, "root.json")
+	latestRoot, err := os.ReadFile(rootPath)
+	if err != nil {
+		return fmt.Errorf("reading updated root.json: %w", err)
+	}
```
RebeccaMahany, line 121: I would probably pull this right from `localStore` instead -- you should be able to do this --

```suggestion
	metadata, err := localStore.GetMeta()
	if err != nil {
		return fmt.Errorf("reading updated metadata from local store: %w", err)
	}
	latestRoot = metadata["root.json"]
```

ee/tuf/generator/generator.go
RebeccaMahany: I was wondering if we'd be able to move either `ee/tuf/assets/tuf/root.json` to a location where we are able to embed it here, or whether we'd be able to move the generator to `ee/tuf` to be able to straightforwardly embed it. Looking at it further, I don't actually think either of those options would work, unfortunately (though maybe worth confirming that just to be sure!).

I don't think it's terrible to have the duplication, either. `ee/tuf/assets/tuf/root.json` is the latest version, while this is the known/validated first version.

ee/tuf/generator/generator.go
RebeccaMahany: It looks like `package main` is how our existing `ee/ui/assets/generator/generator.go` works, fwiw

ee/tuf/generator/generator.go
directionless: I can get behind having one that's checked in and lagging, and one that's generated and used at build time. _Maybe_. 

But I don't like having 3 checked in.

ee/tuf/generator/generator.go
RebeccaMahany: Gah, sorry about the `latestRoot = metadata["root.json"]` / `latestRoot := metadata["root.json"]` typo!

ee/tuf/generator/generator.go
RebeccaMahany: Are we able to remove `ee/tuf/assets/tuf/root.json` and the one in `packaging` from source control and add those files to our .gitignore without anything complaining? If so, sounds like that might just work. 🙂

RebeccaMahany: Looks like it's working! So we now only have one checked in. 🥳

RebeccaMahany (approved):
Nice work, LGTM!

ee/tuf/generator/generator.go
directionless: My feeling is that there's value in having one checked in -- using it as the start point for the build time generation, and then building in whatever this does. 

That ties the chain of trust to something in git, and gets a current version into the build

.gitignore
```diff
@@ -24,6 +24,7 @@ test-cmds
 # Generated assets
 pkg/packagekit/internal/assets.go
 pkg/packagekit/wix/internal/assets.go
+pkg/packaging/assets/tuf/root.json
```
directionless, line 27: Why are we ignoring two root.jsons here?

ee/tuf/generator/generator.go
directionless, line 22: I would put this into an `assets` folder somewhere, and not inline it here. The problem with inlineing it, is that it makes it ugly when we update it.

ee/tuf/generator/generator.go
```diff
@@ -0,0 +1,153 @@
...
+
+func main() {
+	logger := logutil.NewCLILogger(true)
+
+	flagset := flag.NewFlagSet("tuf-generator", flag.ExitOnError)
+	var (
+		flDebug        = flagset.Bool("debug", false, "use a debug logger")
+		flTufURL       = flagset.String("tuf-url", "https://tuf.kolide.com", "TUF repository URL")
+		flMetadataPath = flagset.String("metadata-path", "/repository", "TUF metadata path")
+		flOutputDirs   = flagset.String("output-dirs", "./assets/tuf,../../pkg/packaging/assets/tuf", "comma-separated output directories")
```
directionless, line 32: Do we need to write this into 2 places?

.gitignore
RebeccaMahany: The packaging code and the tuf code both need a validated root.json. (The packaging code uses it to get latest validated TUF metadata, so that it can pick the correct version for the given channel.)

RebeccaMahany: (Also answers https://github.com/kolide/launcher/pull/2142#discussion_r1998947297)

ee/tuf/generator/generator.go
RebeccaMahany: This is version 1, so we shouldn't ever update it

ee/tuf/generator/generator.go
directionless: I guess that makes sense. Feels like a thing to clean up, but maybe not this PR.

directionless (dismissed):
Nice. Let's try!

ee/tuf/generator/generator.go
```diff
@@ -17,9 +17,8 @@ import (
...
+// Path to the initial root.json file
+const initialRootPath = "./assets/tuf/initial_root.json"
```
RebeccaMahany, line 21: Agh, sorry for another nitpick on the format here -- are we able to use an embed to avoid having to do the `os.ReadFile` below? I think it would work if we move `ee/tuf/assets/tuf/initial_root.json` inside this path (i.e. `ee/tuf/generator/assets/initial_root.json`) instead, and then do --

```
//go:embed assets/initial_root.json
var initialRootJSON []byte
```

cesarfda: Done

RebeccaMahany (approved):
🔥 🔥

ee/tuf/generator/generator.go
RebeccaMahany: @directionless FYI -- we ended up having to back part of this change out (https://github.com/kolide/launcher/pull/2154). When we pull in the `pkg/packaging` package in our package building tool, it expects the asset file to already exist, and we can't generate it from inside the other repository. We'll have to revisit this later to see if we can figure out how to remove that one, but this will at least let us build packages with the new changes for now!

PR #2143, opened 2025-03-06
test win arm runner

James-Pickett:


James-Pickett:
I think the arm images are still not ready for prime time. It looks like github / msft is relying on 3rd parties to handle them. The first issue I've encountered is that the windows arm runner seems to be base windows images. It doesn't include the default tools like bash or even powershell core (pwsh) that we rely on in our pipe line. Here are some relevant issues in the actions/partner-runner-images repo.

https://github.com/actions/partner-runner-images/issues/62
https://github.com/actions/partner-runner-images/issues/55

Issue #2055, opened 2025-01-22
Backfill support for all datatypes for Firefox and Chrome extension databases


RebeccaMahany:
We've been adding support for datatypes as we see them come up, which means that by the time we hit a parsing error, we need to do a full launcher release before end users will see the fix reflected.

We should proactively go through the Chrome/Firefox code and add support to parsing for all data types we haven't yet covered.

Go through the Chrome and Firefox code (linked [here](https://github.com/kolide/launcher/tree/main/ee/katc/test_data#general-object-deserialization-process) at the bottom of this section) and compare the data types listed there with our [Firefox](https://github.com/kolide/launcher/blob/main/ee/katc/deserialize_firefox.go) and [Chrome](https://github.com/kolide/launcher/blob/main/ee/indexeddb/values.go) parsing support.

For each data type that we don't yet cover:

- [ ] Add the data type to our test databases for both Chrome and Firefox, following the instructions here: https://github.com/kolide/launcher/tree/main/ee/katc/test_data#instructions-for-generating-new-test-indexeddbs
- [ ] Implement parsing for the data type for Firefox: https://github.com/kolide/launcher/blob/main/ee/katc/deserialize_firefox.go
- [ ] Implement parsing for the data type for Chrome: https://github.com/kolide/launcher/blob/main/ee/indexeddb/values.go

directionless:
I feel like I asked this before, but is there a way to skip the unparsable values?

RebeccaMahany:
@directionless not reliably, no -- when we get to a value with an unsupported data type tag, we aren't able to advance through it confidently to the next key-value pair because values have variable lengths, and may contain within them characters that also denote data type tags.

RebeccaMahany:
If anyone else is looking to grab work on this ticket, I think `Error` might be the next quickest option to grab!

* `kError` for Chrome: https://github.com/v8/v8/blob/master/src/objects/value-serializer.cc#L164-L166
* `SCTAG_ERROR_OBJECT` for Firefox: https://searchfox.org/mozilla-central/source/js/src/vm/StructuredClone.cpp#148
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

PR #2148, opened 2025-03-13
[IndexedDB/KATC] TypedArrays and ArrayBuffers

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/2055

TypedArrays plus ArrayBuffers. Also covers back-reference objects, albeit in an incomplete way.

It might be helpful to skim the [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) and [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) documentation before reviewing this PR, if unfamiliar!


ee/indexeddb/values.go
```diff
@@ -473,6 +487,234 @@ func deserializeDenseArray(ctx context.Context, slogger *slog.Logger, srcReader
...
+	return result, nil
+}
+
+// readUint32Array is suitable for reading TypedArrays: Uint32Array, Int32Array
+func readUint32Array(typedArrayReader *bytes.Reader) ([]uint32, error) {
+	arrayLen := typedArrayReader.Len() / 4
+	result := make([]uint32, arrayLen)
+	for i := 0; i < arrayLen; i++ {
+		if err := binary.Read(typedArrayReader, binary.NativeEndian, &result[i]); err != nil {
+			return nil, fmt.Errorf("reading uint8 at index %d in TypedArray: %w", i, err)
```
zackattack01, line 640: should these error messages all be updated as uint16/32 etc

ee/indexeddb/values.go
```diff
@@ -473,6 +487,234 @@ func deserializeDenseArray(ctx context.Context, slogger *slog.Logger, srcReader
...
+	if byteLength != 0 {
+		return nil, fmt.Errorf("found standalone array buffer view with length %d despite no preceding data", byteLength)
+	}
+
+	// Next, we expect an array termination byte.
+	nextByte, err := srcReader.ReadByte()
+	if err != nil {
+		return nil, fmt.Errorf("reading next byte at end of dense array: %w", err)
+	}
+	if nextByte != 0x03 {
```
zackattack01, line 711: nit- might be nice to have a const for this if it consistently means array termination, seems like we reference and disregard in the other deserializeArray helpers?

RebeccaMahany: I am sort of guessing that it means array termination -- I haven't found it documented anywhere in the source code so I've inferred based on the context of where I've seen this byte pop up. I will make this a const and document that assumption more clearly!

ee/katc/deserialize_firefox.go
```diff
@@ -436,6 +442,222 @@ func deserializeArray(arrayLength uint32, srcReader *bytes.Reader) ([]byte, erro
...
+		uint8Arr := make([]uint8, length)
+		for i := 0; i < int(length); i++ {
+			if err := binary.Read(typedArrayReader, binary.NativeEndian, &uint8Arr[i]); err != nil {
+				return nil, fmt.Errorf("reading %T at index %d in TypedArray: %w", uint8Arr[i], i, err)
+			}
+		}
+		results = uint8Arr
+	case typedArrayInt16:
+		int16Arr := make([]int16, length)
+		for i := 0; i < int(length); i++ {
```
zackattack01, line 564: just a q because i think I'm misunderstanding something- why don't we need to manipulate the length for these to get the expected number of iterations based on the types being read in, similar to e.g. `readFloat64Array`?

zackattack01: ahh okay i see, that is what i missed. thank you!

James-Pickett (approved):

PR #2149, opened 2025-03-13
james/dt4a auth

James-Pickett:
adds chain of trust / NaCl auth middleware to dt4a endpoint

relates to https://github.com/kolide/k2/issues/11530

### DIFF
--- a/ee/localserver/jwk_keys.go
+++ b/ee/localserver/jwk_keys.go
@@ -20,13 +20,19 @@ type jwk struct {
 	KeyID string `json:"kid"`
 }
 
+const (
+	curveP256 string = "P-256"
+	curveP384 string = "P-384"
+	curveP521 string = "P-521"
+)
+
 func parseEllipticCurve(str string) (elliptic.Curve, error) {
-	switch strings.ToLower(str) {
-	case "p-256":
+	switch strings.ToUpper(str) {
+	case curveP256:
 		return elliptic.P256(), nil
-	case "p-384":
+	case curveP384:
 		return elliptic.P384(), nil
-	case "p-521":
+	case curveP521:
 		return elliptic.P521(), nil
 	default:
 		return &elliptic.CurveParams{}, fmt.Errorf("unsupported curve: %s", str)
--- a/ee/localserver/zta_auth_middleware_test.go
+++ b/ee/localserver/zta_auth_middleware_test.go
@@ -358,11 +358,11 @@ func toJWK(key any, kid string) (*jwk, error) {
 		var crv string
 		switch k.Curve {
 		case elliptic.P256():
-			crv = "P-256"
+			crv = curveP256
 		case elliptic.P384():
-			crv = "P-384"
+			crv = curveP384
 		case elliptic.P521():
-			crv = "P-521"
+			crv = curveP521
 		default:
 			return nil, fmt.Errorf("unsupported elliptic curve, %s", crv)
 		}


ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (z *ztaAuthMiddleware) Wrap(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		boxParam := r.URL.Query().Get("payload")
+		if boxParam == "" {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		boxBytes, err := base64.StdEncoding.DecodeString(boxParam)
```
James-Pickett, line 33: should use url decoding

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,233 @@
...
+	// make a copy of the root key so that we can reassign it
+	parentEcdsa := &ecdsa.PublicKey{
+		Curve: rootKey.Curve,
+		X:     rootKey.X,
+		Y:     rootKey.Y,
+	}
+
+	if !ok {
+		return fmt.Errorf("root key with kid %s not found in trusted keys", c.Links[0].SignedBy)
+	}
```
James-Pickett, line 171:
```suggestion
```

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,360 @@
...
+	t.Run("handles missing box param", func(t *testing.T) {
+		t.Parallel()
+		rr := httptest.NewRecorder()
+		handler.ServeHTTP(rr, httptest.NewRequest(http.MethodGet, "/", nil))
+		require.Equal(t, http.StatusBadRequest, rr.Code,
+			"should return bad request when box param is missing",
+		)
+	})
+
+	t.Run("handles missing bad b64", func(t *testing.T) {
```
James-Pickett, line 51:
```suggestion
	t.Run("handles bad b64", func(t *testing.T) {
```

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,360 @@
...
+		t.Parallel()
+		rr := httptest.NewRecorder()
+		encoded := base64.StdEncoding.EncodeToString([]byte("badchain"))
+		handler.ServeHTTP(rr, httptest.NewRequest(http.MethodGet, fmt.Sprintf("/?payload=%s", encoded), nil))
+		require.Equal(t, http.StatusBadRequest, rr.Code,
+			"should return bad request when chain cannot be unmarshalled",
+		)
+	})
+
+	t.Run("handles invalid chain unmarshall failure", func(t *testing.T) {
```
James-Pickett, line 70: I think?
```suggestion
	t.Run("handles invalid chain", func(t *testing.T) {
```

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,360 @@
...
+func newChain(counterPartyPubEncryptionKey *[32]byte, ecdsaKeys ...*ecdsa.PrivateKey) (*chain, error) {
+	if len(ecdsaKeys) == 0 {
+		return nil, errors.New("no keys provided")
+	}
+
+	if len(ecdsaKeys) == 1 {
+		return nil, errors.New("only one key provided")
+	}
+
+	// data will be last link in links
```
James-Pickett, line 282:
```suggestion
	// counterPartyPubEncryptionKey will be last link in links
```

ee/localserver/server.go
```diff
@@ -104,8 +104,17 @@ func New(ctx context.Context, k types.Knapsack, presenceDetector presenceDetecto
...
+		return nil, fmt.Errorf("loading dt4a keys %w", err)
+	}
+
+	ztaAuthMiddleware := &ztaAuthMiddleware{
+		counterPartyKeys: dt4aKeys,
+	}
+
 	// In the future, we will want to make this authenticated; for now, it is not authenticated.
-	mux.Handle("/zta", ls.requestZtaInfoHandler())
+	mux.Handle("/zta", ztaAuthMiddleware.Wrap(ls.requestZtaInfoHandler()))
```
James-Pickett, line 117: rename endpoint to dt4a

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+type ztaResponse struct {
+	Data   []byte    `json:"data"`
+	PubKey *[32]byte `json:"pubKey"`
+}
+
+func (z *ztaAuthMiddleware) Wrap(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		boxParam := r.URL.Query().Get("payload")
```
James-Pickett, line 28: add logging here

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+	// counterPartyPubEncryptionKey is the last public key in the chain, which is a x25519 key
+	// we set this after we extract and verify in the validate method
+	counterPartyPubEncryptionKey *[32]byte
+}
+
+// chainLink represents a link in a chain of trust
+type chainLink struct {
+	// Payload a b64 url encoded json
+	Payload string `json:"payload"`
+	// Signature is the raw bytes of the signature of the b64 url encoded json
```
James-Pickett, line 170:
```suggestion
	// Signature is b64 url encode of signature
```

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,336 @@
...
+		payloadBytes, err := json.Marshal(thisPayload)
+		if err != nil {
+			return nil, err
+		}
+
+		payloadB64 := base64.URLEncoding.EncodeToString(payloadBytes)
+
+		sig, err := echelper.Sign(parentKey, []byte(payloadB64))
+		if err != nil {
+			return nil, fmt.Errorf("failed to self sign root key: %w", err)
```
James-Pickett, line 325:
```suggestion
			return nil, err
```

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+type ztaResponse struct {
+	Data   []byte    `json:"data"`
+	PubKey *[32]byte `json:"pubKey"`
+}
+
+func (z *ztaAuthMiddleware) Wrap(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		boxParam := r.URL.Query().Get("payload")
+		if boxParam == "" {
+			z.slogger.Log(r.Context(), slog.LevelDebug,
```
RebeccaMahany, line 32: Do we potentially want to log these errors at a higher level than debug, so we ship them to the cloud? `LevelWarn` or `LevelError` both seem appropriate to me

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+				"failed to decode payload",
+				"err", err,
+			)
+
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		var myChain chain
```
RebeccaMahany, line 51: Instead of `myChain`, maybe something more specific like `requestTrustChain` or similar?

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+				"err", err,
+			)
+
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		newReq := &http.Request{
+			Method: http.MethodGet,
+			Header: make(http.Header),
```
RebeccaMahany, line 74: I think we want to include the original headers if available here, so that we can perform our `Origin` header check

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+				"failed to validate chain",
+				"err", err,
+			)
+
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		newReq := &http.Request{
+			Method: http.MethodGet,
```
RebeccaMahany, line 73: Given this, I think we can maybe remove the `r.Method != http.MethodGet` check from `zta.go`? I guess fine to leave it, too

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	p.PublicKey = key
+
+	return nil
+}
+
+// chain represents a chain of trust, where each link in the chain has a payload signed by the key in the previous link.
+// The first link in the chain is signed by a trusted root key.
+// There can by any number if keys between the root and the last key.
+// The last key is an∂ x25519 key which is the public half of a NaCl box key.
```
RebeccaMahany, line 158: Nitpick --

```suggestion
// There can be any number of keys between the root and the last key.
// The last key is an x25519 key which is the public half of a NaCl box key.
```

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/kolide/krypto/pkg/echelper"
+	"github.com/lestrrat-go/jwx/jwk"
+)
+
+type ztaAuthMiddleware struct {
+	counterPartyKeys map[string]*ecdsa.PublicKey
```
RebeccaMahany, line 19: Small documentation suggestion, since I see we document the format of `counterPartyKeys` later in chain.validate too --

```suggestion
	counterPartyKeys map[string]*ecdsa.PublicKey // maps KID to pubkey
```

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	// make a copy of the root key so that we can reassign it
+	parentEcdsa := &ecdsa.PublicKey{
+		Curve: rootKey.Curve,
+		X:     rootKey.X,
+		Y:     rootKey.Y,
+	}
+
+	var lastPayload payload
```
RebeccaMahany, line 223: I think this would be more accurate as `currentPayload`, not `lastPayload` -- `lastPayload` to me implies `c.Links[i-1]`, but this is the payload for `c.Links[i]`. (I get that the variable name is more accurate once we exit the for loop and start processing the counter party key, but it's confusing during the for loop -- I thought we were trying to refer to a previous link payload to validate the previous one.)

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+
+	for i := range len(c.Links) {
+
+		signature, err := base64.URLEncoding.DecodeString(c.Links[i].Signature)
+		if err != nil {
+			return fmt.Errorf("failed to decode signature: %w", err)
```
RebeccaMahany, line 229: Do you think it'd be worth it to add `"at index %d", i` to all these error messages? Or not really helpful

ee/localserver/zta_auth_middleware.go
```diff
@@ -0,0 +1,285 @@
...
+			return fmt.Errorf("failed to decode payload: %w", err)
+		}
+
+		if err := json.Unmarshal(payloadBytes, &lastPayload); err != nil {
+			return fmt.Errorf("failed to unmarshal payload: %w", err)
+		}
+
+		if lastPayload.ExpirationDate < time.Now().Unix() {
+			return fmt.Errorf("payload %d has expired", i)
+		}
```
RebeccaMahany, line 247: Do we also need/want to check that `payload.DateTimeSigned` is not too far in the past, or is this time check is sufficient?

RebeccaMahany (commented):
Made it through everything but the test, will come back to look at that later! Looks really clean.

ee/localserver/zta_auth_middleware.go
James-Pickett: Nice catch! I was considering moving the origin checks over to the auth middle ware as well, but decided to wait until we got an E2E, so there is no chance of breaking folks using current /zta endpoint

ee/localserver/zta_auth_middleware.go
James-Pickett: definitely, I added that, as well as the kid when when possible

ee/localserver/zta_auth_middleware.go
James-Pickett: I believe this is sufficient, the server side is just `payload.DateTimeSigned + TTL`

ee/localserver/zta_auth_middleware.go
RebeccaMahany: I think totally reasonable to move the origin checks here in the future too!

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+	t.Run("handles chain unmarshall failure", func(t *testing.T) {
+		t.Parallel()
+		rr := httptest.NewRecorder()
+		encoded := base64.URLEncoding.EncodeToString([]byte("badchain"))
+		handler.ServeHTTP(rr, httptest.NewRequest(http.MethodGet, fmt.Sprintf("/?payload=%s", encoded), nil))
+		require.Equal(t, http.StatusBadRequest, rr.Code,
+			"should return bad request when chain cannot be unmarshalled",
+		)
+	})
```
RebeccaMahany, line 68: Maybe worth adding a similar test confirming we return a `http.StatusBadRequest` for an empty chain? i.e. I think `encoded := base64.URLEncoding.EncodeToString([]byte("[]"))`.

I see there's a test below in `Test_ValidateCertChain` that covers empty chain, so maybe not needed?

ee/localserver/zta_auth_middleware_test.go
```diff
@@ -0,0 +1,336 @@
...
+			"should return ok when chain is valid",
+		)
+
+		bodyBytes, err := io.ReadAll(rr.Body)
+		require.NoError(t, err)
+
+		bodyDecoded, err := base64.StdEncoding.DecodeString(string(bodyBytes))
+		require.NoError(t, err)
+
+		var ztaResponse ztaResponse
```
RebeccaMahany, line 129: Nitpick -- maybe `var z ztaResponse` or `var resp ztaResponse` to avoid shadowing?

ee/localserver/certs.go
```diff
@@ -16,4 +26,162 @@ GINysvEwYrNoGjASt+nqzlFesagt+2A/4W7JR16nE91mbCHn+HV6x+H8gw==
 MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEwowFsPUaOC61LAfDz1hLnsuSDfEx
 SC4TSfHtbHHv3lx2/Bfu+H0szXYZ75GF/qZ5edobq3UkABN6OaFnnJId3w==
 -----END PUBLIC KEY-----`
+
+	dt4aCertsJWK = `
```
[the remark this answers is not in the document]
James-Pickett: I'm in favor of the const string, the keys were provided to me in that format and it just seemed easier to make a mistake by actually instantiating them directly to something else. This only happens once during start up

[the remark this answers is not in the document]
RebeccaMahany: Fair enough!

RebeccaMahany (commented):
Tests look great and really thorough!

ee/localserver/zta_auth_middleware_test.go
James-Pickett: I think it's a legit test, easy to add, done

ee/localserver/certs.go
```diff
@@ -16,4 +26,162 @@ GINysvEwYrNoGjASt+nqzlFesagt+2A/4W7JR16nE91mbCHn+HV6x+H8gw==
...
+		x := new(big.Int).SetBytes(xBytes)
+
+		yBytes, err := base64.RawURLEncoding.DecodeString(v["y"])
+		if err != nil {
+			return nil, err
+		}
+		y := new(big.Int).SetBytes(yBytes)
+
+		pubKey := &ecdsa.PublicKey{
+			Curve: elliptic.P256(),
```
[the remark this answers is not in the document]
zackattack01: oh nice! 🚀

ee/localserver/server.go
```diff
@@ -104,9 +104,23 @@ func New(ctx context.Context, k types.Knapsack, presenceDetector presenceDetecto
 	// /v0/cmd left for transition period
 	mux.Handle("/v1/cmd", ecKryptoMiddleware.Wrap(ls.munemoCheckHandler(ecAuthedMux)))
 
+	dt4aKeys, err := dt4aKeys()
```
RebeccaMahany, line 107: Ah sorry, should've noticed earlier -- if you end up pushing other changes, could you also rename this variable to `trustedDt4aKeys` or `keys` to avoid shadowing?

RebeccaMahany (approved):
LGTM, nice work!

ee/localserver/server.go
James-Pickett: done

James-Pickett: almost wonder if we should look for a "shadow linter"

RebeccaMahany: There are a couple options! I think I looked at enabling the golangci-lint `govet => shadow` option, but didn't end up turning it on because it's idiomatic to shadow `err` and `ctx`, so we have a bunch of violations for that. But maybe it would be simple to add an exception for `err` and `ctx`.

RebeccaMahany (dismissed):

zackattack01 (dismissed):
nice! great docs/comments throughout 🔥

ee/localserver/certs.go
```diff
@@ -1,6 +1,15 @@
 package localserver
 
+import (
+	"crypto/ecdsa"
+	"encoding/json"
+	"fmt"
+
+	"github.com/lestrrat-go/jwx/jwk"
```
directionless, line 8: Do we need to pull in this library? It feels a little sprawling, and I feel very -_- about third parties these days

ee/localserver/server.go
```diff
@@ -104,9 +104,23 @@ func New(ctx context.Context, k types.Knapsack, presenceDetector presenceDetecto
...
+		counterPartyKeys: dt4aKeys,
+		slogger:          k.Slogger().With("component", "dt4aAuthMiddleware"),
+	}
+
 	// In the future, we will want to make this authenticated; for now, it is not authenticated.
+	// TODO: make this authenticated or remove
 	mux.Handle("/zta", ls.requestZtaInfoHandler())
 
+	// mux.Handle("/zta", ztaAuthMiddleware.Wrap(ls.requestZtaInfoHandler()))
+	mux.Handle("/v0/dt4a", ztaAuthMiddleware.Wrap(ls.requestZtaInfoHandler()))
```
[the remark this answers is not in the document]
James-Pickett: yea, it grew on me too, updated to `/v3/dt4a`

ee/localserver/certs.go
James-Pickett: yea, it had quite a few dependencies, I dropped it and now just using std lib. More code to handle key conversions, but once separated, makes core logic code simpler.

ee/localserver/certs.go
```diff
@@ -152,25 +150,18 @@ func dt4aKeys() (map[string]*ecdsa.PublicKey, error) {
...
-		// Note that a successful parsing of any type of key does NOT necessarily guarantee a valid key.
-		jwkKey, err := jwk.ParseKey(v)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse key, kid %s: %w", k, err)
-		}
 
-		var pubKey ecdsa.PublicKey
-		if err := jwkKey.Raw(&pubKey); err != nil {
-			return nil, fmt.Errorf("failed to convert public key to public ecdsa, kid %s: %w", k, err)
+		var jwk jwk
```
RebeccaMahany, line 154: `var j jwk` to avoid shadowing? (I did put in the linter changes -- https://github.com/kolide/launcher/pull/2155 -- but fixing violations is a bit of a sprawling changeset so I was going to wait on pushing those.)

ee/localserver/jwk_keys.go
```diff
@@ -0,0 +1,89 @@
...
+type jwk struct {
+	Type  string `json:"kty"`
+	Curve string `json:"crv"`
+	X     string `json:"x"`
+	Y     string `json:"y"`
+	KeyID string `json:"kid"`
+}
+
+// jwkToECDSAPublicKey converts a JWK JSON string into an ECDSA public key.
+func (jwk jwk) ecdsaPubKey() (*ecdsa.PublicKey, error) {
```
RebeccaMahany, line 21: Same shadowing thing, plus maybe we want a pointer receiver here? `func (j *jwk)`? Ditto for `x25519PubKey` below.

ee/localserver/jwk_keys.go
```diff
@@ -0,0 +1,89 @@
...
+	// under the hood it calls Curve.IsOnCurve(...), but if you call that directly
+	// you get deprecated warnings
+	if _, err := pubKey.ECDH(); err != nil {
+		return nil, fmt.Errorf("invalid public key: %w", err)
+	}
+
+	return pubKey, nil
+}
+
+// jwkToX25519PublicKey converts jwk in to x25519 key (*[32]byte)
```
RebeccaMahany, line 71:
```suggestion
// x25519PubKey converts jwk into x25519 key (*[32]byte)
```

ee/localserver/jwk_keys.go
```diff
@@ -0,0 +1,89 @@
...
+	Curve string `json:"crv"`
+	X     string `json:"x"`
+	Y     string `json:"y"`
+	KeyID string `json:"kid"`
+}
+
+// jwkToECDSAPublicKey converts a JWK JSON string into an ECDSA public key.
+func (jwk jwk) ecdsaPubKey() (*ecdsa.PublicKey, error) {
+	// Ensure the key type is EC.
+	if jwk.Type != "EC" {
```
RebeccaMahany, line 23: For documentation, I think it would be helpful to add a link to https://datatracker.ietf.org/doc/html/rfc7518#section-6.2, and note that the jwk.Type and jwk.Curve values are case-sensitive. Could maybe move `"EC"`, `"P-256"`, `"P-384"`, `"P-521"` to consts so we can document all that in one place?

ee/localserver/jwk_keys.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"math/big"
+)
+
+type jwk struct {
```
RebeccaMahany, line 12: Do you think this could potentially be worth moving to its own `ee/jwk` package, in case we want to use it elsewhere in the future? (I'm fine with leaving it here, or moving it in a later PR.)

James-Pickett: I considered that, but my gut feeling is that if were to do an `ee/jwk` we should probably fully implement the jwk spec and at that point my as well use the pkg?

RebeccaMahany (dismissed):
🔥

ee/localserver/jwk_keys.go
James-Pickett: fixed

PR #2153, opened 2025-03-17
Requests to the control server should have timeouts set

RebeccaMahany:
We are investigating an issue where sometimes we repeatedly see launcher failing to make requests to the control server, with logs containing `fetch is currently executing elsewhere`.

We are not currently setting timeouts when making requests to the control server; this PR updates to always set a timeout of 30s when making requests, so that any one request to the control server can't prevent `fetch` from executing indefinitely.

### DIFF
--- a/ee/control/client_http.go
+++ b/ee/control/client_http.go
@@ -72,23 +72,17 @@ func (c *HTTPClient) GetConfig(ctx context.Context) (io.Reader, error) {
 	ctx, span := traces.StartSpan(ctx)
 	defer span.End()
 
-	challengeCtx, challengeCancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer challengeCancel()
-
-	challengeReq, err := http.NewRequestWithContext(challengeCtx, http.MethodGet, c.url("/api/agent/config").String(), nil)
+	challengeReq, err := http.NewRequestWithContext(ctx, http.MethodGet, c.url("/api/agent/config").String(), nil)
 	if err != nil {
 		return nil, fmt.Errorf("could not create challenge request: %w", err)
 	}
 
-	challenge, err := c.do(challengeCtx, challengeReq)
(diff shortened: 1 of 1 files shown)


ee/control/client_http.go
```diff
@@ -70,17 +72,23 @@ func (c *HTTPClient) GetConfig(ctx context.Context) (io.Reader, error) {
 	ctx, span := traces.StartSpan(ctx)
 	defer span.End()
 
-	challengeReq, err := http.NewRequestWithContext(ctx, http.MethodGet, c.url("/api/agent/config").String(), nil)
+	challengeCtx, challengeCancel := context.WithTimeout(ctx, defaultRequestTimeout)
```
James-Pickett, line 75: this works, but couldn't we simplify just by putting the logic in the `do()` func?

something like
```go

	ctx, cancel := context.WithTimeout(req.Context(), defaultRequestTimeout)
	defer cancel()

	req = req.WithContext(ctx)
```

James-Pickett (approved):

ee/control/client_http.go
RebeccaMahany: I wasn't sure if we might want different timeouts in the future for different endpoints

James-Pickett: fair enough

RebeccaMahany: Decided we probably didn't want different timeouts, so I updated to move it to `do()`!

Missed your other comment -- I think we might've only been passing that ctx in for the traces anyway. 😅 I'll update!

zackattack01 (approved):

ee/control/client_http.go
RebeccaMahany: Okay, fixed both things! LMK how it looks

zackattack01 (approved):

James-Pickett (approved):

PR #2156, opened 2025-03-18
rename zta to dt4a as much as possible

James-Pickett:
Renames everything from zta to dt4a, except a handful of literals where things would break if name changed.

### DIFF
--- a/cmd/launcher/launcher.go
+++ b/cmd/launcher/launcher.go
@@ -74,7 +74,7 @@ const (
 	desktopMenuSubsystemName = "kolide_desktop_menu"
 	authTokensSubsystemName  = "auth_tokens"
 	katcSubsystemName        = "katc_config" // Kolide ATC
-	ztaInfoSubsystemName     = "zta_info"
+	dt4aInfoSubsystemName    = "zta_info"
 )
 
 // runLauncher is the entry point into running launcher. It creates a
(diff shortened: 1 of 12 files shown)


ee/localserver/dt4a.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 var (
-	localserverZtaInfoKey = []byte("localserver_zta_info")
+	localserverDt4aInfoKey = []byte("localserver_dt4a_info")
```
James-Pickett, line 12: I think this needs to stay for now

ee/localserver/server.go
```diff
@@ -109,17 +109,17 @@ func New(ctx context.Context, k types.Knapsack, presenceDetector presenceDetecto
...
-	ztaAuthMiddleware := &ztaAuthMiddleware{
+	dt4aAuthMiddleware := &dt4aAuthMiddleware{
 		counterPartyKeys: trustedDt4aKeys,
 		slogger:          k.Slogger().With("component", "dt4a_auth_middleware"),
 	}
 
 	// In the future, we will want to make this authenticated; for now, it is not authenticated.
 	// TODO: make this authenticated or remove
-	mux.Handle("/zta", ls.requestZtaInfoHandler())
+	mux.Handle("/dt4a", ls.requestDt4aInfoHandler())
```
James-Pickett, line 119: probably needs to stay for now too

cmd/launcher/launcher.go
```diff
@@ -74,7 +74,7 @@ const (
 	desktopMenuSubsystemName = "kolide_desktop_menu"
 	authTokensSubsystemName  = "auth_tokens"
 	katcSubsystemName        = "katc_config" // Kolide ATC
-	ztaInfoSubsystemName     = "zta_info"
+	dt4aInfoSubsystemName    = "zta_info"
```
RebeccaMahany, line 77: For an easier transition between the old and new subsystem names, could we support two subsystems and have them handled by the same consumer and store? i.e. something like --

```golang
ztaInfoSubsystemName     = "zta_info"
dt4aInfoSubsystemName    = "dt4a_info"

// ...

// Set up consumer to receive DT4A info from the control server in both current and legacy subsystem
dt4aInfoConsumer := keyvalueconsumer.NewConfigConsumer(k.Dt4aInfoStore())
if err := controlService.RegisterConsumer(ztaInfoSubsystemName, dt4aInfoConsumer); err != nil {
	return fmt.Errorf("failed to register ZTA info consumer: %w", err)
}
if err := controlService.RegisterConsumer(dt4aInfoSubsystemName, dt4aInfoConsumer); err != nil {
	return fmt.Errorf("failed to register dt4a info consumer: %w", err)
}
```

James-Pickett: sure, that makes sense, done

zackattack01 (approved):

RebeccaMahany (dismissed):

zackattack01 (approved):

PR #2157, opened 2025-03-18
Allow for manipulating osquery's effective distributed_interval within 5 seconds

RebeccaMahany:
We need osquery to respond to distributed queries very quickly -- we think a distributed_interval of 5 seconds is adequate. However, we do not think the cloud can handle receiving `GetQueries` at that rate.

Therefore, this PR updates the osquery configuration to have a `distributed_interval` of 5 seconds, and updates the extension to only forward those requests to the cloud once per minute. There are two exceptions to this rule:

1. If the extension receives a `GetQueries` response that includes `accelerate` in the response, it will forward all requests for the given number of seconds -- this matches how osquery would respond to receiving the `accelerate` field in the response.
2. On startup, the extension will also forward every request for the first two minutes -- this is similar to prior extension behavior (the period was longer than two minutes before).

In the future, there will be a third exception: launcher will be able to additionally toggle this setting, to allow for accelerating the distributed_interval on demand via e.g. localserver.

This PR also updates the `GetQueries` logging middleware to log successes as well as failures. The successes will be logged at the debug level, so they generally won't be shipped to our cloud logs. I think this will be useful to have in flares, to determine whether launcher is requesting distributed queries, but am not sure if it will be too noisy, so I'd like to get input on whether we think it's a good idea to add this logging.

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -68,26 +68,27 @@ const (
 	// (applies per log type).
 	defaultMaxBufferedLogs = 500000
 
-	// How frequently osquery should check for distributed queries to run during
-	// osquery instance startup -- an accelerated interval, where the unit is seconds.
-	// This shorter interval should hopefully help launcher begin to process stale,
-	// slow-running queries as quickly as possible after device startup.
-	startupDistributedInterval = 5
+	// How frequently osquery should check for distributed queries to run.
+	// We set this to 5 seconds, which is more frequent than we think the
+	// server can comfortably handle, so we only forward these requests
+	// to the cloud once every 60 seconds.
+	osqueryDistributedInterval = 5
 )
 
 var (
 	// Osquery configuration options that we set during the osquery instance's startup period.
 	// These are the override, non-standard values.
 	startupOsqueryConfigOptions = map[string]any{
-		"verbose":              true,                       // receive as many osquery logs as we can, in case we need to troubleshoot an issue
-		"distributed_interval": startupDistributedInterval, // request distributed queries more frequently, so stale checks run ASAP
+		"verbose":              true, // receive as many osquery logs as we can, in case we need to troubleshoot an issue
+		"distributed_interval": osqueryDistributedInterval,
 	}
 
 	// Osquery configuration options that we set after the osquery instance has been up and running
-	// for at least 10 minutes. These are the "normal" values. We don't set the distributed interval
-	// here so that the cloud can set it instead.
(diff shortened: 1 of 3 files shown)


James-Pickett (approved):
nice, pretty clean

pkg/osquery/extension.go
```diff
@@ -68,26 +71,27 @@ const (
...
 )
 
 var (
 	// Osquery configuration options that we set during the osquery instance's startup period.
 	// These are the override, non-standard values.
 	startupOsqueryConfigOptions = map[string]any{
-		"verbose":              true,                       // receive as many osquery logs as we can, in case we need to troubleshoot an issue
-		"distributed_interval": startupDistributedInterval, // request distributed queries more frequently, so stale checks run ASAP
+		"verbose":              true, // receive as many osquery logs as we can, in case we need to troubleshoot an issue
+		"distributed_interval": osqueryDistributedInterval,
```
directionless, line 86: Seems a little weird to have this in both `startupOsqueryConfigOptions` and `postStartupOsqueryConfigOptions`. I wonder if we should make some abstraction for the options we just want to be true. _But_ maybe that's premature.

pkg/osquery/extension.go
```diff
@@ -35,17 +35,20 @@ type settingsStoreWriter interface {
...
+	serviceClient                 service.KolideService
+	settingsWriter                settingsStoreWriter
+	enrollMutex                   sync.Mutex
+	done                          chan struct{}
+	interrupted                   atomic.Bool
+	slogger                       *slog.Logger
+	logPublicationState           *logPublicationState
+	lastRequestQueriesTimestamp   *atomic.Int64
+	distributedForwardingInterval *atomic.Int64 // how frequently to forward RequestQueries requests to the cloud, in seconds
+	forwardAllDistributedUntil    *atomic.Int64 // allows for accelerated distributed requests until given timestamp
```
directionless, line 51: I bet this one is going to need a mutex. Oh, or maybe that's what atomic does. I wonder if a mutex is faster. Eh.

directionless (approved):
Looks pretty clean. I'm sure we're going to want to trigger this from localserver, and possible menu interation. But that can be a followup

pkg/osquery/extension.go
RebeccaMahany: Chatted offline -- seph looked this up and determined atomic is (probably) faster! I will revisit in the next PR if necessary.

PR #2159, opened 2025-03-19
Add Exec Test job for containers

RebeccaMahany:
Allows us to smoke test launcher and osquery on older/nonstandard versions, like ubuntu 20.04.

This is basically a copy-paste of the existing exec test job, with a small amount of extra setup so that it will run correctly in the container.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -306,11 +306,15 @@ jobs:
       - version_baseline # version_baseline implies build_linux
       - build # need the other builds too
     steps:
-    # Needed so we can restore cache -- see https://github.com/actions/cache/issues/1455#issuecomment-2328358604
-    - name: Install zstd
+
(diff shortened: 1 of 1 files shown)


James-Pickett (approved):

zackattack01 (approved):

.github/workflows/go.yml
```diff
@@ -292,6 +292,59 @@ jobs:
...
+          exit 1
+        fi
+
+    - name: Download Osquery
+      working-directory: build
+      run: ./launcher download-osquery --directory .
+
+    - name: Osquery Version
+      working-directory: build
+      run: ./osqueryd --version
```
directionless, line 343: This is probably fine, but it may cause our PRs to fail for very unrelated osquery reasons

RebeccaMahany: This is in the existing exec test, so I figured either both should have them or we can remove them in both places

PR #2160, opened 2025-03-19
Add two new flags controlling acceleration

RebeccaMahany:
This PR is a follow-up to https://github.com/kolide/launcher/pull/2157; it gets through the boilerplate of adding two new agent flags to the knapsack.

* `osquery_distributed_forwarding_interval` -- will replace the distributed forwarding interval hardcoded in launcher's extension; this flag is implemented as overridable for a duration
* `allow_overly_broad_dt4a_acceleration` -- if set, will instruct the `/v3/dt4a` endpoint to accelerate osquery and control server intervals

I will put in a subsequent PR that will implement this behavior after.


ee/agent/flags/flag_controller.go
```diff
@@ -341,6 +341,15 @@ func (fc *FlagController) ControlRequestInterval() time.Duration {
 	).get(fc.getControlServerValue(keys.ControlRequestInterval))
 }
 
+func (fc *FlagController) SetAllowDt4aAcceleration(enabled bool) error {
+	return fc.setControlServerValue(keys.AllowDt4aAcceleration, boolToBytes(enabled))
+}
+func (fc *FlagController) AllowDt4aAcceleration() bool {
```
James-Pickett, line 347: I thought we determined that we didn't want to accelerate on every call to `/dt4a`, I this a "just incase minds change"?

RebeccaMahany: Updated!

```
	// AllowDt4aAcceleration enables acceleration via /v3/dt4a localserver endpoint. It is a test flag
	// for development use; it should ultimately be replaced by a call to a new /v3 endpoint that only
	// performs acceleration.
```

Let me know if you think a flag rename would be useful also.

James-Pickett: The comment is good, thank you!

James-Pickett (dismissed):

ee/agent/flags/flag_controller.go
RebeccaMahany: seph suggested a name update to `allow_overly_broad_dt4a_acceleration` so I've done that as well! Hopefully makes it a little clearer 🤞

cesarfda (approved):

James-Pickett (dismissed):

cesarfda (approved):

PR #2161, opened 2025-03-19
Allow dt4a endpoints to accelerate control + osquery request intervals

RebeccaMahany:
Builds on https://github.com/kolide/launcher/pull/2157 and https://github.com/kolide/launcher/pull/2160.

This PR adds new capabilities to the localserver:

1. A new `/v3/accelerate` endpoint, which accelerates both osquery and control server intervals
2. Accelerate when existing endpoint `/v3/dt4a` is called, if `AllowOverlyBroadDt4aAcceleration` flag is set -- this functionality exists for testing/development purposes; it should eventually be replaced by 1

In order to accelerate osquery distributed request intervals, the osquery extension becomes a knapsack change observer for the `DistributedForwardingInterval` flag. This is our first change observer that can possibly be deleted and replaced while launcher is running (since osquery instances may be restarted without restarting launcher). I added a new `DeregisterChangeObserver` function called on extension shutdown, so that knapsack would not forever hold onto a reference to the old/replaced extension.


ee/localserver/dt4a.go
```diff
@@ -78,3 +88,46 @@ func (ls *localServer) requestDt4aInfoHandlerFunc(w http.ResponseWriter, r *http
...
+		if _, ok := allowlistedDt4aOriginsLookup[requestOrigin]; !ok && !strings.HasPrefix(requestOrigin, safariWebExtensionScheme) {
+			escapedOrigin := strings.ReplaceAll(strings.ReplaceAll(requestOrigin, "\n", ""), "\r", "") // remove any newlines
+			ls.slogger.Log(r.Context(), slog.LevelInfo,
+				"received dt4a request with origin not in allowlist",
+				"req_origin", escapedOrigin,
+			)
+			w.WriteHeader(http.StatusForbidden)
+			return
+		}
+	}
```
RebeccaMahany, line 114: @James-Pickett let me know if you want me to move these origin checks into the middleware now -- I went with this to start because I didn't want to cause merge conflicts if you've got something in flight

James-Pickett: I don't have any dt4a stuff in flight at the moment, feel free to move it over.

ee/agent/types/flags.go
```diff
@@ -10,6 +10,8 @@ import (
 type Flags interface {
 	// Registers an observer to receive messages when the specified keys change.
 	RegisterChangeObserver(observer FlagsChangeObserver, flagKeys ...keys.FlagKey)
+	// Deregisters an existing observer
```
James-Pickett, line 13: this is a test

ee/agent/flags/flag_controller_test.go
```diff
@@ -360,3 +360,54 @@ func TestControllerOverride(t *testing.T) {
...
+func TestDeregisterChangeObserver(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name       string
+		valueToSet time.Duration
+		interval   time.Duration
+		duration   time.Duration
+	}{
+		{
```
James-Pickett, line 373: multi line test

ee/localserver/server.go
```diff
@@ -119,6 +119,7 @@ func New(ctx context.Context, k types.Knapsack, presenceDetector presenceDetecto
 	mux.Handle("/zta", ls.requestDt4aInfoHandler())
 
 	mux.Handle("/v3/dt4a", dt4aAuthMiddleware.Wrap(ls.requestDt4aInfoHandler()))
+	mux.Handle("/v3/accelerate", dt4aAuthMiddleware.Wrap(ls.requestDt4aAccelerationHandler()))
```
directionless, line 122:
```suggestion
	mux.Handle("/zta", ls.requestDt4aInfoHandler())
	mux.Handle("/accelerate", ls.requestDt4aInfoHandler())
	
	mux.Handle("/v3/dt4a", dt4aAuthMiddleware.Wrap(ls.requestDt4aInfoHandler()))
	mux.Handle("/v3/accelerate", dt4aAuthMiddleware.Wrap(ls.requestDt4aAccelerationHandler()))
```

Maybe, could skip this?

directionless (commented):
nice

ee/localserver/server.go
James-Pickett: ahh woops, I didn't think about that .... @RebeccaMahany , I think the origin stuff cannot go into the dt4a auth middleware until we transition over to that completely

RebeccaMahany: I just removed the middleware when testing locally -- I figured we didn't need an unauthenticated accelerate endpoint that was more permanent than that

directionless: We might not. Coordinating with other teams is always 🤷 

I'm good merging as is

James-Pickett: disregard my last comment, I was confused, it's still in there

RebeccaMahany: (They can use the regular /zta endpoint with the `OverlyBroad` flag set to do acceleration until they're ready to do auth for both /v3/dt4a and v3/accelerate)

RebeccaMahany: James and I chatted offline re: origin header -- I added the origin header checks to the dt4a auth middleware, but also left it in the handler hit by /zta directly, so we should be good there!

PR #2163, opened 2025-03-20
Support target parsing for non-default arches

RebeccaMahany:
Relates to github.com/kolide/launcher/issues/1149.

Allows for targets like `linux-arm64-systemd-rpm` to exist, while keeping old format `linux-systemd-rpm` for our default arches.

### DIFF
--- a/pkg/packaging/target_test.go
+++ b/pkg/packaging/target_test.go
@@ -143,7 +143,7 @@ func TestTargetParse(t *testing.T) {
 	}{
 		{
 			in:  "darwin-launchd-pkg",
-			out: &Target{Platform: Darwin, Init: LaunchD, Package: Pkg},
+			out: &Target{Platform: Darwin, Init: LaunchD, Package: Pkg, Arch: Universal},
 		},
 		{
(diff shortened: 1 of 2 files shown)


pkg/packaging/target_test.go
```diff
@@ -143,15 +143,23 @@ func TestTargetParse(t *testing.T) {
...
 			in:  "linux-systemd-rpm",
-			out: &Target{Platform: Linux, Init: Systemd, Package: Rpm},
+			out: &Target{Platform: Linux, Init: Systemd, Package: Rpm, Arch: Amd64},
+		},
+		{
+			in:  "linux-arm64-systemd-rpm",
+			out: &Target{Platform: Linux, Init: Systemd, Package: Rpm, Arch: Arm64},
+		},
+		{
+			in:         "linux-notavalidarch-systemd-deb",
```
James-Pickett, line 161: I read this like a proper name "Notava Lidarch" and thought... what the hell kind of obscure linux distro / install format are we trying to support now?

zackattack01 (approved):

PR #2183, opened 2025-03-28
Increase sleep duration in CPU profiling to 5 seconds 

cesarfda:
This pull request includes a change to the `gatherPprofCpu` function in the `ee/debug/checkups/runtime.go` file. The most important change is the modification of the sleep duration to allow for a longer capture period of CPU profiling data specifically to help with windows machine not containing any samples.

* [`ee/debug/checkups/runtime.go`](diffhunk://#diff-2cce11759247660f34767b2c2ba1a03387e7d9bd4863cc2e1f1adcf5cc5a39d5L120-R120): Increased the sleep duration from 100 milliseconds to 5 seconds in the `gatherPprofCpu` function to improve the capture of background thread activity for CPU profiling.

Test flare profiling with local build:
image
image


RebeccaMahany (approved):

cesarfda:
Results after adding it routine:
image
Tree:
image

cesarfda:
Addressed Seph's suggestion of moving it to a routine and retested it with a flare

ee/debug/checkups/runtime.go
```diff
@@ -112,12 +111,19 @@ func gatherPprofCpu(z *zip.Writer) error {
...
+	// Move the sleep and StopCPUProfile into a goroutine
+	go func() {
+		// cpu profile is really meant to run over a period of time, capturing background information
+		time.Sleep(5 * time.Second)
+		pprof.StopCPUProfile()
+		close(done)
+	}()
 
+	// Wait for profiling to complete before returning
+	<-done
```
RebeccaMahany, line 127: Can we have a `select` here so that we don't block indefinitely, just to be safe? Something like --

```
	select {
	case <-done:
		return nil
	case <-time.After(10*time.Second):
		return errors.New("timed out waiting for CPU profile")
	}
```

cesarfda: done

RebeccaMahany (approved):

PR #2184, opened 2025-03-28
routine performance checks part 1

zackattack01:
I'm breaking off this bit from the larger effort to routinely collect cpu/mem profiles so that we can get some discussion going about criteria for triggering these and ensure we're gathering all of the statistics we want correctly.

This does a few things:
- upgrades our gopsutil package to v4. I didn't see any significant changes for our purposes but there were some bug fixes and memory leaks noted that I thought would make it worth upgrading now while we're looking at all of this closely. there were two small backwards compatibility breaks (updated in osquerylogs/logs.go and desktop/runner/runner_linux.go)
- adds a new performance package for us to collect some documentation and helper functionality for handling these statistics. we do a lot of similar things in several places but if we're going to start taking actions based on this data we will likely need some stricter patterns/expectations/types to work off of. Let me know if there is more you'd like to see added here, it was just meant as a starting point
  - this is very intentional in the data it collects for the `PerformanceStats` it exposes. All fields are expected to be supported on all platforms, and it does not include noisier things like the runtime GC stats, to allow us to add this to a log checkpoint without too much noise
- Adds a new checkup which reports this data. I had originally only meant it to be a `logCheckup` since a lot of this information is covered in our existing runtime checkup, but it seemed helpful as a quicker reference while looking at flares/doctor so I included it there too. happy to remove if people disagree, I could have gone either way

In it's current form, this will emit checkup logs like this once an hour:
<details><summary>checkup log</summary>
<p>

```
{
  "time": "2025-03-28T16:21:22.834035Z",
  "level": "DEBUG",
  "source": {
    "function": "github.com/kolide/launcher/ee/debug/checkups.(*logCheckPointer).Once",
    "file": "/Users/zackolson/go/src/launcher/ee/debug/checkups/checkpoint.go",
    "line": 68
  },
  "msg": "ran checkup",
  "run_id": "01JQERXN2V9FF2ZVJ7GMS0G1QE",
  "component": "log_checkpoint",
  "checkup": "Performance",
  "summary": "process 52226 is using 13.84% CPU, RSS: 53.34 MB (0.08% memory). Note CPU will be higher while running flare.",
  "data": {
    "stats": {
      "pid": 52226,
      "exe": "/Users/zackolson/go/src/launcher/build/launcher",
      "mem_info": {
        "rss_bytes": 55934976,
        "vms_bytes": 422118621184,
        "heap_total_bytes": 11067392,
        "go_mem_bytes": 21505288,
        "non_go_mem_bytes": 34429688,
        "mem_percent": 0.0813961
      },
      "cpu_percent": 13.839578399624044
    }
  },
  "status": "Informational"
}
```

</p>
</details> 

An informational line will also be added to doctor like so:
`Performance: process 50853 is using 16.88% CPU, RSS: 40.66 MB (0.06% memory). Note CPU will be higher while running flare.`


zackattack01:
there seems to be a data race introduced from the gopsutil update, it looks related to their shift from cgo to purego and difficult to avoid. I am going to try upgrading us to the release just before that happened instead

ee/debug/checkups/checkups.go
```diff
@@ -109,6 +109,7 @@ func checkupsFor(k types.Knapsack, target targetBits) []checkupInt {
 		{&desktopMenu{k: k}, flareSupported},
 		{&coredumpCheckup{}, doctorSupported | flareSupported},
 		{&downloadDirectory{}, flareSupported},
+		{&perfCheckup{}, doctorSupported | flareSupported | logSupported},
```
RebeccaMahany, line 112: I think we maybe either don't want this in doctor, or would only want a modified version in doctor? Since doctor is always run via the command line, the runtime stats will never be accurate, right?

ee/performance/stats.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/shirou/gopsutil/v4/process"
+)
+
+/*
+MemInfo is a helper struct with memory stats derived from both:
+- process.MemoryInfoStat (e.g. RSS, VMS). these help to give a picture of the usage from the OS perspective
+- runtime.MemStats (e.g. heap stats). these help to give a picture of the go runtime usage
+Used together, we are able to estimate things like allocations outside of our golang runtime (see nonGoMemUsage).
+The runtime MemStats can be confusing - here is an attempt at outlining some of the fields we're interested in:
```
RebeccaMahany, line 17: Thank you for this, really helpful

ee/debug/checkups/checkups.go
zackattack01: was thinking it would be nice in the doctor output in flares and didn't even consider that 🤦 you're absolutely right, will pull it

RebeccaMahany (approved):

ee/performance/stats.go
```diff
@@ -0,0 +1,114 @@
...
+
+	Sys is total bytes of memory obtained from the OS. The virtual address space reserved by the Go runtime for:
```
directionless, line 19: love the comment, the whitespace is a little fishy...

ee/performance/stats.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type PerformanceStats struct {
+	Pid        int      `json:"pid"`
+	Exe        string   `json:"exe"`
+	MemInfo    *MemInfo `json:"mem_info"`
+	CPUPercent float64  `json:"cpu_percent"`
+}
+
+func CurrentProcessStats(ctx context.Context) (*PerformanceStats, error) {
```
directionless, line 47: One caveat here, is that this gathers stats for the current process. Which _omits_ any desktop processes, or any osquery processes.

directionless (commented):
nice

PR #2185, opened 2025-03-28
Add `launcher query-windowsupdates` subcommand to avoid memory leak temporarily

RebeccaMahany:
Through testing, we've noticed non-go memory usage increasing whenever we call `IUpdateSearcher.Search` -- this memory is not released after the table generate function completes. I have not been able to make any meaningful changes to this behavior via `Release` or `Clear` calls.

We need a temporary solution to this memory leak while we continue to investigate. In this PR, we move the functionality that interfaces directly with go-ole into a launcher subcommand, so that we can query in a separate process (thereby releasing the non-go memory once the process terminates).

I have left the original code intact so that we can switch back to it in the future.


directionless (approved):
Looks good enough

PR #2154, opened 2025-03-17
checking in root.json for packaging

cesarfda:
checking in root.json for packaging to help fix kworker build failing.
Ex: https://github.com/kolide/monorepo/pull/327

PR #2189, opened 2025-04-01
check in updated root.json

zackattack01:
we suspect this needs to be updated to prevent the CI jobs from reading launcher version as `dirty`


RebeccaMahany (approved):

RebeccaMahany:
See https://github.com/kolide/launcher/pull/2154, https://github.com/kolide/launcher/pull/2142#discussion_r1999545946 for additional details about why this file is committed

PR #2196, opened 2025-04-08
macOS run disclaimed

zackattack01:
docs on their way!


James-Pickett (approved):
Nice, impressive work around!

cmd/launcher/main.go
```diff
@@ -45,6 +46,13 @@ func runMain() int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// these systemslogger logs will go to stderr, which will dirty the results of our
+	// rundisclaimed operations. the only log we would see before running the subcommand
+	// should be "launcher starting up", so override here with a no-op logger
+	if len(os.Args) > 1 && os.Args[1] == "rundisclaimed" {
```
RebeccaMahany, line 52: Maybe we would want to formalize the subcommands that should run "silently" because we are expecting a particular output? Since technically `query-windowsupdates` would also need this, we just got lucky that the systemSlogger writes to the event log instead of stdout. I don't think it's worth holding up this PR for, but if it seems like a good idea, maybe we file an issue to tackle it later?

ee/allowedcmd/cmd.go
```diff
@@ -110,3 +110,20 @@ func IsNixOS() bool {
 	checkedIsNixOS = true
 	return isNixOS
 }
+
+func Launcher(ctx context.Context, arg ...string) (*TracedCmd, error) {
```
RebeccaMahany, line 114: I think there are a couple places (the interactive checkup, the desktop runner, the windows update table) that could use this new function, maybe worth filing an issue to update 'em?

ee/allowedcmd/run_disclaimed_darwin.go
```diff
@@ -0,0 +1,144 @@
...
+	switch subcommand {
+	case "brew":
+		return generateBrewCommand(ctx)
+	default:
+		return nil, errors.New("unsupported command for rundisclaimed")
+	}
+}
+
+func generateBrewCommand(ctx context.Context) (*TracedCmd, error) {
+	cmd, err := Brew(ctx, "outdated", "--json")
```
RebeccaMahany, line 129: We don't care about not being able to be able to control these brew args via command-line args to rundisclaimed?

ee/allowedcmd/cmd.go
zackattack01: sounds like a good idea to me, will do!

cmd/launcher/main.go
```diff
@@ -45,6 +46,13 @@ func runMain() int {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// these systemslogger logs will go to stderr, which will dirty the results of our
+	// rundisclaimed operations. the only log we would see before running the subcommand
+	// should be "launcher starting up", so override here with a no-op logger
```
directionless, line 51: Why do we lose the split between stdout and stderr?

directionless (commented):
This seems great, I love how small it is. A couple of questions, though I would approve it as is and iterate when we need to

ee/allowedcmd/run_disclaimed_darwin.go
zackattack01: not for brew currently (this was the only usage i saw for that allowedcmd) but I think that's a good callout in general (others will have varied usage)- we can just whitelist the command itself as the first arg and then allow arbitrary args through. since this is running disclaimed I don't see any additional risk there if that makes sense to you?

cmd/launcher/main.go
zackattack01: yeah are there others you think would benefit? i.e. worth a full audit in separate issue or should I just add `query-windowsupdates` to the check here?

RebeccaMahany: That was the only one where I thought it'd actually matter -- but maybe it would be less confusing to end users if we also didn't print all these extra logs during `version`, `flare`, `doctor`?

cmd/launcher/main.go
zackattack01: at least brew usage just uses the same output buffer for both, I could update but had assumed we did more of that or used `CombinedOutput` elsewhere enough to make this useful. happy to fix that up if you don't like this approach!

ee/allowedcmd/run_disclaimed_darwin.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, fmt.Errorf("rundisclaimed expects 1 subcommand, received args of length %d", len(args))
+	}
+
+	subcommand := args[0]
+	switch subcommand {
+	case "brew":
+		return generateBrewCommand(ctx)
+	default:
+		return nil, errors.New("unsupported command for rundisclaimed")
+	}
```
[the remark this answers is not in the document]
zackattack01: I just wanted to get reviews on the pattern before adding in others- I am a little less nervous about complications with the way this has turned out but assume we'll want to explicitly retest everything we move over (and there is a small amount of code needed for each depending on extra env or setup requirements, etc.)

ee/allowedcmd/run_disclaimed_darwin.go
directionless: I think we're going to want the be able to specify these.

ee/allowedcmd/run_disclaimed_darwin.go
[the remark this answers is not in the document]
directionless: I'm good with that -- I think we're going to want to move basiclly every exec to using this

cmd/launcher/main.go
directionless: I think we should aim to make this a generalized exec tool, and thus I think we should aim to preserve stderr stdout.

zackattack01: makes sense to me! I will update

cmd/launcher/main.go
zackattack01: discussed this more [here](https://github.com/kolide/launcher/pull/2196/files#r2035680933) - I think we're going to remove this and encourage callers to handle stdout/stderr separately, this was only an issue for brew because we were using the same buffer for both. let me know if you still think worth adding an issue for flare/doctor? I could go either way

RebeccaMahany: Nah I think this is fine as-is!

ee/allowedcmd/run_disclaimed_darwin.go
```diff
@@ -0,0 +1,174 @@
+//go:build darwin
+// +build darwin
+
+package allowedcmd
```
directionless, line 4: This is an interesting place to hang this. I don't know if it's wrong, but it'd definitely breaking my model of the world. 😆 

I'd probably ee/disclaim or something

ee/tables/homebrew/upgradeable.go
```diff
@@ -74,12 +75,24 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
...
+		var stderr bytes.Buffer
+
+		targetCmd := allowedcmd.Brew
+		cmdArgs := []string{"outdated", "--json"}
+
+		// override cmd and args for darwin because we must run disclaimed
+		if runtime.GOOS == "darwin" {
+			targetCmd = allowedcmd.Launcher
+			cmdArgs = []string{"rundisclaimed", "brew", "outdated", "--json"}
+		}
```
directionless, line 87: I wonder if we can bake this into `tablehelpers.Run`

ee/allowedcmd/run_disclaimed_darwin.go
zackattack01: yeah this did end up less coupled than I had thought, I can break it out

ee/allowedcmd/run_disclaimed_darwin.go
```diff
@@ -0,0 +1,174 @@
...
+func getCmdGenerator(cmd string) (*allowedCmdGenerator, error) {
+	allowedCmdGenerators := map[string]allowedCmdGenerator{
+		"brew": {
+			allowedOpts: map[string]struct{}{
+				"outdated": {},
+				"--json":   {},
+			},
+			generate: generateBrewCommand,
+		},
+	}
```
RebeccaMahany, line 149: Sorry if I missed this on the last pass -- might as well make this a `var allowedCmdGenerators = map...` at the top so we don't have to instantiate it on every call to `getCmdGenerator`?

RebeccaMahany (approved):
One nitpick, lgtm overall! Really tidy

ee/tables/homebrew/upgradeable.go
zackattack01: I added a `tablehelpers.Disclaimed` for reusability within Run, let me know what you think!

ee/allowedcmd/run_disclaimed_darwin.go
zackattack01: oh yep! will do

ee/tables/tablehelpers/disclaim_darwin.go
```diff
@@ -0,0 +1,37 @@
...
+	"os/exec"
+
+	"github.com/kolide/launcher/ee/allowedcmd"
+)
+
+// Disclaimed (darwin only) replaces the cmd path and args for use via
+// our rundisclaimed subcommand
+func Disclaimed(disclaimCmdName string) ExecOps {
+	return func(cmd *exec.Cmd) error {
+		launcherCmd, err := allowedcmd.Launcher(context.TODO(), "rundisclaimed", disclaimCmdName)
```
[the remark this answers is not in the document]
zackattack01: done!

RebeccaMahany (approved):

ee/disclaim/run_disclaimed_darwin.go
```diff
@@ -0,0 +1,175 @@
...
+	"brew": {
+		allowedOpts: map[string]struct{}{
+			"outdated": {},
+			"--json":   {},
+		},
+		generate: generateBrewCommand,
+	},
+}
+
+func RunDisclaimed(_ *multislogger.MultiSlogger, args []string) error {
```
directionless, line 82: This is effectively a new `main`, and thus a new `ctx` it's going to break traces. I assume there's no way to pass `ctx`, but we might be able to pass whatever we need to connect traces. (either via args, or in the ENV).

I don't think we need to block this PR on this though.

ee/disclaim/run_disclaimed_darwin_test.go
```diff
@@ -0,0 +1,29 @@
...
+package disclaim
+
+import (
+	"testing"
+
+	"github.com/kolide/launcher/pkg/log/multislogger"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRunDisclaimedDoesNotRunArbitraryCommands(t *testing.T) {
```
directionless, line 13: Love the negative tests. Should we get a positive one? (Even if we can't really test the _disclaim_ part, we can probably test the exec and output passing.

(No need to block PR for it)

ee/tables/tablehelpers/disclaim_darwin.go
```diff
@@ -0,0 +1,37 @@
...
+	"errors"
+	"fmt"
+	"os/exec"
+
+	"github.com/kolide/launcher/ee/allowedcmd"
+)
+
+// Disclaimed (darwin only) replaces the cmd path and args for use via
+// our rundisclaimed subcommand
+func Disclaimed(ctx context.Context, disclaimCmdName string) ExecOps {
```
directionless, line 17: Hrm. It feels awkward to use `disclaimCmdName` here, and not `cmd.Path` inside the anonymous function. But I assume it's like this because there's no other way to get the right arg for `allowedcmd` inside disclaim. Probably the simplest tradeoff.

directionless (approved):
This looks super clean to me. A couple of things for followup, but I think reasonable for another PR

PR #2206, opened 2025-04-15
return dt4a data matching dt4a account id

James-Pickett:
DT4A auth middleware set Account & User Uuid headers

Look up dt4a data using account uuid

  if we have an account uuid in the request, only return if we have a match
  if we don't have an account uuid in the request, return legacy data

More tracing

relates to https://github.com/kolide/k2/pull/11915


cesarfda (dismissed):

RebeccaMahany (dismissed):
🔥

cesarfda (dismissed):

ee/localserver/dt4a.go
```diff
@@ -48,17 +50,34 @@ func (ls *localServer) requestDt4aInfoHandlerFunc(w http.ResponseWriter, r *http
 		ls.accelerate(r.Context())
 	}
 
-	dt4aInfo, err := ls.knapsack.Dt4aInfoStore().Get(localserverDt4aInfoKey)
+	if r.Header.Get(dt4aAccountUuidHeaderKey) != "" {
+		dt4aInfo, err := ls.knapsack.Dt4aInfoStore().Get([]byte(r.Header.Get(dt4aAccountUuidHeaderKey)))
```
directionless, line 54: If I read this right, it's storing in a hash with `<accountUUID>` as the key. How does this work with the possible addition of `<userUUID>`

James-Pickett: This PR makes no provisions for user uuid other than setting it the request header. My understanding is that user uuid is purely informational from launcher perspective at this point. I plan to determine where to discover that information on the k2 side and log mis-matches in a follow on PR.

RebeccaMahany (dismissed):

ee/localserver/dt4a_auth_middleware.go
```diff
@@ -127,6 +139,12 @@ func (d *dt4aAuthMiddleware) Wrap(next http.Handler) http.Handler {
 		bhr := &bufferedHttpResponse{}
 		next.ServeHTTP(bhr, newReq)
 
+		if bhr.code < 200 || bhr.code >= 300 {
+			// hacky sleep here so that it's likely a response form another launcher
+			// on a different port will return with success faster than this failure
+			time.Sleep(1 * time.Second)
```
James-Pickett, line 145: now that we have bumped the rate limit, we could end up with max 200 threads I think 🤔

ee/localserver/dt4a.go
```diff
@@ -48,18 +50,44 @@ func (ls *localServer) requestDt4aInfoHandlerFunc(w http.ResponseWriter, r *http
...
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(dt4aInfo)
+		return
+	}
+
+	// dt4aAccountUuid is set, so we will try to get the dt4a info using the account uuid
+	dt4aInfo, err := ls.knapsack.Dt4aInfoStore().Get([]byte(r.Header.Get(dt4aAccountUuidHeaderKey)))
```
directionless, line 79: This is storing as a hash by account uuid? How do we handle useruuid if/when we want that?

I wonder if there's a pattern where we store as an array of blobs, and iterate it to find a match? That would be weird if we had multiple users in the same account, but...

I guess this also depends on how K2 ultimately serves it

James-Pickett: > This is storing as a hash by account uuid?

Yes

> How do we handle useruuid if/when we want that?

We would need K2 to do something like you suggested.

Our original thought was to make the keys be `<accountUuid>/<userUuid>`, but we realized `userUuid` my not be available when we needed, so we decided that starting with `accountUuid`  was enough (per our meeting with k2/dt4a devs).

I suppose we could also just go with `acocuntUuid` when `userUuid` is unavailable.

So we would first try to find with `<accountUuid>/<userUuid>` then fall back to just `accountUuid`.

ee/localserver/dt4a_auth_middleware.go
directionless: Can we lower to 100ms? That should be enough for the successful path, and I don't know what the extension will experience with the long sleep.

ee/localserver/dt4a.go
directionless: I'm thinking about what kind of structure this assumes k2 will emit today, and in the future. And I feel pretty unsure

ee/localserver/dt4a.go
James-Pickett: > This is storing as a hash by account uuid?

When you say hash, I'm assuming you mean the ruby data structure, go map equivalent

ee/localserver/dt4a.go
directionless: Eh. I'm letting perfect distracted from good enough. James says this is how k2 works, so let's leave it

ee/localserver/dt4a_auth_middleware.go
James-Pickett: done

PR #2207, opened 2025-04-15
disclaim falconctl command for macos

zackattack01:
includes a few small changes to support disclaiming our `falconctl` execs for macos:
- updates command generation signatures to allow direct use of allowedcmds where additional logic is not needed
- adds `falconctl` to `allowedCmdGenerators`
- adds an early return without error for missing binaries from rundisclaimed to match behavior within our tablehelpers/run logic


James-Pickett (approved):

directionless (approved):
I love how clean this looks

RebeccaMahany (approved):

PR #2215, opened 2025-04-23
Rename pkg/traces to ee/observability

RebeccaMahany:
I am going to add metrics shipping and want to keep it together with the work in pkg/traces and pkg/traces/exporter, so this PR renames the package in anticipation.


James-Pickett (approved):
LGTM .... I didn't read it all 😄

PR #2221, opened 2025-04-23
Add column "online" to kolide_windows_updates that allows for performing online vs offline windows update searches

RebeccaMahany:
We suspect that performing online searches in the background, but offline searches during distributed queries, will speed up the distributed queries enough that we aren't timing out.

This PR adds a new column `online`; when set to 0, it will perform an offline search. This PR also adds tracing data so that we can validate more comprehensively whether offline searches are significantly faster.


cmd/launcher/query_windowsupdates_windows.go
```diff
@@ -24,7 +24,7 @@ func runQueryWindowsUpdates(systemMultiSlogger *multislogger.MultiSlogger, args
 	var (
 		flagset     = flag.NewFlagSet("query-windowsupdates", flag.ExitOnError)
 		flLocale    = flagset.String("locale", "_default", "search locale")
-		flTableMode = flagset.Int("table_mode", int(windowsupdatetable.UpdatesTable), fmt.Sprintf("updates table (%d); history table (%d)", windowsupdatetable.UpdatesTable, windowsupdatetable.HistoryTable))
+		flTableMode = flagset.Int("table_mode", int(windowsupdatetable.UpdatesTable), fmt.Sprintf("updates table (%d); history table (%d); offline updates (%d)", windowsupdatetable.UpdatesTable, windowsupdatetable.HistoryTable, windowsupdatetable.UpdatesOfflineTable))
```
directionless, line 27: I'm not sure if this should be a new table. 

Possible a `where` condition on the existing table? Maybe even defaulting to offline?

RebeccaMahany: I was thinking about following a pattern similar to `kolide_softwareupdate` vs `kolide_softwareupdate_scan` -- but this is also just for initial testing! If this setting does end up helping, then I can figure out a better option for the existing table if that's what we prefer

directionless: My inclination is to use one table. `select * from xxx where force_fetch=true` or something.

RebeccaMahany: Will update!

RebeccaMahany:
Concluded that data differs between online and offline searches too much to pursue this route

PR #2225, opened 2025-04-24
Deprecate grpc transport option

RebeccaMahany:
We have not used this option in several years, and we no longer want to maintain it.


James-Pickett (approved):
🎉

PR #2229, opened 2025-04-28
Add rungroup actor to periodically cache results of querying Windows Update Agent API

RebeccaMahany:
Sometimes, the Windows Update Agent API's `Search` function is untenably slow. We cap all of our tables at a 4-minute timeout and cannot reasonably extend the timeout for this table without preventing auth attempts for going through, since those have a 5-minute timeout.

We haven't found any way to speed up this API call, nor any alternatives -- so we are going to try periodically calling `Search` in the background and caching the response data. This PR performs that portion of the work.

We tentatively plan to use the response to speed up scheduled query responses to the `kolide_windows_updates` table, to prevent checks using this table from going stale due to table timeouts. That work is TBD in a future PR.


ee/tables/windowsupdatetable/cache_windows.go
```diff
@@ -0,0 +1,125 @@
...
+		slogger:       slogger.With("component", "windows_updates_cacher"),
+		interrupt:     make(chan struct{}, 1),
+	}
+}
+
+func (w *windowsUpdatesCacher) Execute() (err error) {
+	cacheTicker := time.NewTicker(w.cacheInterval)
+	defer cacheTicker.Stop()
+
+	var ctx context.Context
```
zackattack01, line 54: should this be initialized somewhere?

zackattack01: 🤦 i see

zackattack01 (dismissed):

ee/tables/windowsupdatetable/cache_windows.go
```diff
@@ -0,0 +1,125 @@
...
+		Results   *QueryResults `json:"results"`
+	}
+)
+
+func NewWindowsUpdatesCacher(cacheStore types.GetterSetter, cacheInterval time.Duration, slogger *slog.Logger) *windowsUpdatesCacher {
+	return &windowsUpdatesCacher{
+		cacheStore:    cacheStore,
+		cacheInterval: cacheInterval,
+		slogger:       slogger.With("component", "windows_updates_cacher"),
+		interrupt:     make(chan struct{}, 1),
```
James-Pickett, line 46: Can you comment on why the chan has a buffer?

ee/tables/windowsupdatetable/cache_windows.go
James-Pickett: hmm ... I'm waiting for my 🤦 moment, I can't figure out why the var is needed, why not just init on line 60?

RebeccaMahany: We can do

```golang
var ctx context.Context
ctx, w.queryCancel = context.WithTimeout(context.Background(), 10*time.Minute)
```

inside the for loop, if that's clearer! (Can't do it in one line on line 60 because we're setting `w.queryCancel`.) I figured might as well only init it once outside the loop, but if it tripped up both of you then I think it's not very readable as-is -- I will update!

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: I think I started doing this at some point when tracking down rungroup actors that were getting stuck and not shutting down...let me refresh my memory, and I will update here plus add a comment in the code.

RebeccaMahany: I documented this terribly 😅 . This is related to https://github.com/kolide/launcher/issues/1205. I think the point of this was to allow `w.interrupt <- struct{}{}` in `Interrupt` to return right away, instead of blocking waiting for `Execute` to read from the channel.

RebeccaMahany: (I have updated to add a comment here to that effect as well)

ee/tables/windowsupdatetable/cache_windows.go
James-Pickett: I also see it now =) ... the set of w.queryCancel

James-Pickett (approved):

ee/agent/knapsack/knapsack.go
```diff
@@ -184,6 +184,10 @@ func (k *knapsack) Dt4aInfoStore() types.KVStore {
 	return k.getKVStore(storage.Dt4aInfoStore)
 }
 
+func (k *knapsack) WindowsUpdatesCacheStore() types.KVStore {
```
directionless, line 187: Hrm. Given that this is only exposed inside `tables/windowsupdatetable` it feels weird to plumb it through knapsack -- it's not crossing any package boundaries. 

It _feels_ like a package level variable would be simpler. But I know sometimes that's messy to test around. (Since it's sorta global state)

Unclear to me if this is cleaner cached on the table or in the underlying windoewsupdate pkg

directionless: I think if we really do want to push through knapsack, I think we should make a general `cache` thing, and not make a specific one per cache. But maybe I'm premature

ee/tables/windowsupdatetable/cache_windows.go
```diff
@@ -0,0 +1,125 @@
...
+	defer cacheTicker.Stop()
+
+	for {
+		select {
+		case <-cacheTicker.C:
+			// Since this query happens in the background and will not block auth, we can use
+			// a much longer timeout than we use for our tables.
+			var ctx context.Context
+			ctx, w.queryCancel = context.WithTimeout(context.Background(), 10*time.Minute)
+			if err := w.queryAndStoreData(ctx); err != nil {
```
directionless, line 61: This might be an interesting place for a metric

ee/tables/windowsupdatetable/cache_windows.go
```diff
@@ -0,0 +1,125 @@
...
+				w.slogger.Log(ctx, slog.LevelWarn,
```
directionless, line 62: Maybe this is an error -- if we're seeing timeouts, we need to know asap, so we can change that number / behaviors

ee/tables/windowsupdatetable/cache_windows.go
```diff
@@ -0,0 +1,125 @@
...
+	launcherPath, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("getting path to launcher: %w", err)
+	}
+	if !strings.HasSuffix(launcherPath, "launcher.exe") {
+		return errors.New("cannot run generate for non-launcher executable (is this running in a test context?)")
+	}
+
+	queryTime := time.Now()
+	res, err := callQueryWindowsUpdatesSubcommand(ctx, launcherPath, defaultLocale, UpdatesTable)
```
directionless, line 107: Having stared at this code, I'm fairly certain that the timeout works because it's in the context which is pushed all the way down to the underling exec.CommandContext. But a comment to that effect would be nice

ee/tables/windowsupdatetable/windowsupdate.go
```diff
@@ -105,38 +107,12 @@ func (t *Table) generateWithLauncherExec(ctx context.Context, queryContext table
...
-	for _, locale := range tablehelpers.GetConstraints(queryContext, "locale", tablehelpers.WithDefaults("_default")) {
-		args := []string{
-			"query-windowsupdates",
-			"-locale", locale,
-			"-table_mode", strconv.Itoa(int(t.mode)),
-		}
-		cmd := exec.CommandContext(ctx, launcherPath, args...) //nolint:forbidigo // We can exec the current executable safely
-		cmd.Env = append(cmd.Env, "LAUNCHER_SKIP_UPDATES=TRUE")
-		out, err := cmd.CombinedOutput()
+	for _, locale := range tablehelpers.GetConstraints(queryContext, "locale", tablehelpers.WithDefaults(defaultLocale)) {
```
directionless, line 110: hrm. We're not really doing the right caching thing for the different locales. But I think to solve that we'd need to introspect the

RebeccaMahany: I went the lazy route because we appear to always use this default at the moment (no queries specifying `locale` that I could see, anyway) -- if that's subject to change, let me know?

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: Yes! Will add!

ee/agent/knapsack/knapsack.go
RebeccaMahany: I wanted the cache to persist between launcher restarts, which is why I didn't put it in the table or the windowsupdate package. I agree it's weird to have it available everywhere when it's Windows-specific...I will think on a general `cache` (maybe something similar to `InitialResultsStore`?) or maybe moving it to kv.sqlite or something!

cmd/launcher/launcher.go
```diff
@@ -354,6 +355,9 @@ func runLauncher(ctx context.Context, cancel func(), multiSlogger, systemMultiSl
 		runGroup.Add("powerEventWatcher", powerEventWatcher.Execute, powerEventWatcher.Interrupt)
 	}
 
+	windowsUpdatesCacher := windowsupdatetable.NewWindowsUpdatesCacher(k.WindowsUpdatesCacheStore(), 1*time.Hour, slogger)
```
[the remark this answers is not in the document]
RebeccaMahany: I think it doesn't hurt to run the cacher itself all the time -- gets launcher in a good state in case this check becomes enabled, for example. But I do think it's a good question to consider as we get farther into this with e.g. https://github.com/kolide/launcher/pull/2230.

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: A query success/failure metric? Or something else?

ee/tables/windowsupdatetable/windowsupdate.go
directionless: I am unlikely to remember. But presumably we'll hear about the performance

ee/tables/windowsupdatetable/cache_windows.go
directionless: That it timed out. I'm thinking about how we know if there's an issue

ee/agent/knapsack/knapsack.go
directionless: Sorry, I was being confusing...

I think there are 2 questions here. One is what the underlying storage is (and I assumed you went with bbolt to persist across reboots)

The other is how we connect all the dots. One way us to expose the bbolt in knapsack, that feels like it's adding a lot of boilerplate. Another way would be to pass a bbolt store to the actor, and have that stash a package level variable.

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: Added: https://github.com/kolide/launcher/pull/2230/commits/2175f9ef9c52957735ec61ddd1b1384fd0cb44f5

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: Added: https://github.com/kolide/launcher/pull/2230/commits/2175f9ef9c52957735ec61ddd1b1384fd0cb44f5

ee/agent/knapsack/knapsack.go
RebeccaMahany: I think if we do the latter (pass a bbolt store to the actor) then we have a direct bbolt usage, which means we lose the abstraction that we've built on top of bbolt in order to eventually replace it? Or am I misunderstanding the suggestion? (We can chat about it in our meeting tomorrow if that's easier!)

ee/tables/windowsupdatetable/cache_windows.go
RebeccaMahany: Added: https://github.com/kolide/launcher/pull/2230/commits/e33d0db00ae9bbd86ec6e88ff70074da3b03d01c

ee/agent/knapsack/knapsack.go
directionless: Misunderstanding -- could pass a store to the actor, the actor can set a package level var, and we can skip knapsack.

RebeccaMahany: We currently are passing the store to the actor: `func NewWindowsUpdatesCacher(cacheStore types.GetterSetter, cacheInterval time.Duration, slogger *slog.Logger) *windowsUpdatesCacher`

PR #2230, opened 2025-04-28
Add kolide_windows_updates_cached table

RebeccaMahany:
Builds on https://github.com/kolide/launcher/pull/2229


ee/tables/windowsupdatetable/windowsupdate_cached.go
```diff
@@ -0,0 +1,116 @@
...
+			flatData, err := dataflatten.Json(res.Results.RawResults, flattenOpts...)
+			if err != nil {
+				c.slogger.Log(ctx, slog.LevelInfo,
+					"flatten failed",
+					"err", err,
+				)
+				continue
+			}
+
+			rowData := map[string]string{
```
directionless, line 106: This should merge in the cached timestamp. Or maybe as a `age` style int. And probably a `cached=true` or something

ee/tables/windowsupdatetable/windowsupdate_cached.go
```diff
@@ -0,0 +1,116 @@
...
+func CachedWindowsUpdatesTablePlugin(flags types.Flags, slogger *slog.Logger, cacheStore types.Getter) *table.Plugin {
+	columns := dataflattentable.Columns(
+		table.TextColumn("locale"),
+		table.IntegerColumn("is_default"),
+	)
+
+	t := &CachedWindowsUpdatesTable{
+		slogger:    slogger.With("name", "kolide_windows_updates_cached"),
+		cacheStore: cacheStore,
+		name:       "kolide_windows_updates_cached",
```
directionless, line 42: I still feel a bit biased away from a different table name.

RebeccaMahany: Do you mean you want to name this table something else, or that you prefer to keep this inside `kolide_windows_updates` with a `cached` column? If the former, happy to rename -- let me know what you don't like about this one or what you'd prefer. If the latter -- I chose a separate table to make it easier for us to rewrite SQL on the fly, if that's the route we end up taking. (Rewriting the SQL to add/remove a column constraint sounds painful, but replacing `kolide_windows_updates_cached` with `kolide_windows_updates` or vice versa should be easier.)

directionless: I had been thinking about column constraints, but that's a compelling arguement in how we handle re-writing it.

ee/tables/windowsupdatetable/windowsupdate_cached.go
RebeccaMahany: Done!

James-Pickett (approved):
Nice!

ee/tables/windowsupdatetable/windowsupdate_cached.go
```diff
@@ -0,0 +1,122 @@
...
+					"flatten failed",
+					"err", err,
+				)
+				continue
+			}
+
+			rowData := map[string]string{
+				"locale":     res.Results.Locale,
+				"is_default": strconv.Itoa(res.Results.IsDefaultLocale),
+				"is_cached":  "1",
```
James-Pickett, line 113: if I'm reading this correctly, this will always be "1"? We need this and `age`? I can live with it =)

James-Pickett: ehh nvm .... I see @directionless suggested it, I assume there is a downstream sql thingy that this helps with

RebeccaMahany: seph asked for em here! https://github.com/kolide/launcher/pull/2230#discussion_r2064509118

We don't need them at the moment, but implementation is still a bit up in the air so we might want them later

RebeccaMahany: Ah sorry, I only saw your first comment when I responded!

directionless: `1` is true. 

I don't have an exact need, but it feels _correct_ to set it, even though it's a bit superfluous with `age`. :shrug:

PR #2231, opened 2025-04-29
Add flag `use_cached_data_for_scheduled_queries`, and rewrite osquery config accordingly when set

RebeccaMahany:
Relates to https://github.com/kolide/launcher/pull/2229 and https://github.com/kolide/launcher/pull/2230.

This PR adds a new agent flag `use_cached_data_for_scheduled_queries`. When it is set, we rewrite the osquery config (at the same time that we perform other config modifications) to query `kolide_windows_updates_cached` instead of `kolide_windows_updates`. This should prevent timeouts when scheduled queries need Windows Update Agent API data, therefore preventing relevant checks from going stale.

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log/slog"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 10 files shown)


James-Pickett (approved):

pkg/osquery/extension.go
```diff
@@ -556,6 +557,13 @@ func (e *Extension) generateConfigsWithReenroll(ctx context.Context, reenroll bo
 
 	config = e.setOsqueryOptions(config, configOptsToSet)
 
+	// If this feature flag is set, we want to use cached data for scheduled queries that might otherwise
+	// time out. We achieve this by updating the config to point to the tables holding the cached data.
+	if e.knapsack.UseCachedDataForScheduledQueries() {
+		// The kolide_windows_updates table is currently the only table we use this feature for.
+		config = strings.ReplaceAll(config, "kolide_windows_updates", "kolide_windows_updates_cached")
```
directionless, line 564: worth remembering that this is now a magic string that will explode if a customer uses it.

directionless (approved):

pkg/osquery/extension.go
RebeccaMahany: Maybe worth going through and only updating this string if it's in a pack => query section of the config?

RebeccaMahany: Discussed offline, concluded not worth it!

PR #2248, opened 2025-05-07
Migrate plist type tables in darwin_platform_tables

cesarfda:
This pull request introduces a new `mapplist` parser for handling plist data on Darwin-based systems, replacing the previous `dataflattentable`-based approach. It includes the implementation of the parser, test cases, and updates to integrate the parser into the platform-specific tables.

### New `mapplist` Parser Implementation:
* Added a `parser` struct and its methods (`Parse` and `parsePlist`) to handle plist data parsing. The implementation uses the `howett.net/plist` library for unmarshalling plist data. (`ee/tables/execparsers/mapplist/mapplist.go`, `ee/tables/execparsers/mapplist/parser.go`) [[1]](diffhunk://#diff-bede0a3a1149cf254e96e27380086cdb520726770064087d8b8a6d5fd52deb4cR1-R26) [[2]](diffhunk://#diff-f2b6fac567491253e3c29acd669e1d43797757a04aaf19f150fdc9b53aa6cb03R1-R36)

### Test Suite for `mapplist` Parser:
* Introduced comprehensive test cases in `parse_test.go` to validate the parser's handling of various plist data scenarios, including empty input, valid plist data, malformed data, and singleton behavior. Embedded test data files are used for diskutil and powermetrics examples. (`ee/tables/execparsers/mapplist/parse_test.go`)

### Integration with Platform-Specific Tables:
* Replaced `dataflattentable.TablePluginExec` calls with `dataflattentable.NewExecAndParseTable`, using the new `mapplist.Parser` for plist-based commands in `platform_tables_darwin.go`. (`pkg/osquery/table/platform_tables_darwin.go`) [[1]](diffhunk://#diff-2187e568192469c7026a87117ad42c37e533d7c5d71ec1092e69b15da6a70043R17) [[2]](diffhunk://#diff-2187e568192469c7026a87117ad42c37e533d7c5d71ec1092e69b15da6a70043L109-R115)
* kolide_diskutil_list - [livequery](https://app.kolide.com/1/live_query/campaigns/25979/results/162021492)
* kolide_falconctl_stats - [livequery](https://app.kolide.com/1/live_query/campaigns/25979/results/162021501)
* kolide_apfs_list - [livequery](https://app.kolide.com/1/live_query/campaigns/25979/results/162021503)
* kolide_apfs_users - [livequery](https://app.kolide.com/1/live_query/campaigns/25979/results/162021505)
* kolide_tmutil_destinationinfo - No results for both before and after
* kolide_powermetrics - [livequery](https://app.kolide.com/1/live_query/campaigns/25979/results/162021509)

### DIFF
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -128,10 +128,6 @@ func (k *knapsack) KatcConfigStore() types.KVStore {
 	return k.getKVStore(storage.KatcConfigStore)
 }
 
-func (k *knapsack) AutoupdateErrorsStore() types.KVStore {
-	return k.getKVStore(storage.AutoupdateErrorsStore)
-}
-
 func (k *knapsack) ConfigStore() types.KVStore {
 	return k.getKVStore(storage.ConfigStore)
 }
--- a/ee/agent/storage/bbolt/stores_bbolt.go
+++ b/ee/agent/storage/bbolt/stores_bbolt.go
@@ -21,7 +21,6 @@ func MakeStores(ctx context.Context, slogger *slog.Logger, db *bbolt.DB) (map[st
 	var storeNames = []storage.Store{
 		storage.AgentFlagsStore,
 		storage.KatcConfigStore,
-		storage.AutoupdateErrorsStore,
 		storage.ConfigStore,
 		storage.ControlStore,
 		storage.PersistentHostDataStore,
--- a/ee/agent/storage/ci/stores_ci.go
+++ b/ee/agent/storage/ci/stores_ci.go
@@ -20,7 +20,6 @@ func MakeStores(t *testing.T, slogger *slog.Logger, db *bbolt.DB) (map[storage.S
 	var storeNames = []storage.Store{
 		storage.AgentFlagsStore,
 		storage.KatcConfigStore,
-		storage.AutoupdateErrorsStore,
 		storage.ConfigStore,
 		storage.ControlStore,
 		storage.PersistentHostDataStore,
--- a/ee/agent/storage/stores.go
+++ b/ee/agent/storage/stores.go
@@ -6,7 +6,6 @@ type Store string
 const (
 	AgentFlagsStore             Store = "agent_flags"              // The store used for agent control flags.
 	KatcConfigStore             Store = "katc_config"              // The store used for Kolide custom ATC configuration
-	AutoupdateErrorsStore       Store = "tuf_autoupdate_errors"    // The store used for tracking new autoupdater errors.
 	ConfigStore                 Store = "config"                   // The store used for launcher configuration.
 	ControlStore                Store = "control_service_data"     // The store used for control service caching data.
 	PersistentHostDataStore     Store = "persistent_host_data"     // The store used for data about this host.
--- a/ee/agent/types/mocks/knapsack.go
+++ b/ee/agent/types/mocks/knapsack.go
@@ -92,26 +92,6 @@ func (_m *Knapsack) Autoupdate() bool {
 	return r0
 }
 
-// AutoupdateErrorsStore provides a mock function with no fields
-func (_m *Knapsack) AutoupdateErrorsStore() types.GetterSetterDeleterIteratorUpdaterCounterAppender {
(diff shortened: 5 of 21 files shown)


ee/tables/execparsers/mapplist/parser.go
```diff
@@ -0,0 +1,36 @@
...
+		return nil, errors.New("empty input")
+	}
+
+	// Create a variable to hold the parsed data
+	var result interface{}
+
+	// Unmarshal the plist data
+	if _, err := plist.Unmarshal(data, &result); err != nil {
+		return nil, err
+	}
```
RebeccaMahany, line 32: Do we definitely need to read in all the data from the reader first? I am wondering if we could cut out that extra step by using the plist decoder instead, if we know that we expect XML rather than binary data --

```golang
var result interface{}
if err := plist.NewXMLDecoder(reader).Decode(&result); err != nil {
	return nil, fmt.Errorf("decoding plist: %w", err)
}
```

ee/tables/execparsers/mapplist/parse_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+//go:embed test-data/diskutil-list.txt
+var diskutilListData string
+
+//go:embed test-data/powermetrics.txt
+var powermetricsData string
```
RebeccaMahany, line 20: Sorry if I'm just missing it, I looked a couple times but couldn't find it -- is `test-data/diskutil-apfs.txt` in use?

RebeccaMahany (commented):
Small comments, LGTM overall! Sorry for the delay in reviewing

ee/tables/execparsers/mapplist/mapplist.go
```diff
@@ -0,0 +1,26 @@
+//go:build darwin
+// +build darwin
+
+package mapplist
```
directionless, line 4: Why name this `mapplist` and not `plist` ?

ee/tables/execparsers/mapplist/parser.go
directionless: We don't know that. But we might be able to detect it from the first bytes

directionless: But Unmarshal just causes Decode so there's clearly a stream way https://github.com/DHowett/go-plist/blob/v1.0.1/decode.go#L113-L118

cesarfda: NewDecoder takes and io.ReadSeeker type[[NewDecoder](https://github.com/DHowett/go-plist/blob/v1.0.1/decode.go#L82)](url) and we use io.Reader [https://github.com/kolide/launcher/blob/migrate-tables/ee/tables/dataflattentable/flattener_from_parser.go#L13](url). Is there a way around this?

ee/tables/execparsers/mapplist/mapplist.go
cesarfda: I used the XML parser to start but I guess this one doesn't use a map so I'll go ahead and rename it

ee/tables/execparsers/mapplist/parse_test.go
cesarfda: I missed that one, added with latest commit

ee/tables/execparsers/mapplist/parser.go
RebeccaMahany: Ah, I was looking at `github.com/groob/plist`, which we also use in launcher -- that one takes `io.Reader`, but it doesn't detect binary vs xml and just assumes xml.

Since we can't assume XML, I don't think it's worth updating the entire parser interface for this! I'm fine with it as-is.

directionless: groob's plist and dhowett's behave slightly differently. We should use whatever the original used.

Oh, I hadn't noticed the skew between `Reader` and `ReadSeeker`. That's a bummer. It's probably worth an hour seeing how hard it would be to change the interface to `ReadSeeker`, but I think it's reasonable to kick that can down the road and merge as is

PR #2252, opened 2025-05-09
Track checkups "score" and error count as metrics

RebeccaMahany:
First pass at adding a "score" that launcher will report up during the log checkpoint, allowing us to assess launcher health over time.

### DIFF
--- a/ee/debug/checkups/checkpoint.go
+++ b/ee/debug/checkups/checkpoint.go
@@ -66,6 +66,7 @@ func (c *logCheckPointer) Once(ctx context.Context) {
 	warningCount := 0
 	failingCount := 0
 	passingCount := 0
+	erroringCount := 0
 
 	for _, checkup := range checkups {
 		checkup.Run(ctx, io.Discard)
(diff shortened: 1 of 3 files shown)


ee/debug/checkups/checkpoint.go
```diff
@@ -72,5 +77,25 @@ func (c *logCheckPointer) Once(ctx context.Context) {
...
+
+		switch checkup.Status() {
+		case Warning:
+			warningCount += 1
+		case Failing:
+			failingCount += 1
+		case Passing:
+			passingCount += 1
+		case Erroring, Informational, Unknown:
+			// Nothing to do here
```
directionless, line 89: I think we should track Erroring and Unknown. Possible as two more metrics. They represent problem with our system that we _ought_ address. (But not a device health score)

RebeccaMahany: Looks like we only use `Unknown` right now when the launcher installation is not Kolide-hosted -- maybe not worth tracking yet? I'll add a gauge for `Erroring`!

RebeccaMahany: Oh, ha, so currently none of our `logSupported` checkups set `Erroring` status. Added anyway, we'll use it at some point!

James-Pickett (approved):

directionless (approved):

PR #2255, opened 2025-05-12
Make performance checkup a scored checkup

RebeccaMahany:
Updates the performance checkup to return a non-informational status so that it will count toward the checkup score that we report as a metric.

Also, fixes the float math when calculating the checkup so that warnings do actually count for something, instead of 0.

### DIFF
--- a/ee/debug/checkups/checkpoint.go
+++ b/ee/debug/checkups/checkpoint.go
@@ -63,10 +63,10 @@ func (c *logCheckPointer) Interrupt(_ error) {
 func (c *logCheckPointer) Once(ctx context.Context) {
 	checkups := checkupsFor(c.knapsack, logSupported)
 
-	warningCount := 0
-	failingCount := 0
-	passingCount := 0
-	erroringCount := 0
+	warningCount := 0.0
+	failingCount := 0.0
+	passingCount := 0.0
+	erroringCount := 0.0
 
 	for _, checkup := range checkups {
 		checkup.Run(ctx, io.Discard)
@@ -95,12 +95,16 @@ func (c *logCheckPointer) Once(ctx context.Context) {
 
 	// Compute score from warning, passing, and failing counts
 	scoredCheckups := warningCount + failingCount + passingCount
-	score := (float64(passingCount+(warningCount/2)) / float64(scoredCheckups)) * 100
+	score := ((passingCount + (warningCount / 2)) / scoredCheckups) * 100
 	observability.CheckupScoreGauge.Record(ctx, score)
 
 	c.slogger.Log(ctx, slog.LevelDebug,
 		"computed checkup score",
 		"score", score,
+		"failing_count", failingCount,
+		"warning_count", warningCount,
+		"passing_count", passingCount,
+		"total_scored_checkups", scoredCheckups,
 	)
 
 	// Record number of errors separately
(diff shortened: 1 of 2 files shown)


ee/debug/checkups/performance.go
```diff
@@ -8,9 +8,15 @@ import (
 	"github.com/kolide/launcher/ee/performance"
 )
 
+const (
+	cpuPercentThreshold = 30
+	memPercentThreshold = 30
```
directionless, line 13: I don't think memory percentage is a meaningful number. We should pick some numbers for go and non-go

RebeccaMahany: Took a quick peek at our metrics -- how about over 200MB for go, over 100MB for non-go?

RebeccaMahany: Went with those, can adjust as needed!

ee/debug/checkups/checkpoint.go
```diff
@@ -95,12 +128,16 @@ func (c *logCheckPointer) Once(ctx context.Context) {
...
 	// Compute score from warning, passing, and failing counts
 	scoredCheckups := warningCount + failingCount + passingCount
-	score := (float64(passingCount+(warningCount/2)) / float64(scoredCheckups)) * 100
+	score := ((passingCount + (warningCount / 2)) / scoredCheckups) * 100
 	observability.CheckupScoreGauge.Record(ctx, score)
 
 	c.slogger.Log(ctx, slog.LevelDebug,
 		"computed checkup score",
 		"score", score,
+		"failing_count", failingCount,
```
zackattack01, line 137: thinking about how we're going to hone to this logic (and any thresholds we're working on)- I see we're recording the metric but do we have a quick way to get an idea of why devices have that score? I'm wondering if we should at least start with this log being Info level, and including the names for failing and passing checkups. Just so we can log search on this and see any trends worth digging into, rather than investigating one device at a time based on score

RebeccaMahany: How about `Warn` level if the score isn't 100%, `Debug` (or `Info`) level otherwise, and then adding the names of the checkups?

zackattack01: works for me!

ee/debug/checkups/performance.go
directionless: My guess is we'll see a lot of errors here, but I'm game to start with those numbers.

ee/debug/checkups/performance.go
```diff
@@ -21,9 +28,20 @@ func (p *perfCheckup) Run(ctx context.Context, _ io.Writer) error {
...
 
+	memOver := stats.MemInfo.GoMemUsage > golangMemUsageThreshold || stats.MemInfo.NonGoMemUsage > nonGolangMemUsageThreshold
+	cpuOver := stats.CPUPercent > cpuPercentThreshold
+	if cpuOver && memOver {
+		p.status = Failing
+	} else if cpuOver || memOver {
+		p.status = Warning
+	} else {
+		p.status = Passing
+	}
```
directionless, line 43: Logic feels weird.

I would say that if we exceed _either_ threshold we're Failing. 

If we want a warning state, we could define those thresholds.

ee/debug/checkups/checkpoint.go
RebeccaMahany: Updated

PR #2258, opened 2025-05-14
Run WMI ExecQuery synchronously instead of semisynchronously

RebeccaMahany:


ee/wmi/wmi.go
```diff
@@ -170,10 +173,13 @@ func Query(ctx context.Context, slogger *slog.Logger, className string, properti
 		"query", queryString,
 	)
 
-	// result is a SWBemObjectSet
-	resultRaw, err := oleutil.CallMethod(service, "ExecQuery", queryString)
+	// ExecQuery runs semi-synchronously by default. To ensure we aren't missing any results,
+	// we prefer synchronous mode, which we achieve by setting iFlags to wbemFlagForwardOnly+wbemFlagReturnWhenComplete
+	// instead of the default wbemFlagReturnImmediately. See https://learn.microsoft.com/en-us/windows/win32/wmisdk/calling-a-method#semisynchronous-mode.
+	// The result is a SWBemObjectSet.
+	resultRaw, err := oleutil.CallMethod(service, "ExecQuery", queryString, "WQL", WBEM_FLAG_FORWARD_ONLY+WBEM_FLAG_RETURN_WHEN_COMPLETE)
```
zackattack01, line 180: Is the `wbemFlagForwardOnly` for the performance reasons noted when running semisynchronously? I didn't see where/how that should be used for synchronous mode

RebeccaMahany: Yeah exactly, it seemed safer to include it given that they recommended it specifically! Updated the comment to make it clearer

zackattack01: I'm having a hard time understanding the docs for this :/ not saying it should block but wondering if we're misunderstanding something- I cannot figure out how these flags work as an integer when there are multiple flags with the value 0. i.e. how does the ExecQuery know that `WBEM_FLAG_FORWARD_ONLY+WBEM_FLAG_RETURN_WHEN_COMPLETE` is different than just `WBEM_FLAG_FORWARD_ONLY` and that either of those are different than `wbemFlagBidirectional` + forward only?

zackattack01: okay yeah I think you're right 😅 thank you for explaining!

zackattack01 (dismissed):

ee/wmi/wmi.go
RebeccaMahany: No prob, I definitely should've documented that the first time around 😅 -- updated the comment w an abbreviated version of the above

zackattack01 (approved):

directionless (approved):
I'm game to try. But I wonder if there are cases this will be worse, and if we might need roll back.

PR #2259, opened 2025-05-14
If unenrolled and k2 provides enrollment secret in localserver request, save it

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1473


RebeccaMahany:
Going with https://github.com/kolide/launcher/pull/2291 instead

PR #2262, opened 2025-05-19
Store registrations in new registration store

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1473. Start to store enrollment data (default registration ID, munemo, node key, optionally enrollment secret) in persistent store so that we don't need to require an on-disk enrollment secret.

The general idea is:

* Start to store munemo and enrollment secret in RegistrationStore (this PR)
* In other locations where we read the enrollment secret directly just to get the munemo, read the munemo from the RegistrationStore instead
* In the one place where we use the enrollment secret for something besides the munemo (flare upload), find a suitable replacement
* We should then be in a good place to support an alternative enrollment method for secretless installations that doesn't require receiving an enrollment secret from k2


pkg/osquery/extension.go
```diff
@@ -304,6 +305,124 @@ func isNodeInvalidErr(err error) bool {
...
+		)
+	}
+
+	e.slogger.Log(context.TODO(), slog.LevelInfo,
+		"successfully stored new registration",
+		"munemo", r.Munemo,
+	)
+}
+
+func (e *Extension) updateRegistration(nodeKey string) {
```
zackattack01, line 362: just a few non-blocking thoughts/comments if any seem worth it to you- I don't have as strong of a vision of what this will look like next so might be off here:
- should `updateRegistration`/`addRegistration` at least have an error return? It doesn't seem necessary (i.e. an error shouldn't change any of the caller flows) currently so I could go either way, but this seemed like it could be an important enough path down the road to check
- it took me a little to read through and understand the update vs add Registration methods and grasp what flows should trigger which and when. would it at least sound clearer to make `updateRegistration` something like `ensureRegistrationNodeKeySet`? maybe with a func comment that explains it reads enroll secret and triggers `addRegistration` if needed

RebeccaMahany: Both add and update are under-commented, I will fix that! I could go either way on the error return too -- I think it sounds like a good idea for the future, so I will update.

zackattack01 (approved):

pkg/osquery/extension.go
RebeccaMahany: Added comments, renamed function!

zackattack01 (dismissed):

ee/debug/checkups/enroll-secret.go
```diff
@@ -26,6 +26,18 @@ func (c *enrollSecretCheckup) Run(_ context.Context, extraFH io.Writer) error {
 	secretStatus := make(map[string]Status, 0)
 	secretSummary := make(map[string]string, 0)
 
+	// Check registrations first -- will only be available if we're running in situ
+	if registrations, err := c.k.Registrations(); err == nil {
+		for i, registration := range registrations {
```
James-Pickett, line 31: Should we log something if there are no registrations?

pkg/osquery/extension.go
```diff
@@ -304,6 +305,106 @@ func isNodeInvalidErr(err error) bool {
 	}
 }
 
+// addRegistration should be called after enrollment to store enrollment/registration details
+// in our persistent store under e.registrationId. We frequently use the munemo associated with
+// the registration (e.g. in checkups), so we parse the enrollment secret in order to extract
+// the munemo from the claims and store it as well.
+func (e *Extension) addRegistration(nodeKey string, enrollmentSecret string) error {
```
James-Pickett, line 312: how gnarly would it be to to add some assertions to an enrollment test to make sure these get stored?

RebeccaMahany: I think not bad, actually! Let me see 👀

ee/debug/checkups/enroll-secret.go
RebeccaMahany: Updated!

pkg/osquery/extension.go
RebeccaMahany: Not bad at all! Updated, thank you for the suggestion!

PR #2264, opened 2025-05-20
Get munemo for header checks from registration store

RebeccaMahany:
Like https://github.com/kolide/launcher/pull/2263, this PR also uses the new registration store to get the munemo associated with the launcher installation, rather than reading and parsing the enrollment secret.

This means that there's the possibility that localserver will attempt to read from the store before the registration exists in it. To address this, I added a goroutine in the localserver similar to the async workers one, which will periodically check for registrations in the store until enrollment is complete, and then store the munemo in the middleware and exit when done. I liked this a little better than the alternative (injecting the knapsack into the middleware and having the middleware handle periodically checking instead) because this option allows us to keep the middleware more focused/lightweight, and fits better with our existing patterns.

Since this means the munemo can be altered in the middleware now, I opted for `atomic.String` to avoid data races. I chose `atomic.String` over a mutex to avoid the possibility of a lock holding up localserver requests.

Builds on https://github.com/kolide/launcher/pull/2262 -- tackles part of the second bullet point. Relates to https://github.com/kolide/launcher/issues/1473.

### DIFF
--- a/ee/localserver/krypto-ec-middleware.go
+++ b/ee/localserver/krypto-ec-middleware.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kolide/launcher/ee/presencedetection"
 	"github.com/kolide/launcher/pkg/log/multislogger"
 	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/atomic"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


James-Pickett (approved):
Nice

ee/localserver/krypto-ec-middleware.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/kolide/launcher/ee/presencedetection"
 	"github.com/kolide/launcher/pkg/log/multislogger"
 	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/atomic"
```
directionless, line 29: huh. I think we usually use `sync/atomic`

RebeccaMahany: Oh, super weird, that is definitely the one I meant to use! Thank you, I will update

PR #2273, opened 2025-05-22
New Table! The deadly sleeper

directionless:
Sometimes, while debugging, I want to make a query that takes a long time to return. Perhaps this will fit the bill....

### DIFF
--- a/ee/tables/sleeper/table.go
+++ b/ee/tables/sleeper/table.go
@@ -35,6 +35,8 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	var results []map[string]string
 
 	for _, durationStr := range tablehelpers.GetConstraints(queryContext, "duration") {
+		// This is using an anonymous function so that the ticker.Stop can cleanly be deferred. Elsewise, we end up
+		// stacking them, and the linter complains.
 		if err := func() error {
 			duration, err := strconv.Atoi(durationStr)
 			if err != nil {
--- a/ee/tables/sleeper/table.go
+++ b/ee/tables/sleeper/table.go
@@ -12,6 +12,7 @@ import (
 	"github.com/osquery/osquery-go/plugin/table"
 )
 
+// sleeper.Table is a debugging table, used to test queries that take a long time to run.
 type Table struct {
 	slogger *slog.Logger
 }
@@ -43,7 +44,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 			t.slogger.Log(ctx, slog.LevelWarn, "The deadly sleeper table sleeps!", "duration", duration)
 
 			ticker := time.NewTicker(time.Duration(duration) * time.Second)
-			defer ticker.Stop() // nolint:revive // we want to
+			defer ticker.Stop()
 
 			select {
 			case <-ticker.C:
--- a/ee/tables/sleeper/table.go
+++ b/ee/tables/sleeper/table.go
@@ -34,23 +34,28 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 	var results []map[string]string
 
 	for _, durationStr := range tablehelpers.GetConstraints(queryContext, "duration") {
-		duration, err := strconv.Atoi(durationStr)
-		if err != nil {
+		if err := func() error {
+			duration, err := strconv.Atoi(durationStr)
+			if err != nil {
+				return err
+			}
+
+			t.slogger.Log(ctx, slog.LevelWarn, "The deadly sleeper table sleeps!", "duration", duration)
+
+			ticker := time.NewTicker(time.Duration(duration) * time.Second)
+			defer ticker.Stop() // nolint:revive // we want to
+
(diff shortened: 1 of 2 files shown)


ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,57 @@
...
+
+	columns := []table.ColumnDefinition{
+		table.IntegerColumn("duration"),
+	}
+
+	t := &Table{
+		slogger: slogger.With("table", "kolide_deadly_sleeper"),
+	}
+
+	return tablewrapper.New(flags, slogger, "kolide_deadly_sleeper", columns, t.generate, tablewrapper.WithTableGenerateTimeout(10*time.Minute))
```
directionless, line 29: I wrapped it, to ensure traces. Even though this should only show up in debugging, I wanted to make sure we saw it if someone calls it in prod.

ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,57 @@
...
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	var results []map[string]string
+
+	for _, durationStr := range tablehelpers.GetConstraints(queryContext, "duration") {
+		duration, err := strconv.Atoi(durationStr)
+		if err != nil {
+			return results, err
+		}
+
+		t.slogger.Log(ctx, slog.LevelWarn, "The deadly sleeper table sleeps!", "duration", duration)
```
directionless, line 42: `WARN` might be overkill, but same. I wanted a clear trail here.

ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,57 @@
...
+
+		ticker := time.NewTicker(time.Duration(duration) * time.Second)
+		defer ticker.Stop()
+
+		select {
```
directionless, line 47: Maybe select and ticket means we don't block the thread?

blaedj:
ahh this would be much nicer/simpler than my convoluted "query curl table to hit an http server that sleeps before responding" setup!

ee/tables/sleeper/table.go
RebeccaMahany: This seems correct to me!

ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,62 @@
...
+		if err := func() error {
+			duration, err := strconv.Atoi(durationStr)
+			if err != nil {
+				return err
+			}
+
+			t.slogger.Log(ctx, slog.LevelWarn, "The deadly sleeper table sleeps!", "duration", duration)
+
+			ticker := time.NewTicker(time.Duration(duration) * time.Second)
+			defer ticker.Stop() // nolint:revive // we want to
```
RebeccaMahany, line 46: For additional clarity, maybe --

```suggestion
			defer ticker.Stop() // nolint:revive // we don't expect a large number of duration strings in the query, so we think it is fine to accumulate defers inside the for loop here to keep the code simpler
```

...or something like that, but less wordy 😅

ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,62 @@
...
+	"strconv"
+	"time"
+
+	"github.com/kolide/launcher/ee/agent/types"
+	"github.com/kolide/launcher/ee/tables/tablehelpers"
+	"github.com/kolide/launcher/ee/tables/tablewrapper"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+type Table struct {
```
RebeccaMahany, line 15: Could we add a small amount of documentation for the table?

```suggestion
Table is a debugging table, used to craft queries that take a long time to run.
type Table struct {
```

RebeccaMahany (approved):
LGTM, just two requests for extra documentation. Neat!

James-Pickett (commented):
LGTM, pending Becca's suggestions

ee/tables/sleeper/table.go
directionless: Oh. Ha. I didn't mean to commit that. There's a anonymous function to make the linter happy 😆

RebeccaMahany: Oh yup, I see it now! Might be worth a comment at the top of the anonymous func to explain that we use it to avoid stacking up defers, then

ee/tables/sleeper/table.go
```diff
@@ -0,0 +1,62 @@
...
+
+	columns := []table.ColumnDefinition{
+		table.IntegerColumn("duration"),
+	}
+
+	t := &Table{
+		slogger: slogger.With("table", "kolide_deadly_sleeper"),
+	}
+
+	return tablewrapper.New(flags, slogger, "kolide_deadly_sleeper", columns, t.generate, tablewrapper.WithTableGenerateTimeout(10*time.Minute))
```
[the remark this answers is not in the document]
directionless: I think that's okay, and I don't want to make bigger changes for this relatively narrow debugging tool

cesarfda (approved):

Issue #2181, opened 2025-03-26
Update comshim usage to prevent panics


zackattack01:
We use comshim to manage our calls to `ole.CoInitializeEx` in our [WMI queries](https://github.com/kolide/launcher/blob/b06b3a2de3c280a9164bb22b015790dda4d15cc5/ee/wmi/wmi.go#L134) and [windows update table](https://github.com/kolide/launcher/blob/b06b3a2de3c280a9164bb22b015790dda4d15cc5/ee/tables/windowsupdatetable/windowsupdate.go#L99).

Unfortunately there are a few potential avenues where this can introduce a panic, at least when called multiple times without being uninitialized, or when changing thread modes (watchdog calls `ole.CoInitialize(0)`, not multithreaded, which can result in race condition here).

We should fork and update, or somehow avoid the panic [here](https://github.com/scjalliance/comshim/blob/b2ef9d8d7e0fe205a6f89f706c48da27411b9cd9/shim.go#L64-L65).

We should also update watchdog to use the same pattern (or at least use `ole.CoInitializeEx` with `ole.COINIT_MULTITHREADED`) [here](https://github.com/kolide/launcher/blob/main/ee/watchdog/controller_windows.go)

PR #2275, opened 2025-05-22
prevent comshim panics

zackattack01:
- brings in new fork of comshim package which has already addressed the exact use case we were looking at with an additional `comshim.TryAdd` method
- updates existing comshim usages to utilize `TryAdd` with error handling
- updates watchdog controller's direct CoInitialize invocations to use comshim

I looked through the changes between the new and old fork and don't see anything we wouldn't want, this is pretty similar to how I was going to approach this with our own fork (+ some nice unrelated updates and tests)

fixes https://github.com/kolide/launcher/issues/2181

### DIFF
--- a/cmd/launcher/query_windowsupdates_windows.go
+++ b/cmd/launcher/query_windowsupdates_windows.go
@@ -10,11 +10,11 @@ import (
 	"os"
 	"strconv"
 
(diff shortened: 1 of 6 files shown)


RebeccaMahany (approved):
Nice! 🤞

James-Pickett (approved):
🔥

PR #2277, opened 2025-05-23
Gather performance stats for launcher child processes

RebeccaMahany:
* Gather and log performance stats for osquery; additionally, report these metrics via our metrics exporter
* Gather and log performance stats for launcher desktop -- not doing metrics yet
* Increase metrics exporting interval to 15m

Keeping this in the performance, rather than process, checkup, so that we can ultimately remove the process checkup from the log checkpointer, since the process checkpoint is more performance-intensive to run.

### DIFF
--- a/ee/performance/stats.go
+++ b/ee/performance/stats.go
@@ -92,7 +92,11 @@ func ChildProcessStatsForPid(ctx context.Context, pid int32) ([]*PerformanceStat
 	}
 
 	childProcesses, err := proc.ChildrenWithContext(ctx)
-	if err != nil {
+	// ChildrenWithContext uses pgrep, which will exit with exit status 1 if there were no matching processes
+	// (i.e. no child processes). This is unexpected for us -- launcher should typically have child processes --
+	// but it's not necessarily an error. Only return an error if we got an actual error string back here;
+	// callers can handle an empty list of children appropriately in the case that we received exit status 1.
+	if err != nil && !strings.Contains(err.Error(), "exit status 1") {
 		return nil, fmt.Errorf("getting child processes for pid %d: %w", pid, err)
 	}
 
--- a/ee/performance/stats.go
+++ b/ee/performance/stats.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/kolide/launcher/ee/observability"
 	"github.com/shirou/gopsutil/v4/process"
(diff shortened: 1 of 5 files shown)


ee/performance/stats.go
```diff
@@ -56,51 +58,113 @@ func ProcessStatsForPid(ctx context.Context, pid int) (*PerformanceStats, error)
...
+	stats := make([]*PerformanceStats, 0)
+	for _, childProcess := range childProcesses {
+		ps, _, err := statsForProcess(ctx, childProcess)
+		if err != nil {
+			continue
+		}
+		stats = append(stats, ps)
+
+		if strings.Contains(ps.Cmdline, "osquery") {
+			observability.OsqueryRssHistogram.Record(ctx, int64(ps.MemInfo.RSS))
```
zackattack01, line 108: Do we want to keep these as a side effect of stats collection? It does seem clean here but would feel a little more flexible to me if the caller was in charge of setting this stuff after the stats collection took place

RebeccaMahany: My thought was that any time we're doing the work to gather this data, we'd want to emit the metrics associated with it too. I am pretty sure I read somewhere in the [otel spec](https://opentelemetry.io/docs/specs/otel/metrics/api/) that instruments should only ever be used in one place -- I can't find it there now, but that's also why I originally wanted to record the metrics as close to where we gather them as possible. (Plus we'd need to rework the log checkpointer to conditionally record metrics.)

I agree it feels a little weird/not quite correct as a side effect, though...at the very least, I can document these side effects better.

ee/performance/stats.go
```diff
@@ -56,51 +62,121 @@ func ProcessStatsForPid(ctx context.Context, pid int) (*PerformanceStats, error)
...
+			observability.OsqueryCpuPercentHistogram.Record(ctx, ps.CPUPercent)
+		}
+
+		// We want to grab one more level of child processes, to account for the desktop process
```
zackattack01, line 124: just verifying- are we sure the CPU usage is not bubbled up/accounted for by the parent process stats? like if we have launcher proper -> launcher desktop sudo wrapper -> launcher desktop, those will all report separate CPU usages that we can sum together later?

RebeccaMahany: I think yes -- I consistently see the `sudo --preserve-env -u <username> /path/to/launcher desktop` process reporting a smaller CPU percent than its child desktop process. e.g. in a recent log, the parent desktop process had cpu_percent 0.017492090009365427 while the child had 0.39365630681368013. I can document this.

RebeccaMahany: Added a comment!

zackattack01 (approved):

PR #2282, opened 2025-05-28
add windows machine GUID to hardware identifiers

zackattack01:
- adds logic to our reset code to collect and store `SOFTWARE\Microsoft\Cryptography\MachineGuid` for use during hardware change detection

### DIFF
--- a/ee/agent/hardware_identifiers_other.go
+++ b/ee/agent/hardware_identifiers_other.go
@@ -1,0 +1,16 @@
+//go:build !windows
+// +build !windows
+
+package agent
+
+import (
+	"context"
+
+	"github.com/kolide/launcher/ee/agent/types"
+)
+
+// currentMachineGuid is only implemented for windows, where the hardware_uuid
+// cannot be used as a stable identifier
+func currentMachineGuid(_ context.Context, _ types.Knapsack) (string, error) {
+	return "", nil
+}
--- a/ee/agent/reset.go
+++ b/ee/agent/reset.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


ee/agent/hardware_identifiers_other.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"context"
+
+	"github.com/kolide/launcher/ee/agent/types"
+)
+
+// currentMachineGuid is only implemented for windows, where the hardware_uuid
+// cannot be used as a stable identifier
+func currentMachineGuid(_ context.Context, _ types.Knapsack) (string, error) {
+	return "", nil
```
zackattack01, line 15: unsure whether to return an empty string or some const value here for non-windows. empty string seems more correct to me, but will have the effect of causing every startup to rewrite this empty string: https://github.com/kolide/launcher/blob/66a737764cbe8c92fd03cf6eb1e49f26e33bdb5e/ee/agent/reset.go#L243-L248
the end result will still be what we want either way, and we could also change that logic to just not write out empty strings but maybe there's something I'm not considering? curious what others think here

RebeccaMahany: ```if len(storedValue) == 0 && len(currentValue) > 0 {```

seems fine to me? Would be nice to not have the extra log spam, and I agree that an empty string seems better than a const

ee/agent/reset.go
```diff
@@ -87,11 +99,14 @@ func DetectAndRemediateHardwareChange(ctx context.Context, k types.Knapsack) {
 		munemoChanged = valueChanged(ctx, k, slogger, currentTenantMunemo, hostDataKeyMunemo)
 	}
 
-	if serialChanged && hardwareUUIDChanged {
+	// note that machineGuid is only collected for windows. machineGuidChanged will only ever be true there
```
RebeccaMahany, line 102:
```suggestion
	// note that machineGuid is only collected for windows. machineGuidChanged will only ever have a meaningful value for Windows
```

I originally interpreted "machineGuidChanged will only ever be true there" as "machineGuidChanged can never false on Windows" -- I think this update or similar avoids that interpretation 😅

RebeccaMahany (commented):
Looks good to me!

ee/agent/reset.go
zackattack01: oh yep! good call ty

RebeccaMahany (approved):

Issue #2190, opened 2025-04-02
launcher should not be running after executing remote uninstall request


RebeccaMahany:
When testing v1.18.3, @James-Pickett and I both noticed that after remotely uninstalling launcher, the launcher service would start back up again.

@James-Pickett suggested the issue is that launcher exits before the Windows Service Manager fully receives the update that the launcher service should be in "manual" start mode, so the Service Manager restarts launcher -- i.e., we have a race between [setting start mode to manual](https://github.com/kolide/launcher/blob/main/ee/uninstall/uninstall_windows.go#L33) and calling [os.Exit](https://github.com/kolide/launcher/blob/main/ee/uninstall/uninstall.go#L59). We should update `disableAutoStart` on Windows to confirm with the Service Manager that the start mode has updated to "manual" before returning.

We don't think this bug is new to this release, so it won't prevent v1.18.3 from going out.

RebeccaMahany:
Tested this out today, and even with an additional verification (backoff for 30sec querying new config and confirming start type), I still see this issue. Additionally, I see that the start type is correctly set to manual in the service manager as well.

directionless:
mmmmm. In a secretless (or multitenant) world, I think the remote uninstall should _unenroll_ but leave launcher running. I'm not sure we'll still want a remote _uninstall_. What do you think?

directionless:
mmm, that's a good question.

I can tell you that when we originally build this, it was meant as a "reset for the next enrollment" but since secretless didn't exist, it was a disable. The intent is from the perspective of K2 or the admin. `This device is no longer part of this tenant, please reset for the next`

It feels somewhat wrong to _uninstall_.

RebeccaMahany:
When a duplicate device is identified, k2 sends down an uninstall request to the "older" device. Is it correct for that device to uninstall launcher, or to unenroll?

RebeccaMahany:
KOL-267

PR #2288, opened 2025-05-30
Reset RecoveryActions during uninstall to ensure manual start will be respected

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2190.

I tested and our original data race theory didn't bear out -- I updated the uninstall function to confirm that the launcher service's start type was set to manual start, and even with backoff and a fresh service handle, the behavior didn't change. I looked in the event logs, and those confirmed that the service's start type was set to demand start (manual) before the service shut down. I also saw in the event logs immediately afterwards that the service manager noted that the launcher svc had terminated unexpectedly _and that it would restart in 5 seconds_.

Updating the uninstall function to also remove our recovery actions fixed the issue. Now, launcher remains stopped after remote uninstall.

### DIFF
--- a/ee/uninstall/uninstall_windows.go
+++ b/ee/uninstall/uninstall_windows.go
@@ -3,6 +3,7 @@ package uninstall
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/kolide/launcher/ee/agent/types"
 	"github.com/kolide/launcher/ee/watchdog"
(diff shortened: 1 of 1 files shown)


cesarfda (approved):

directionless (approved):

directionless:
Actually, we could pause slight. Let's chat in #2190

RebeccaMahany:
Might want to take this in a different direction, closing for now 🤷‍♀️

PR #2289, opened 2025-06-02
Zscale table

cesarfda:
Adding support for new command-line utilities (`zscli`) on macOS and Windows

### Command-line utility support:

* Added `func Zscli` to `cmd_darwin.go` for macOS, enabling execution of the Zscaler CLI utility.
* Added `func Zscli` to `cmd_windows.go` for Windows, enabling execution of the Zscaler CLI utility.


ee/allowedcmd/cmd_darwin.go
```diff
@@ -150,3 +150,7 @@ func Zfs(ctx context.Context, arg ...string) (*TracedCmd, error) {
 func Zpool(ctx context.Context, arg ...string) (*TracedCmd, error) {
 	return validatedCommand(ctx, "/usr/sbin/zpool", arg...)
 }
+
+func Zscli(ctx context.Context, arg ...string) (*TracedCmd, error) {
+	return validatedCommand(context.Background(), "/Applications/Zscaler/Zscaler.app/Contents/PlugIns/zscli", arg...)
```
RebeccaMahany: Unless there's a particular reason we don't want to use `ctx` from the args, we should thread it through here so that our traces connect as expected, and so timeouts set on `ctx` are respected.

```suggestion
	return validatedCommand(ctx, "/Applications/Zscaler/Zscaler.app/Contents/PlugIns/zscli", arg...)
```

ee/allowedcmd/cmd_windows.go
```diff
@@ -54,3 +54,7 @@ func ZerotierCli(ctx context.Context, arg ...string) (*TracedCmd, error) {
 	// For windows, "-q" should be prepended before all other args
 	return validatedCommand(ctx, filepath.Join(os.Getenv("SYSTEMROOT"), "ProgramData", "ZeroTier", "One", "zerotier-one_x64.exe"), append([]string{"-q"}, arg...)...)
 }
+
+func Zscli(ctx context.Context, arg ...string) (*TracedCmd, error) {
+	return validatedCommand(context.Background(), filepath.Join(os.Getenv("PROGRAMFILES"), "Zscaler", "ZSACli", "ZSACli.exe"), arg...)
```
RebeccaMahany: Same here -- 

```suggestion
	return validatedCommand(ctx, filepath.Join(os.Getenv("PROGRAMFILES"), "Zscaler", "ZSACli", "ZSACli.exe"), arg...)
```

Issue #2285, opened 2025-05-30
BufSpanProcessor sometimes triggers a panic


RebeccaMahany:
Stack trace:

```
runtime error: invalid memory address or nil pointer dereference
github.com/kolide/launcher/ee/gowrapper.GoWithRecoveryAction.func1.1
	/Users/runner/work/launcher/launcher/ee/gowrapper/goroutine.go:31
runtime.gopanic
	/Users/runner/hostedtoolcache/go/1.23.0/x64/src/runtime/panic.go:785
runtime.panicmem
	/Users/runner/hostedtoolcache/go/1.23.0/x64/src/runtime/panic.go:262
runtime.sigpanic
	/Users/runner/hostedtoolcache/go/1.23.0/x64/src/runtime/signal_unix.go:900
go.opentelemetry.io/otel/sdk/trace.(*batchSpanProcessor).enqueueDrop
	/Users/runner/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.35.0/trace/batch_span_processor.go:389
go.opentelemetry.io/otel/sdk/trace.(*batchSpanProcessor).enqueue
	/Users/runner/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.35.0/trace/batch_span_processor.go:371
go.opentelemetry.io/otel/sdk/trace.(*batchSpanProcessor).OnEnd
	/Users/runner/go/pkg/mod/go.opentelemetry.io/otel/sdk@v1.35.0/trace/batch_span_processor.go:138
github.com/kolide/launcher/ee/observability/bufspanprocessor.(*BufSpanProcessor).SetChildProcessor
	/Users/runner/work/launcher/launcher/ee/observability/bufspanprocessor/bufspanprocessor.go:41
github.com/kolide/launcher/ee/observability/exporter.(*TelemetryExporter).setNewGlobalTracerProvider
	/Users/runner/work/launcher/launcher/ee/observability/exporter/exporter.go:317
github.com/kolide/launcher/ee/observability/exporter.(*TelemetryExporter).setNewGlobalProvider
	/Users/runner/work/launcher/launcher/ee/observability/exporter/exporter.go:248
github.com/kolide/launcher/ee/observability/exporter.(*TelemetryExporter).Execute
	/Users/runner/work/launcher/launcher/ee/observability/exporter/exporter.go:369
github.com/kolide/launcher/pkg/rungroup.(*Group).Run.func1
[...]

PR #2290, opened 2025-06-02
Skip ending spans for nil spans; log nil spans

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/2285

Attempts to avoid the panic seen in the linked issue, and to log information about the potential panic so we can understand how it's happening.

I noticed that we were acquiring `b.bufMu.Lock()` _after_ iterating over `b.bufferedSpans` -- this seemed incorrect to me, so I moved the lock above the iteration; it seemed to work fine in test.


ee/observability/bufspanprocessor/bufspanprocessor.go
```diff
@@ -70,6 +105,12 @@ func (b *BufSpanProcessor) OnEnd(s sdktrace.ReadOnlySpan) {
 	defer b.bufMu.Unlock()
 
 	if len(b.bufferedSpans) >= b.MaxBufferedSpans {
+		// This log might get a little verbose, but we should only be using this functionality in the
+		// BufSpanProcessor for under a minute -- we should be setting the child processor within a very
+		// short time window on launcher startup.
+		b.slogger.Log(context.TODO(), slog.LevelDebug,
+			"hit max number of buffered spans, dropping",
+		)
```
RebeccaMahany, line 113: In test, I currently only see 61 to 64 spans created before we set the child processor, whereas our capacity here is 500 -- so we shouldn't see this log at all right now. But I figured if anything changes in the future, we'd want to know that we're dropping spans.

PR #2291, opened 2025-06-02
Store node key and munemo if received from callback response

RebeccaMahany:

### DIFF
--- a/ee/localserver/krypto-ec-middleware.go
+++ b/ee/localserver/krypto-ec-middleware.go
@@ -21,6 +21,8 @@ import (
 	"github.com/kolide/krypto"
 	"github.com/kolide/krypto/pkg/challenge"
 	"github.com/kolide/launcher/ee/agent"
+	"github.com/kolide/launcher/ee/agent/storage"
+	"github.com/kolide/launcher/ee/agent/types"
 	"github.com/kolide/launcher/ee/gowrapper"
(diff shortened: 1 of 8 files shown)


ee/localserver/krypto-ec-middleware.go
```diff
@@ -45,6 +47,10 @@ const (
 	kolideMunemoHeaderKey                             = "X-Kolide-Munemo"
 )
 
+var (
+	nodeKeyKey = []byte("nodeKey")
```
directionless: This feels wrong.  I think it's here, to make a constant for how to talk to the store. But passing in a store feels wrong. 

Thinking aloud... This feels like the kind of thing we should do in knapsack -- Something exposes the getter/setter. I might say use knapsack, but don't take the whole thing, declare a smaller interface here?

What do you think?

RebeccaMahany: We have a `RegistrationTracker` interface in knapsack and I was thinking about adding a method there to handle adding registrations, especially because the code in this PR (mostly) duplicates what the extension is already doing to add registrations. I think that update should make the access here feel cleaner. (Mostly I am waiting on doing any larger refactoring till we are sure this is the direction we are going.)

RebeccaMahany: Addressed!

PR #2294, opened 2025-06-03
add usetesting linter and fix test cleanups

zackattack01:
While looking into some new linting I noticed our tenv linter was deprecated already. The suggested replacement is `usetesting`, which also flagged some cases where we are not using `t.TempDir()`.

I opted to just update these, as it will handle cleanup automatically as well as ensure uniqueness across tests. Let me know if there are any cases where we think we should be avoiding.

Update: swapping to using `t.TempDir()` introduced several test failures, all on windows and all in the form of `TempDir RemoveAll cleanup: remove C:\x\y\z.txt: The process cannot access the file because it is being used by another process.`

Digging in, all of these failures were legitimate- we were previously swallowing any errors by attempted cleanups (e.g. `os.RemoveAll`, but the failed cleanup from `t.TempDir()` loudly fails the test.

The reasons were different for each, several of them are addressed [here](https://github.com/kolide/launcher/pull/2296), where our `makeTempDB` helper was closing the new db connection but missed the temp file handle we had created.

The only one that required a non-test change (also non-functional for now) was in `locallogger` - we do not expose a Close function or hold a reference to our underlying `lumberjack.Logger` which keeps a file handle open until `Close` is called, preventing cleanup. For now I've added a close function to fix the test, I will test out using that on shutdown in a separate PR

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -255,8 +255,8 @@ jobs:
           - macos-13
           - macos-14
           - macos-15
-          - windows-2019
           - windows-2022
+          - windows-2025
     needs:
       - version_baseline # version_baseline implies build_linux
       - build # need the other builds too
--- a/ee/allowedcmd/cmd_darwin.go
+++ b/ee/allowedcmd/cmd_darwin.go
@@ -150,3 +150,7 @@ func Zfs(ctx context.Context, arg ...string) (*TracedCmd, error) {
 func Zpool(ctx context.Context, arg ...string) (*TracedCmd, error) {
 	return validatedCommand(ctx, "/usr/sbin/zpool", arg...)
 }
+
+func Zscli(ctx context.Context, arg ...string) (*TracedCmd, error) {
+	return validatedCommand(ctx, "/Applications/Zscaler/Zscaler.app/Contents/PlugIns/zscli", arg...)
+}
--- a/ee/allowedcmd/cmd_windows.go
+++ b/ee/allowedcmd/cmd_windows.go
@@ -54,3 +54,7 @@ func ZerotierCli(ctx context.Context, arg ...string) (*TracedCmd, error) {
 	// For windows, "-q" should be prepended before all other args
 	return validatedCommand(ctx, filepath.Join(os.Getenv("SYSTEMROOT"), "ProgramData", "ZeroTier", "One", "zerotier-one_x64.exe"), append([]string{"-q"}, arg...)...)
 }
+
+func Zscli(ctx context.Context, arg ...string) (*TracedCmd, error) {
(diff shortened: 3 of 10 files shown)


ee/agent/storage/ci/keyvalue_store_ci.go
```diff
@@ -30,32 +29,14 @@ func NewStore(t *testing.T, slogger *slog.Logger, bucketName string) (types.KVSt
 // SetupDB is used for creating bbolt databases for testing
 func SetupDB(t *testing.T) *bbolt.DB {
 	// Create a temp directory to hold our bbolt db
-	var dbDir string
```
zackattack01, line 33: I couldn't find any usages which could/should result in t being nil so simplified this method to only use the recommended `t.TempDir()` approach. let me know if this seems like a mistake

pkg/debug/debug_test.go
```diff
@@ -25,7 +25,7 @@ func getDebugURL(t *testing.T, tokenPath string) string {
 
 func TestStartDebugServer(t *testing.T) {
 	t.Parallel()
-	tokenFile, err := os.CreateTemp("", "kolide_debug_test")
+	tokenFile, err := os.CreateTemp(t.TempDir(), "kolide_debug_test")
```
zackattack01, line 28: the usages in this file were missing cleanups already anyway

pkg/osquery/extension_test.go
```diff
@@ -96,25 +95,18 @@ func TestNewExtensionEmptyEnrollSecret(t *testing.T) {
 }
 
 func TestNewExtensionDatabaseError(t *testing.T) {
-
```
zackattack01: i think this test was just confused as this shouldn't be needed at all, since `makeTempDB` handles this logic all over again we were creating two temp dirs/files to only ever delete the first

pkg/osquery/extension_test.go
```diff
@@ -79,35 +79,29 @@ func TestNewExtensionEmptyEnrollSecret(t *testing.T) {
...
-	if err != nil {
-		t.Fatalf("opening bolt DB: %s", err.Error())
-	}
-	defer func() {
-		db.Close()
-		os.Remove(file.Name())
-	}()
 
 	m := mocks.NewKnapsack(t)
-	m.On("ConfigStore").Return(agentbbolt.NewStore(context.TODO(), multislogger.NewNopLogger(), db, storage.ConfigStore.String()))
```
zackattack01, line 106: this was actually just failing the entire time and returning nil, error due to the attempted bucket creation from a read only connection (not mocking as intended). closing the DB before creating the extension seems to be more in line with the type of Database Error I think this was trying to test

RebeccaMahany (approved):
🔥

PR #2295, opened 2025-06-03
Proposal for standardizing log levels in launcher

RebeccaMahany:
Follows up from discussion here: https://github.com/kolide/launcher/pull/2283#discussion_r2114675163.

I think this largely already aligns with what we've been doing lately, but I think it's helpful to document our standard. Also, older logs don't necessarily fall in line with this proposal as we used to treat log levels differently prior to the introduction of slog and logshipper, so this document will help us update older logs. Finally, `LevelReportedError` is a new log level for our code base and this will document how we plan to use it.

### DIFF
--- a/docs/architecture/log_levels.md
+++ b/docs/architecture/log_levels.md
@@ -11,11 +11,11 @@ When troubleshooting, we want to be able to distinguish between and easily find 
 * Error information (any time launcher tries to perform an action but cannot)
 * Critical error information (any time an actor exits and must be restarted due to an error)
 
-We want logs to be stored by level in different locations:
+We want logs to be stored by level in different locations according to purpose:
 
-* Stored locally and collected via flare, to troubleshoot individual devices
-* Shipped to Cloud Log, to troubleshoot devices across a tenant or multiple tenants
-* Shipped to Cloud Log and Error Reporting, to identify error trends that might indicate launcher bugs
+* Stored locally and collected via flare; used to troubleshoot individual devices; may be more verbose
+* Shipped to Cloud Log; used to troubleshoot devices across a tenant or multiple tenants, or to look for trends across a tenant or multiple tenants
+* Shipped to Cloud Log and Error Reporting; used to identify errors and highlight trends that might indicate launcher bugs
 
 ## Proposal
 
--- a/docs/architecture/log_levels.md
+++ b/docs/architecture/log_levels.md
@@ -1,0 +1,122 @@
+# Log levels
+
+Using "actor" here to describe a repeated or ongoing routine in launcher. It includes all rungroup actors, plus launcher itself, plus some other one-offs.
+
+## What we need logs and log levels for
+
+When troubleshooting, we want to be able to distinguish between and easily find the following categories of logs:
+
+* Positive confirmation that launcher is operating as expected (routine actions are succeeding)
+* Contextual status information (how launcher is performing at a given time)
+* Error information (any time launcher tries to perform an action but cannot)
+* Critical error information (any time an actor exits and must be restarted due to an error)
+
+We want logs to be stored by level in different locations:
+
+* Stored locally and collected via flare, to troubleshoot individual devices
+* Shipped to Cloud Log, to troubleshoot devices across a tenant or multiple tenants
+* Shipped to Cloud Log and Error Reporting, to identify error trends that might indicate launcher bugs
+
+## Proposal
+
+We should use the same five levels we already have -- we don't need to add additional complexity by adding more, especially since we don't have more categories of logs to cover. We should use the following definitions to level our logs:
+
+* LevelDebug: Routine information, such as ongoing status or performance
+* LevelInfo: Normal but significant events, such as start up or shut down
+* LevelWarn: Error occurred that is temporary or that does not prevent the action from continuing
+* LevelError: Error occurred that prevents action from continuing
+* LevelReportedError: Error occurred that results in actor shutdown
+
+We could consider reporting "Error occurred that prevents _significant_ action from continuing" logs on the LevelReportedError level.
+
+We should do the following with respect to log storage:
+
+* We should continue to ship logs at LevelInfo and above to Cloud Log
+* We should consider increasing the amount of logs that we keep locally, as sometimes we don't have enough history available there to troubleshoot
+
+### Example categorized logs in launcher
+
+I went through a selection of logs, categorized them, and then applied our new levels to them:
+
+| Level | Category description |
+|---|---|
+| LevelDebug | Status info |
+| LevelDebug | Configuration info |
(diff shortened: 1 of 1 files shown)


docs/architecture/log_levels.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Proposal
+
+We should use the same five levels we already have -- we don't need to add additional complexity by adding more, especially since we don't have more categories of logs to cover. We should use the following definitions to level our logs:
+
+* LevelDebug: Routine information, such as ongoing status or performance
+* LevelInfo: Normal but significant events, such as start up or shut down
+* LevelWarn: Error occurred that is temporary or that does not prevent the action from continuing
+* LevelError: Error occurred that prevents action from continuing
+* LevelReportedError: Error occurred that results in actor shutdown
```
RebeccaMahany, line 28: Adding this after making https://github.com/kolide/launcher/pull/2305 --

```suggestion
* LevelReportedError: Error occurred that results in actor shutdown, or error occurred that prevents very important action from continuing
```

docs/architecture/log_levels.md
```diff
@@ -0,0 +1,122 @@
...
+* Positive confirmation that launcher is operating as expected (routine actions are succeeding)
+* Contextual status information (how launcher is performing at a given time)
+* Error information (any time launcher tries to perform an action but cannot)
+* Critical error information (any time an actor exits and must be restarted due to an error)
+
+We want logs to be stored by level in different locations:
+
+* Stored locally and collected via flare, to troubleshoot individual devices
+* Shipped to Cloud Log, to troubleshoot devices across a tenant or multiple tenants
+* Shipped to Cloud Log and Error Reporting, to identify error trends that might indicate launcher bugs
```
directionless: I'm not wholly sure what the difference here is. I know what it is technically, but the words here don't quite line up. They're sort of additive, but it's not quite explicit. 

```suggestion
* Stored locally and collected via flare: To troubleshoot individual devices, may be more verbose
* Shipped to Cloud Log: To troubleshoot devices, or look for trends across a tenant or multiple tenants
* Error Reporting: Identify errors and highlite trends that might indicate launcher bugs
```

docs/architecture/log_levels.md
directionless: I remain skeptical about whether we need the `LevelReportedError` category. Something feels very squirrely about warn vs error vs reportederror. Are there examples of what things are clearly in each?

docs/architecture/log_levels.md
```diff
@@ -0,0 +1,122 @@
...
+* LevelWarn: Error occurred that is temporary or that does not prevent the action from continuing
+* LevelError: Error occurred that prevents action from continuing
+* LevelReportedError: Error occurred that results in actor shutdown
+
+We could consider reporting "Error occurred that prevents _significant_ action from continuing" logs on the LevelReportedError level.
+
+We should do the following with respect to log storage:
+
+* We should continue to ship logs at LevelInfo and above to Cloud Log
+* We should consider increasing the amount of logs that we keep locally, as sometimes we don't have enough history available there to troubleshoot
```
directionless, line 35: I've been thinking about this.... because we're logging all the localserver access, it's getting really noisy. I wonder if there's a dedup style "Recived X times in the last minute" approach. Or if we should just up the collection instead.

docs/architecture/log_levels.md
```diff
@@ -0,0 +1,122 @@
...
+| LevelDebug | Status info |
+| LevelDebug | Configuration info |
+| LevelDebug or LevelInfo | Successfully performed action |
+| LevelInfo | Actor startup |
+| LevelInfo | Actor shutdown |
+| LevelInfo | Launcher detected error state and is performing self-corrective action |
+| LevelWarn | Could not create non-critical actor, proceeding without it |
+| LevelWarn | Potentially temporary error performing action, will retry |
+| LevelWarn | Error performing action, can proceed |
+| LevelError | Error performing action, cannot proceed |
```
directionless, line 52: Why wouldn't we push this error to reporting?

docs/architecture/log_levels.md
```diff
@@ -0,0 +1,122 @@
...
+
+| Level | Meaning |
+|---|---|
+| DEFAULT 	(0) 	| The log entry has no assigned severity level. |
+| DEBUG 	(100)	| Debug or trace information. |
+| INFO 		(200)	| Routine information, such as ongoing status or performance. |
+| NOTICE 	(300)	| Normal but significant events, such as start up, shut down, or a configuration change. |
+| WARNING 	(400)	| Warning events might cause problems. |
+| ERROR 	(500)	| Error events are likely to cause problems. |
+| CRITICAL 	(600)	| Critical events cause more severe problems or outages. |
```
directionless, line 118: huh.. _this_ highlites what feels off to me.

You're kinda defining `error` and `critical`, but `critical` is getting conflated with the reporting path. 

So maybe we want to use these, and then decide, separately, what level to report on?

docs/architecture/log_levels.md
RebeccaMahany: Fwiw I find the (very verbose) localserver logs super useful to have in flares because they can help us determine whether something is a launcher vs k2/browser issue

docs/architecture/log_levels.md
RebeccaMahany: https://github.com/kolide/launcher/pull/2305/files has all of the `LevelError` messages that I upgraded to `LevelReportedError` -- I reviewed all the `LevelError` logs in launcher and those were the ones I decided to upgrade. I will put actual examples into `Example categorized logs in launcher` section as well.

docs/architecture/log_levels.md
RebeccaMahany: I'm not really sure we need a `critical`?

I have been thinking of `LevelReportedError` as basically "error, and also we want to put this error into the reporting path" -- that's why we re-level `LevelReportedError` to `LevelError` when shipping logs.

docs/architecture/log_levels.md
directionless: Would they be as useful if they said "message repeated 60 times in the last minute"

docs/architecture/log_levels.md
directionless: You're description elsewhere was:
> * LevelError: Error occurred that prevents action from continuing
> * LevelReportedError: Error occurred that results in actor shutdown

And sounds like ERROR vs CRITICAL.

But  I think the real question is we denote what needs reporting. And what I keep circling around is "why not all errors"

docs/architecture/log_levels.md
RebeccaMahany: Updated in https://github.com/kolide/launcher/pull/2295/commits/3c1fa0321fa4cace2a319d48b15e5cd553675ccd

Issue #1346, opened 2023-09-09
Launcher should detect, and remediate, migration assistent


directionless:
We have a consistent, if infrequent, issue that happens when migration assistant is used to copy between hosts. When this happens, the agent databases are copied between the nodes, and we end up with two hardware things with the same keys.

The work on TPMs (and the Apple Secure Enclave) _would_ have fixed this, but alas we're not currently using the secure enclave.

I think there are a couple of viable paths. And we may need to try all of them.
1. https://github.com/kolide/launcher/issues/554
2. detect serial number change, and _remove the database on startup_

RebeccaMahany:
Some Windows machines (specifically seen on a small number of Windows 10 devices so far) appear to have their hardware UUID changing more frequently than expected, up to as frequently as on every single query to the `system_info` table.

directionless:
We shipped the detection code, but we never enabled the deletion code....

https://github.com/kolide/launcher/blob/7396a0c61bb4f7df4d52d2d30db44debe8d6800e/ee/agent/reset.go#L86-L93

We need to understand if this is safe to enable, and what needs to change to make it safe.

RebeccaMahany:
We should try enabling for macOS only

James-Pickett:
We should consider using multiple factors to determine when to re-enroll. ie. Only re-enroll if `serialChanged && hardwareUuidChanged && tpmKeysNotFound` etc...

We should also be sure to make this a server controlled flag that we can roll out to a percentage of machines and turn off quickly if needed.

RebeccaMahany:
Moved to KOL-906

PR #2297, opened 2025-06-03
Add ResetOnHardwareChangeEnabled feature flag; reset the db if change is detected and flag is enabled

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1346

To avoid a data race where the node key is cached by the osquery extension [around here](https://github.com/kolide/launcher/blob/main/pkg/osquery/extension.go#L178) before detection/remediation happens, we have to switch back to running DetectAndRemediateHardwareChange as a blocking operation rather than a background operation.

This is not ideal -- our P50 for DetectAndRemediateHardwareChange is 2.4s; our P95 is 8.8s. Since this is operation is now blocking, launcher startup will be delayed by an additional ~2-10 seconds.

We _could_ run DetectAndRemediateHardwareChange and update the osquery extension to wipe its cached node key in addition to the node key in the database. We'd also need to do similar with the agent keys that are stored in the database and held in `agent.localDbKeys` -- and possibly similar in other locations as well. This seems tedious, hard to track, and hard to reason about.

I propose making the detect-and-remediate function into a rungroup actor, where the `Execute` function checks for hardware change, and if it detects one, wipes the database _and returns an error_. This will allow the check to be non-blocking, since it will run alongside all the other actors -- but also avoids the data race by effectively restarting launcher and wiping the node key and any other internal state related to the previous device enrollment.


zackattack01 (approved):

directionless:
> but also avoids the data race by effectively restarting launcher and wiping the node key and any other internal state related to the previous device enrollment

This feels clever, but maybe too clever. Similar to autoupdate, it puts us in a situation of needing to try to tell why something semi-routine is happening

PR #2299, opened 2025-06-03
Remove windows-2019 in preparation for deprecation; add windows-2025

RebeccaMahany:
windows-2019 will be EOL'ed at the end of this month: https://github.com/actions/runner-images/issues/12045

Accordingly, removing windows-2019 and adding windows-2025 for our exec test job.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -255,8 +255,8 @@ jobs:
           - macos-13
           - macos-14
           - macos-15
-          - windows-2019
           - windows-2022
+          - windows-2025
     needs:
       - version_baseline # version_baseline implies build_linux
       - build # need the other builds too


.github/workflows/go.yml
```diff
@@ -255,8 +255,8 @@ jobs:
           - macos-13
           - macos-14
           - macos-15
-          - windows-2019
```
directionless, line 258: Is there any way we can test this in a different pattern?

RebeccaMahany: I don't think we can do the docker workaround like we did for older versions of ubuntu, looking at https://hub.docker.com/r/microsoft/windows-server.

The automated nightly tests run against windows-2022 right now, but it looks like we could do windows-2019 or even windows-2016. So maybe expanding our test setup there could be an easier alternative.

Issue #2271, opened 2025-05-21
kolide_program_icons sometimes panics


RebeccaMahany:
Example stack trace:

```
runtime error: invalid memory address or nil pointer dereference
github.com/kolide/launcher/ee/gowrapper.GoWithRecoveryAction.func1.1
	D:/a/launcher/launcher/ee/gowrapper/goroutine.go:31
runtime.gopanic
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/runtime/panic.go:785
runtime.panicmem
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/runtime/panic.go:262
runtime.sigpanic
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/runtime/signal_windows.go:401
compress/flate.(*huffmanBitWriter).writeTokens
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/huffman_bit_writer.go:578
compress/flate.(*huffmanBitWriter).writeBlock
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/huffman_bit_writer.go:492
compress/flate.(*compressor).writeBlock
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/deflate.go:171
compress/flate.(*compressor).deflate
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/deflate.go:409
compress/flate.(*compressor).close
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/deflate.go:639
compress/flate.(*Writer).Close
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/flate/deflate.go:731
compress/zlib.(*Writer).Close
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/compress/zlib/writer.go:184
image/png.(*encoder).writeImage
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/image/png/writer.go:543
image/png.(*encoder).writeIDATs
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/image/png/writer.go:556
image/png.(*Encoder).Encode
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/image/png/writer.go:657
image/png.Encode
	C:/hostedtoolcache/windows/go/1.23.0/x64/src/image/png/writer.go:586
github.com/kolide/launcher/pkg/osquery/table.parseIcoFile
[...]

PR #2304, opened 2025-06-04
Close files after opening

RebeccaMahany:
While investigating https://github.com/kolide/launcher/issues/2271, I noticed we have a handful of places where we're opening a file but not closing it afterward. This PR updates those locations to close the file when we're done with it.

I took a brief look for golangci-linters or revive rules that could have caught this issue, but did not find any.

I also added more verbose error messages to the kolide_program_icons `parseIcoFile`, which will not help with the panic but seemed useful.


zackattack01 (approved):
🔥

PR #2310, opened 2025-06-12
Add Microsoft Defender ATP health tables

Micah-Kolide:
We've got some prospection around having checks for Microsoft Defender ATP on MacOS and Linux. This small PR is to add the health stats of devices through Microsoft Defender.

### DIFF
--- a/ee/disclaim/run_disclaimed_darwin.go
+++ b/ee/disclaim/run_disclaimed_darwin.go
@@ -100,8 +100,9 @@ var allowedCmdGenerators = map[string]allowedCmdGenerator{
 	},
 	"microsoft_defender_atp": {
 		allowedOpts: map[string]struct{}{
-			"health":        {},
-			"--output json": {},
+			"health":   {},
+			"--output": {},
+			"json":     {},
 		},
 		generate: allowedcmd.MicrosoftDefenderATP,
 	},
--- a/ee/disclaim/run_disclaimed_darwin.go
+++ b/ee/disclaim/run_disclaimed_darwin.go
@@ -98,6 +98,13 @@ var allowedCmdGenerators = map[string]allowedCmdGenerator{
 		},
 		generate: allowedcmd.Zscli,
 	},
+	"microsoft_defender_atp": {
+		allowedOpts: map[string]struct{}{
+			"health":        {},
+			"--output json": {},
+		},
+		generate: allowedcmd.MicrosoftDefenderATP,
+	},
 }
 
 func RunDisclaimed(_ *multislogger.MultiSlogger, args []string) error {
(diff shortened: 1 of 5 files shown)


pkg/osquery/table/platform_tables_linux.go
```diff
@@ -73,5 +73,6 @@ func platformSpecificTables(k types.Knapsack, slogger *slog.Logger, currentOsque
 		dataflattentable.NewExecAndParseTable(k, slogger, "kolide_nftables", json.Parser, allowedcmd.Nftables, []string{"-jat", "list", "ruleset"}), // -j (json) -a (show object handles) -t (terse, omit set contents)
 		zfs.ZfsPropertiesPlugin(k, slogger),
 		zfs.ZpoolPropertiesPlugin(k, slogger),
+		dataflattentable.NewExecAndParseTable(k, slogger, "kolide_microsoft_defender_atp_health", json.Parser, allowedcmd.MicrosoftDefenderATP, []string{"health", "--output json"}),
```
RebeccaMahany: I think usually we'd see this as `[]string{"health", "--output", "json"}` -- is there a reason we have `"--output json"` instead? (I took a peek in https://pkg.go.dev/os/exec + stackoverflow but couldn't find much direction one way or another.)

RebeccaMahany (dismissed):
LGTM! Had one question but I'm mostly just curious about it, it's not blocking.

pkg/osquery/table/platform_tables_linux.go
Micah-Kolide: Ahh my bad. There was no reason for it like that. I'll switch it over

ee/disclaim/run_disclaimed_darwin.go
```diff
@@ -98,6 +98,13 @@ var allowedCmdGenerators = map[string]allowedCmdGenerator{
 		},
 		generate: allowedcmd.Zscli,
 	},
+	"microsoft_defender_atp": {
+		allowedOpts: map[string]struct{}{
+			"health":        {},
+			"--output json": {},
```
directionless: If this works like I expect it to, this should probably be 2 array entries, or `"--output": "json"` Two array entries is probably cleaner.

pkg/osquery/table/platform_tables_darwin.go
```diff
@@ -129,5 +129,6 @@ func platformSpecificTables(k types.Knapsack, slogger *slog.Logger, currentOsque
 		dataflattentable.NewExecAndParseTable(k, slogger, "kolide_zscaler", json.Parser, allowedcmd.Launcher, []string{"rundisclaimed", "zscaler", "status", "-s", "all"}),
 		zfs.ZfsPropertiesPlugin(k, slogger),
 		zfs.ZpoolPropertiesPlugin(k, slogger),
+		dataflattentable.NewExecAndParseTable(k, slogger, "kolide_microsoft_defender_atp_health", json.Parser, allowedcmd.Launcher, []string{"rundisclaimed", "microsoft_defender_atp", "health", "--output json"}),
```
directionless: Same.

pkg/osquery/table/platform_tables_linux.go
RebeccaMahany: Sorry, should've mentioned this comment applied to everywhere we have `"--output json"`!

pkg/osquery/table/platform_tables_darwin.go
Micah-Kolide: Updated

RebeccaMahany (approved):

PR #2305, opened 2025-06-05
Upgrade some log levels to LevelReportedError

RebeccaMahany:
Upgrading any logs that seem either particularly interesting or devastating to `LevelReportedError`. Also did a small amount of log cleanup in the process.

PR #2311, opened 2025-06-12
Downgrade some log levels from error to warn

RebeccaMahany:
Relates to https://github.com/kolide/launcher/pull/2295, https://github.com/kolide/launcher/pull/2283, https://github.com/kolide/launcher/pull/2305, https://github.com/kolide/launcher/pull/2309

Now that we can see errors getting reported, downgrade the noisy ones that aren't actionable.

### DIFF
--- a/ee/control/control.go
+++ b/ee/control/control.go
@@ -345,7 +345,7 @@ func (cs *ControlService) fetchAndUpdate(ctx context.Context, subsystem, hash st
 	// Consumer and subscriber(s) notified now
 	if err := cs.update(ctx, subsystem, data); err != nil {
 		// Returning the error so we don't store the hash and we can try again next time
-		slogger.Log(ctx, slog.LevelError,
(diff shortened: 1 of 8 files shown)


directionless (approved):

ee/control/control.go
```diff
@@ -345,7 +345,7 @@ func (cs *ControlService) fetchAndUpdate(ctx context.Context, subsystem, hash st
 	// Consumer and subscriber(s) notified now
 	if err := cs.update(ctx, subsystem, data); err != nil {
 		// Returning the error so we don't store the hash and we can try again next time
-		slogger.Log(ctx, slog.LevelError,
+		slogger.Log(ctx, slog.LevelWarn,
```
RebeccaMahany, line 348: I thiiiiiink this is generally a warning, and if it's a true (actionable/reportable) error then the subsystem itself should be logging it at the error level for reporting.

zackattack01 (approved):

Issue #2302, opened 2025-06-04
kolide_zscaler improvements


RebeccaMahany:
The `zscaler status -s all` command can return various errors, like:

* Please log into ZCC to enable CLI usage by policy.
* The command passed has exceeded the maximum limit allowed. Please try again after some time.
* Unable to fetch status of any service

If we encounter these errors, we should add a row with column `fullkey` set to "error" and the value set to the error message.

PR #2314, opened 2025-06-13
add table options for reporting errors

zackattack01:
this adds 2 new `execTableV2Opts`:
- `WithReportStderr` to allow configuration of tables where we want to see stderr reported separately from stdout. see inline comments for additional details
- `WithReportMissingBinary` to allow configuration of tables where we want to report that the device is missing the target binary, instead of our typical empty response

```
[*]osquery> select fullkey, value from kolide_zscaler;
+---------+---------------------------------+
| fullkey | value                           |
+---------+---------------------------------+
| error   | binary is not present on device |
+---------+---------------------------------+
[*]osquery> select fullkey, value from kolide_zscaler;
+---------+----------------------------------------------------+
| fullkey | value                                              |
+---------+----------------------------------------------------+
| error   | Please log into ZCC to enable CLI usage by policy. |
+---------+----------------------------------------------------+
```

This got a little messy with RunDisclaimed on macos, and we did need to revisit the choice made [here](https://github.com/kolide/launcher/pull/2196/files/6baf22278f32e11f33c1d4d26ad31ba0bf2f3da7#r2035680933) in order to keep stderr clean for reporting in results.

resolves https://github.com/kolide/launcher/issues/2302

### DIFF
--- a/ee/tables/dataflattentable/exec_and_parse.go
+++ b/ee/tables/dataflattentable/exec_and_parse.go
@@ -23,15 +23,16 @@ type bytesFlattener interface {
 
 // execTableV2 is the next iteration of the dataflattentable wrapper. Aim to migrate exec based tables to this.
 type execTableV2 struct {
-	slogger        *slog.Logger
-	tableName      string
-	flattener      bytesFlattener
-	timeoutSeconds int
-	tabledebug     bool
-	includeStderr  bool
-	reportStderr   bool
-	cmd            allowedcmd.AllowedCommand
-	execArgs       []string
+	slogger             *slog.Logger
+	tableName           string
+	flattener           bytesFlattener
+	timeoutSeconds      int
+	tabledebug          bool
+	includeStderr       bool
+	reportStderr        bool
+	reportMissingBinary bool
+	cmd                 allowedcmd.AllowedCommand
+	execArgs            []string
 }
 
 type execTableV2Opt func(*execTableV2)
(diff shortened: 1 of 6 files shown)


ee/tables/dataflattentable/exec_and_parse.go
```diff
@@ -48,12 +49,34 @@ func WithTableDebug() execTableV2Opt {
...
+}
+
+// WithReportMissingBinary will include an error row in the results
+// indicating that the binary is missing. Without this option, queries
+// against missing binaries typically return no results, with the error
+// being ignored.
+// Note that for tables that run through macos RunDisclaimed, we cannot pass our missing
+// binary errors back through- this information is conveyed through stderr, so callers
+// should also include the WithReportStderr option to see the same behavior there.
+func WithReportMissingBinary() execTableV2Opt {
```
directionless, line 74: Do you think we'll ever want not use `WithReportMissingBinary` and `WithReportStderr` together?

directionless: Honestly, I wonder if we want to make them the defaults.

zackattack01: I was surprised we didn't have a need for this before, but don't understand the upstream impacts enough to make a call on default behavior. There seems to be a pattern of no binary == no data everywhere else, but I wasn't sure what (if anything) depends on that assumption from checks.

Either way i'm fine with making this one option if that feels cleaner, our only documented use case for this (zscaler) does want both

Issue #2060, opened 2025-01-23
Table Request: Certificate Trust Settings (macOS)


FritzX6:
## What is being requested?

A new table `kolide_certificate_trust`, which invokes the macOS `security` binary for both the system and the admin certs as described in the help text here:

```
security dump-trust-settings -h
Usage: dump-trust-settings [-s] [-d]
    -s                  Display trusted system certs (default is user)
    -d                  Display trusted admin certs (default is user)
```

These values roughly correspond to the trust settings displayed in the macOS Keychain Access app:

<img width="989" alt="Image" src="https://github.com/user-attachments/assets/0953ebb2-e128-49c2-84fe-cbbbda29afc4" />

<img width="625" alt="Image" src="https://github.com/user-attachments/assets/811bb714-616f-4c46-8a1d-ff3da8f5b61b" />

## What is the purpose of this table?
Several products rely upon intermediate root certificates to inspect web traffic and provide VPN services. Without a properly installed and trusted root certificate, these products may not function as expected or intended.

Root Certificates can also be used in a similar capacity by a malicious actor to man-in-the-middle (MitM) attack a device and exfiltrate sensitive information and web traffic.

Understanding what root certificates are installed on a device (both desirable and undesirable) allows administrators to better assess its security and compliance posture.

## What format does this binary output?

Output of each is a semi-structured format. An example of several different outputs is shown below:

**System Certs (No Trust Settings)**
```
security dump-trust-settings -d
Number of trusted certs = 1
Cert 0: SimpleMDM
   Number of trust settings : 0
```

**System Certs (Trust Settings Enabled)**
```
security dump-trust-settings -d
Number of trusted certs = 1
Cert 0: Example Root CA
   Number of trust settings : 10
   Trust Setting 0:
      Policy OID            : SSL
      Allowed Error         : CSSMERR_TP_CERT_EXPIRED
      Result Type           : kSecTrustSettingsResultTrustRoot
   Trust Setting 1:
      Policy OID            : SSL
      Allowed Error         : Host name mismatch
      Result Type           : kSecTrustSettingsResultTrustRoot
   Trust Setting 2:
      Policy OID            : SMIME
      Allowed Error         : CSSMERR_TP_CERT_EXPIRED
      Result Type           : kSecTrustSettingsResultTrustRoot
   Trust Setting 3:
      Policy OID            : SMIME
      Allowed Error         : S/MIME Email address mismatch
      Result Type           : kSecTrustSettingsResultTrustRoot
[...]

PR #2316, opened 2025-06-17
Add kolide_certificate_trust table to present output of security dump-trust-settings

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2060.

### DIFF
--- a/ee/tables/security/cert_trust_settings.go
+++ b/ee/tables/security/cert_trust_settings.go
@@ -1,0 +1,361 @@
+//go:build darwin
+// +build darwin
+
+package security
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/kolide/launcher/ee/agent/types"
+	"github.com/kolide/launcher/ee/allowedcmd"
+	"github.com/kolide/launcher/ee/dataflatten"
+	"github.com/kolide/launcher/ee/observability"
+	"github.com/kolide/launcher/ee/tables/dataflattentable"
+	"github.com/kolide/launcher/ee/tables/tablehelpers"
+	"github.com/kolide/launcher/ee/tables/tablewrapper"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+type certTrustSettingsTable struct {
+	name    string
+	slogger *slog.Logger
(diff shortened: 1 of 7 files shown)


ee/tables/security/cert_trust_settings.go
```diff
@@ -0,0 +1,361 @@
...
+// parseTrustSetting parses the upcoming trust setting from the given reader.
+// The trust setting, in context, will look like this:
+//
+//	Cert 0: Example Root CA
+//	   Number of trust settings : 1
+//	   Trust Setting 0:
+//	      Policy OID            : SSL
+//	      Allowed Error         : CSSMERR_TP_CERT_EXPIRED
+//	      Result Type           : kSecTrustSettingsResultTrustRoot
+//	Cert 1: ...
```
zackattack01, line 291: docs are great 🔥

ee/tables/security/cert_trust_settings.go
```diff
@@ -0,0 +1,361 @@
...
+	t := trustSetting{}
+	for {
+		// We peek ahead here to see whether we've already finished reading this setting.
+		// We check for a line starting with `certHeaderPrefix` (indicating this was the last
+		// trust setting in our list of trust settings, and we've advanced to the next cert), or
+		// `trustSettingHeaderPrefix` (indicating this was not the last trust setting in our list, and
+		// we've advanced to the next trust setting). The longer of these strings is `trustSettingHeaderPrefix`,
+		// typically with 3 spaces before it. So, we peek ahead len(trustSettingHeaderPrefix)+10, to be safe.
+		nextBytes, err := reader.Peek(len(trustSettingHeaderPrefix) + 10)
+		if err != nil && !errors.Is(err, bufio.ErrBufferFull) && !errors.Is(err, io.EOF) {
```
zackattack01, line 322: if there is an `io.EOF` err here should we break before attempting to parse nextBytes?

RebeccaMahany: No, because we might have one more line that's under `len(trustSettingHeaderPrefix) + 10` bytes long left to read -- I will leave a comment

RebeccaMahany: (i.e., next line is `SomethingShort: 1` and it's the last line in the output, and its length is under `len(trustSettingHeaderPrefix) + 10` -- so we'd get an EOF peeking ahead, but we'd still want to parse that last setting below)

zackattack01 (approved):

PR #2335, opened 2025-06-25
Add .exit command to input in TestProc to ensure that process will shut down

RebeccaMahany:
New theory about test flakiness:

* `loadExtensions` is actually a blocking operation, since it calls `extensionManagerServer.Start()`
* This means that `StartProcess` doesn't actually return _until the process has shut down_
* My theory is that on Ubuntu, the process is actually running happily. This is supported by logs in recent failures (https://github.com/kolide/launcher/actions/runs/15878752198/job/44773402854?pr=2334), showing that the osquery process is responding to queries with results.

I added some documentation to interactive.go to make it clearer that `loadExtensions` isn't just loading extensions but also running the server until it is shut down.

To attempt to address the test flakiness, I have added a `.exit` command to our input, to ensure that the process will shut down so the test can proceed.

### DIFF
--- a/pkg/osquery/interactive/interactive.go
+++ b/pkg/osquery/interactive/interactive.go
@@ -109,7 +109,8 @@ func StartProcess(knapsack types.Knapsack, interactiveRootDir string, inFile, ou
 		"osquery socket file created",
 	)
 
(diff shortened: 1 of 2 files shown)


zackattack01 (approved):
🔥 🔥 🔥

PR #2337, opened 2025-06-30
autoupdate download splay

zackattack01:
- adds a new flag key `autoupdate_download_splay` which defines the duration of time over which launcher will choose a random delay before downloading newly promoted updates and restarting
  - setting to `0s` will effectively disable this new logic, ensuring that launcher downloads any new updates as soon as possible
- adds new logic to `findReleasePromoteTime`: this will search through a given set of tuf targets and find the release by target binary, os, arch, and update channel
  - within this metadata we now add a `promote_time` field which will be unmarshalled from custom metadata and used to determine the beginning of the download splay window
- if at any point launcher sees a promote_time which is older than our autoupdate splay duration is long, it will download and update immediately like it currently does (e.g. any devices which do not see a new update the day it is promoted)


ee/tuf/autoupdate.go
```diff
@@ -479,7 +488,9 @@ func (ta *TufAutoupdater) currentRunningVersion(binary autoupdatableBinary) (str
 
 // checkForUpdate fetches latest metadata from the TUF server, then checks to see if there's
 // a new release that we should download. If so, it will add the release to our updates library.
-func (ta *TufAutoupdater) checkForUpdate(ctx context.Context, binariesToCheck []autoupdatableBinary) error {
+// If allowDelay is set to false, our knapsack.AutoupdateDownloadSplay will be ignored and the update
+// will be downloaded immediately, regardless of promotion time.
+func (ta *TufAutoupdater) checkForUpdate(ctx context.Context, binariesToCheck []autoupdatableBinary, allowDelay bool) error {
```
zackattack01, line 498: if anyone has better ideas for preventing this logic during autoupdate now/channel change requests please let me know- I don't love the bool flag here but trying to pull all of this logic apart was starting to seem like overkill for this

ee/tuf/autoupdate.go
```diff
@@ -586,13 +597,19 @@ func (ta *TufAutoupdater) checkForUpdate(ctx context.Context, binariesToCheck []
...
-func (ta *TufAutoupdater) downloadUpdate(binary autoupdatableBinary, targets data.TargetFiles) (string, error) {
+// and not already downloaded. If allowDelay is true, the download may be delayed according to
+// the promotion time and the knapsack.AutoupdateDownloadSplay.
+func (ta *TufAutoupdater) downloadUpdate(binary autoupdatableBinary, targets data.TargetFiles, allowDelay bool) (string, error) {
 	target, targetMetadata, err := findTarget(context.Background(), binary, targets, ta.pinnedVersions[binary], ta.updateChannel, ta.slogger)
 	if err != nil {
 		return "", fmt.Errorf("could not find appropriate target: %w", err)
 	}
 
+	// TODO move this below currentRunningVersion check below once done testing
```
zackattack01: still testing, will handle this

ee/tuf/autoupdate.go
```diff
@@ -718,3 +761,71 @@ func PlatformArch() string {
...
+	if time.Since(promoteStart) > ta.knapsack.AutoupdateDownloadSplay() {
+		slogger.Log(context.TODO(), slog.LevelDebug,
+			"promote start was longer ago than download splay, will not delay download",
+			"promote_start", promoteStart,
+		)
+		return false
+	}
+
+	delayWindow := int64(ta.knapsack.AutoupdateDownloadSplay().Seconds())
+	splayHash := getSplayHash(ta.knapsack.GetEnrollmentDetails().HardwareUUID)
```
zackattack01: this might be needlessly complicated, curious what others think. According to the current logic we generate a seed for the hash generation per launcher run (set as `splayHashSeed` variable)- I had thought about hardcoding the seed but there is really no need/benefit for this to be consistent across runs given our backstop logic with `promote_time` and the fact that launchers will all be running on different timer interval cycles anyway. which makes me wonder if we really even care about hashing based on an identifier at all... it is probably cleaner to just select a random time?

ee/tuf/autoupdate.go
```diff
@@ -247,7 +253,8 @@ func (ta *TufAutoupdater) Execute() (err error) {
 		ta.slogger.Log(context.TODO(), slog.LevelInfo,
 			"checking for updates",
 		)
-		if err := ta.checkForUpdate(context.TODO(), binaries); err != nil {
+		// always allow delay during routine checks for autoupdates
```
RebeccaMahany: Here and in comments later (e.g. 345, 407) I would specify "splay delay" or something to distinguish this from "AutoupdateInitialDelay" above

ee/tuf/autoupdate.go
```diff
@@ -709,6 +726,32 @@ func findRelease(ctx context.Context, binary autoupdatableBinary, targets data.T
 	return "", data.TargetFileMeta{}, fmt.Errorf("could not find metadata for release target %s for binary %s", releaseTarget, binary)
 }
 
+// findReleasePromoteTime extracts the promotion timestamp from the release file metadata for a given binary and channel.
+// It searches the TUF targets for the appropriate release.json file based on the binary, OS, architecture, and channel,
+// then unmarshals the custom metadata to retrieve the PromoteTime field.
+// Returns the Unix timestamp of when the release was promoted, or 0 if the release file is not found or cannot be parsed.
+func findReleasePromoteTime(ctx context.Context, binary autoupdatableBinary, targets data.TargetFiles, channel string) int64 {
```
RebeccaMahany, line 738: Why do we have to find the release metadata and unmarshall it separately here, instead of doing this in `findRelease`?

ee/tuf/autoupdate.go
RebeccaMahany: Am in favor of the simpler approach w selecting a random time

ee/tuf/autoupdate.go
RebeccaMahany: I am fine w the bool flag, I agree that refactoring to avoid it would be a nontrivial amount of work

directionless: What if K2 just set `autoupdate_download_splay` to 0 for some duration?

ee/tuf/autoupdate.go
zackattack01: it seemed like `findRelease` was returning the `TargetFileMeta` associated with the target itself (just using the release.json entry to find that). but the new metadata is stored in the release.json entries custom blob- I suppose we could take the promote time while unmarshalling for the custom `Target` there and add it to the actual TargetFileMeta.Custom returned, but that didn't really seem right to me either

ee/tuf/autoupdate.go
zackattack01: Discussed offline, we would likely run into some ordering issues if depending on the updated download splay values from K2 and are opting to leave this as is. thanks all!

Issue #2348, opened 2025-07-08
Update allowlist for dt4a endpoint request origins


RebeccaMahany:
Requests should also be allowed to have an origin that matches `https://*.1password.com`. This is the production equivalent of https://github.com/kolide/launcher/pull/2345.

PR #2352, opened 2025-07-09
Add 1p origin pattern to allowlist

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/2348.

Since we now have two one-offs outside of our lookup map `allowlistedDt4aOriginsLookup` (the safari extension prefix, and the new regex for 1p URLs), I've refactored the origin check into its own function.

### DIFF
--- a/ee/localserver/dt4a_auth_middleware_test.go
+++ b/ee/localserver/dt4a_auth_middleware_test.go
@@ -350,6 +350,16 @@ func Test_originIsAllowlisted(t *testing.T) {
 			expectAllowlisted: true,
 		},
 		{
+			testCaseName:      "1p with .ca",
+			requestOrigin:     "https://example2.1password.ca",
+			expectAllowlisted: true,
+		},
+		{
+			testCaseName:      "1p with .eu",
+			requestOrigin:     "https://example3.1password.eu",
+			expectAllowlisted: true,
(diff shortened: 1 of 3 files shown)


cesarfda (dismissed):

ee/localserver/dt4a_auth_middleware_test.go
```diff
@@ -324,6 +324,54 @@ func Test_ValidateCertChain(t *testing.T) {
...
+			testCaseName:      allowlistedOrigin,
+			requestOrigin:     allowlistedOrigin,
+			expectAllowlisted: true,
+		})
+	}
+
+	for _, tt := range testCases {
+		tt := tt
+		t.Run(tt.testCaseName, func(t *testing.T) {
+			t.Parallel()
```
zackattack01, line 380: should we call `originIsAllowlisted` here?

RebeccaMahany: oh my goodness

RebeccaMahany: lol yes, thank you 🤦‍♀️ updated

zackattack01 (approved):

zackattack01 (approved):

PR #2353, opened 2025-07-09
remove user permissions for windows secret file (runtime)

zackattack01:
This does the runtime equivalent of https://github.com/kolide/launcher/pull/2346, for launchers which have already been installed and need to have these permissions updated at runtime.

For some reason I am not seeing the CREATOR/OWNER grant being added while testing (was also the case when inspecting the unit tests). But we are setting the new security info without error, and the result is functionally equivalent (we install as SYSTEM which will be equivalent to CREATOR/OWNER anyway), so I opted to keep the logic as is for now- similar to how we set the root dir permissions.

This accomplishes the reduced ACL scope by omitting any user grant when resetting the security info for the file, we should expect that there will be full permissions granted to the builtin administrators and localsystem groups only


cmd/launcher/svc_config_windows.go
```diff
@@ -456,3 +458,113 @@ func checkRootDirACLs(logger *slog.Logger, rootDirectory string) {
...
+func checkEnrollSecretACLs(logger *slog.Logger, enrollSecretPath string) {
+	logger = logger.With(
+		"component", "checkEnrollSecretACLs",
+		"enroll_secret_path", enrollSecretPath,
+	)
+
+	if strings.TrimSpace(enrollSecretPath) == "" {
+		logger.Log(context.TODO(), slog.LevelError,
+			"unable to check permissions without enroll secret path set, skipping",
+		)
```
RebeccaMahany: I would do a debug level here, or at least not an error level -- it's not really an error we can do something about, and once we support secretless it won't be an error at all

cmd/launcher/svc_config_windows.go
```diff
@@ -456,3 +458,113 @@ func checkRootDirACLs(logger *slog.Logger, rootDirectory string) {
...
+	if err != nil {
+		logger.Log(context.TODO(), slog.LevelError,
+			"generating new DACL from access entries",
+			"err", err,
+		)
+
+		return
+	}
+
+	// apply the new DACL to the root directory
```
RebeccaMahany:
```suggestion
	// apply the new DACL to the secret file
```

cmd/launcher/svc_config_windows.go
zackattack01: 👍 will do!

RebeccaMahany (approved):
🌟

PR #2370, opened 2025-07-22
Replace osquery client in TUF autoupdater with runsimple query instead

RebeccaMahany:


ee/tuf/ci/valid_executable.go
```diff
@@ -71,3 +71,57 @@ func CopyBinary(t *testing.T, executablePath string) {
 	require.NoError(t, os.MkdirAll(filepath.Dir(executablePath), 0755))
 	require.NoError(t, os.Symlink(testOsqueryBinary, executablePath))
 }
+
+var testOlderOsqueryBinary string
+
+const OlderBinaryVersion = "5.17.0"
+
+var downloadOlderOnceFunc = sync.OnceFunc(func() {
+	downloadDir, err := os.MkdirTemp("", "tufci")
```
zackattack01, line 80: do we ever clean this up? I get why we wouldn't want to between tests but I think this will end up with a new osquery download in a random directory each run right?

RebeccaMahany: I'm not sure if we can avoid that? Last time I looked into it, Golang didn't have a test cleanup function that would execute after all other tests in the file were executed.

zackattack01: maybe there's a way we could remove the random aspect of tmp dir creation? I don't mind leaving it after the tests i'm just wondering if there's a way to avoid downloading and leaving behind a new one for each run when i'm running tests locally

zackattack01: i'm fine with addressing this as a separate improvement, there's a few places we can update if we figure something out

zackattack01 (approved):

ee/tuf/ci/valid_executable.go
RebeccaMahany: Sure thing, let me file an issue for that!

RebeccaMahany: https://github.com/kolide/launcher/issues/2373

PR #2375, opened 2025-07-23
Windows system cert

cesarfda:
This pull request introduces support for handling CA certificates on Windows systems, including extracting system certificates, exporting them to a file, and providing fallback mechanisms. It also includes unit tests to ensure robustness. The most important changes are grouped into functionality and testing.

### Functionality:

* Added `extractSystemCerts` and `extractCertsFromStore` functions in `pkg/osquery/extract_windows_certs.go` to extract CA certificates from Windows system certificate stores (`ROOT` and `CA`). These functions use Windows APIs via `syscall` to enumerate and filter certificates.
* Implemented `InstallCaCerts` in `pkg/osquery/install_ca_certs_windows.go` to export system certificates to a file for use by `osquery`. If system certificates cannot be exported, it falls back to an embedded CA bundle.
* Added `exportSystemCaCerts` in `pkg/osquery/install_ca_certs_windows.go` to handle the creation of a PEM file containing system certificates, ensuring reusability by checking for existing files with the same content.

### Testing:

* Added unit tests in `pkg/osquery/extract_certs_windows_test.go` for `extractSystemCerts` and `extractCertsFromStore`, covering both valid and invalid scenarios for certificate extraction.
* Added unit tests in `pkg/osquery/install_ca_certs_windows_test.go` for `InstallCaCerts` and `exportSystemCaCerts`, verifying correct behavior, file reuse, and fallback mechanisms.

### Cross-Platform Support:

* Updated `pkg/osquery/install_ca_certs.go` with a build tag to exclude it from Windows builds, ensuring platform-specific implementations.


pkg/osquery/extract_certs_windows_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Try to extract system certificates
+	certs, err := extractSystemCerts()
+
+	// This might fail in restricted environments
+	if err != nil {
+		require.Error(t, err)
+		require.Nil(t, certs)
+		return
+	}
```
RebeccaMahany: Do we know when specifically we expect this test to fail? If we know when we expect it to fail, we can make the test more specific -- e.g. if we expect this to fail in CI but not locally:

```
if os.Getenv("CI") == "true" {
    		require.Error(t, err)
		require.Nil(t, certs)
		return
}
```

pkg/osquery/extract_certs_windows_test.go
```diff
@@ -0,0 +1,82 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			certs, err := extractCertsFromStore(tc.storeName)
+
+			if tc.storeName == "INVALID_STORE_NAME" {
+				// Should fail for invalid store
+				require.Error(t, err)
+				return
+			}
+
+			// ROOT and CA stores might be empty or inaccessible in some environments
```
RebeccaMahany: Same q as above -- do we know which ones? Is this call failing locally or in CI?

pkg/osquery/extract_windows_certs.go
```diff
@@ -0,0 +1,108 @@
...
+// extractCertsFromStore extracts certificates from a specific Windows certificate store
+func extractCertsFromStore(storeName string) ([]*x509.Certificate, error) {
+	// Convert store name to UTF16
+	storeNamePtr, err := syscall.UTF16PtrFromString(storeName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert store name: %w", err)
+	}
+
+	// Open the certificate store
+	store, _, err := certOpenSystemStore.Call(0, uintptr(unsafe.Pointer(storeNamePtr)))
```
RebeccaMahany, line 75: Could you add a comment with a link to https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certopensystemstorew here? Same goes for the other windows API calls -- it's really helpful to have the documentation handy to e.g. know why the first argument has to be `0`

pkg/osquery/extract_windows_certs.go
```diff
@@ -0,0 +1,108 @@
...
+	var certs []*x509.Certificate
+
+	// Certificate store names to check
+	storeNames := []string{"ROOT", "CA"}
+
+	for _, storeName := range storeNames {
+		storeCerts, err := extractCertsFromStore(storeName)
+		if err != nil {
+			// Log error but continue with other stores
+			fmt.Printf("Warning: Failed to extract certificates from %s store: %v\n", storeName, err)
```
RebeccaMahany: We should pass in a slogger and use it here rather than `fmt.Printf`

pkg/osquery/install_ca_certs_windows.go
```diff
@@ -0,0 +1,78 @@
...
+// On Windows, it exports system certificates to a file for osquery to use.
+func InstallCaCerts(directory string) (string, error) {
+	// Try to export Windows system certificates first
+	systemCertsPath, err := exportSystemCaCerts(directory)
+	if err == nil {
+		return systemCertsPath, nil
+	}
+
+	// If exporting system certs fails, fall back to embedded bundle
+	fmt.Printf("Warning: Failed to export Windows system certificates: %v. Using embedded bundle.\n", err)
```
RebeccaMahany: Same here as before -- we should pass in a slogger and log using that

pkg/osquery/install_ca_certs_windows.go
```diff
@@ -0,0 +1,78 @@
...
+
+	sum := sha256.Sum256(defaultCaCerts)
+	caFile := filepath.Join(directory, fmt.Sprintf("ca-certs-embedded-%x.crt", sum))
+
+	_, err = os.Stat(caFile)
+	switch {
+	case os.IsNotExist(err):
+		return caFile, os.WriteFile(caFile, defaultCaCerts, 0444)
```
RebeccaMahany, line 43: I think these file permissions won't be respected on Windows -- I'm not sure whether we care? If we do, we'd need to do something like what @zackattack01 did in https://github.com/kolide/launcher/pull/2353 to adjust permissions. Maybe feels like overkill for just the curl table here. Thoughts?

pkg/osquery/install_ca_certs_windows_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestExportSystemCaCerts(t *testing.T) {
+	t.Parallel()
+
+	tempDir := t.TempDir()
+
+	// Try to export system certs
+	certsPath, err := exportSystemCaCerts(tempDir)
+
+	// This might fail on some Windows systems without proper permissions
+	// or in restricted environments, so we just verify behavior
```
RebeccaMahany: Same q as in the other test file re which environments we expect this test to fail on

pkg/osquery/extract_windows_certs.go
```diff
@@ -0,0 +1,111 @@
...
+	var certs []*x509.Certificate
+
+	// Certificate store names to check
+	storeNames := []string{"ROOT", "CA"}
+
+	for _, storeName := range storeNames {
+		storeCerts, err := extractCertsFromStore(storeName)
+		if err != nil {
+			// Log error but continue with other stores
+			slog.Warn("Warning: Failed to extract certificates from %s store: %v\n", storeName, err)
```
RebeccaMahany: I know lint says `WarnContext`, but we actually prefer `slogger.Log` (which also takes a ctx -- can use `context.TODO()` here).

We also prefer to not use interpolation for log messages -- sloglint will probably catch that as well on a future check, or at least it should. The reason is that when we have a consistent log message like "failed to extract certificates from store", a) they're easier to grep for in the codebase and b) our error reporting system can group them.

So here we would want something like:

```golang
slog.Log(context.TODO(), slog.LevelWarn,
	"failed to extract certificates from store",
	"store_name", storeName
	"err", err,
)
```

cesarfda: fixed logging on both new files to match the pattern

RebeccaMahany (approved):
Nice work!

Issue #2072, opened 2025-01-28
support periodic generation of pprof heap profiles and other mem info


zackattack01:
While recently debugging some memory issues and looking into the memprofile we generate for flares (see code [here](https://github.com/kolide/launcher/blob/98c5e92d4f2637d2ab507b07b5a480aec70a533d/ee/debug/checkups/runtime.go#L92)) I noticed a few things:

- there are several code paths that are specifically busy while generating a flare, and these tend to overwhelm the results in a way that isn't indicative of what we'd normally expect launcher to be doing (e.g. reading/copying/writing files and zipping for the flare, etc.)
- it is very difficult to find anything meaningful from these taken only at a single point in time

We could likely put ourselves in a much better position to catch real memory issues by supporting a debug mode that generates a collection of these every so often (e.g. once an hour?). I'm not sure what the best way to gate this behavior would be, if we use the existing debug flag we should probably add some sort of rotation/cap for the generated profiles. I'd think we could just pick some temp directory to store them. this could look something like this:
- if (existing or new) debug flag is set:
  - every hour or so we create some artifacts that include:
    - device's runtime when the report is generated
    - launcher's runtime when the report is generated
    - some basic memory stats (e.g. we collect process mem info [here](https://github.com/kolide/launcher/blob/98c5e92d4f2637d2ab507b07b5a480aec70a533d/ee/debug/checkups/processes.go#L47))
    - the heap profile noted above
    - anything else fun that people can think of

directionless:
We sorta used to do this... https://github.com/kolide/launcher/blob/main/pkg/debug listens for a signal, and enables an http interface to `net/http/pprof`. Two problems...
1. Local only -- it was really designed for the developer working locally, not for debugging remote customers
2. no windows -- because it's triggered off a signal, there's no support on windows for it. I took a couple stabs at making a filewatch style twigger, but never got it performant. 

I think it's reasonable to explore creating these periodically. Some brainstorming:
- Needing to put a machine into debug mode sounds hard, and potentially like we'd miss something. What's the impact to generating every hour?
- How big are would they be?
- Do we have the ability to detect anomalous ones? That can feed into retention, or even an automatic submission mode.

directionless:
@RebeccaMahany Did you sneak this in via a different PR?

directionless:
I think we can close this then?

directionless:
Fixed by https://github.com/kolide/launcher/pull/2376

PR #2376, opened 2025-07-23
Add feature-flagged performance monitoring to automatically upload flares

RebeccaMahany:
If the new feature flag is enabled, then launcher will check its performance stats every fifteen minutes. If it notices egregiously high Golang memory usage, then it will trigger a flare upload. Launcher will do this at most once a day.

For now, this only triggers on high Golang memory, but we could trigger on high CPU in the future as well.


RebeccaMahany:
Resolves https://github.com/kolide/launcher/issues/2072

PR #2386, opened 2025-07-31
Extract log level from osquery logs

RebeccaMahany:
Allows us to log osquery error logs at the error level, so they can be aggregated by our error reporting system

### DIFF
--- a/ee/log/osquerylogs/log.go
+++ b/ee/log/osquerylogs/log.go
@@ -77,12 +77,31 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
 
 	msg := strings.TrimSpace(string(p))
 	caller := extractOsqueryCaller(msg)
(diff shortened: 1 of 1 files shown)


ee/log/osquerylogs/log.go
```diff
@@ -77,14 +77,33 @@ func (l *OsqueryLogAdapter) Write(p []byte) (int, error) {
...
 	return len(p), nil
 }
 
+func (l *OsqueryLogAdapter) extractLogLevel(msg string) slog.Level {
+	if len(msg) == 0 {
+		// Use default level
+		return l.level
+	}
+
+	switch msg[0] {
```
directionless, line 95: I know osquery has a verbose log, but I don't remember where it is

RebeccaMahany: In test, I didn't see any levels besides info, warn, and error. We have the "verbose" flag that we set in the extension for the first ten minutes of osquery execution and turn off thereafter, but I don't see those logs marked any differently.

PR #2388, opened 2025-08-04
Log dedup

cesarfda:
This pull request introduces a new content-based deduplication system for local file logging, significantly improving how duplicate log messages are handled and reducing log noise. The main changes include the addition of a `DedupHandler` for `slog` logging, updates to the `localLogger` to support deduplication, and comprehensive tests to ensure correct deduplication behavior.

**Deduplication functionality:**

* Added `DedupHandler` in `pkg/log/locallogger/dedup_handler.go` to wrap `slog.Handler` and deduplicate log messages based on their content, excluding timestamp and source fields. It tracks duplicates, suppresses frequent repeats, and periodically logs the count of suppressed duplicates. (F8060b2aR1)
* Updated `localLogger` in `pkg/log/locallogger/locallogger.go` to use the new `DedupHandler` for `slog` logging, and exposed a `SlogHandler()` method to return the deduplication-enabled handler.
* Introduced configuration constants for deduplication cache expiry, size, and cleanup intervals, and implemented a hash function to generate content-based keys for log entries. [[1]](diffhunk://#diff-5c8f54d4bdc97c16a56d40efd78d8d0c02bc43e225429be05200597b57fe2755R4-R108) [[2]](diffhunk://#diff-5c8f54d4bdc97c16a56d40efd78d8d0c02bc43e225429be05200597b57fe2755R122-R148)

**Integration and usage:**

* Changed logger initialization in `cmd/launcher/main.go` and `cmd/launcher/svc_windows.go` to use the deduplication-enabled slog handler via the new `SlogHandler()` method, replacing manual handler construction. [[1]](diffhunk://#diff-621f76db694d94390921f84a5f98a9560ac41d5448a2d886671baf1e051a9abeL137-R137) [[2]](diffhunk://#diff-78805256d862be689f892e9c4abc9718d9496ec2b16dfd01e93604d8e49e4883L67-R67)

**Testing and validation:**

* Added extensive tests in `pkg/log/locallogger/locallogger_test.go` to verify deduplication logic, including scenarios for identical and different messages, correct suppression and counting, and hash consistency. Also updated existing tests to use the new logger API. [[1]](diffhunk://#diff-e555333ec393aba7bbad98fa247b28938919c08b69aee51ef569134fc2ff1a35R4-R12) [[2]](diffhunk://#diff-e555333ec393aba7bbad98fa247b28938919c08b69aee51ef569134fc2ff1a35L22-R212)

These changes collectively enhance local logging by making debug logs more readable and actionable, especially in environments where repeated log messages can obscure important information.

### DIFF
--- a/pkg/log/locallogger/dedup_handler.go
+++ b/pkg/log/locallogger/dedup_handler.go
@@ -140,8 +140,8 @@ func (dh *DedupHandler) hashRecord(record slog.Record) string {
 func (dh *DedupHandler) shouldSkipDuplicate(hash string) (bool, int) {
 	now := time.Now()
 
-	// Always attempt cleanup - the cleanup function decides if it's needed
-	dh.cleanupCache(now) // This call is non-blocking and triggers async cleanup
+	// Trigger background cleanup if needed (non-blocking)
+	dh.triggerCleanupIfNeeded()
 
 	dh.dedupMutex.Lock()
 	defer dh.dedupMutex.Unlock()
@@ -172,7 +172,7 @@ func (dh *DedupHandler) shouldSkipDuplicate(hash string) (bool, int) {
 }
 
 // triggerCleanupIfNeeded triggers background cleanup if needed
-func (dh *DedupHandler) triggerCleanupIfNeeded(now time.Time) {
+func (dh *DedupHandler) triggerCleanupIfNeeded() {
 	select {
 	case dh.cleanupTrigger <- struct{}{}:
 		// Cleanup triggered successfully
@@ -272,9 +272,4 @@ func (dh *DedupHandler) performCleanup() {
 			"remaining_entries", len(dh.dedupCache),
 		)
 	}
-}
-
-// cleanupCache is a wrapper that calls performCleanup with proper locking
-func (dh *DedupHandler) cleanupCache(now time.Time) {
-	dh.triggerCleanupIfNeeded(now)
 }


pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,124 @@
...
+	cleanupInterval      = 1 * time.Minute // How often to clean up expired entries
+	duplicateLogInterval = 1 * time.Minute // How long to wait before logging a duplicate again
 )
 
+// Fields to exclude when creating content hash for deduplication
+var excludedHashFields = map[string]bool{
+	"ts":     true, // go-kit timestamp
+	"time":   true, // slog timestamp
+	"caller": true, // go-kit caller info
+	"source": true, // slog source info
```
RebeccaMahany: I would also add `original.time` and `original.source` here for desktop and watchdog logs

pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,124 @@
...
 	"github.com/go-kit/kit/log"
 	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 const (
 	truncatedFormatString = "%s[TRUNCATED]"
+
+	// Deduplication configuration
+	defaultCacheExpiry   = 5 * time.Minute // How long to remember log entries
+	defaultMaxCacheSize  = 1000            // Maximum number of unique log entries to track
```
RebeccaMahany: With a very quick search in launcher, I see maybe ~820 different calls to slogger.Log -- I think we might want a larger default here to give us plenty of breathing room for adding more logs? Maybe 2k?

pkg/log/locallogger/locallogger.go
```diff
@@ -69,5 +194,122 @@ func filterResults(keyvals ...interface{}) {
 			}
 		}
 	}
+}
+
+// hashKeyvals creates a hash of the key-value pairs for deduplication
+// We exclude timestamp and caller fields since they change for identical log content
+func (ll *localLogger) hashKeyvals(keyvals ...interface{}) string {
+	return hashKeyValuePairs(keyvals...)
+}
```
RebeccaMahany: What is this wrapper around `hashKeyValuePairs` for?

pkg/log/locallogger/locallogger.go
```diff
@@ -69,5 +194,122 @@ func filterResults(keyvals ...interface{}) {
...
+func (ll *localLogger) shouldSkipDuplicate(hash string) (bool, int) {
+	ll.dedupMutex.Lock()
+	defer ll.dedupMutex.Unlock()
+
+	now := time.Now()
+
+	// Perform cleanup if needed
+	if now.Sub(ll.lastCleanup) > cleanupInterval {
+		ll.cleanupCacheUnsafe(now)
+		ll.lastCleanup = now
```
RebeccaMahany: I think we should set `ll.lastCleanup = now` inside `ll.cleanupCacheUnsafe` -- then we don't have to worry about data races with the `ll.lastCleanup` field if we ever have more routines calling `ll.cleanupCacheUnsafe`, because it will be effectively protected by `ll.dedupMutex`. Plus, responsibility-wise, I think it makes sense for the cache-cleaning routine to own the cache clean timestamp.

pkg/log/locallogger/locallogger.go
```diff
@@ -69,5 +194,122 @@ func filterResults(keyvals ...interface{}) {
...
+		entry.lastLogged = now
+		return false, entry.count // Don't skip, log it with count
+	}
+
+	// Skip this duplicate
+	return true, entry.count
+}
+
+// cleanupCacheUnsafe removes expired entries from the deduplication cache
+// Must be called with dedupMutex held
```
RebeccaMahany: I think we should instead lock `dedupMutex` inside this function (we'd also have to update `shouldSkipDuplicate` to wait to lock `dedupMutex` until after it's finished calling `cleanupCacheUnsafe`). That way, we can guarantee that this function is handling the cache safely, instead of relying on the caller not making a mistake with the mutex.

pkg/log/locallogger/locallogger.go
```diff
@@ -69,5 +194,122 @@ func filterResults(keyvals ...interface{}) {
...
+func (ll *localLogger) cleanupCacheUnsafe(now time.Time) {
+	// Remove entries older than cacheExpiry
+	for hash, entry := range ll.dedupCache {
+		if now.Sub(entry.lastSeen) > ll.cacheExpiry {
+			delete(ll.dedupCache, hash)
+		}
+	}
+
+	// If cache is still too large, remove oldest entries
+	if len(ll.dedupCache) > ll.maxCacheSize {
```
RebeccaMahany: Nitpick -- prefer the early return here

```golang
if len(ll.dedupCache) <= ll.maxCacheSize {
	return
}

// Create a slice of hashes sorted by lastSeen time

```

pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,124 @@
...
+
+// logEntry tracks information about seen log messages for deduplication
+type logEntry struct {
+	firstSeen  time.Time
+	lastSeen   time.Time
+	count      int
+	lastLogged time.Time
+}
+
+// dedupWriter wraps the actual writer and deduplicates JSON log lines
```
RebeccaMahany: I think I'm a little confused on which part owns what. The documentation here says that `dedupWriter` deduplicates JSON log lines, but it's actually `localLogger` that holds all the deduplication logic (in `localLogger.shouldSkipDuplicate`), plus the dedup cache and related items.

Could you explain how the log messages travel through these structs? I think we can probably make the documentation a little easier to follow, and I'm wondering if we're able to refactor/simplify so that we don't have `localLogger` holding a `dedupWriter` _and_ the same `dedupWriter` holding a reference to its containing `localLogger`.

pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,124 @@
...
+	cacheExpiry  time.Duration
+	maxCacheSize int
+	lastCleanup  time.Time
+
+	// Deduplicating writer for slog and other direct writes
+	dedupWriter *dedupWriter
 }
 
-func NewKitLogger(logFilePath string) localLogger {
+func NewKitLogger(logFilePath string) *localLogger {
```
RebeccaMahany, line 54: I don't remember exactly how all the slog handlers work. Does this dedup logic living here mean that we won't be deduplicating the log entries that we ship to the cloud?

RebeccaMahany (commented):
Nice, excited to make the logs easier to wade through!

pkg/log/locallogger/locallogger.go
cesarfda: Done

pkg/log/locallogger/locallogger.go
cesarfda: Good call, done

pkg/log/locallogger/locallogger.go
cesarfda: yeah this would only apply to local logger and not Logshipper or Multislogger

pkg/log/locallogger/locallogger.go
cesarfda: flowchart_localogger
My first attempt was without the dedupWritter and I forgot to update the comments.

I've updated the structure a little bit so I put together this diagram to clarify the call chain

pkg/log/locallogger/locallogger.go
cesarfda: Done

pkg/log/locallogger/locallogger.go
cesarfda: Done

pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,248 @@
...
+
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+
+		// Try to parse as JSON to extract fields for deduplication
+		var logData map[string]interface{}
+		if err := json.Unmarshal([]byte(line), &logData); err != nil {
```
directionless: Needing to unmarshal and remarshal feels weird. 

I wonder if this would be any easier if you did this at the slog layer. I forget how wrappers there work. Maybe as handlers?

pkg/log/locallogger/locallogger.go
```diff
@@ -1,24 +1,248 @@
...
+}
+
+// shouldSkipDuplicate checks if this log message is a duplicate and should be skipped
+// Returns (skip, duplicateCount) where skip indicates if the message should be skipped
+// and duplicateCount is the number of times this message has been seen
+func (w *dedupWriter) shouldSkipDuplicate(hash string) (bool, int) {
+	now := time.Now()
+
+	// Always attempt cleanup - the cleanup function decides if it's needed
+	w.cleanupCache(now)
```
directionless: I think this puts it into the write path? Does it make sense to make this a go routine?

cesarfda: moved into ta goroutine

pkg/log/locallogger/locallogger.go
cesarfda: I'll look into the handler approach

pkg/log/locallogger/dedup_handler.go
```diff
@@ -0,0 +1,280 @@
...
+	if now.Sub(entry.lastLogged) >= duplicateLogInterval {
+		entry.lastLogged = now
+		return false, entry.count // Log with duplicate count
+	}
+
+	return true, entry.count // Skip this duplicate
+}
+
+// triggerCleanupIfNeeded triggers background cleanup if needed
+func (dh *DedupHandler) triggerCleanupIfNeeded(now time.Time) {
```
zackattack01: why does this take a time?

pkg/log/locallogger/dedup_handler.go
```diff
@@ -0,0 +1,280 @@
...
+		dh.slogger.Log(dh.ctx, slog.LevelDebug,
+			"deduplication cache cleanup completed",
+			"expired_entries", expiredCount,
+			"remaining_entries", len(dh.dedupCache),
+		)
+	}
+}
+
+// cleanupCache is a wrapper that calls performCleanup with proper locking
+func (dh *DedupHandler) cleanupCache(now time.Time) {
```
zackattack01: what's the purpose for wrapping this? vs just using triggerCleanupIfNeeded from callers?

pkg/log/locallogger/dedup_handler.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// Handle implements slog.Handler.Handle
+func (dh *DedupHandler) Handle(ctx context.Context, record slog.Record) error {
+	// Create a hash of the log record content (excluding time and source)
+	hash := dh.hashRecord(record)
+
+	// Check if we should skip this duplicate
+	if skip, duplicateCount := dh.shouldSkipDuplicate(hash); skip {
+		return nil // Skip this duplicate
```
zackattack01, line 68: if something logs 10 times in 50 seconds, we'll see the first log without a `duplicate_count` attr, and then nothing else? but if it is logged again after 1 minute and before 5 minutes then we'll get the record with `duplicate_count`?

zackattack01: (not sure that's a big problem, just trying to understand)

cesarfda: Yeah that would reduce the noise on a 1 minute window, we can tweak those as well.
The main gap is there could be a silent expiration, the assumption is that if we are seeing a lot of copies of the same log in a minute we would see it again before 5 minutes. If we are worried about losing information I could add a forced summary on expiration but then I'd need to keep the message instead of just the hash

cesarfda:
Will move changes up to multislogger chain and create a new pr

PR #2389, opened 2025-08-08
Log dedup slogger

cesarfda:
This pull request introduces a deduplication middleware for logging, integrates its lifecycle management into `MultiSlogger`, and adds comprehensive tests for the deduplication logic. The main goal is to prevent duplicate log messages from being emitted within a configurable time window, and to emit summary records for suppressed duplicates. Additionally, `MultiSlogger` now supports proper lifecycle handling, making it compatible with rungroup orchestration. Below are the most important changes grouped by theme.

### Logging Deduplication Middleware

* Added a new deduplication middleware (`dedup.Engine`) to the `MultiSlogger` logging pipeline, ensuring duplicate log messages are suppressed within a configurable window and summary records are emitted for suppressed duplicates. (`pkg/log/multislogger/multislogger.go`, [pkg/log/multislogger/multislogger.goR82-R155](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaR82-R155))
* Comprehensive unit tests for the deduplication middleware were added, covering suppression, relogging, summary emission, disabling deduplication, and cleanup behavior. (`pkg/log/dedup/dedup_middleware_test.go`, [pkg/log/dedup/dedup_middleware_test.goR1-R298](diffhunk://#diff-2ec5a87351595ffe2665913386778860f599963b6df2878be750c374b7ab8fd4R1-R298))

### MultiSlogger Lifecycle Management

* Integrated lifecycle management methods (`Start`, `Stop`, `ExecuteWithContext`, `Interrupt`) into `MultiSlogger` to properly manage background resources (such as deduplication cleanup) and allow seamless integration with rungroup orchestration. (`pkg/log/multislogger/multislogger.go`, [[1]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaR39-R45) [[2]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaR56-R60) [[3]](diffhunk://#diff-8c990b07ab0b7c1c4922b94725cb79d6b3778c5b862c22649749977b769ebadaR82-R155)
* Updated the launcher to manage `multiSlogger` and `systemMultiSlogger` lifecycles via rungroup, ensuring proper startup and shutdown. (`cmd/launcher/launcher.go`, [cmd/launcher/launcher.goR282-R287](diffhunk://#diff-dd532c327d56a5d3623a61b7983e174e88bf93ba87f785ddfe9811a8029a9a63R282-R287))

### Miscellaneous

* Minor test fix: changed the debug message in `TestMultiSlogger` to avoid confusion with deduplication tests. (`pkg/log/multislogger/multislogger_test.go`, [pkg/log/multislogger/multislogger_test.goL49-R52](diffhunk://#diff-d3651b6f5c9c5f8d26d37272682d70100db565cb30fc6bd0ec6f2763a585a818L49-R52))

**Example running locally**
image

### DIFF
--- a/pkg/log/dedup/dedup_middleware.go
+++ b/pkg/log/dedup/dedup_middleware.go
@@ -134,6 +134,7 @@ type Engine struct {
 	started atomic.Bool
 
 	// background cleanup machinery
+	lifecycleLock                sync.Mutex // protects cancel field for Start/Stop operations
 	cancel                       context.CancelFunc
 	backGroundCleanUpWorkerGroup sync.WaitGroup
 
@@ -244,8 +245,12 @@ func (d *Engine) Stop() {
 	if d == nil {
 		return
 	}
-	if d.cancel != nil {
-		d.cancel()
+	d.lifecycleLock.Lock()
+	cancel := d.cancel
+	d.lifecycleLock.Unlock()
+
+	if cancel != nil {
+		cancel()
 	}
 	d.backGroundCleanUpWorkerGroup.Wait()
 	d.started.Store(false)
@@ -263,7 +268,11 @@ func (d *Engine) Start(ctx context.Context) {
 		return
 	}
 	runCtx, cancel := context.WithCancel(ctx)
+
+	d.lifecycleLock.Lock()
 	d.cancel = cancel
+	d.lifecycleLock.Unlock()
+
 	d.started.Store(true)
 	d.backGroundCleanUpWorkerGroup.Add(1)
 	go d.periodicCleanupLoop(runCtx)
--- a/pkg/log/multislogger/multislogger.go
+++ b/pkg/log/multislogger/multislogger.go
@@ -51,6 +51,7 @@ func New(h ...slog.Handler) *MultiSlogger {
 	ms := &MultiSlogger{
 		// setting to fanout with no handlers is noop
 		Logger: slog.New(slogmulti.Fanout()),
+		stopCh: make(chan struct{}),
 	}
 
 	// Initialize deduper once at construction; it will emit summaries using the
@@ -106,9 +107,13 @@ func (m *MultiSlogger) ExecuteWithContext(ctx context.Context) func() error {
 		}
 		// reset interruption state on each run
 		m.interrupted.Store(false)
-		// lazily initialize stop channel
-		if m.stopCh == nil {
+		// recreate stop channel if it was previously closed
+		select {
+		case <-m.stopCh:
+			// channel was closed, create a new one
 			m.stopCh = make(chan struct{})
+		default:
+			// channel is open, use existing one
 		}
 		// start background middleware (dedup cleanup loop)
 		m.Start(ctx)
--- a/pkg/log/multislogger/multislogger_test.go
+++ b/pkg/log/multislogger/multislogger_test.go
@@ -108,7 +108,7 @@ func TestInterrupt_Multiple(t *testing.T) {
 
 	// Start and then interrupt
 	go executeFunc()
-	time.Sleep(3 * time.Second) // Give it time to start
+	time.Sleep(100 * time.Millisecond) // Give it time to start
 	interruptStart := time.Now()
 	multislogger.Interrupt(errors.New("test error"))
 


pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+	wg     sync.WaitGroup
+
+	// logger used to emit records on cleanup
+	logger *slog.Logger
+}
+
+// NewEngine creates a new deduplication engine. The provided logger is used to
+// emit summary records on expiration; pass the pipeline's logger so records go
+// through the same handler chain.
+func NewEngine(logger *slog.Logger, opts ...Option) *Engine {
```
directionless: Why does this need the `logger` passed in? I think the primary usage is to emit the last occurrence on cleanup, right?

I wonder if you can add `next` to the cache?

directionless:
```suggestion
func New(logger *slog.Logger, opts ...Option) *Engine {
```

Prefer `New` in a dedicated package like this

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+		cancel:      cancel,
+		logger:      logger,
+	}
+
+	d.startBackgroundCleanup()
+	return d
+}
+
+// Middleware returns the inline middleware function using the provided engine.
+func Middleware(d *Engine) func(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
```
directionless: This feels a little weird. If you have a struct for the engine, can `Engine.Middleware` have the expected signature and functionality?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+
+	// logger used to emit records on cleanup
+	logger *slog.Logger
+}
+
+// NewEngine creates a new deduplication engine. The provided logger is used to
+// emit summary records on expiration; pass the pipeline's logger so records go
+// through the same handler chain.
+func NewEngine(logger *slog.Logger, opts ...Option) *Engine {
+	cfg := Config{
```
directionless, line 152: Why have a separate struct for `Config` (vs using these as unexpected attributes on engine)

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+// Middleware returns the inline middleware function using the provided engine.
+func Middleware(d *Engine) func(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
+	return func(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
+		// Do not deduplicate debug and lower logs; keep developer visibility.
+		if record.Level < slog.LevelInfo {
+			return next(ctx, record)
+		}
+		// Skip dedup if this is an internally emitted record
```
directionless: I don't understand this code block. I think `EmittedAttrKey` is set to true when it comes out of dedup. So why is it being checked up here?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+				pc:         record.PC,
+			}
+			d.mu.Unlock()
+			// First occurrence: let it through unmodified
+			return next(ctx, record)
+		}
+
+		entry.lastSeen = now
+		entry.count++
+		if now.Sub(entry.lastLogged) >= d.cfg.DuplicateLogWindow {
```
directionless: This block makes sense to me -- I read it as logging the current count if it's been a given interval. 

But you're calculating based on `lastLogged`, which I think will do something a little weird if there's a contineous stream of events. Imagine we're getting a message every 10s. I think we'd want to:
1. log the first one
2. `DuplicateLogWindow` later, log that it's been repeated N times in the last `DuplicateLogWindow`
3. `DuplicateLogWindow` later,  repeat (2)

But I think this code will cause the N count to increase from first seen, and I think that should be reset. (Bonus: I'm not sure you really need to track `lastLogged`, probably just `firstSeen`)

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,384 @@
...
+		)
+		// Emit using the provided logger's handler so it traverses the pipeline
+		if d.logger != nil {
+			_ = d.logger.Handler().Handle(context.Background(), rec)
+		}
+	}
+}
+
+// maybeCleanup triggers cleanup based on time since last cleanup.
+func (d *Engine) maybeCleanup() {
```
directionless: What does this add over `periodicCleanupLoop` ?

pkg/log/dedup/dedup_middleware.go
cesarfda: Under config it's all the levers we can change, there's no functional benefit to using the config struct over those consts being attributes

pkg/log/dedup/dedup_middleware.go
cesarfda: updated the signature to be a method of the engine

pkg/log/dedup/dedup_middleware.go
directionless: You have the same levers if they're struct variables on the engine itself. A common construction would be to make the engine, then pass the engine to the functions.

pkg/log/dedup/dedup_middleware.go
cesarfda: Updated the window to fixed at firstSeen instead of rolling and updating every time we see a duplicate

pkg/log/dedup/dedup_middleware.go
cesarfda: The expired messages go through the logger pipeline so they are skipped to avoid looping over them again

pkg/log/dedup/dedup_middleware.go
cesarfda: During cleanup there's no next present, if we cached the next from when the message was last seen we could mess up the order the messages are coming out

pkg/log/dedup/dedup_middleware.go
cesarfda: maybeCleanup is a speculative trigger called each time the Middleware is used, it would trigger before the periodicCleanup in the case where there's scheduler delays or GC pause

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+const duplicateSummaryMsg = "duplicate_summary"
+
+// excludedHashFields are the attribute keys that should not affect the content hash.
+var excludedHashFields = map[string]bool{
+	"ts":              true, // go-kit timestamp
+	"time":            true, // slog timestamp
+	"caller":          true, // go-kit caller info
+	"source":          true, // slog source info
+	"original.time":   true,
+	"original.source": true,
```
RebeccaMahany: If you want to document these two fields like the others for completeness, something like `// slog timestamp forwarded from desktop/watchdog process (see ee/log package)` and `// slog source info forwarded from desktop/watchdog process (see ee/log package)` would be accurate

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	EmittedAttrKey:    true, // internal marker
+}
+
+// Config controls the behavior of the dedup middleware.
+type Config struct {
+	CacheExpiry        time.Duration
+	MaxCacheSize       int
+	CleanupInterval    time.Duration
+	DuplicateLogWindow time.Duration
```
RebeccaMahany, line 90: I'm thinking we might want a control server flag that would let us control the most important setting(s) here _and_ enable/disable the dedup behavior, just in case. This part of the config seems like a good candidate -- it looks to me like setting `DuplicateLogWindow` to `0s` would effectively disable log deduplication? If yes, what do you think about adding this to the `agent_flags` subsystem? (I think it would be totally fine to do that in a follow-up PR, since there's a bunch of boilerplate involved with adding a new flag.)

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+type Engine struct {
+	// configuration
+	cfg Config
+
+	// runtime state
+	mu          sync.RWMutex
+	cache       map[string]*logEntry
+	lastCleanup time.Time
+	// ensure only one cleanup runs at a time
+	cleanupRunning int32
```
RebeccaMahany: Is there a reason we want an `int32/atomic.Int32` versus an `atomic.Bool` here? My initial thought is that an atomic.Bool might be a little bit easier to immediately understand how it's being used in the engine

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// Engine is a stateful deduplication engine. It is safe for concurrent use.
+type Engine struct {
+	// configuration
+	cfg Config
+
+	// runtime state
+	mu          sync.RWMutex
+	cache       map[string]*logEntry
```
RebeccaMahany: I would add a comment like --

```suggestion
	cache       map[string]*logEntry // maps log hash to corresponding tracked entry
```

or similar, to preemptively explain what the `string` key is.

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	lastCleanup time.Time
+	// ensure only one cleanup runs at a time
+	cleanupRunning int32
+
+	// background cleanup machinery
```
RebeccaMahany, line 136: The organization and documentation in this struct is great, makes this really readable 👍

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	return d
+}
+
+// Middleware is an inline slog middleware method bound to this Engine instance.
+// It matches slog-multi's inline middleware signature.
+func (d *Engine) Middleware(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
+	// Do not deduplicate debug and lower logs; keep developer visibility.
+	if record.Level < slog.LevelInfo {
+		return next(ctx, record)
+	}
```
RebeccaMahany: I would tend toward a) deduplicating logs at all levels, including debug-level logs, and b) allowing devs to disable this via a flag-file setting (e.g. setting `DuplicateLogWindow` to `0s` like I asked about in another comment). My reasoning is that it seems like we have enough noisy debug-level logs that we might want those deduplicated in a flare too, to make it easier to review log files. What do you think?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	pc      uintptr
+}
+
+// Engine is a stateful deduplication engine. It is safe for concurrent use.
+type Engine struct {
+	// configuration
+	cfg Config
+
+	// runtime state
+	mu          sync.RWMutex
```
RebeccaMahany: Could we rename this from `mu` to `cacheLock` or `cacheMutex`  -- or any similar name that explains a) that it's a mutex and b) what it's protecting? I find that makes it easier to reason about how it should be used when I'm reading code.

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	// Create a content hash for this record
+	hash := d.hashRecord(record)
+
+	// Possibly trigger cleanup in the background (non-blocking)
+	d.maybeCleanup()
+
+	// Update dedup state and decide whether to log
+	now := time.Now()
+
+	d.mu.Lock()
```
RebeccaMahany: It _looks_ like we could do a `defer d.mu.Unlock()` here, instead of having to remember it before the 3 separate returns below -- could we make that update or is there something I'm missing?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+			pc:        record.PC,
+		}
+		d.mu.Unlock()
+		// First occurrence: let it through unmodified
+		return next(ctx, record)
+	}
+
+	entry.lastSeen = now
+	entry.count++
+	if now.Sub(entry.firstSeen) >= d.cfg.DuplicateLogWindow {
```
RebeccaMahany: Making sure I'm understanding this correctly -- this effectively means that any duplicated log will get logged at most twice (within the constraints of `DuplicateLogWindow` / `CacheExpiry`)?

1. The first time, the log goes through unmodified
2. The last time is either triggered by a log showing up here after `DuplicateLogWindow` has passed, _or_ by the cleanup routine noticing that this log has not been seen in the `CacheExpiry` window; in both cases, the log goes through with `duplicate_count` added

If that's correct, I think it'd be useful to add a comment here `// Window for tracking this particular log has elapsed -- log the log with its duplicate count` or similar. And I think it'd also be useful to document this behavior more broadly at the top of this file -- you could put it in the docblock for the `Engine` struct, or maybe a mermaid js state diagram would be useful?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+		d.mu.Unlock()
+
+		record.Add("duplicate_count", slog.IntValue(entry.count))
```
RebeccaMahany: What do you think about setting `first_seen` and `last_seen` as attrs on this record also?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+	mu          sync.RWMutex
+	cache       map[string]*logEntry
+	lastCleanup time.Time
+	// ensure only one cleanup runs at a time
+	cleanupRunning int32
+
+	// background cleanup machinery
+	ctx    context.Context //nolint:containedctx // Used for background goroutine lifecycle
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
```
RebeccaMahany: Renaming `wg` to `backgroundCleanupWorker` or similar would help with readability inside `Stop()` especially

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+// performCleanup removes expired entries and emits a summary record for each.
+func (d *Engine) performCleanup(now time.Time) {
+	d.mu.Lock()
+
+	// Remove expired entries
+	if now.Sub(d.lastCleanup) < d.cfg.CleanupInterval {
+		d.mu.Unlock()
+		return
+	}
+	d.lastCleanup = now
```
RebeccaMahany: If we had to wait a while for the lock to be available, `now` could actually be a little bit in the past. Is there a reason we have to pass in `now time.Time` instead of calculating it here?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// runCleanup executes performCleanup but ensures only one cleanup is running concurrently.
+func (d *Engine) runCleanup(now time.Time) {
+	if !atomic.CompareAndSwapInt32(&d.cleanupRunning, 0, 1) {
+		return
+	}
+	defer atomic.StoreInt32(&d.cleanupRunning, 0)
+	d.performCleanup(now)
+}
```
RebeccaMahany: I would suggest getting rid of this function and having `d.performCleanup()` own checking and setting `d.cleanupRunning`. Right now, calling `d.runCleanup()` isn't supposed to run standalone -- it should be called by `d.performCleanup()` to avoid duplicate runs -- but the names of these functions don't indicate that distinction, and it would be really easy for someone to introduce a bug by accidentally calling the wrong one. (Case in point, I have had to edit this comment twice because I swapped the names of the functions incorrectly 😅 .) `d.performCleanup()` runs the cleanup so it should own `d.cleanupRunning` too.

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+			delete(d.cache, items[i].hash)
+		}
+	}
+
+	d.mu.Unlock()
+
+	// Emit outside the lock to avoid re-entrancy deadlocks
+	for _, e := range toEmit {
+		// Build a record that preserves the original call site via PC
+		rec := slog.NewRecord(time.Now(), e.level, duplicateSummaryMsg, 0)
```
RebeccaMahany: I think we want to use `e.message` as the message here, not `duplicateSummaryMsg`. Otherwise, this could group every duplicated error-level log in our error reporting system (even ones with totally different `original_msg`s), and it will make it harder for us to use tools like `jq`. We've already got `duplicate_count` and `EmittedAttrKey` to indicate that this was a duplicate.

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+		rec.PC = e.pc
+		for _, a := range e.attrs {
+			rec.AddAttrs(a)
+		}
+		rec.AddAttrs(
+			slog.Int("duplicate_count", e.count),
+			slog.Bool(EmittedAttrKey, true),
+			slog.String("original_msg", e.message),
```
RebeccaMahany, line 381: Similar question as above -- what do you think about adding first_seen and last_seen attrs here?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+					attrs:   append([]slog.Attr(nil), entry.attrs...),
+					count:   entry.count,
+					pc:      entry.pc,
+				})
+			}
+			delete(d.cache, hash)
+		}
+	}
+
+	// Enforce max cache size by removing oldest without emission
```
RebeccaMahany: Why don't we want to emit here?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,379 @@
...
+func (d *Engine) runCleanup(now time.Time) {
+	if !atomic.CompareAndSwapInt32(&d.cleanupRunning, 0, 1) {
+		return
+	}
+	defer atomic.StoreInt32(&d.cleanupRunning, 0)
+	d.performCleanup(now)
+}
+
+// hashRecord creates a hash of the log record content, excluding time and source information.
+func (d *Engine) hashRecord(record slog.Record) string {
```
RebeccaMahany: I think this can be `func hashRecord(record slog.Record) string {` -- I don't think you're using `d *Engine`.

pkg/log/dedup/dedup_middleware_test.go
```diff
@@ -0,0 +1,280 @@
+package dedup
```
RebeccaMahany, line 1: I will come back and review these tests more thoroughly in a bit, just wanted to add a note that I think we should add a test for `Stop` returning correctly in a timely manner!

pkg/log/multislogger/multislogger.go
```diff
@@ -59,15 +63,36 @@ func NewNopLogger() *slog.Logger {
 func (m *MultiSlogger) AddHandler(handler ...slog.Handler) {
 	m.handlers = append(m.handlers, handler...)
 
+	// Initialize deduper once and keep it persistent
+	if m.dedupEngine == nil {
+		// Temporary logger; will be replaced after construction
+		m.dedupEngine = dedup.New(slog.New(slogmulti.Fanout()))
+	}
```
RebeccaMahany: Should this move inside `New(h ...slog.Handler) *MultiSlogger` instead? That would allow us to set it once, and initialize it in an expected place.

pkg/log/multislogger/syslogger_posix.go
```diff
@@ -8,5 +8,11 @@ import (
 )
 
 func SystemSlogger() (*MultiSlogger, io.Closer, error) {
-	return defaultSystemSlogger(), io.NopCloser(nil), nil
+	ms := defaultSystemSlogger()
+	// Ensure the dedup engine is stopped when the closer is closed
+	closer := closerFunc(func() error {
```
RebeccaMahany: I think relying on this `closer` for shutting down the dedup engine isn't going to be sufficient. For context, the "closer" is set on the system slogger _only_, and it currently exists solely to call `Close` on our event log writer for Windows. So, for our regular multislogger that _isn't_ our syslogger, I don't think we're going to ever call `Stop`.

* It looks like [here](https://github.com/kolide/launcher/blob/main/cmd/launcher/main.go#L55), when we override the system slogger with a new no-op logger, we'd get a new dedup engine that then won't get shut down.
* [Here's](https://github.com/kolide/launcher/blob/main/cmd/launcher/main.go#L130) where we init our main slogger and I don't think that one is getting closed either.

I think we'll need to audit everywhere that the multislogger is instantiated and make sure we're deferring calls to `Stop` in all those areas.

pkg/log/dedup/dedup_middleware.go
cesarfda: Added a specific short-circuit to skip the logic when the window is 0.
Will add a follow up pr for the flag

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: I think that makes sense, having the flag will allow us to easily debug if needed so we won't need to filter it by level.
Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
cesarfda: Added emission for cache size eviction as well

pkg/log/dedup/dedup_middleware.go
cesarfda: updated

pkg/log/multislogger/multislogger.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,453 @@
...
+
+	// runtime state
+	cacheLock   sync.RWMutex
+	cache       map[string]*logEntry // maps log hash to corresponding tracked entry
+	lastCleanup time.Time
+	// ensure only one cleanup runs at a time
+	cleanupRunning atomic.Bool
+
+	// background cleanup machinery
+	ctx                          context.Context //nolint:containedctx // Used for background goroutine lifecycle
```
directionless: Hrm. That seems weird. Normally one would pass this in around some `Start(ctx)` function

pkg/log/dedup/dedup_middleware.go
cesarfda: If we defer, the function would still be holding on the lock as we call next which might also be a duplicate message and reach the middleware causing a deadlock.
I can refactor so we can do everything in a closure and add defer there before returning to next

pkg/log/dedup/dedup_middleware_test.go
cesarfda: Added

pkg/log/multislogger/syslogger_posix.go
cesarfda: Added defer stop calls to:

- cmd/launcher/main.go
- cmd/launcher/svc_windows.go
- cmd/launcher/launcher.go
- cmd/launcher/doctor.go
- cmd/launcher/flare.go
- cmd/launcher/interactive.go
- ee/watchdog/watchdog_task_windows.go
- ee/agent/knapsack/knapsack.go
- ee/agent/types/knapsack.go
- ee/agent/types/mocks/knapsack.go
- pkg/log/multislogger/syslogger_posix.go

cesarfda: Will remove these and look into moving start stop functions being part of a run group

pkg/log/dedup/dedup_middleware.go
cesarfda: Removed the dependency and used the last seen next instead

pkg/log/dedup/dedup_middleware.go
cesarfda: Added Start method and rungroup to handle the lifecycle

pkg/log/multislogger/syslogger_posix.go
cesarfda: Done

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,493 @@
...
+// Middleware is an inline slog middleware method bound to this Engine instance.
+// It matches slog-multi's inline middleware signature.
+func (d *Engine) Middleware(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
+	// If the engine hasn't been started, act as a no-op middleware.
+	if !d.started.Load() {
+		return next(ctx, record)
+	}
+	// Remember the latest downstream 'next' so background cleanup can emit
+	// summary records through the same pipeline.
+	d.lastNext.Store(nextFunc(next))
```
zackattack01, line 180: any reason to do this if disabled? just wondering if the `d.cfg.DuplicateLogWindow` check below should go first here

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,493 @@
...
+	})
+	if skip {
+		return next(ctx, record)
+	}
+
+	// Create a content hash for this record
+	hash := hashRecord(record)
+
+	// Possibly trigger cleanup in the background (non-blocking)
+	d.maybeCleanup()
```
zackattack01: what's the point of calling this here vs relying on the `periodicCleanupLoop`? other than some edge cases coming out of sleep i'd expect this to mostly be a no-op because that timer is firing on the same interval that we check against lastCleanup, but i'm guessing i'm just missing something

cesarfda: That's right maybeCleanup is a speculative trigger called each time the Middleware is used, it would trigger before the periodicCleanup in the case where there's scheduler delays or GC pause

pkg/log/dedup/dedup_middleware.go
cesarfda: moved that check up to avoid storing that reference when the window is 0

pkg/log/dedup/dedup_middleware.go
zackattack01: is this the only reason we have to store and handle `lastCleanup`? just wondering if this could be a little simpler if there was only one thing responsible for the cleanup, and it was outside of the middleware flow. my thinking is if there is a GC pause/scheduler delay, we're also probably not logging anything and it's not a big deal if we take a little more than 5 minutes for whatever reason. if we could get rid of `lastCleanup` entirely there might be a little less mutex contention and fewer code paths to reason about here

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,478 @@
...
+	var toEmit []expired
+
+	for hash, entry := range d.cache {
+		if now.Sub(entry.lastSeen) > d.cfg.CacheExpiry {
+			// Only emit a summary when there were actual duplicates
+			if entry.count > 1 {
+				toEmit = append(toEmit, expired{
+					level:     entry.level,
+					message:   entry.message,
+					attrs:     append([]slog.Attr(nil), entry.attrs...),
```
zackattack01: what's the `[]slog.Attr(nil)` piece for? vs just setting to `entry.attrs`

pkg/log/dedup/dedup_middleware.go
cesarfda: Yeah that will simplify the locks a lot, removed it

pkg/log/dedup/dedup_middleware.go
cesarfda: It's making a copy of the slice to provide clear intent that the data here should be maintained if the underlying changes. Even though in the engine this shouldn't change it's a cheap copy to make it concurrent-safe and future proof if we ever add any change to the attrs in the engine

pkg/log/multislogger/multislogger.go
```diff
@@ -65,9 +79,80 @@ func (m *MultiSlogger) AddHandler(handler ...slog.Handler) {
...
+		case <-m.stopCh:
+		}
+		return nil
+	}
+}
+
+// Interrupt implements a rungroup-compatible interrupt handler. It is safe to
+// call multiple times; only the first call closes the internal stop channel and
+// stops background resources.
+func (m *MultiSlogger) Interrupt(_ error) {
```
RebeccaMahany, line 132: Could we add a `TestInterrupt_Multiple` test for this? There are a bunch of examples in the codebase that you can basically just copy-paste into the tests for multislogger. That test helps us validate that any actor we add to the rungroup is safe to add.

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
+// Package dedup provides a stateful slog middleware that suppresses bursts of
```
RebeccaMahany, line 1: Thank you for adding the docs, they are really useful!

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+// Default configuration values.
+const (
+	DefaultCacheExpiry        = 5 * time.Minute
+	DefaultMaxCacheSize       = 2000
+	DefaultCleanupInterval    = 1 * time.Minute
+	DefaultDuplicateLogWindow = 0
+)
+
+// duplicateSummaryMsg is no longer used; we emit the original message on summary
+// to preserve grouping in external log processing systems.
```
RebeccaMahany: I think you don't need this comment

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+// It matches slog-multi's inline middleware signature.
+func (d *Engine) Middleware(ctx context.Context, record slog.Record, next func(context.Context, slog.Record) error) error {
+	// If the engine hasn't been started, act as a no-op middleware.
+	if !d.started.Load() {
+		return next(ctx, record)
+	}
+	// If the duplicate window is disabled (<= 0), short-circuit and skip all dedup logic
+	if d.cfg.DuplicateLogWindow <= 0 {
+		return next(ctx, record)
+	}
```
RebeccaMahany: Nitpick -- you could combine these to make them a little more concise, since they're both pretty short conditions and lead to the same exit path.

```suggestion
	// If the engine hasn't been started, or is disabled, act as a no-op middleware.
	if !d.started.Load() || d.cfg.DuplicateLogWindow <= 0 {
		return next(ctx, record)
	}
```

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+// Start launches the periodic cleanup worker bound to the provided context.
+// Subsequent calls are no-ops until Stop is called.
+func (d *Engine) Start(ctx context.Context) {
+	if d == nil {
+		return
+	}
+	// If already started, do nothing.
+	if d.cancel != nil {
+		return
+	}
```
RebeccaMahany: Why do we check `d.cancel` instead of `d.started`?

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+
+// performCleanup removes expired entries and emits a summary record for each.
+func (d *Engine) performCleanup() {
+	now := time.Now()
+	// Ensure only one cleanup runs at a time
+	if !d.cleanupRunning.CompareAndSwap(false, true) {
+		return
+	}
+	defer d.cleanupRunning.Store(false)
+	d.cacheLock.Lock()
```
RebeccaMahany, line 306: Super nitpick -- would move this below the comment `// Build a list to avoid holding the lock while emitting` below to make it immediately obvious why we aren't calling `defer unlock` right away

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+		attrs     []slog.Attr
+		count     int
+		pc        uintptr
+		firstSeen time.Time
+		lastSeen  time.Time
+	}
+	var toEmit []expired
+
+	for hash, entry := range d.cache {
+		if now.Sub(entry.lastSeen) > d.cfg.CacheExpiry {
```
RebeccaMahany: Another nitpick -- we would usually do an early continue to reduce nesting below, like this:

```golang
		if now.Sub(entry.lastSeen) <= d.cfg.CacheExpiry {
				continue
		}
```

pkg/log/dedup/dedup_middleware.go
```diff
@@ -0,0 +1,455 @@
...
+		// Build a record that preserves the original call site via PC and uses the original message
+		rec := slog.NewRecord(time.Now(), e.level, e.message, 0)
+		rec.PC = e.pc
+		for _, a := range e.attrs {
+			rec.AddAttrs(a)
+		}
+		rec.AddAttrs(
+			slog.Int("duplicate_count", e.count),
+			slog.Bool(EmittedAttrKey, true),
+			slog.String("original_msg", e.message),
```
RebeccaMahany, line 381: Do we need this attr?

pkg/log/dedup/dedup_middleware.go
cesarfda: That was how I was checking before I added started to use as the noop check in the middleware. Updated to check started

pkg/log/dedup/dedup_middleware.go
cesarfda: removed

pkg/log/dedup/dedup_middleware.go
cesarfda: added

pkg/log/dedup/dedup_middleware.go
cesarfda: done

pkg/log/multislogger/multislogger.go
cesarfda: done

PR #2391, opened 2025-08-12
Add new `kolide_find_my` table

RebeccaMahany:
On Tahoe, we cannot query for whether "Find My Mac" is enabled using the previous method of querying for the FMM token in NVRAM. In this PR, we add a new table to support querying whether "Find My Mac" is enabled using the `FindMyDevice` framework.

### DIFF
--- a/ee/tables/find_my/FMNSXPCConnection.h
+++ b/ee/tables/find_my/FMNSXPCConnection.h
@@ -18,7 +18,7 @@
 @property (retain, nonatomic) FMFuture *future;
 @property (nonatomic) _Bool _testMockConnection;
 @property (readonly, nonatomic) long long state;
-@property (readonly) unsigned long long hash;
+@property (readonly) unsigned long hash;
 @property (readonly) Class superclass;
 @property (readonly, copy) NSString *description;
 @property (readonly, copy) NSString *debugDescription;


Micah-Kolide (approved):
Looks good to me and it will be easy to add the Find my Network setting if we end up wanting to include that into the check. I do think there will be eventual interest in that setting.

ee/tables/find_my/FMNSXPCConnection.h
```diff
@@ -0,0 +1,37 @@
...
+
+@class FMFuture;
+
+@interface FMNSXPCConnection : NSObject <FMNSXPCConnecting>
+
+@property (retain, nonatomic) NSXPCConnection *connection;
+@property (retain, nonatomic) FMFuture *future;
+@property (nonatomic) _Bool _testMockConnection;
+@property (readonly, nonatomic) long long state;
+@property (readonly) unsigned long long hash;
```
Micah-Kolide: I guess this just needs to be an unsigned long?

Micah-Kolide: Or perhaps xcode needs to be updated for the CI runs

RebeccaMahany: I see the same warning locally too -- updating to `unsigned long` does fix this, but I was hoping to avoid editing the generated file if possible. No luck so far with re-generating these header files with different settings, or updating build flags.

Micah-Kolide: Hmm that's unfortunate that you didn't have any luck with re-generating. I definitely would love to avoid editing it too. Are you using Xcode 16.2 by chance? I see CI is currently on 15.2

RebeccaMahany: Should be `Command Line Tools for Xcode 26.0 beta 5`, based on my software update history, I think. Here's the other relevant bits --

```
% xcode-select --version     
xcode-select version 2412.

% clang --version
Apple clang version 17.0.0 (clang-1700.3.16.4)
Target: arm64-apple-darwin25.0.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

% pkgutil --pkg-info=com.apple.pkg.CLTools_Executables
package-id: com.apple.pkg.CLTools_Executables
version: 26.0.0.0.1.1754022347
volume: /
location: /
install-time: 1754615089
```

Micah-Kolide: Hmm yeah I think that's the newest, same one I'm on as well.

RebeccaMahany: If I can't figure it out this afternoon I'll just update to `unsigned long` so we can get this in the 1.24 release!

PR #2398, opened 2025-08-20
Switch to non-deprecated setup-zig action

RebeccaMahany:
Per https://github.com/goto-bus-stop/setup-zig/issues/88, the setup-zig action we're using is deprecated; this PR updates to use a new action.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -136,7 +136,7 @@ jobs:
       run: make deps
 
     - name: Set up zig
-      uses: goto-bus-stop/setup-zig@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -136,7 +136,7 @@ jobs:
       run: make deps
 
     - name: Set up zig
-      uses: goto-bus-stop/setup-zig@v2
+      uses: mlugg/setup-zig@v1
```
directionless: IIRC we only use this for cross compilation.

zackattack01 (approved):

PR #2400, opened 2025-08-26
bug fix for duration flag value

### LINKED ISSUE #2175: Control server agent flags of type `time.Duration` not stored or parsed as expected
When we receive a time like `4s` or `10m`, the control service stores this as a raw string. But when we pull it out of the data store, the control service expects it to be in nanoseconds. We should update how we store it or how we parse it from storage.

Fix must be backwards-compatible with current control server workaround of sending down the duration in nanoseconds instead of using a duration.

cesarfda:
This pull request enhances the handling of duration flag values by allowing support for both human-readable duration strings (like "4s", "10m") and legacy nanosecond integer formats. It refactors the parsing logic into a dedicated helper method and adds comprehensive tests to ensure correct behavior and backward compatibility.

Closes #2175 

### Improved duration flag parsing

* Added a new helper method `parseControlServerValue` to handle parsing of control server values as either duration strings or nanoseconds, improving code clarity and maintainability.
* Updated the main parsing logic in `get` to use the new helper method, ensuring that both formats are supported and errors are handled consistently.

### Expanded test coverage

* Added multiple new test cases to `TestFlagValueDuration` to verify support for duration strings, fallback to defaults, and continued compatibility with legacy nanosecond values.

### DIFF
--- a/ee/agent/flags/flag_value_duration.go
+++ b/ee/agent/flags/flag_value_duration.go
@@ -62,19 +62,9 @@ func NewDurationFlagValue(slogger *slog.Logger, key keys.FlagKey, opts ...durati
 
 func (d *durationFlagValue) get(controlServerValue []byte) time.Duration {
 	int64Value := d.defaultVal
+
 	if controlServerValue != nil {
-		// Control server provided integers are stored as strings and need to be converted back
-		var err error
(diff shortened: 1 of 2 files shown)


RebeccaMahany (approved):
Thank you for fixing this! 🔥

zackattack01 (approved):
nice!

PR #2402, opened 2025-08-28
Add ProgramIconChecksums table

### LINKED ISSUE #2401: Add kolide_program_icon_checksums table
We are noticing high memory usage on Windows devices when querying the `kolide_program_icons` table -- up to 400MB. This seems to be because we are reading the entire icon into memory in order to return it in the `icon` column. The code path given to us by pprof is `table.GenerateProgramIcons => table.GenerateUninstallerProgramIcons => table.parseIcoFile => ico.Decode => ioutil.ReadAll => io.ReadAll`.

Currently, the table will iterate through all uninstaller program icons and all installer program icons, read each one into memory, and return a row for each with all columns: `icon`, `hash`, `name`, and `version`. However, usually when we are querying the `kolide_program_icons` table, we only care about two, smaller columns: `hash` (the checksum), and `name` (to perform joins against `programs`). Osquery will always query for all of the columns regardless, so we aren't able to optimize within `kolide_program_icons`.

We should add a new table `kolide_program_icon_checksums` that only has columns `hash`, `name`, and `version`.

cesarfda:
This pull request introduces a new osquery table for Windows called `kolide_program_icon_checksums`, which provides checksums for program icons without including the icon data itself. The implementation refactors existing code to support this new functionality and improves the flexibility of icon data collection.

Closes #2401 

**New Table Addition:**

* Added the `kolide_program_icon_checksums` table to the platform-specific tables for Windows, allowing users to query program icon hashes alongside program names and versions, but without the icon image data. [[1]](diffhunk://#diff-509fce9bc1a12231004e66cc7ac9dbcda803034495fc9732b12cc8980f12aca4R26) [[2]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cR41-R74)

**Refactoring for Data Collection:**

* Refactored the icon data generation functions (`generateUninstallerProgramIcons` and `generateInstallersProgramIcons`) to accept an `includeIcon` parameter, enabling reuse of logic for both tables and controlling whether icon image data is included in the results. [[1]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cR41-R74) [[2]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cL77-R108) [[3]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cL116-R142) [[4]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cL139-R174)

**Implementation Details:**

* Updated the logic in both icon generation functions to conditionally include the `"icon"` field in the result map based on the `includeIcon` parameter, ensuring the correct schema for each table. [[1]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cL77-R108) [[2]](diffhunk://#diff-7004279a92ce1f57116ef50af15b2519927912f21c44315006bc1214b5d0227cL139-R174)
* Created the `generateProgramIconChecksums` function to aggregate program icon hashes for the new table, leveraging the refactored icon generation functions with `includeIcon` set to `false`.

These changes improve the flexibility and maintainability of the program icon tables, allowing for both detailed icon data and lightweight checksum queries.

### DIFF
--- a/pkg/osquery/table/program_icons_windows.go
+++ b/pkg/osquery/table/program_icons_windows.go
@@ -90,7 +90,7 @@ func generateUninstallerProgramIcons(ctx context.Context, includeIcon bool) []ma
 				continue
 			}
 
-			icon, err := parseIcoFile(ctx, iconPath)
+			icon, err := parseIcoFile(ctx, iconPath, includeIcon)
 			if err != nil {
 				continue
 			}
@@ -157,7 +157,7 @@ func generateInstallersProgramIcons(ctx context.Context, includeIcon bool) []map
 				continue
 			}
 
-			icon, err := parseIcoFile(ctx, iconPath)
+			icon, err := parseIcoFile(ctx, iconPath, includeIcon)
 			if err != nil {
 				continue
 			}
@@ -205,7 +205,7 @@ func getRegistryKeyProductData(ctx context.Context, key registry.Key, path strin
 //
 // This doesn't support extracting an icon from a exe. Windows stores some icon in
 // the exe like 'OneDriveSetup.exe,-101'
-func parseIcoFile(ctx context.Context, fullPath string) (icon, error) {
+func parseIcoFile(ctx context.Context, fullPath string, includeIcon bool) (icon, error) {
 	_, span := observability.StartSpan(ctx, "icon_path", fullPath)
 	defer span.End()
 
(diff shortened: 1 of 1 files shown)


pkg/osquery/table/program_icons_windows.go
```diff
@@ -74,12 +95,17 @@ func generateUninstallerProgramIcons(ctx context.Context) []map[string]string {
...
-				"icon":    icon.base64,
+			result := map[string]string{
 				"hash":    fmt.Sprintf("%x", icon.hash),
 				"name":    name,
 				"version": version,
-			})
+			}
+
+			if includeIcon {
+				result["icon"] = icon.base64
```
RebeccaMahany, line 105: Can we also handle this further down, inside `parseIcoFile`? i.e., also pass `includeIcon` to `parseIcoFile`, and have that function skip even performing the base64 encoding --

```
checksum := crc64.Checksum(buf.Bytes(), crcTable)
if includeIcon {
	return icon{base64: base64.StdEncoding.EncodeToString(buf.Bytes()), hash: checksum}, nil
}
return icon{base64: "", hash: checksum}, nil
```

cesarfda: Updated

RebeccaMahany (approved):
Nice!!

Issue #2192, opened 2025-04-03
Should wait until Windows exits Modern Standby to reload KATC extension


RebeccaMahany:
https://github.com/kolide/launcher/pull/2128 was a helpful improvement to avoid restarting the osquery instance if we can help it when the KATC configuration changes. However, it didn't take into account modern standby on Windows.

@zackattack01 noticed that when his Windows device is in modern standby and launcher receives an updated KATC configuration, the osquery instance can die and not be able to be restarted by launcher.

The logs report that the KATC extension was successfully updated in place -- `KATC configuration changed, successfully updated in-place without restarting instances` -- but then the osquery instance receives an i/o timeout when registering the KATC extension, and this causes the instance to shut down. Then, because we're in modern standby, the osquery runner is unable to successfully restart the osquery instance.

We should delay the KATC extension reload until the device exits modern standby.

PR #2408, opened 2025-09-11
delay osq restarts while in modern standby

James-Pickett:
Add flag to osquery runner for needed restart, if restart called while where in modern standby, flag is set and no restart performed.

Subscribed osquery runner to get notification when InModernStandby flag changes and executes restart if restart flag is set or flag other than InModernStandby is set.

Closes [1292](https://github.com/kolide/launcher/issues/2192)

### DIFF
--- a/pkg/osquery/runtime/runner.go
+++ b/pkg/osquery/runtime/runner.go
@@ -37,7 +37,7 @@ type Runner struct {
 	shutdown        chan struct{}
 	interrupted     *atomic.Bool
 	needsRestart    *atomic.Bool
-	restartLock     sync.Mutex // use a restart lock to ensure we don't get multiple quick succesion restarts due to in modern standy flapping
+	restartLock     sync.Mutex // use a restart lock to ensure we don't get multiple quick succession restarts due to in modern standy flapping
 }
 
 func New(k types.Knapsack, serviceClient service.KolideService, settingsWriter settingsStoreWriter, opts ...OsqueryInstanceOption) *Runner {


pkg/osquery/runtime/runner.go
```diff
@@ -269,20 +275,32 @@ func (r *Runner) triggerShutdownForInstances(ctx context.Context) error {
...
+
 	r.slogger.Log(ctx, slog.LevelDebug,
-		"control server flags changed, restarting instance to apply",
+		"restarting osquery instances due to flag changes or needed restart",
 		"flags", fmt.Sprintf("%+v", flagKeys),
+		"in_modern_standby", r.knapsack.InModernStandby(),
+		"needs_restart", r.needsRestart.Load(),
 	)
 
 	if err := r.Restart(ctx); err != nil {
```
RebeccaMahany, line 310: I think maybe worth adding a comment that `r.Restart` will check `r.knapsack.InModernStandby` and decline to restart if it's true -- I expected the check to be here at first

pkg/osquery/runtime/runner.go
```diff
@@ -337,6 +357,20 @@ func (r *Runner) Restart(ctx context.Context) error {
 		"runner.Restart called",
 	)
 
+	// check to see if we are in modern standby; if so, we cannot restart now
```
RebeccaMahany, line 372: I wanted to call out that sometimes we see the `InModernStandby` value flap rapidly when the device changes state. `triggerShutdownForInstances` has a lock in it, so I don't think we'll get simultaneous restart requests, but I think it might be possible for us to end up doing two restart requests in quick succession when we only need one...?

pkg/osquery/runtime/runner.go
James-Pickett: added

pkg/osquery/runtime/runner.go
James-Pickett: Yea, I could see that happening if InModernStandby flapped before the restart func set needsRestart to false, the restart requests would queue up.

I added a restart lock so that FlagsChanged will always have to wait on any pending restarts to finish before it decides if it needs to call Restart.

RebeccaMahany: I think that waiting for the lock here would also delay the flags controller from proceeding to notify any other change observers about any other flags that may have changed -- it'd have to wait until the restart is complete, which might take up to ~30sec.

Maybe no big deal, since it seems pretty unlikely that a control server flag changes at the exact same time that we come out of modern standby with a restart request also already queued up. Or maybe we want to handle it by updating the flags controller to fire off FlagsChanged requests async instead of one by one (in a future PR, not in this one). I think I'm most inclined to try it like you have it now, and see how it goes?

James-Pickett: Looking at the code, we are already delaying the flags controller while waiting restart due to instance lock in triggerShutdownForInstances. However, it previously it would only happen when the watchdog flags changed. So yea, I agree we should try it as is.

PR #2412, opened 2025-09-23
Assess performance impact of tables (benchmark)

RebeccaMahany:
Adds our first benchmark tests for tables, plus a package to allow us to report non-Golang memory allocations during benchmarking. Also adds Makefile + CI support, saving our benchmark output as artifacts. In the future, we can e.g. compare the benchmark output from the PR against the benchmark output from main, to see if we have regressions in performance.

### DIFF
--- a/ee/agent/dbdump.go
+++ b/ee/agent/dbdump.go
@@ -14,8 +14,10 @@ type bucketStatsHolder struct {
 }
 
 type dbStatsHolder struct {
-	Stats bbolt.TxStats
-	Size  int64
+	Stats           bbolt.TxStats
+	Size            int64
+	ReadTxCount     int
+	OpenReadTxCount int
 }
 
 type Stats struct {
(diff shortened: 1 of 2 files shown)


ee/tables/app-icons/app_icons_darwin_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Set up table
+	appIconsTable := AppIcons(mockFlags, slogger)
+
+	// Report memory allocations
+	b.ReportAllocs()
+
+	for range b.N {
+		// Confirm we can call the table successfully
+		response := appIconsTable.Call(context.TODO(), map[string]string{
```
directionless: FYI I probably write a helper to generate that map.

ee/tables/app-icons/app_icons_darwin_test.go
```diff
@@ -0,0 +1,52 @@
...
+			"action": "generate",
+			"context": `{
+	"constraints": [
+		{
+			"name": "path",
+			"list": [
+				{
+					"op": 2,
+					"expr": "not a real path"
```
directionless: This may cause it to miss the things that leak memory

ee/tables/app-icons/app_icons_darwin_test.go
RebeccaMahany: Added in https://github.com/kolide/launcher/pull/2412/commits/25f739dab3f962c38dc90d866b8b25e230d37bc6

PR #2413, opened 2025-09-23
support build options for debug builds with symbols

zackattack01:
- finishes wiring up our WithoutStripped flag to allow us to build binaries with debug symbols
- adds gcflags suggested for debug builds
- can be used like: `MAKE_debugsymbols=true make build`

### DIFF
--- a/pkg/make/builder.go
+++ b/pkg/make/builder.go
@@ -293,7 +293,11 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 		if !b.notStripped {
 			ldFlags = append(ldFlags, "-w -s")
 		} else {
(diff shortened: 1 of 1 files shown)


pkg/make/builder.go
```diff
@@ -292,6 +292,10 @@ func (b *Builder) BuildCmd(src, appName string) func(context.Context) error {
 
 		if !b.notStripped {
 			ldFlags = append(ldFlags, "-w -s")
+		} else {
+			fmt.Println("compiling debug build without stripping symbols")
```
RebeccaMahany: I think we have a logger available here instead

zackattack01: thank you!

PR #2418, opened 2025-09-26
Updates to auto triggered flares

directionless:
When the autoflares trigger in a high CPU environment, sometimes they don't complete. This is a couple of attempts at improving that. And in generally trying to get better CPU profiling

1. Create a separate flare environment for perf triggers. Having written this, I think the environment and type are overlapping, and it's messy
2. Limit which flares run in the perf environment
3. Add some logging in the hopes that we see _something_
4. tweak the cpu profiling

### DIFF
--- a/ee/debug/checkups/checkups.go
+++ b/ee/debug/checkups/checkups.go
@@ -304,7 +304,7 @@ func RunFlare(ctx context.Context, k types.Knapsack, flareStream io.WriteCloser,
 
 	for _, c := range checkupsFor(k, flareType) {
 		// Log that we're doing this, because sometimes we seem to hang, and we want to debug it.
-		k.Slogger().Log(ctx, slog.LevelDebug,
+		k.Slogger().Log(ctx, slog.LevelInfo,
 			"running flare checkup",
 			"name", c.Name(),
 			"runtime_environment", runtimeEnvironment,
--- a/ee/debug/checkups/checkups.go
+++ b/ee/debug/checkups/checkups.go
@@ -75,7 +75,7 @@ const (
 func checkupsFor(k types.Knapsack, target targetBits) []checkupInt {
 	// This encodes what checkups run in which contexts. This could be pushed down into the checkups directly,
 	// but it seems nice to have it here. TBD
-	  	var runEverywhere targetBits = 255
+	var runEverywhere targetBits = 255
 
 	var potentialCheckups = []struct {
 		c       checkupInt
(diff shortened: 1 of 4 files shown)


ee/debug/checkups/checkups.go
```diff
@@ -292,7 +302,13 @@ func RunFlare(ctx context.Context, k types.Knapsack, flareStream io.WriteCloser,
 		return errors.Join(fmt.Errorf("writing flare environment: %w", err), finalize())
 	}
 
-	for _, c := range checkupsFor(k, flareSupported) {
+	for _, c := range checkupsFor(k, flareType) {
+		// Log that we're doing this, because sometimes we seem to hang, and we want to debug it.
+		k.Slogger().Log(ctx, slog.LevelDebug,
```
directionless: Might want to bump these to Info, debug doesn't ship to the cloud, right?

ee/debug/checkups/runtime.go
```diff
@@ -118,6 +118,12 @@ func gatherPprofCpu(z *zip.Writer) error {
 		return fmt.Errorf("creating cpuprofile: %w", err)
 	}
 
+	// Increase the CPU profiling rate, in the hopes we can capture more / better detail. This
+	// is quite unclear, there's a long discussion at https://github.com/golang/go/issues/40094 and
+	// linked issues. But my takeaway is that in _some_ cases, there's value in raising this. But it
+	// does produce a spurious error -- "runtime: cannot set cpu profile rate until previous profile has finished"
+	runtime.SetCPUProfileRate(200)
```
RebeccaMahany: I think this might actually have a detrimental effect on Windows -- https://github.com/golang/go/issues/61665 (especially https://github.com/golang/go/issues/61665#issuecomment-1659714314)

ee/debug/checkups/checkups.go
RebeccaMahany: Correct -- bumping to Info seems fine

ee/debug/checkups/runtime.go
RebeccaMahany: I have been wondering if we want to try lowering to 60 instead, but haven't tested yet. Can test on Windows with 200 vs 60 if you'd like

RebeccaMahany: Results from testing showed that we didn't see much of a difference between the two:

```
200: Duration: 15.20s, Total samples = 18.98s (124.92%)
64: Duration: 15.19s, Total samples = 19.88s (130.81%)
```

ee/debug/checkups/checkups.go
directionless: Upped it to info. I didn't add much logging, so we won't really know if these completed, or something happened. But it might be enough

ee/debug/checkups/runtime.go
directionless: I replaced this with a giant comment. Because we have no idea what correct is.

directionless: (Thank you)

PR #2422, opened 2025-09-29
Evaluate alternatives to consoleusers on Windows

RebeccaMahany:


ee/consoleuser/consoleuser_windows.go
```diff
@@ -7,44 +7,54 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"slices"
 
+	"github.com/carlpett/winlsa"
```
directionless: We should fork this, and add CI and whatnot.

ee/consoleuser/consoleuser_windows.go
```diff
@@ -7,44 +7,54 @@ import (
...
-	explorerProcs, err := explorerProcesses(ctx)
+	luids, err := winlsa.GetLogonSessions()
 	if err != nil {
-		return nil, fmt.Errorf("getting explorer processes: %w", err)
+		return nil, fmt.Errorf("getting logon sessions: %w", err)
 	}
 
-	// unclear if windows will ever have more than one explorer process for a single user
-	// guard against this by forcing uniqueness
+	activeSids := make([]string, 0)
```
directionless: If there are dups, a `map[string]interface{}` may be slightly more efficient. But whatev. I'm code golfing

directionless (commented):
I think we should fork winlsa before merging

ee/consoleuser/consoleuser_windows.go
directionless:
```suggestion
	"https://github.com/kolide/go-winlsa"
```

ee/consoleuser/consoleuser_windows.go
```diff
@@ -8,43 +8,70 @@ import (
...
-	uids := make([]string, len(uidsMap))
-	uidCount := 0
-	for uid := range uidsMap {
-		uids[uidCount] = uid
-		uidCount++
+	activeUsernameList := make([]string, len(activeUsernames))
+	i := 0
+	for sid := range activeUsernames {
+		activeUsernameList[i] = sid
+		i += 1
```
directionless: Is this just [`map.Keys`](https://pkg.go.dev/maps#Keys)?

directionless: I guess thats an `iter`. Not sure if it would just work though

zackattack01: updated!

ee/consoleuser/consoleuser_windows.go
```diff
@@ -58,12 +60,7 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 		activeUsernames[usernameFromSessionData(sessionData)] = struct{}{}
 	}
 
-	activeUsernameList := make([]string, len(activeUsernames))
-	i := 0
-	for sid := range activeUsernames {
-		activeUsernameList[i] = sid
-		i += 1
-	}
+	activeUsernameList := slices.Collect(maps.Keys(activeUsernames))
```
directionless, line 63: I wonder if we could have changed the signature to iter, instead of bringing in slices.Collect, but sure.

PR #2423, opened 2025-09-30
Limit the maximum number of go procs

directionless:


cmd/launcher/launcher.go
```diff
@@ -736,3 +739,25 @@ func runOsqueryVersionCheckAndAddToKnapsack(ctx context.Context, slogger *slog.L
 		"osqueryd_path", osquerydPath,
 	)
 }
+
+
+// gomaxprocsLimiter sets a limit on the number of OS threads that can be used at a given time.
+func gomaxprocsLimiter(ctx context.Context,  slogger *slog.Logger) {
+	const MAX = 10 // FIXME(seph): Not sure what this should be, or if we need remote control functionality
```
RebeccaMahany: If we're not sure, I feel like probably worth making it an agent flag (at least, every time I'm not sure and decide to _not_ make it an agent flag, I tend to regret it later 😅 )

RebeccaMahany: Discussed offline -- fine for a first pass for testing, can make it an agent flag later if we want

Issue #2415, opened 2025-09-24
`kolide_app_icons` does not work for most apps on Tahoe


RebeccaMahany:
Querying for most apps results in `Error: error generating table: decoding tiff bytes: tiff: unsupported feature: sample format`.

PR #2427, opened 2025-10-01
fix tiff decoding on kolide_app_icons table

James-Pickett:
fix tiff decoding, add tests to kolide_app_icons table

closes https://github.com/kolide/launcher/issues/2415


RebeccaMahany (approved):
Nice!

PR #2436, opened 2025-10-02
Presence Detection Tahoe Crash Fix

James-Pickett:
use C-string instead of NSString for error message so it wont be auto released

closes https://github.com/kolide/launcher/issues/2383


zackattack01 (approved):
nice find!

PR #2437, opened 2025-10-08
refactor download of osquery binary for tests

### LINKED ISSUE #2373: Improve handling of downloaded binaries in tests
Right now, we download the osquery binary for a couple different tests (see `ee/tuf/ci/valid_executable.go`, `pkg/osquery/interactive/interactive_test.go`, `ee/agent/reset_test.go`, `pkg/osquery/runsimple/osqueryrunner_test.go`, `pkg/osquery/runtime/runtime_test.go`). This means six separate downloads of the osquery (`ee/tuf/ci/valid_executable.go` now downloads both an older version of osquery and latest nightly) -- and although these downloads go into the temp directory, they are not cleaned up by the tests.

We would like to achieve:

* Fewer downloads of osquery
* Cleanup of the osquery binary post-test run

Improvements:

- [ ] We should pick one place (maybe somewhere inside pkg/osquery) to hold the logic to download the osquery binary for tests, instead of duplicating it across all the files above.
- [ ] We should either A) reuse the osquery binary on different test runs (by standardizing the filepath name), or B) figure out how to remove the osquery binary after all tests in the given package have run.
- [ ] If we use option A), we should at least clean up the downloaded tarballs in the temp directory, leaving only the bare binary or app bundle.

A: We can standardize the filepath name so that when we re-run tests locally, we don't have to re-download osquery for each test run -- i.e., after you run tests, you will have 1-2 temporary directories holding binaries that will be re-used on subsequent test runs. If we choose this approach, we will probably want to add the osquery version to the filepath so that we _will_ download a new osquery when there's a new nightly available.

B: We can investigate removing the temporary directory holding the osquery binaries after all tests complete. Last I looked, Golang tests didn't have great support for this. If we want to do this, then I think we'd need to go back to using `TestMain` rather than `sync.OnceFunc`. This would mean still doing six separate osquery downloads for one full `make test` run, but would not leave behind any artifacts after test run.

cesarfda:
This pull request refactors how test code downloads and manages the `osqueryd` binary across multiple test suites. Instead of duplicating logic for downloading and preparing the binary in each test, the code now uses a shared utility function, greatly simplifying the setup in test files. Additionally, two new command wrappers are introduced for macOS command-line utilities.

closes #2373.

Key changes include:

**Test binary management simplification:**

* Replaced custom logic for downloading and preparing the `osqueryd` binary in multiple test files (`ee/agent/reset_test.go`, `ee/tuf/ci/valid_executable.go`, `pkg/osquery/interactive/interactive_test.go`, `pkg/osquery/runsimple/osqueryrunner_test.go`, `pkg/osquery/runtime/runtime_test.go`) with a call to the new `testutil.DownloadOsqueryOrDie` helper, reducing boilerplate and improving maintainability. [[1]](diffhunk://#diff-a5aa8a2b5a6fe274cd605c8d132a7e673a5645e5728d1b23b2ea0c226f4c49beL4) [[2]](diffhunk://#diff-a5aa8a2b5a6fe274cd605c8d132a7e673a5645e5728d1b23b2ea0c226f4c49beL17-R22) [[3]](diffhunk://#diff-a5aa8a2b5a6fe274cd605c8d132a7e673a5645e5728d1b23b2ea0c226f4c49beL35-L78) [[4]](diffhunk://#diff-cf3aaa3759feb59da75e4704db300b4f59e6713046e420b321cdf08db9f5b0a0L4-R9) [[5]](diffhunk://#diff-cf3aaa3759feb59da75e4704db300b4f59e6713046e420b321cdf08db9f5b0a0L24-R24) [[6]](diffhunk://#diff-cf3aaa3759feb59da75e4704db300b4f59e6713046e420b321cdf08db9f5b0a0L80-R37) [[7]](diffhunk://#diff-1fc3a92c51148e943d7c505297f11bbd27349b5f8761cbc9b089494e144ee268L7) [[8]](diffhunk://#diff-1fc3a92c51148e943d7c505297f11bbd27349b5f8761cbc9b089494e144ee268L18-R24) [[9]](diffhunk://#diff-1fc3a92c51148e943d7c505297f11bbd27349b5f8761cbc9b089494e144ee268L38-R41) [[10]](diffhunk://#diff-f64bdefcc277fb19989354ea8d47bb1338a680288ca573e3c73ab55acce723e1L6-L45) [[11]](diffhunk://#diff-f64bdefcc277fb19989354ea8d47bb1338a680288ca573e3c73ab55acce723e1L157-L183) [[12]](diffhunk://#diff-6ba56a9b0c77cc308419a9ccf0cbccdaf8fd74aef730f118f37af5dbf2f88c1aL21-L23) [[13]](diffhunk://#diff-6ba56a9b0c77cc308419a9ccf0cbccdaf8fd74aef730f118f37af5dbf2f88c1aL34-R32) [[14]](diffhunk://#diff-6ba56a9b0c77cc308419a9ccf0cbccdaf8fd74aef730f118f37af5dbf2f88c1aL56-L115)

**New allowed command wrappers for macOS:**

* Added `Codesign` and `Xattr` command wrappers to `ee/allowedcmd/cmd_darwin.go`, allowing validated execution of `/usr/bin/codesign` and `/usr/bin/xattr` from Go code. [[1]](diffhunk://#diff-b97509fc6b71cb4e194dbaaf75f63f04bca4790c2ab24d9daaf6058dd5c694b0R23-R26) [[2]](diffhunk://#diff-b97509fc6b71cb4e194dbaaf75f63f04bca4790c2ab24d9daaf6058dd5c694b0R154-R157)

These changes improve test reliability, reduce duplication, and add support for additional system command wrappers on macOS.

### DIFF
--- a/ee/agent/reset_test.go
+++ b/ee/agent/reset_test.go
@@ -7,9 +7,9 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"os"
 	"path/filepath"
 	"runtime"
+	"sync"
 	"testing"
 	"time"
 
@@ -27,24 +27,16 @@ import (
 
 var testOsqueryBinary string
 
-// TestMain allows us to download osquery once, for use in all tests, instead of
-// downloading once per test case.
-func TestMain(m *testing.M) {
-	var err error
-	testOsqueryBinary, _, err = testutil.DownloadOsquery("nightly")
-	if err != nil {
-		fmt.Printf("failed to download osquery binary for tests: %v\n", err)
-		os.Exit(1) //nolint:forbidigo // Fine to use os.Exit inside tests
-	}
-
-	// Run the tests
-	retCode := m.Run()
-
-	os.Exit(retCode) //nolint:forbidigo // Fine to use os.Exit inside tests
-}
+// downloadOnceFunc downloads a real osquery binary for use in tests. This function
+// can be called multiple times but will only execute once -- the osquery binary is
+// stored at path `testOsqueryBinary` and can be reused by all subsequent tests.
+var downloadOnceFunc = sync.OnceFunc(func() {
+	testOsqueryBinary, _, _ = testutil.DownloadOsquery("nightly")
+})
 
 func TestDetectAndRemediateHardwareChange(t *testing.T) {
 	t.Parallel()
+	downloadOnceFunc()
 
 	testCases := []struct {
 		name                         string
@@ -503,6 +495,7 @@ func TestDetectAndRemediateHardwareChange(t *testing.T) {
 
 func TestDetectAndRemediateHardwareChange_SavesDataOverMultipleResets(t *testing.T) {
 	t.Parallel()
+	downloadOnceFunc()
 
 	slogger := multislogger.NewNopLogger()
 
@@ -599,6 +592,7 @@ func TestDetectAndRemediateHardwareChange_SavesDataOverMultipleResets(t *testing
 
 func TestExecute(t *testing.T) {
 	t.Parallel()
+	downloadOnceFunc()
 
 	var logBytes threadsafebuffer.ThreadSafeBuffer
 	slogger := slog.New(slog.NewTextHandler(&logBytes, &slog.HandlerOptions{
@@ -654,6 +648,7 @@ func TestExecute(t *testing.T) {
 
 func TestInterrupt_Multiple(t *testing.T) {
 	t.Parallel()
+	downloadOnceFunc()
 
 	var logBytes threadsafebuffer.ThreadSafeBuffer
 	slogger := slog.New(slog.NewTextHandler(&logBytes, &slog.HandlerOptions{
--- a/pkg/osquery/runsimple/osqueryrunner_test.go
+++ b/pkg/osquery/runsimple/osqueryrunner_test.go
@@ -7,7 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"os"
+	"sync"
 	"testing"
 
 	"github.com/kolide/launcher/pkg/osquery/testutil"
@@ -16,22 +16,16 @@ import (
 
 var testOsqueryBinary string
 
-// TestMain overrides the default test main function. This allows us to share setup/teardown.
-func TestMain(m *testing.M) {
-	var err error
-	testOsqueryBinary, _, err = testutil.DownloadOsquery("stable")
-	if err != nil {
-		fmt.Printf("failed to download osquery binary for tests: %v\n", err)
-		os.Exit(1) //nolint:forbidigo // Fine to use os.Exit in tests
-	}
-
-	// Run the tests!
-	retCode := m.Run()
-
-	os.Exit(retCode) //nolint:forbidigo // Fine to use os.Exit in tests
-}
+// downloadOnceFunc downloads a real osquery binary for use in tests. This function
+// can be called multiple times but will only execute once -- the osquery binary is
+// stored at path `testOsqueryBinary` and can be reused by all subsequent tests.
+var downloadOnceFunc = sync.OnceFunc(func() {
+	testOsqueryBinary, _, _ = testutil.DownloadOsquery("stable")
+})
 
 func Test_OsqueryRunSqlNoIO(t *testing.T) {
+	downloadOnceFunc()
+
 	osq, err := NewOsqueryProcess(testOsqueryBinary)
 	require.NoError(t, err)
 
@@ -39,6 +33,8 @@ func Test_OsqueryRunSqlNoIO(t *testing.T) {
 }
 
 func Test_OsqueryRunSql(t *testing.T) {
+	downloadOnceFunc()
+
 	tests := []struct {
 		name      string
 		sql       string
--- a/pkg/osquery/runtime/runtime_test.go
+++ b/pkg/osquery/runtime/runtime_test.go
@@ -14,6 +14,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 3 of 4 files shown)


pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,117 @@
...
+//   - version: The osquery version to download (e.g., "nightly", "stable", "5.17.0")
+//
+// Returns:
+//   - The path to the osquery binary
+//   - A cleanup function that can be called to remove the binary (optional)
+func DownloadOsqueryOrDie(version string) (binaryPath string, cleanup func() error) {
+	binaryPath, cleanup, err := DownloadOsquery(version)
+	if err != nil {
+		fmt.Printf("failed to download osqueryd binary for tests: %v\n", err)
+		os.Exit(1) //nolint:forbidigo // Fine to use os.Exit inside tests
```
directionless: I think os.Exit might skip cleanup things. I'd recommend taking a `testing.TB` and using `tb.Fail(...)`. -- Same thing we'd do inside a test.

Or you could define a simpler interface 🤷

pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"time"
+
+	"github.com/kolide/kit/fsutil"
+	"github.com/kolide/launcher/pkg/packaging"
```
directionless, line 12: Love that reuse

pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,117 @@
...
+
+		// Copy to our standardized cache location
+		if err := fsutil.CopyFile(dlPath, binaryPath); err != nil {
+			return "", nil, fmt.Errorf("copying osqueryd binary from %s to %s: %w", dlPath, binaryPath, err)
+		}
+	}
+
+	// Always ensure the binary is executable and has no quarantine attributes,
+	// even when reusing a cached binary (these attributes might get re-applied)
+	if runtime.GOOS != "windows" {
```
directionless: If you're using platform specific `prepareBinaryForExecution` (like this PR does), you should not have GOOS if blocks here. Just stick this into `prepareBinaryForExecution`

pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,117 @@
...
+		if err := os.Chmod(binaryPath, 0755); err != nil {
+			return "", nil, fmt.Errorf("setting executable permissions: %w", err)
+		}
+	}
+
+	// Remove ALL extended attributes on macOS that might prevent execution
+	// and ad-hoc sign the binary to bypass Gatekeeper restrictions
+	if runtime.GOOS == "darwin" {
```
directionless: Same. if `prepareBinaryForExecution` is platform specific. you don't need the if.

pkg/osquery/testutil/testutil_darwin.go
```diff
@@ -0,0 +1,72 @@
...
+	if _, err := os.Stat(binaryPath); err != nil {
+		return fmt.Errorf("binary does not exist at %s: %w", binaryPath, err)
+	}
+
+	// Clear extended attributes (best effort)
+	_ = removeAllExtendedAttributes(binaryPath)
+
+	// Ad-hoc sign the binary to bypass Gatekeeper restrictions
+	// This is required because com.apple.provenance cannot be removed with xattr
+	signCmd, err := allowedcmd.Codesign(context.Background(), "--force", "--sign", "-", binaryPath)
```
directionless: I thought we could download the signed osquery application bundle?

pkg/osquery/testutil/testutil.go
directionless: You could also pass said interface to `DownloadOsquery` and drop `DownloadOsqueryOrDie`

pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,117 @@
...
+		return "", nil, fmt.Errorf("parsing platform %s: %w", runtime.GOOS, err)
+	}
+	target.Arch = packaging.ArchFlavor(runtime.GOARCH)
+	if runtime.GOOS == "darwin" {
+		target.Arch = packaging.Universal
+	}
+
+	// Create a predictable directory in temp for storing test osquery binaries
+	// This allows reuse across test runs
+	cacheDir := filepath.Join(os.TempDir(), "launcher-test-osquery")
```
directionless, line 41: Hrm. I see why you're doing this, but I don't love it. OTOH, I think it's either that or a package level var. Not wholly sure which makes more sense,.

pkg/osquery/testutil/testutil_darwin.go
cesarfda: I think you are right, I had some issues before I added this but it was for a different approach

pkg/osquery/testutil/testutil.go
cesarfda: I simplified the interface and let the callers handle the error as needed since sometimes it's called inside a test main and sometimes it's not

pkg/osquery/testutil/testutil.go
```diff
@@ -0,0 +1,91 @@
...
+
+		// Copy to our standardized cache location
+		if err := fsutil.CopyFile(dlPath, binaryPath); err != nil {
+			return "", nil, fmt.Errorf("copying osqueryd binary from %s to %s: %w", dlPath, binaryPath, err)
+		}
+	}
+
+	// Always ensure the binary is executable and has no quarantine attributes,
+	// even when reusing a cached binary (these attributes might get re-applied)
+	if runtime.GOOS != "windows" {
```
directionless: You shouldn't need the `runtime.GOOS != "windows"`. Windows supports `os.Chmod` (it's mostly a no-op)

pkg/osquery/interactive/interactive_test.go
```diff
@@ -15,71 +14,36 @@ import (
...
+// so the symlink will point to an executable binary.
 func copyBinary(t *testing.T, executablePath string) {
-	downloadOnceFunc()
+	// Download osquery binary once, but fail the test if download fails
+	downloadOsqueryOnce.Do(func() {
+		testOsqueryBinary, _, downloadOsqueryErr = testutil.DownloadOsquery("nightly")
+	})
+	if downloadOsqueryErr != nil {
+		t.Fatalf("failed to download osquery binary: %v", downloadOsqueryErr)
+	}
```
directionless: Why have a sync and package vars?

pkg/osquery/testutil/testutil.go
cesarfda: Removed it

pkg/osquery/interactive/interactive_test.go
cesarfda: Fixed

ee/agent/reset_test.go
```diff
@@ -32,51 +30,17 @@ var testOsqueryBinary string
 // TestMain allows us to download osquery once, for use in all tests, instead of
 // downloading once per test case.
 func TestMain(m *testing.M) {
```
RebeccaMahany: I think we can now get rid of this `TestMain` wrapper and replace it with -- 

```
var downloadOnceFunc = sync.OnceFunc(func() {
	testOsqueryBinary, _, _ = testutil.DownloadOsquery("nightly")
})
```

-- like how you have it in e.g. `tuf/ci/valid_executable.go`.

pkg/osquery/runsimple/osqueryrunner_test.go
```diff
@@ -3,47 +3,32 @@ package runsimple
...
-	"github.com/kolide/kit/fsutil"
-	"github.com/kolide/launcher/pkg/packaging"
+	"github.com/kolide/launcher/pkg/osquery/testutil"
 	"github.com/stretchr/testify/require"
 )
 
 var testOsqueryBinary string
 
 // TestMain overrides the default test main function. This allows us to share setup/teardown.
 func TestMain(m *testing.M) {
```
RebeccaMahany: Same comment here, I think we can get rid of this in favor of the sync.OnceFunc approach 🙂

pkg/osquery/runtime/runtime_test.go
RebeccaMahany: Same suggestion re removing this in favor of sync.OnceFunc

PR #2438, opened 2025-10-09
add idb_cmp1 comparer for indexeddb iteration

zackattack01:
Updates our indexeddb iteration logic to use a full implementation of the `idb_cmp1` comparator, relying on the default bitwiseComparator can result in iteration issues while reading from indexeddb-backed leveldb stores.

There is an existing cli utility (referenced in code comments) which has a nice golang implementation, but after writing tests and attempting to debug I found that it was extremely difficult to rely on the recovered panics. So we've pulled in the logic and reworked things to use our slogger when issues are encountered, as well as adding some tests and general cleanup/docs

### DIFF
--- a/ee/agent/reset_test.go
+++ b/ee/agent/reset_test.go
@@ -1,27 +1,25 @@
 package agent
 
 import (
-	"context"
 	"crypto/ecdsa"
 	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
-	"os"
 	"path/filepath"
 	"runtime"
+	"sync"
 	"testing"
 	"time"
 
-	"github.com/kolide/kit/fsutil"
 	"github.com/kolide/krypto/pkg/echelper"
 	"github.com/kolide/launcher/ee/agent/storage"
 	storageci "github.com/kolide/launcher/ee/agent/storage/ci"
 	"github.com/kolide/launcher/ee/agent/types"
 	typesmocks "github.com/kolide/launcher/ee/agent/types/mocks"
 	"github.com/kolide/launcher/pkg/log/multislogger"
-	"github.com/kolide/launcher/pkg/packaging"
+	"github.com/kolide/launcher/pkg/osquery/testutil"
 	"github.com/kolide/launcher/pkg/threadsafebuffer"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,686 @@
+// Package indexdbcomparator provides an implementation of the chrome indexeddb comparator, idb_cmp1. This contains
```
directionless, line 4: Should probably start with
```
// Copyright (c) 2021-2025 cions
// Licensed under the MIT License.
//
```

Or if it's diverged, instrea of just "logic is largely taken from" use something like "Portions of this file are copyright and distributed under the MIT license"

zackattack01: makes sense, thank you! i'll add in the copyright at the top

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+	if len(a) == 0 {
+		return nil, nil, errors.New("invalid empty key provided to decodeKeyPrefix")
+	}
+
+	firstByte := a[0]
+	a = a[1:]
+
+	databaseIdBytes := int((((firstByte >> 5) & 0x07) + 1))
+	objectStoreIdBytes := int(((firstByte >> 2) & 0x07) + 1)
+	indexIdBytes := int((firstByte & 0x03) + 1)
```
RebeccaMahany, line 682: This could probably use some documentation, if at all possible

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+
+	if len(a) < databaseIdBytes+objectStoreIdBytes+indexIdBytes {
+		return nil, nil, errors.New("invalid key provided to decodeKeyPrefix (insufficient length for prefix bytes)")
```
RebeccaMahany: Would be nice to include lengths in case we ever find an error in the calculations -- 

```golang
fmt.Errorf("invalid key provided to decodeKeyPrefix (insufficient length for prefix bytes): key length %d, expected at least %d + %d + %d", len(a), databaseIdBytes, objectStoreIdBytes, indexIdBytes)
```

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+
+	databaseId, err := decodeInt(a[:databaseIdBytes])
+	if err != nil {
+		return nil, nil, err
```
RebeccaMahany: Could we add additional context to these errors? Same goes for below and anywhere else the original code is just returning an error without wrapping it first.

```golang
fmt.Errorf("decoding database ID as int: %w", err)
```

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+}
+
+func (c *idbCmp1Comparer) Compare(a, b []byte) int {
+	slogger := c.slogger.With(
+		"key_a", fmt.Sprintf("%x", a),
+		"key_b", fmt.Sprintf("%x", b),
+	)
+
+	a, prefixA, err := decodeKeyPrefix(a)
```
RebeccaMahany: Maybe worth doing `aWithoutPrefix, prefixA, err := decodeKeyPrefix(a)` to make it clearer that `a` is being transformed, and how it's being transformed? (And/or maybe switching the order of the returned args would help too, so the prefix is first -- `prefixA, a, err := decodeKeyPrefix(a)`.)

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+	b, prefixB, err := decodeKeyPrefix(b)
+	if err != nil {
+		slogger.Log(context.TODO(), slog.LevelError,
+			"error decoding key prefix b",
+			"err", err,
+		)
+		return 0
+	}
+
+	if ret := compareKeyPrefix(prefixA, prefixB); ret != 0 {
```
RebeccaMahany, line 156: So this covers the case where the key prefixes are different, and the rest of this function covers the case where the key prefixes are the same so we have to compare the remainder of the keys? Maybe worth a quick comment here. For the future, for readability, might be worth moving the big switch statement below to its own function/splitting it up further -- it becomes hard to keep track of where we are, especially with the nested switches.

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+	c.slogger.Log(context.TODO(), slog.LevelError,
+		"invalid key provided for decodeVarInt",
+		"invalid_key", fmt.Sprintf("%x", a),
+	)
+
+	return nil, 0, errors.New("invalid key provided for decodeVarInt")
+}
+
+func (c *idbCmp1Comparer) compareBinary(a, b []byte) ([]byte, []byte, int, error) {
+	a, len1, err := c.decodeVarInt(a)
```
RebeccaMahany: Nit -- `len1` => `lenA` and `len2` => `lenB`

RebeccaMahany: There are a couple other places where we're doing this, would be nice to standardize those too

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+			if err != nil {
+				return 0 // error already logged by compareStringWithLength via decodeVarInt
+			}
+			if ret != 0 {
+				return ret
+			}
+
+			if len(a) == 0 || len(b) == 0 {
+				return cmp.Compare(len(a), len(b))
+			}
```
RebeccaMahany: Can we comment why we need this again? At first glance, it looks redundant with 191-193, but I am guessing we have to redo this check because `c.compareStringWithLength` has modified `a` and `b`?

pkg/indexeddbcomparator/comparer.go
```diff
@@ -0,0 +1,689 @@
...
+		if ret := cmp.Compare(a[0], b[0]); ret != 0 {
+			return ret
+		}
+
+		typeByte := a[0]
+		a, b = a[1:], b[1:]
+
+		if typeByte < maxSimpleDatabaseMetaDataTypeByte {
+			return 0
+		}
```
RebeccaMahany, line 243: Do we know why we do this?

pkg/indexeddbcomparator/comparer.go
RebeccaMahany: Same q for other sections of the switch below, e.g. line 235-237 + 250-252

pkg/indexeddbcomparator/comparer.go
zackattack01: I could not find anything that explicitly explained this but it does seem correct per the chromium implementation: https://chromium.googlesource.com/chromium/src/+/main/content/browser/indexed_db/indexed_db_leveldb_coding.cc#1365. best guess is that by that point we already know that key prefixes are equal, and that that type byte is equal. so maybe simple global metadata types are just considered equal for comparison purposes and there'd be no reason to go further

pkg/indexeddbcomparator/comparer.go
zackattack01: good call, on it!

pkg/indexeddbcomparator/comparer.go
zackattack01: that is correct! I'm adding docs to the helper funcs for these to make this easier to figure out and will document a couple of these sections

pkg/indexeddbcomparator/comparer.go
zackattack01: added some comments! good call

PR #2445, opened 2025-10-15
Enhance gomaxprocsObserver to capture initial GOMAXPROCS value

cesarfda:
This pull request improves the way the application manages and limits the Go runtime's `GOMAXPROCS` value by capturing the initial value at startup and using it as an upper bound for later adjustments. The changes ensure that the number of OS threads can be safely reset to the original value and prevent exceeding the initial configuration.

Enhancements to GOMAXPROCS management:

* Added a mechanism to capture the initial `GOMAXPROCS` value at process startup using a `sync.Once` guard, ensuring it is only set once and is available for future resets. (`cmd/launcher/gomaxprocs_observer.go`)
* Updated the `gomaxprocsLimiter` function to:
  * Reset `GOMAXPROCS` to the initial value if the requested limit is zero or exceeds the initial value.
  * Only change `GOMAXPROCS` if the new value differs from the current one.
  * Add detailed logging for capturing, resetting, and changing the `GOMAXPROCS` value. (`cmd/launcher/gomaxprocs_observer.go`)

### DIFF
--- a/cmd/launcher/gomaxprocs_observer.go
+++ b/cmd/launcher/gomaxprocs_observer.go
@@ -60,26 +60,21 @@ func gomaxprocsLimiter(ctx context.Context, slogger *slog.Logger, maxProcs int) 
 
 	current := runtime.GOMAXPROCS(0)
 
-	// If maxProcs is 0 or exceeds the initial value, reset to initial
+	// Determine target: reset to initial if invalid or exceeds initial, otherwise use maxProcs
+	target := maxProcs
 	if maxProcs <= 0 || maxProcs > initialGOMAXPROCS {
-		if current != initialGOMAXPROCS {
-			slogger.Log(ctx, slog.LevelInfo,
-				"resetting GOMAXPROCS to initial value",
-				"from", current,
-				"to", initialGOMAXPROCS,
-			)
-			runtime.GOMAXPROCS(initialGOMAXPROCS)
-		}
+		target = initialGOMAXPROCS
+	}
+
+	// Early return if already at target
+	if current == target {
 		return
 	}
 
-	// Apply the limit if it differs from current
-	if current != maxProcs {
-		slogger.Log(ctx, slog.LevelInfo,
-			"changing GOMAXPROCS",
-			"from", current,
-			"to", maxProcs,
-		)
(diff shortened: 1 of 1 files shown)


cmd/launcher/gomaxprocs_observer.go
```diff
@@ -41,20 +49,37 @@ func (g *gomaxprocsObserver) FlagsChanged(ctx context.Context, flagKeys ...keys.
...
+		slogger.Log(ctx, slog.LevelDebug,
+			"captured initial GOMAXPROCS",
+			"value", initialGOMAXPROCS,
 		)
+	})
+
+	current := runtime.GOMAXPROCS(0)
+
+	// If maxProcs is 0 or exceeds the initial value, reset to initial
+	if maxProcs <= 0 || maxProcs > initialGOMAXPROCS {
```
RebeccaMahany: If we want to use a value of 0 to signal resetting to the initial gomaxprocs value, I think we need to also update flag controller's DesktopGoMaxProcs/LauncherGoMaxProcs to allow the min value to be 0 instead of 1?

cmd/launcher/gomaxprocs_observer.go
```diff
@@ -41,20 +49,37 @@ func (g *gomaxprocsObserver) FlagsChanged(ctx context.Context, flagKeys ...keys.
...
 	}
 
-	slogger.Log(ctx, slog.LevelInfo,
-		"limiting GOMAXPROCS",
-		"from", cur,
-		"to", maxProcs,
-	)
-	runtime.GOMAXPROCS(maxProcs)
+	// Apply the limit if it differs from current
+	if current != maxProcs {
```
directionless: The prior logic runs even if there's no change. Don't know if it makes sense reordering. This is rare enough I think it's fine as is

cesarfda: Changed to an early return when it's equal

cmd/launcher/gomaxprocs_observer.go
cesarfda: I updated the flag controller to allow the min to be 0 so we have a way of resetting the maxprocs to default

PR #2453, opened 2025-10-17
Add osquery variables data store and table

Micah-Kolide:
This PR adds a new data store for osquery variables to be sent down by the server to place into a table for easier check queries and will allow our customers to query directly from as well.

I started with a simple implementation of the table just using our `dataflattentable` as the data coming in is in JSON format.

### DIFF
--- a/pkg/osquery/runtime/osqueryinstance_test.go
+++ b/pkg/osquery/runtime/osqueryinstance_test.go
@@ -484,6 +484,7 @@ func TestReloadKatcExtension(t *testing.T) {
 	s := settingsstoremock.NewSettingsStoreWriter(t)
 	s.On("WriteSettings").Return(nil)
 	osqHistory := setupHistory(t, k)
+	k.On("ServerReleaseTrackerDataStore").Return(inmemory.NewStore()).Maybe()
 
 	// Create an instance and launch it
 	i := newInstance(types.DefaultRegistrationID, k, mockServiceClient(t), s)


cmd/launcher/launcher.go
```diff
@@ -71,13 +71,14 @@ import (
...
-	ztaInfoSubsystemName     = "zta_info"    // legacy name for dt4aInfo subsystem
-	dt4aInfoSubsystemName    = "dt4a_info"
+	agentFlagsSubsystemName       = "agent_flags"
+	serverDataSubsystemName       = "kolide_server_data"
+	desktopMenuSubsystemName      = "kolide_desktop_menu"
+	authTokensSubsystemName       = "auth_tokens"
+	katcSubsystemName             = "katc_config" // Kolide ATC
+	ztaInfoSubsystemName          = "zta_info"    // legacy name for dt4aInfo subsystem
+	dt4aInfoSubsystemName         = "dt4a_info"
+	osqueryVariablesSubsystemName = "osquery_variables"
```
RebeccaMahany: I think I might need a little more context for this one -- what kind of osquery variables are we going to start sending down via the control server?

RebeccaMahany: I did slightly more research and now I understand! This approach seems fine to me. I am good with the dataflattentable. Think you just have to update `setUpMockStores` to include this new store and then the tests will pass too 🙂

Micah-Kolide: Yeah sorry about not including the results from the table. I had ran them in LQ, so I didn't think to grab them, but as I'm sure you found it is just the variables we use in checks to look for browser up to date versions etc.

I've added the store to the `setUpMockStores`. Thank you for pointing me to that one!

directionless: I don't know what we want to call these. My hunch is that it's really much closer to "Server Data" than the thing we've already named that. (That other thing is really closer to "Device Data"

"Osquery Variables" might be okay, it's a little jargon." Shared Server Data" might be okay (this is shared across pretty much everything). dunno really. What would be clear to someone in a year?

pkg/osquery/table/table.go
```diff
@@ -39,6 +40,7 @@ func LauncherTables(k types.Knapsack, slogger *slog.Logger) []osquery.OsqueryPlu
 		launcher_db.TablePlugin(k, slogger, "kolide_control_flags", k.AgentFlagsStore()),
 		LauncherAutoupdateConfigTable(slogger, k),
 		osquery_instance_history.TablePlugin(k, slogger),
+		osquery_variables.TablePlugin(k, slogger, k.OsqueryVariablesStore()),
```
directionless: Why pass knapsack, and a store?

ee/tables/osquery_variables/table.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/kolide/launcher/ee/tables/tablewrapper"
+	"github.com/osquery/osquery-go/plugin/table"
+)
+
+func TablePlugin(flags types.Flags, slogger *slog.Logger, getter types.Getter) *table.Plugin {
+	columns := dataflattentable.Columns()
+	return tablewrapper.New(flags, slogger, "kolide_osquery_variables", columns, generate(slogger, getter))
+}
+
+func generate(slogger *slog.Logger, getter types.Getter) table.GenerateFunc {
```
directionless: This is an unusual `generate` function. I think all our other `generate` functions are the table generate, and when there's some state, we usually track that in a struct, and not via an anonymous generator. 

I think the anonymous pattern isn't inherently bad, it's just not customary here. If you do go that way, I would not call this `generate`. It's more like `generate_generator` which is horrific.

pkg/osquery/table/table.go
directionless: Oh, I see. It's because one is used a types.Flag, and the other as a Getter. Hrm.

ee/tables/osquery_variables/table.go
```diff
@@ -0,0 +1,43 @@
...
+	return func(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+		ctx, span := observability.StartSpan(ctx, "table_name", "kolide_osquery_variables")
+		defer span.End()
+
+		results := make([]map[string]string, 0)
+
+		variables, err := getter.Get([]byte("releases"))
```
directionless: Not convinced I would call this `releases`

ee/tables/osquery_variables/table.go
```diff
@@ -0,0 +1,43 @@
...
+		if err != nil {
+			slogger.Log(ctx, slog.LevelInfo, "failure getting data from store", "err", err)
+			return nil, err
+		}
+
+		flattened, err := dataflatten.Json(variables)
```
directionless: This works, and seems quite clever. 

I don't entirely understand if we're incurring extra marshalling overhead (vs storing it in a more structured fashion) or if it would matter.

ee/tables/osquery_variables/table.go
Micah-Kolide: I updated it to follow the pattern used by launcher_db, osquery_instances, and tufinfo, etc.

ee/tables/osquery_variables/table.go
Micah-Kolide: That's just to grab the data from K2 as it's currently `releases` in the osquery variables sent down.

cmd/launcher/launcher.go
Micah-Kolide: I changed the naming around a bit to make it less osquery-esque

RebeccaMahany (approved):

cmd/launcher/launcher.go
directionless: I suspect we may want to evolve this name. And how it intersects with the other server provided data field works. But, I think I'm being fussy and just getting this out is a win

Issue #1746, opened 2024-06-10
`launcher.exe version` should not return exit code 1 when bootstrapping an autoupdate


RebeccaMahany:
I noticed that the Windows end-to-end tests were failing:

```
Test Update: FAILED: launcher did not update within 6m0s: still on 1.6.7-7-g8e5c472: timeout after 6m0s (7 attempts): checking launcher version: running launcher --version: exit status 1; output `launcher - version 1.7.0-1-ge13781c
  branch: 	main
  revision: 	e13781cadcb35808ca4a3a6a612133df06970c67
  build date: 	2024-06-10
  build user: 	 (fv-az1111-111\runneradmin)
  go version: 	go1.21.11
```

Running locally, I see this additional log:

```
{"time":"2024-06-10T18:40:31.0761251Z","level":"ERROR","msg":"error execing newer version of launcher","launcher_run_id":"01J01Q7B7K8FNJN4M6479G9900","new_binary":"C:\\ProgramData\\Kolide\\Launcher-kolide-k2\\data\\updates\\launcher\\1.7.1\\launcher.exe","err":"exec completed with exit code 0"}
```

Whenever launcher has an autoupdate available that the execwrapper attempts to run for `launcher.exe version`, we will run into this bug, where we return an error when we should not.

Since this does not affect autoupdate, due to [this error code suppression function](https://github.com/kolide/launcher/blob/main/ee/tuf/util.go#L71), this is not critical to fix for this release -- but it should get fixed so the end-to-end automated tests will run again.

I think the source of this bug is likely the changes in https://github.com/kolide/launcher/pull/1693.

PR #2461, opened 2025-10-27
Zero exit codes for exits without error for commands expected to exit

James-Pickett:
Adds bool "commandExpectedToExit" to execwrapper.Exec and updates to not return an error if exits cleanly.
Updates main func to exit with 0 if exec did not return an error.
Updates tests.

closes https://github.com/kolide/launcher/issues/1746

### DIFF
--- a/ee/agent/knapsack/knapsack.go
+++ b/ee/agent/knapsack/knapsack.go
@@ -153,12 +153,20 @@ func (k *knapsack) SaveRegistration(registrationId, munemo, nodeKey, enrollmentS
 	}
 
 	// Now, store our data
-	if err := nodeKeyStore.Set(storage.KeyByIdentifier(nodeKeyKey, storage.IdentifierTypeRegistration, []byte(registrationId)), []byte(nodeKey)); err != nil {
+	nodeKeyKeyForRegistration := storage.KeyByIdentifier(nodeKeyKey, storage.IdentifierTypeRegistration, []byte(registrationId))
+	if err := nodeKeyStore.Set(nodeKeyKeyForRegistration, []byte(nodeKey)); err != nil {
 		return fmt.Errorf("setting node key in store: %w", err)
 	}
 	if err := registrationStore.Set([]byte(registrationId), rawRegistration); err != nil {
 		return fmt.Errorf("adding registration to store: %w", err)
 	}
+
+	k.Slogger().Log(context.Background(), slog.LevelInfo,
+		"successfully saved registration",
+		"node_key_key", string(nodeKeyKeyForRegistration),
+		"registration_id", registrationId,
+		"munemo", munemo,
+	)
 
 	return nil
 }
--- a/ee/localserver/krypto-ec-middleware.go
+++ b/ee/localserver/krypto-ec-middleware.go
@@ -196,6 +196,11 @@ func (e *kryptoEcMiddleware) callbackWorker() {
 				return nil
 			}
 
+			e.slogger.Log(req.Context(), slog.LevelInfo,
+				"launcher received new node key -- proceeding with secretless enrollment",
+				"munemo", r.Munemo,
+			)
+
 			if r.Munemo != "" {
 				e.tenantMunemo.Store(r.Munemo)
 			}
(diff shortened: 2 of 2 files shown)


cmd/launcher/main.go
```diff
@@ -187,7 +207,8 @@ func runMain() int {
 			level.Debug(logger).Log("msg", "could not get current executable to perform restart", "err", err.Error())
 			return 1
 		}
-		if err := execwrapper.Exec(ctx, systemSlogger.Logger, currentExecutable, os.Args, os.Environ()); err != nil {
+
+		if err := execwrapper.Exec(ctx, systemSlogger.Logger, currentExecutable, os.Args, os.Environ(), isNonSvcSubcommand()); err != nil {
```
RebeccaMahany: If we're here, we know we got a restart command, which means we know we're running launcher/launcher svc -- so we could do `execwrapper.Exec(ctx, systemSlogger.Logger, currentExecutable, os.Args, os.Environ(), false)` and skip doing the calculation. Not a big deal either way, though!

cmd/launcher/main.go
```diff
@@ -88,13 +88,33 @@ func runMain() int {
 	// fork-bombing itself. This is an ENV, because there's no
 	// good way to pass it through the flags.
 	if !env.Bool("LAUNCHER_SKIP_UPDATES", false) && !inBuildDir {
-		if err := runNewerLauncherIfAvailable(ctx, systemSlogger.Logger); err != nil {
-			systemSlogger.Log(ctx, slog.LevelInfo,
-				"could not run newer version of launcher",
+		lastestLauncherPath, err := latestLauncherPath(ctx, systemSlogger.Logger)
```
RebeccaMahany, line 93: Could we throw a quick comment in here saying that we do this rather than calling `runNewerLauncherIfAvailable` so that we can distinguish between "exited with error" and "exited without error"? I could see us forgetting why we made this change and trying to tidy up/refactor, and accidentally consolidating back into `runNewerLauncherIfAvailable` again.

pkg/execwrapper/exec_windows.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/kolide/launcher/ee/gowrapper"
 )
 
-func Exec(ctx context.Context, slogger *slog.Logger, argv0 string, argv []string, envv []string) error {
+func Exec(ctx context.Context, slogger *slog.Logger, argv0 string, argv []string, envv []string, isNonSvcSubCommand bool) error {
```
RebeccaMahany: Maybe `isNonSvcSubCommand` => `cmdExpectedToExit`, to make it clearer what its purpose is for this function?

pkg/execwrapper/exec_test.go
```diff
@@ -18,13 +18,55 @@ func TestExec(t *testing.T) {
...
+	if runtime.GOOS == "windows" {
+		command = "cmd.exe"
+		// Exec removes the first argument, so add a blank to make sure our args get processed
+		args = []string{"", "/c", "echo", "test string"}
+	}
+
+	// flag this call as a non-svc subcommand so if it exist with out error, we do not return an error
+	err := Exec(t.Context(), slogger, command, args, os.Environ(), true)
+
+	// on non-windows, we never actually get to this line, because syscall.Exec replaces the current process
```
zackattack01: still worth running this for non-windows?

zackattack01: seems like it must be passing I'm just confused what actually happens test wise

zackattack01 (dismissed):

cmd/launcher/main.go
James-Pickett: good call, updated

pkg/execwrapper/exec_windows.go
James-Pickett: renamed all references to commandExpectedToExit, easier to understand

cmd/launcher/main.go
James-Pickett: added comment

pkg/execwrapper/exec_test.go
James-Pickett: probably not worth it, skipping on non-windows

Issue #2463, opened 2025-10-28
Improve efficiency of `consoleusers.CurrentUids` and `consoleusers.ExplorerProcess`


RebeccaMahany:
Right now, on Windows, the new `consoleusers.CurrentUids` implementation added in https://github.com/kolide/launcher/pull/2422 is incorrectly identifying additional users that we should start desktop processes for. This means that the desktop runner is, every five seconds, trying to start a desktop process for these users, which means it calls `consoleusers.ExplorerProcess` and enumerates every single process. This is very noisy in the logs, and we suspect there's a performance impact. (This issue will, though, be invisible to the end user otherwise.)

For the next release:

* Figure out how to correctly filter out users that we shouldn't create login sessions for (or, at the very least, hardcode known exceptions into `consoleusers.CurrentUids`): https://github.com/kolide/launcher/pull/2465

Over the next two releases:

* Complete https://github.com/kolide/launcher/issues/2462, so that we can increase the interval on the cloud side
* Figure out an alternative to `consoleusers.ExplorerProcess` that doesn't require enumerating every single process
* Evaluate performance impact of current `consoleusers.CurrentUids` implementation on macOS and Linux

In the long term:

* Investigate subscribing to logon events so that we don't have to run this check quite so frequently (or maybe we can phase it out altogether)

RebeccaMahany:
We haven't had complaints about performance issues in this area in some time; the changes we made are likely sufficient.

PR #2468, opened 2025-10-31
Revert LSA changes, and consolidate

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/2463.

First, this PR reverts the changes in https://github.com/kolide/launcher/pull/2422. That PR was intended to improve the performance of `CurrentUids` by avoiding iterating over every process to identify explorer.exe processes. That implementation _was_ faster -- but it also incorrectly identified user accounts without explorer.exe processes frequently. Therefore, every 5s, we were calling `ExplorerProcess` anyway when attempting to launch desktop processes for these accounts via `runAsUser`, and doing the very iteration over every process that we were hoping to avoid.

So, I've reverted us back to the previous implementation, iterating over every process to identify explorer.exe processes. This will be a little slower, but will stop identifying extra accounts. I also updated the return signature for `CurrentUids` to return the explorer.exe PID as well as the UID. This allows us to skip our second `ExplorerProcess` call in `runAsUser` -- so we ultimately only iterate over all processes once, instead of twice.

### DIFF
--- a/ee/desktop/runner/runner_test.go
+++ b/ee/desktop/runner/runner_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/ee/agent/flags/keys"
 	"github.com/kolide/launcher/ee/agent/types/mocks"
-	"github.com/kolide/launcher/ee/consoleuser"
 	"github.com/kolide/launcher/ee/desktop/user/notify"
 	"github.com/kolide/launcher/ee/presencedetection"
 	"github.com/kolide/launcher/pkg/backoff"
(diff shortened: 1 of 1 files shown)


ee/consoleuser/consoleuser_windows.go
```diff
@@ -6,116 +6,48 @@ package consoleuser
 import (
 	"context"
 	"fmt"
-	"maps"
 	"path/filepath"
-	"slices"
-	"strings"
 
-	winlsa "github.com/kolide/go-winlsa"
```
zackattack01, line 14: do we need a `go mod tidy` or still using this?

RebeccaMahany: Updated, thank you!

ee/consoleuser/consoleuser_windows.go
```diff
@@ -6,116 +6,48 @@ package consoleuser
...
+			Uid:            uid,
+			UserProcessPid: explorerProc.Pid,
 		}
 	}
 
-	return nil, nil
+	// convert map keys to slice
+	uids := make([]*ConsoleUser, len(uidsMap))
+	uidCount := 0
+	for _, consoleUser := range uidsMap {
```
zackattack01, line 45: could use `slices.Collect(maps.Values(uidsMap))` here

PR #2470, opened 2025-11-03
Track known invalid usernames and ignore them

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/2463; implements suggestion from here: https://github.com/kolide/launcher/pull/2468#issuecomment-3474896699.

https://github.com/kolide/launcher/pull/2465 excluded some known usernames, but wasn't enough to cover all usernames that we aren't able to find explorer.exe processes for. This PR adds a lookup map to track usernames that we know we are unable to find explorer.exe processes for. Once we've failed to find an explorer.exe process for a username 3 times within 1 minute, we assume that this is a username we should not be starting a desktop process for.

### DIFF
--- a/ee/consoleuser/consoleuser_windows.go
+++ b/ee/consoleuser/consoleuser_windows.go
@@ -167,17 +167,11 @@ func updateInvalidUsernameMaps(username string, windowSeconds int64) {
 	// within the lookup window.
 	lookupWindowStart := lookupTimestamp - windowSeconds
 	failureCount := 0
-	lastTimestampOutsideWindowIdx := -1
-	for i, ts := range potentialInvalidUsernamesMap[username] {
+	for _, ts := range potentialInvalidUsernamesMap[username] {
 		if ts >= lookupWindowStart {
 			failureCount += 1
 			continue
 		}
-		lastTimestampOutsideWindowIdx = i
-	}
-	// Remove old timestamps outside of window, if any
-	if lastTimestampOutsideWindowIdx > -1 {
-		potentialInvalidUsernamesMap[username] = potentialInvalidUsernamesMap[username][lastTimestampOutsideWindowIdx+1:]
 	}
 
 	// Too many failures -- move to knownInvalidUsernamesMap.
@@ -187,6 +181,15 @@ func updateInvalidUsernameMaps(username string, windowSeconds int64) {
 		knownInvalidUsernamesMapLock.Unlock()
 
 		delete(potentialInvalidUsernamesMap, username)
+		return
+	}
+
+	// We haven't yet hit maxUsernameLookupFailureCount failures within the lookup window.
+	// Make sure we don't accumulate a bunch of entries outside of the window --
+	// we can safely keep just the last maxUsernameLookupFailureCount entries.
+	if len(potentialInvalidUsernamesMap[username]) > maxUsernameLookupFailureCount {
+		newStartIdx := len(potentialInvalidUsernamesMap[username]) - maxUsernameLookupFailureCount
+		potentialInvalidUsernamesMap[username] = potentialInvalidUsernamesMap[username][newStartIdx:]
 	}
 }
 
--- a/ee/consoleuser/consoleuser_windows_test.go
+++ b/ee/consoleuser/consoleuser_windows_test.go
@@ -175,6 +175,6 @@ func Test_updateInvalidUsernameMaps_RequiresFailuresWithinWindow(t *testing.T) {
 	// Confirm that we do still have some timestamps recorded
 	potentialInvalidUsernamesMapLock.Lock()
 	require.Contains(t, potentialInvalidUsernamesMap, invalidTestUsername)
-	require.Greater(t, len(potentialInvalidUsernamesMap[invalidTestUsername]), 0)
+	require.Equal(t, maxUsernameLookupFailureCount, len(potentialInvalidUsernamesMap[invalidTestUsername]))
 	potentialInvalidUsernamesMapLock.Unlock()
 }


ee/consoleuser/consoleuser_windows.go
```diff
@@ -115,9 +144,45 @@ func ExplorerProcess(ctx context.Context, uid string) (*process.Process, error)
...
+	lookupWindowStart := lookupTimestamp - windowSeconds
+	failureCount := 0
+	for _, ts := range potentialInvalidUsernamesMap[username] {
+		if ts >= lookupWindowStart {
+			failureCount += 1
+		}
+	}
+
+	// Too many failures -- move to knownInvalidUsernamesMap.
+	if failureCount >= maxUsernameLookupFailureCount {
```
zackattack01, line 178: anything additional we should log here to be able to track and eventually exclude these from `shouldIgnoreSession`? or is the `could not run console user desktop process` log enough for that

ee/consoleuser/consoleuser_windows.go
```diff
@@ -115,9 +144,45 @@ func ExplorerProcess(ctx context.Context, uid string) (*process.Process, error)
...
+	} else {
+		potentialInvalidUsernamesMap[username] = append(potentialInvalidUsernamesMap[username], lookupTimestamp)
+	}
+
+	// We know we've seen this username before. Check to see if we have maxUsernameLookupFailureCount failures
+	// within the lookup window.
+	lookupWindowStart := lookupTimestamp - windowSeconds
+	failureCount := 0
+	for _, ts := range potentialInvalidUsernamesMap[username] {
+		if ts >= lookupWindowStart {
```
zackattack01, line 171: do we think worth deleting the expired ones? in case there's some failure pattern where this could just keep growing but not hit the 1 minute threshold for deletion

ee/consoleuser/consoleuser_windows.go
RebeccaMahany: Yeah that log will be enough for that!

RebeccaMahany: (We also don't have logging access here atm)

ee/consoleuser/consoleuser_windows.go
zackattack01: or probably more efficient to only ever keep the latest 3 in the first place

RebeccaMahany: Ah, saw your first comment but not the second, and updated to deleting the expired ones. I will update to keep the last 3, that's a little simpler!

RebeccaMahany: I guess technically we could keep the last 2 and not 3 -- if it's 3, then we've hit the max failure count and are about to delete the list anyway. But it feels easier to keep the last 3 and not have to leave a whole comment about why it's actually maxUsernameLookupFailureCount-1. 😅

ee/consoleuser/consoleuser_windows.go
```diff
@@ -49,7 +68,8 @@ func CurrentUids(ctx context.Context) ([]string, error) {
 		}
 
 		// Check for a couple well-known accounts that we know we don't want to create desktop processes for.
-		if shouldIgnoreSession(sessionData) {
+		sessionUsername := usernameFromSessionData(sessionData)
+		if shouldIgnoreSession(sessionData, sessionUsername) {
```
directionless, line 72: We should make sure this gets logged.

ee/consoleuser/consoleuser_windows.go
```diff
@@ -10,13 +10,32 @@ import (
...
 	winlsa "github.com/kolide/go-winlsa"
 	"github.com/kolide/launcher/ee/observability"
 	"github.com/shirou/gopsutil/v4/process"
 	"golang.org/x/sys/windows"
 )
 
+var (
+	// knownInvalidUsernamesMap tracks usernames that we are repeatedly unable to find explorer.exe processes for,
+	// indicating that we should not include them in the return for `CurrentUids`.
+	knownInvalidUsernamesMap     = make(map[string]struct{})
```
directionless, line 25: Should we track a timestamp and expire this daily or anything?

directionless (commented):
nice

ee/consoleuser/consoleuser_windows.go
directionless: Or more. I think of this as less of a permanent block-list, and mostly a fix for "don't check every 5 seconds"

ee/consoleuser/consoleuser_windows.go
RebeccaMahany: We'll have the `no user explorer process found for <username>` logs from the runner!

ee/consoleuser/consoleuser_windows.go
zackattack01: ah yeah I see, this works for me!

zackattack01 (approved):

ee/consoleuser/consoleuser_windows.go
directionless: True... Though we need to infer that it's why we didn't start a desktop

directionless: And I guess I really _don't_ want a log every 5s

ee/consoleuser/consoleuser_windows.go
RebeccaMahany: I don't think that tracking the timestamp and expiring daily is worth it for this. If I've gotten our consts wrong here (3 failures in 1 minute) and launcher incorrectly flags a username as invalid when it shouldn't, and a user notices -- launcher self-remediating a day later isn't going to feel like a fast enough fix.

directionless: Okay, I guess we go with it

RebeccaMahany: We can always adjust depending on how internal rollout goes 💯

PR #2473, opened 2025-11-03
Ensure we start to use node key set by secretless enrollment

RebeccaMahany:

When secretless enrollment completes, the node key is set in the data store, but it is not yet set on the osquery extension struct.

When the osquery extension was calling device server endpoints, it was using `e.NodeKey`, which is the still-empty node key set on the struct. The extension therefore got back an invalid response from the device server, and that triggered our `RequireReenroll` flow -- which deletes the node key and registration from the database, thereby undoing the secretless enrollment.

This PR updates the extension so that when it calls device server endpoints, it uses `e.NodeKey` if set, and falls back to grabbing the node key from the database.

I verified that it is still possible to enroll with these changes, so they do not prevent the regular reenrollment flow from occurring.

### DIFF
--- a/pkg/osquery/extension_test.go
+++ b/pkg/osquery/extension_test.go
@@ -418,6 +418,69 @@ func TestGenerateConfigs_CannotEnrollYet(t *testing.T) {
 	assert.False(t, s.RequestEnrollmentFuncInvoked)
 }
 
+func TestGenerateConfigs_WorksAfterSecretlessEnrollment(t *testing.T) {
+	nodeKeyFromSecretlessEnrollment := "new_node_key_from_secretless_enrollment"
+	configVal := `{"foo":"bar","options":{"distributed_interval":5,"verbose":true}}`
+	s := &mock.KolideService{
+		RequestConfigFunc: func(ctx context.Context, nodeKey string) (string, bool, error) {
+			if nodeKey != nodeKeyFromSecretlessEnrollment {
+				// Returns node_invalid
+				return "", true, nil
+			}
+			return configVal, false, nil
+		},
+	}
+
+	k := mocks.NewKnapsack(t)
+	k.On("OsquerydPath").Maybe().Return("")
+	k.On("LatestOsquerydPath", testifymock.Anything).Maybe().Return("")
+	configStore, err := storageci.NewStore(t, multislogger.NewNopLogger(), storage.ConfigStore.String())
+	require.NoError(t, err, configStore)
+	k.On("ConfigStore").Return(configStore, nil)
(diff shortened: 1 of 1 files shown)


pkg/osquery/extension.go
```diff
@@ -563,6 +563,31 @@ func (e *Extension) enrolled() bool {
...
+	nodeKey := e.NodeKey
+	if nodeKey != "" {
+		return nodeKey
+	}
+
+	// Not yet set on the extension, but may be available from the database
+	nodeKey, _ = NodeKey(e.knapsack.ConfigStore(), e.registrationId)
+	if nodeKey != "" {
+		// Set it on the extension so we don't have to fetch it from the database in the future
+		e.enrollMutex.Lock()
```
zackattack01: wouldn't this already be locked?

RebeccaMahany: Whoops yes, missed in refactor

RebeccaMahany: Updated, and added a test to cover this case too

zackattack01 (approved):

pkg/osquery/extension.go
```diff
@@ -563,6 +563,29 @@ func (e *Extension) enrolled() bool {
 	return nodeKey != ""
 }
 
+// nodeKey returns the current node key for this registration. It ensures that if the
+// node key is set in the database but not in the extension, we will pull it. We should therefore
+// always use this function when we need the node key to make requests to the device server,
+// to avoid making requests with an empty node key and attempting to reenroll instead.
+func (e *Extension) nodeKey() string {
```
directionless, line 570: Do we need it in both places? Can we move it out of the extension?

RebeccaMahany: If we pulled it out of the extension struct, then we'd have to be making a db call to fetch the node key for every single call to the device server -- I assumed we wouldn't want this, but maybe it's no big deal

RebeccaMahany: I'm going to merge this to unblock the folks waiting on this fix, but let me know what you think about this -- I can make a follow-up PR to pull the node key out of the struct!

directionless: You're seeing the extension struct as a local cache to minimize the unknown cost on the DB? Okay, I think that could make sense. 

But if we do that, I think we should rename/recomment things to make it clear thats the intent. I don't know if it shifts other parts of the enroll flow as well

RebeccaMahany: I don't believe it should affect other parts of the enroll flow, but maybe I'm missing something. I'll put in a PR to adjust naming/comments

RebeccaMahany: https://github.com/kolide/launcher/pull/2474 lmk!

PR #2475, opened 2025-11-04
enable watchdog by default

zackattack01:
This removes the old launcher watchdog enabled flag in favor of our new opt-out based model. We expect that devices will install the watchdog task by default after the first time the controller's `FlagsChanged` is called.

- also adds some caching around install state to prevent excessive COM initializations and/or installations
- small refactors for readability and testability, pulling the actual task management functions into a `taskManager` interface for mocking so we can add unit tests around the caching and `FlagsChanged` behavior

#### Overview of expected `installState` caching flow:

This is intended to be lightweight and reduce our `CoInitializeEx` calls wherever possible (and prevent excessive installations), but ensure we will always get to the desired state (worst case scenario even with major bugs here the next launcher startup/config fetch should self-heal).
- whenever launcher boots it will reset to default value of `installStateUnset`
- whenever `FlagsChanged` is called:
  - if the current state is `installStateUnset`, check if the task exists and update the state to match reality
  - otherwise, compare that state to the desired state from `LauncherWatchdogDisabled`
- if we install or remove without error, update the cached `installState` accordingly
- if an error is ever encountered, reset the state to `installStateUnset` so we will re-evaluate things on the next `FlagsChanged`

### DIFF
--- a/ee/watchdog/controller_windows.go
+++ b/ee/watchdog/controller_windows.go
@@ -89,7 +89,7 @@ func NewController(ctx context.Context, k types.Knapsack, configFilePath string)
 		configFilePath: configFilePath,
 	}
 
-	tm.taskInstallState.Store(installStateUnset)
+	tm.setCurrentInstallState(installStateUnset)
 
 	wc := &WatchdogController{
 		slogger:      slogger,


ee/watchdog/controller_windows.go
```diff
@@ -52,21 +81,76 @@ func NewController(ctx context.Context, k types.Knapsack, configFilePath string)
...
+	tm := watchdogTaskManager{
+		slogger:        slogger,
 		knapsack:       k,
-		interrupt:      make(chan struct{}, 1),
-		logPublisher:   logPublisher,
 		configFilePath: configFilePath,
-	}, nil
+	}
+
+	tm.taskInstallState.Store(installStateUnset)
```
RebeccaMahany: Can we call the interface method `setCurrentInstallState` instead?

zackattack01: oh yep!

ee/watchdog/controller_windows.go
```diff
@@ -52,21 +81,76 @@ func NewController(ctx context.Context, k types.Knapsack, configFilePath string)
...
+		)
+		return
+	}
+
+	currentInstallState := wc.taskManager.currentInstallState()
+	nowDisabled := wc.knapsack.LauncherWatchdogDisabled()
+
+	if currentInstallState == installStateUnset {
+		taskExists, err := watchdogTaskExists(wc.knapsack.Identifier())
+		if !taskExists || err != nil { // if we err for any reason assume not installed
```
James-Pickett, line 129: Do we want to log the error here?

zackattack01: oh that's not a bad idea, was thinking it would usually error for not installed but forgot we squash that in `watchdogTaskExists`. i'll add in a new PR, thanks!

PR #2476, opened 2025-11-04
Do not permit reenrollment for secretless installations

RebeccaMahany:

Reenrollment via device server requires an enrollment secret, which secretless launcher installations will never have. Therefore, secretless launcher installations should never attempt to perform reenrollment via device server -- we will have to find another (likely localserver) mechanism for reenrollment in the case of an invalid node key.

This PR ensures that secretless installations won't perform reenrollment. It will also thereby keep the extension from wiping valid-but-not-yet-used node keys.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false # Consider changing this sometime
       matrix:
         os:
-          - macos-13
+          - macos-14
           - windows-latest
     steps:
     - name: Check out code
@@ -200,7 +200,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
-          - macos-13
+          - macos-14
           - windows-latest
     steps:
     - name: Check out code
@@ -263,7 +263,6 @@ jobs:
           # See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
           - ubuntu-22.04
           - ubuntu-24.04
-          - macos-13
           - macos-14
           - macos-15
           - windows-2022
@@ -419,7 +418,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
-          - macos-13
+          - macos-14
           - windows-latest
     steps:
     - uses: actions/checkout@v4


pkg/osquery/extension.go
```diff
@@ -41,6 +41,7 @@ type Extension struct {
 	NodeKey                       string
 	Opts                          ExtensionOpts
 	registrationId                string
+	isSecretless                  bool
```
directionless: Do you think we should track this as a boolean, or as the absence of a secret?

pkg/osquery/extension.go
```diff
@@ -589,6 +594,15 @@ func (e *Extension) nodeKey() string {
...
 // next call to Enroll will cause the enrollment process to take place.
 func (e *Extension) RequireReenroll(ctx context.Context) {
+	// We cannot perform reenrollment via this extension for secretless installations,
+	// because reenrollment requires an enroll secret. Instead, launcher will have to reenroll
+	// via localserver.
+	if e.isSecretless {
+		e.slogger.Log(ctx, slog.LevelWarn,
+			"RequireReenroll called for secretless installation, will not reenroll",
+		)
+		return
```
directionless, line 604: I think what we do here is an _interesting_ question. Maybe ignoring is good enough? If we knew enough, we might want to treat it as an unenroll event?

pkg/osquery/extension.go
RebeccaMahany: I'm not sure I understand the question. I wanted to store the status in `isSecretless` so we don't have to check the filesystem every time we need to know whether reenrollment is acceptable.

pkg/osquery/extension.go
RebeccaMahany: In this particular case, we don't want to treat it as an unenroll event.

For more context, here's the race condition we're currently running into:

1. Extension makes a request to the device server with an empty node key (it is always constantly forwarding requests from osquery, regardless of enrollment status)
2. Launcher completes secretless enrollment and stores node key
3. The device server request from 1 fails with invalid node, since the node key was empty, so the extension calls RequireReenroll before trying again -- which (before the changes in this PR) would delete the node key stored in step 2

This leaves launcher in a limbo where k2 thinks that launcher is enrolled, and launcher does not. So, for now, we must ignore RequireReenroll when we're a secretless installation.

Re: unenroll, maybe we'll want to have k2 send down an unenroll command (or reuse the uninstall command) for that instead?

RebeccaMahany: TL;DR I think handling unenroll + reenroll for real are definitely open questions and I am punting on them a little bit for now 😅

pkg/osquery/extension.go
RebeccaMahany: Oh, hm. I am actually wondering if we _do_ want to check `ReadEnrollmentSecret` every time we want to know about `isSecretless`, to allow for admins to distribute secrets via files (or registry keys, etc...) post-installation.

directionless: Yes, that. Diskio shouldn't be bad -- it's not like this is getting called all the time.

But I think there's a lot of unknowns about how secretless and multitenancy works.

pkg/osquery/extension.go
directionless: I support kicking the unenroll/reenroll can forward, yes. 

I wonder if we want the extension to _not_ send things if there's no node key. I suspect there are artfacts here of when we were mostly just osquery, and the usual enroll path was to have osquery do something, and get the re-enroll event.

pkg/osquery/extension.go
RebeccaMahany: A couple tests later and I think I'm overthinking it, and that enrollment would still proceed (we call `RequireReenroll`, then always call `Enroll` after anyway -- so even if we're secretless, the check in `RequireReenroll` will only prevent us from clearing empty data, and we'd still proceed to `Enroll`, pick up the new secret from disk, and proceed with enrollment). But it feels tidier to always directly check, so I've updated to do that. Added a test for that case, too.

pkg/osquery/extension.go
RebeccaMahany: I think the extension could not send things IFF there's no node key and no enrollment secret

Issue #1412, opened 2023-10-18
TUF autoupdater should subscribe to changes in AutoupdateInterval and AutoupdaterInitialDelay


RebeccaMahany:
`TufAutoupdater.Execute` would not currently respond to a change in autoupdate interval or autoupdater initial delay from the control server. We should update it so that it subscribes to updates for `AutoupdateInterval` and `AutoupdaterInitialDelay` (similar to how we do temporary interval overrides for localserver and the control server).

Relates to https://github.com/kolide/launcher/issues/954

PR #2477, opened 2025-11-04
Tuf autoupdate interval

cesarfda:
This pull request enhances the autoupdate system by introducing runtime overrides for the autoupdate interval and initial delay, and ensures that changes to these flags are immediately reflected in the updater's behavior. It also updates the relevant interfaces, mock implementations, and tests to support these new override methods and behaviors.

**Autoupdate flag override support:**

* Added `SetAutoupdateIntervalOverride` and `SetAutoupdateInitialDelayOverride` methods to the `Flags` interface and their implementations in `flag_controller.go`, allowing runtime overrides of autoupdate interval and initial delay. [[1]](diffhunk://#diff-d4072ec5e95c96a74b54248830c08732d0ff7bd69e5c516be1541cee4b439430R181) [[2]](diffhunk://#diff-d4072ec5e95c96a74b54248830c08732d0ff7bd69e5c516be1541cee4b439430R190) [[3]](diffhunk://#diff-73705c5a4d5b76f004593cbf1cb84a1fb6966ccef4f2a990e1fded7d453306f2R549-R560) [[4]](diffhunk://#diff-73705c5a4d5b76f004593cbf1cb84a1fb6966ccef4f2a990e1fded7d453306f2R580-R591)
* Updated mocks in `flags.go` and `knapsack.go` to support the new override methods for testing. [[1]](diffhunk://#diff-c0d4e1b9c036b43e815c7ab1c26f8bb4f14023b0606127751fed7f04a2940e34R1066-R1070) [[2]](diffhunk://#diff-c0d4e1b9c036b43e815c7ab1c26f8bb4f14023b0606127751fed7f04a2940e34R1089-R1093) [[3]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1R1654-R1658) [[4]](diffhunk://#diff-ba742f0ddc02b42b772f5c78fdd4ec47d1e623161dbfb07763f60d0804357fa1R1677-R1681)

**Autoupdater runtime behavior improvements:**

* Modified the `TufAutoupdater` to subscribe to changes in `AutoupdateInterval` and `AutoupdateInitialDelay` flags, and to immediately reset the update ticker and optionally perform an update check when these flags change. [[1]](diffhunk://#diff-34982e7fbd6048577137937a666019c6c6c4f5ff114530d09a3fd19ad4481065R96) [[2]](diffhunk://#diff-34982e7fbd6048577137937a666019c6c6c4f5ff114530d09a3fd19ad4481065L162-R163) [[3]](diffhunk://#diff-34982e7fbd6048577137937a666019c6c6c4f5ff114530d09a3fd19ad4481065L244-R246) [[4]](diffhunk://#diff-34982e7fbd6048577137937a666019c6c6c4f5ff114530d09a3fd19ad4481065L275-R276) [[5]](diffhunk://#diff-34982e7fbd6048577137937a666019c6c6c4f5ff114530d09a3fd19ad4481065R429-R460)

### DIFF
--- a/ee/tuf/autoupdate.go
+++ b/ee/tuf/autoupdate.go
@@ -91,6 +91,7 @@ type TufAutoupdater struct {
 	updateChannel        string
 	pinnedVersions       map[autoupdatableBinary]string        // maps the binaries to their pinned versions
 	pinnedVersionGetters map[autoupdatableBinary]func() string // maps the binaries to the knapsack function to retrieve updated pinned versions
+	initialDelayStart    time.Time                             // when the autoupdater was created
 	initialDelayEnd      atomic.Value                          // stores time.Time for thread-safe access
 	updateLock           *sync.Mutex
 	checkTicker          *time.Ticker
@@ -118,11 +119,13 @@ func NewTufAutoupdater(ctx context.Context, k types.Knapsack, metadataHttpClient
 	ctx, span := observability.StartSpan(ctx)
 	defer span.End()
 
+	startTime := time.Now()
 	ta := &TufAutoupdater{
-		knapsack:      k,
-		interrupt:     make(chan struct{}, 10), // We have a buffer so we don't block on sending to this channel
-		signalRestart: make(chan error, 10),    // We have a buffer so we don't block on sending to this channel
-		updateChannel: k.UpdateChannel(),
+		knapsack:          k,
+		interrupt:         make(chan struct{}, 10), // We have a buffer so we don't block on sending to this channel
+		signalRestart:     make(chan error, 10),    // We have a buffer so we don't block on sending to this channel
+		updateChannel:     k.UpdateChannel(),
+		initialDelayStart: startTime,
 		pinnedVersions: map[autoupdatableBinary]string{
 			binaryLauncher: k.PinnedLauncherVersion(), // empty string if not pinned
 			binaryOsqueryd: k.PinnedOsquerydVersion(), // ditto
@@ -138,8 +141,8 @@ func NewTufAutoupdater(ctx context.Context, k types.Knapsack, metadataHttpClient
 		calculatedSplayDelay: &atomic.Int64{},
 	}
 
-	// Set initial delay end time atomically
-	ta.initialDelayEnd.Store(time.Now().Add(k.AutoupdateInitialDelay()))
+	// Set initial delay end time atomically (calculated from start time)
+	ta.initialDelayEnd.Store(startTime.Add(k.AutoupdateInitialDelay()))
 
 	for _, opt := range opts {
 		opt(ta)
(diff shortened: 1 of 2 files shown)


ee/tuf/autoupdate.go
RebeccaMahany: I think we would need a change somewhere around here as well for the autoupdater to completely change its initial delay value? I see you updated `ta.initialDelayEnd`, but I think only `Do` and `FlagsChanged` use that value.

ee/tuf/autoupdate.go
```diff
@@ -425,6 +426,38 @@ func (ta *TufAutoupdater) FlagsChanged(ctx context.Context, flagKeys ...keys.Fla
 		"pinned_launcher_version", ta.knapsack.PinnedLauncherVersion(),
 		"pinned_osqueryd_version", ta.knapsack.PinnedOsquerydVersion(),
 	)
+
+	// Check for AutoupdateInterval changes
```
RebeccaMahany: I would move all of this to the top of this function, I think -- otherwise, the early return if `len(binariesToCheckForUpdate) == 0 || time.Now().Before(ta.initialDelayEnd)` could prevent us from getting here.

RebeccaMahany: I think I'd also reorder to check for changes in the `AutoupdateInitialDelay` flag first, since the other flag changes rely on `ta.initialDelayEnd` to know whether they are allowed to perform an autoupdate.

So we can reorder `FlagsChanged` to make sure that we're processing all the flags, and only performing an autoupdate a maximum of once --

* Check `AutoupdateDownloadSplay` (I think it is fine to leave this first)
* Check `AutoupdateInitialDelay`, set `ta.initialDelayEnd`
* Check `AutoupdateInterval`, and if it changed, set a bool flag `autoupdateIntervalChanged` (or whatever name seems reasonable) -- but don't perform an autoupdate yet
* Keep current code to check if UpdateChannel or pinned versions changed
* Update `if len(binariesToCheckForUpdate) == 0 || time.Now().Before(ta.initialDelayEnd) {return}` to also take into account `autoupdateIntervalChanged`
* Proceed with call to `ta.checkForUpdate`

ee/tuf/autoupdate_test.go
RebeccaMahany, line 948: Can we update these existing `TestFlagsChanged_` tests, or add a new test, to test that `FlagsChanged` behaves the way we expect when these two new flags change?

RebeccaMahany:
Closes https://github.com/kolide/launcher/issues/1412

ee/tuf/autoupdate.go
```diff
@@ -371,6 +378,32 @@ func (ta *TufAutoupdater) FlagsChanged(ctx context.Context, flagKeys ...keys.Fla
 		ta.calculatedSplayDelay.Store(0)
 	}
 
+	// Check for AutoupdateInitialDelay changes
+	if slices.Contains(flagKeys, keys.AutoupdateInitialDelay) {
+		if time.Now().Before(ta.initialDelayEnd) {
+			newDelay := ta.knapsack.AutoupdateInitialDelay()
+			ta.initialDelayEnd = time.Now().Add(newDelay)
```
RebeccaMahany: I think this doesn't work quite the way I would expect it to? E.g.:

* launcher starts up at noon, and has an initial 1-hour delay, so the initial delay will be over at 1pm
* at 12:30pm, we send down a new delay of 35 min
* I would expect that that means the delay should end at 12:35 pm (5 mins from now, 35 mins after initial startup) -- but instead, we'd add another 35 mins now, and the initial delay would end at 1:05pm

We probably need to add a new `ta.initialDelayStart`, so that we could instead do `ta.initialDelayEnd = ta.initialDelayStart.Add(newDelay)` here?

cesarfda: Sounds good, I updated that and added a test case to capture this behavior as well

RebeccaMahany (approved):
🔥

PR #2481, opened 2025-11-06
Update `commandExpectedToExit` to return true for `launcher --version`

RebeccaMahany:
We can run the `runVersion` subcommand in two ways: via `launcher version` or `launcher --version` (`launcher -version`). This PR updates `commandExpectedToExit` to account for the latter as well.

Hopefully fixes e.g.

```
Run binaries/Linux-build/linux.amd64/launcher --version
{"time":"2025-11-06T04:27:28.928490431Z","level":"INFO","msg":"launcher starting up","version":"1.28.2-8-g42c6d297","revision":"42c6d2971d722274c88ad5a656782abf4e4ecf3d"}
{"time":"2025-11-06T04:27:28.928602249Z","level":"ERROR","msg":"could not check out latest launcher","err":"checking out latest launcher: could not get autoupdate config: could not read config file because it does not exist at /etc/kolide-k2/launcher.flags: stat /etc/kolide-k2/launcher.flags: no such file or directory","@type":"type.googleapis.com/google.devtools.clouderrorreporting.v1beta1.ReportedErrorEvent","message":"could not check out latest launcher"}
Error: Process completed with exit code 1.
```

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"text/tabwriter"
 	"time"
@@ -355,5 +356,8 @@ func runVersion(_ *multislogger.MultiSlogger, args []string) error {
 // svc is used on windows to run as a service and we need a non-zero exit code for those no matter
 // the reason for the exit so the service manager will auto restart launcher
 func commandExpectedToExit(osArgs []string) bool {
-	return len(osArgs) > 1 && !strings.HasPrefix(osArgs[1], "-") && !strings.HasPrefix(osArgs[1], "svc")
+	return len(osArgs) > 1 &&
+		// Either first arg is not a flag, OR the flag is `--version` or `-version`
+		(!strings.HasPrefix(osArgs[1], "-") || slices.Contains(osArgs, "--version") || slices.Contains(osArgs, "-version")) &&
+		!strings.HasPrefix(osArgs[1], "svc")
 }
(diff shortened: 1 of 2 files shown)


cmd/launcher/main.go
```diff
@@ -354,6 +355,9 @@ func runVersion(_ *multislogger.MultiSlogger, args []string) error {
 // commandExpectedToExit determines if we're running a subcommand that is expected to exit (excluding those starting with "svc")
 // svc is used on windows to run as a service and we need a non-zero exit code for those no matter
 // the reason for the exit so the service manager will auto restart launcher
-func commandExpectedToExit() bool {
-	return len(os.Args) > 1 && !strings.HasPrefix(os.Args[1], "-") && !strings.HasPrefix(os.Args[1], "svc")
+func commandExpectedToExit(osArgs []string) bool {
+	return len(osArgs) > 1 &&
```
zackattack01, line 359: just a nit (not blocking) but this might be easier to read broken up into separate if blocks with the expected cases for each individually documented. i.e. 
```
if len(osArgs) <= 1 { return false } // we don't expect naked launcher call to exit
if strings.HasPrefix(osArgs[1], "svc") { return false } // we don't expect svc calls to exit
if slices.Contains(osArgs, "--version") || slices.Contains(osArgs, "-version") { return true } // we do expect version calls to exit
return !strings.HasPrefix(os.Args[1], "-") // we do expect other subcommands to exit
```
(untested)

zackattack01 (approved):

cmd/launcher/main.go
RebeccaMahany: I agree re: readability -- will get this updated in a subsequent PR

PR #2485, opened 2025-11-10
Complete RegistrationTracker interface

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1473; this PR and a follow-up will supersede https://github.com/kolide/launcher/pull/2474.

This PR adds the following functionality to the `RegistrationTracker` interface and the knapsack's implementation of `RegistrationTracker`:

* `EnsureRegistrationStored`: makes sure that for older launcher installations that only stored the node key in the config store, the registration will be stored in the registration store (including the node key)
* `NodeKey`: retrieves the node key
* `DeleteRegistration`: deletes registration and node key

These additions will allow us to remove all direct node key and registration access from the osquery extension (which I will do in a follow-up PR). This will let us standardize how registrations and node keys are created, managed, and deleted.

### DIFF
--- a/ee/agent/knapsack/knapsack_test.go
+++ b/ee/agent/knapsack/knapsack_test.go
@@ -247,7 +247,6 @@ func TestSaveRegistration(t *testing.T) {
 	}
 
 	for _, tt := range testCases {
-		tt := tt
 		t.Run(tt.testCaseName, func(t *testing.T) {
 			t.Parallel()
 
@@ -363,7 +362,6 @@ func TestEnsureRegistrationStored(t *testing.T) {
 	}
 
 	for _, tt := range testCases {
-		tt := tt
 		t.Run(tt.testCaseName, func(t *testing.T) {
 			t.Parallel()
 


ee/agent/knapsack/knapsack.go
```diff
@@ -140,6 +142,34 @@ func (k *knapsack) SaveRegistration(registrationId, munemo, nodeKey, enrollmentS
 		return errors.New("no registration store")
 	}
 
+	// Ensure we have the minimum information required to save a registration
+	if munemo == "" && enrollmentSecret == "" {
```
zackattack01, line 146: does it make sense to stick these cheaper checks right at the top before doing anything else?

zackattack01: oh getKVStore is just a map lookup, feel free to ignore

ee/agent/knapsack/knapsack_test.go
```diff
@@ -170,6 +173,350 @@ func TestMergeEnrollmentDetails(t *testing.T) {
...
+				expectedMunemo:       testMunemo,
+				expectedNodeKey:      "",
+				expectedEnrollSecret: enrollSecret,
+				errorExpected:        true,
+			},
+		}...)
+	}
+
+	for _, tt := range testCases {
+		tt := tt
```
zackattack01: nit- don't need these anymore

zackattack01 (approved):
🔥

ee/agent/knapsack/knapsack_test.go
RebeccaMahany: ah thank you, updated!

ee/agent/knapsack/knapsack.go
RebeccaMahany: Had kept these a little lower so the munemo/secret check was closer to the munemo extraction from the secret

zackattack01 (approved):

PR #2486, opened 2025-11-10
add dual publication for osquery log cutover

zackattack01:
This adds logic to be able to publish logs separately from our existing pipeline, there should be no change to default behavior and dual publication is fully opt-in at this point. 

### new flags
These can be configured manually via config, and we will add control server logic to be able to set these up in bulk.

- `osquery_log_publish_url`: string url base for the target domain
- `osquery_log_publish_api_key`: temporary auth mechanism used for the new service
- `osquery_log_publish_percent_enabled`: integer percentage (0-100) to determine the amount of cutover to enable.
  - 0 is default, disabled
  - 100 would be full dual publication
  - anything in between will be evaluated for eligibility during each publication batch

### new client
I've opted to introduce a new client and new interface for this, it felt much cleaner this way and should allow us to drop large amounts of code once we've cutover.

- `LogPublisherClient` adheres to our new `Publisher` interface [here](https://github.com/kolide/launcher/blob/zack/dual_log_pubs/ee/osquerylogpublisher/publisher.go#L12)
- idea is that we will also add `PublishResults` and the relevant functionality the same way


ee/osquerypublisher/client.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/kolide/launcher/pkg/service"
+	osqlog "github.com/osquery/osquery-go/plugin/logger"
+)
+
+type (
+	// LogPublisherClient adheres to the Publisher interface. It handles log publication
+	// to the agent-ingester microservice
+	LogPublisherClient struct {
+		logger   *slog.Logger
+		knapsack types.Knapsack
```
RebeccaMahany: If you wanted, I think this could be `types.Flags` instead of `types.Knapsack`, to allow for a narrower/more specific interface.

ee/osquerylogpublisher/publisher.go
```diff
@@ -0,0 +1,50 @@
+package osquerylogpublisher
```
RebeccaMahany: If this is going to also handle publishing distributed query results in the future, do we want a more generic name that doesn't specify `log`? We could do just `osquerypublisher` if it's going to be just a substitute for PublishLogs and PublishResults. If we think we might have RequestQueries/RequestConfig/RequestEnrollment here eventually too, that's a little harder for naming -- maybe like...`osqueryadapter` or `osqueryconnector`.

ee/osquerylogpublisher/publisher.go
```diff
@@ -0,0 +1,50 @@
...
+		IngestedBytes int64  `json:"ingested_bytes"`
+		LogCount      int    `json:"log_count"`
+		Message       string `json:"message"`
+	}
+)
+
+// levelForError returns slog.LevelError if err != nil, else slog.LevelDebug
+func levelForError(err error) slog.Level {
+	if err != nil {
+		return slog.LevelWarn
```
RebeccaMahany: Comment says slog.LevelError, but function returns slog.LevelWarn?

ee/osquerylogpublisher/publisher.go
RebeccaMahany: Same comment goes for the agent flags `OsqueryLogPublishURL` etc

pkg/launcher/options.go
```diff
@@ -225,6 +230,10 @@ func ParseOptions(subcommandName string, args []string) (*Options, error) {
 		flLogIngestServerURL              = flagset.String("log_ingest_url", "", "Where to export logs")
 		flTraceIngestServerURL            = flagset.String("trace_ingest_url", "", "Where to export traces")
 		flDisableIngestTLS                = flagset.Bool("disable_trace_ingest_tls", false, "Disable TLS for observability ingest server communication")
+		// Osquery log ingest configuration for dual publication cutover
+		flOsqueryLogPublishURL            = flagset.String("osquery_log_publish_url", "", "URL base for publishing osquery logs and status")
+		flOsqueryLogPublishAPIKey         = flagset.String("osquery_log_publish_api_key", "", "TEMPORARY- API key for osquery log ingest")
```
RebeccaMahany: Is there a reason we need to allow this to be set via command line?

zackattack01: it's just how I've been testing, I can pull if you think that's a bad idea or just plan to remove during clean up at the end

ee/osquerylogpublisher/publisher.go
zackattack01: oh copy paste error, will fix both. good catch thanks!

ee/osquerylogpublisher/publisher.go
zackattack01: 👍 makes sense, I'll do `osquerypublisher` for now and we can re-evaluate something more broad if needed later

ee/osquerypublisher/client.go
zackattack01: will do 👍

ee/osquerylogpublisher/publisher.go
zackattack01: i'll get the flag names updated as well, probably tomorrow AM

zackattack01: done!

pkg/launcher/options.go
RebeccaMahany: I figured we probably don't want the option to have the api key in plaintext in e.g. launcher.flags, but I'm fine w removing it during cleanup at the end if that's easiest

RebeccaMahany (dismissed):
LGTM, your call re command line flag

ee/osquerypublisher/client.go
```diff
@@ -0,0 +1,164 @@
...
+	return &LogPublisherClient{
+		logger:   logger.With("component", "osquery_log_publisher"),
+		knapsack: k,
+		client:   client,
+	}
+}
+
+// helper method to allow us to make any http client tweaks as we learn realistic
+// parameters for interacting with the agent-ingester service
+func NewPublisherHTTPClient() PublisherHTTPClient {
```
directionless:
```suggestion
// NewPublisherHTTPClient is a helper method to allow us to make any http client tweaks as we learn realistic
// parameters for interacting with the agent-ingester service
func NewPublisherHTTPClient() PublisherHTTPClient {
```

godoc style

ee/osquerypublisher/client.go
```diff
@@ -0,0 +1,164 @@
...
+			"status_code", resp.StatusCode,
+			"err", err,
+		)
+		return nil, fmt.Errorf("unable to unmarshal agent-ingester response: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		logger.Log(ctx, slog.LevelError, "received non-200 response from agent-ingester",
+			"status_code", resp.StatusCode,
+		)
```
directionless: Hrm. I can't quite tell, but is this duplicated with the deferred logging?

ee/osquerypublisher/publisher.go
```diff
@@ -0,0 +1,50 @@
...
+	// This is exported for use by agent-ingester
+	PublishLogsRequest struct {
+		LogType osqlog.LogType `json:"log_type"`
+		Logs    []string       `json:"logs"`
+	}
+
+	// PublishLogsResponse represents the response structure from agent-ingester for a given batch of log publications.
+	// Note that this can also be used to unmarshal and errorResponse from agent-ingester, in that case
+	// only the status and message fields will be populated. This is exported here for use by agent-ingester.
+	PublishLogsResponse struct {
```
directionless: This is a protocol, yes?
1. I wonder if we should stick it in a types.go file, and import it on the ingestion server
2. I wonder if we should define it using any of the API definition tools (jsonscema, openapi. etc). Something to enforce and test the boundary

ee/osquerypublisher/client.go
zackattack01: oh yeah good point, I will pull this

ee/osquerypublisher/publisher.go
zackattack01: I had left it exported to use from agent-ingester, I can move it to a types.go file. and will look into incorporating a tool here!

ee/osquerypublisher/client.go
directionless: TBH all the logging seems a little rough. But I'm not sure that's wrong, we'll evolve. 🚀

ee/osquerypublisher/publisher.go
directionless: >  had left it exported to use from agent-ingester

That's like 80% good enough 😆 

I think the advantage to using a tool is that it gives us space for better testing and system changes. But it's a little premature -- we don't do much with that yet. This could be good enough for now, and we can think about a large effort to move all the APIs into some schema oriented tool

zackattack01: cool, I've moved the Request/Response structs to a new types file and added a comment about our goals here! going to merge this early to avoid more extension_test conflicts 😄

PR #2487, opened 2025-11-11
Update osquery extension to use RegistrationTracker exclusively for managing registrations and node keys

RebeccaMahany:
Relates to https://github.com/kolide/launcher/issues/1473; this PR and https://github.com/kolide/launcher/pull/2485 supersede https://github.com/kolide/launcher/pull/2474.

Uses updates in previous PR to move registration management entirely out of the osquery extension, so that management of the registration (and node keys) is centralized in the knapsack.

### DIFF
--- a/pkg/osquery/extension.go
+++ b/pkg/osquery/extension.go
@@ -42,7 +42,7 @@ type Extension struct {
 	knapsack                      types.Knapsack
 	serviceClient                 service.KolideService
 	settingsWriter                settingsStoreWriter
-	enrollMutex                   *sync.Mutex
+	enrollMutex                   *sync.Mutex // ensures that for non-secretless installations, we never have simultaneous RequestEnrollment requests
 	done                          chan struct{}
 	interrupted                   atomic.Bool
 	slogger                       *slog.Logger


pkg/osquery/extension.go
```diff
@@ -554,38 +410,19 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
+	return e.nodeKey() != ""
 }
 
-// nodeKey returns the current node key for this registration. It ensures that if the
-// node key is set in the database but not in the extension, we will pull it. We should therefore
-// always use this function when we need the node key to make requests to the device server,
-// to avoid making requests with an empty node key and attempting to reenroll instead.
+// nodeKey returns the current node key for this registration.
 func (e *Extension) nodeKey() string {
 	e.enrollMutex.Lock()
```
zackattack01: do we still need to perform this locking?

pkg/osquery/extension.go
```diff
@@ -425,48 +304,25 @@ func (e *Extension) Enroll(ctx context.Context) (string, bool, error) {
...
-		e.NodeKey = key
-		if err := e.ensureNodeKeyStored(key); err != nil {
-			e.slogger.Log(ctx, slog.LevelError,
-				"could not update registration",
-				"err", err,
-			)
-		}
-		return e.NodeKey, false, nil
-	}
+	// Only one thread should ever be allowed to attempt enrollment at the
```
zackattack01: should we consider moving this locking to knapsack? i don't see anything that wouldn't work but it feels a little wrong and like it might be difficult to maintain/understand here

RebeccaMahany: Consolidating with https://github.com/kolide/launcher/pull/2487#discussion_r2515885698 --

My best guess is that we do need the enrollment lock inside `Enroll` itself, because we only want one thread to attempt enrollment at a time -- for that reason, I think we should keep it inside the extension. But maybe we _don't_ need the enrollment lock around the node key access anymore, since we won't have data race issues trying to access it anymore since it's not cached in the struct. I am going to try removing usage of the enrollment lock from everywhere except `Enroll`, and moving it to the top of `Enroll` instead, and see how that goes.

RebeccaMahany: > for that reason, I think we should keep it inside the extension

To clarify -- I mean we mostly care about the enroll mutex because we don't want the extension firing off multiple calls to `Enroll` at the same time, so that it's not requesting/overwriting multiple node keys. We don't have to worry about enroll races with secretless enrollment, because the localserver callback worker processes callbacks one at a time. So, since we only have to worry about it here, and we need to worry about it for an operation that's longer than just the one call to `SaveRegistration` etc, I think it makes sense for the enroll mutex to live here.

RebeccaMahany: ~~And -- waffling on whether we still want the enroll mutex lock inside `RequireReenroll`, but I think I'm leaning toward keeping that to preserve current behavior. I've pushed those changes, lmk what you think!~~ 

JK -- when testing, this forced a double enrollment, where `RequireReenroll` was called because a function that used to have to wait to get a node key until the already-ongoing `Enroll` was finished _now_ got an empty node key right away, failed, and queued up its `RequireEnroll` call before `Enroll` finished. I am going to think on this a little longer.

zackattack01: ahh okay so it sounds like we definitely do want to lock on reads then? I think that the localserver callback queue and enrollment happening here now are all good arguments but this smells like if anything else ever uses these (`k.NodeKey` included) we'd be subject to the same race conditions, vs just locking on what we expose from knapsack should never allow this to happen and would allow us to clean a lot of this mutex management up.

I could go either way here but I think if we had written this logic from scratch with knapsack being responsible, it would seem obvious that that is where the locking should happen. maybe i'm missing something though, are there other side effects outside of what we do through knapsack now that we're locking to protect against?

zackattack01: ahh this

> and we need to worry about it for an operation that's longer than just the one call to SaveRegistration

i'm torn because now I feel like there should be the lock on NodeKey anyway 😬

RebeccaMahany: Okay, so here's what I think we have to manage:

* For non-secretless installations, enrollment happens via the extension. If an enrollment is ongoing via the extension, _all other_ extension calls must wait for it to complete, to avoid triggering reenrolls. The enrollMutex, since it also protected node key access, solved this.
  * There are two paths to enrollment here. The first is the initial call to `Enroll` right after we create the extension, inside the osquery instance. Ideally, if the installation isn't yet enrolled, this first call to `Enroll` will enroll it. The second path is the `withReenroll` functions inside the extension -- if the first call to `Enroll` failed, and then a call to e.g. `RequestConfig` fails with a "node invalid" response, this triggers `RequireReenroll` and then `Enroll`.
* For secretless installations, enrollment happens via the localserver; reenrolls via the extension will _never_ work. We must therefore always avoid triggering reenrolls, because `RequireReenroll` can delete the valid registration+node key.

Those cases look like this:

Is secretless | Node key is non-empty | Node key is valid | Enrollment ongoing | Should delete existing node key (RequireReenroll) | Enroll should run
-- | -- | -- | -- | -- | --
Yes | N/A | N/A | N/A | No | No
No | No | N/A | No | No (but it wouldn't hurt?) | Yes
No | No | N/A | Yes | No | No
No | Yes | No | No | Yes | Yes
No | Yes | No | Yes | No | No
No | Yes | Yes | N/A | No | No

Proposal:

* Keep the `enrollMutex` and use it only for enrollment -- do not lock node key.
* For all `doXWithReenroll` functions, check if node key is empty immediately after retrieving it. If it's empty, queue up an `Enroll` call right then (rather than waiting until after we get a "node invalid" response from `doX`). We won't have simultaneous `Enroll` calls from different `doXWithReenroll` functions because of the `enrollMutex`; any extraneous `Enroll` calls will return early, when they notice the node key is non-empty; and `Enroll` will also bail out safely if it notices we're a secretless enrollment. (The improvement here over our current implementation is that we won't call `RequireReenroll`, and therefore won't accidentally delete a node key that got set by a different thread.)
* We should therefore only get to the `RequireReenroll` call if we have a truly invalid node key, not just an empty one during first-time install. (This should be a pretty rare case.) We can go ahead and call `RequireReenroll` and then `Enroll`. No point in locking the `enrollMutex` in `RequireReenroll`.

zackattack01: okay, that is really helpful thank you 💯 
I was not understanding that there is actually zero overlap possible between these paths. that proposal makes sense to me

RebeccaMahany: > I was not understanding that there is actually zero overlap possible between these paths

I have been figuring this out as we were working through testing secretless enrollment over last week 😅 so it's new info, and also not well-documented yet. I think I probably need to write something up in https://github.com/kolide/launcher/tree/main/docs.

RebeccaMahany: Okay, these changes appear to work in practice and not just in theory 😅  -- I've pushed them up. I will handle docs in a separate PR.

zackattack01: looks great, thanks again!

RebeccaMahany: Thank you for raising it!

PR #2676, opened 2026-04-09
Avoid execing for version detection in packaging

RebeccaMahany:
We can't use version detection when building cross-arch packages (e.g. when building an ARM Linux package on AMD), and it is not workable for us to always pass in PackageVersion in this case.

This PR updates to pull the version from `LauncherVersion`. This value can be either an actual version (like "2.0.0"), or a channel (like "stable"). If it's an actual version, then we use that version (after confirming formatting). If it's a channel, then we pull the release version for that channel from TUF. To avoid repeated calls to TUF when building many packages in sequence, we save TUF metadata inside the cache directory.

If you're testing package-builder by providing a local launcher build (via `-launcher_path`), you now must set the launcher and package version via the `-package_version` flag. That felt like a better tradeoff than maintaining the exec-version-and-parse logic, given that we only use `-launcher_path` for testing purposes and not for our actual build.


pkg/packaging/helpers.go
```diff
@@ -18,45 +18,105 @@ func sanitizeHostname(hostname string) string {
...
-func (p *PackageOptions) osqueryLocation() string {
-	if p.target.Platform == Darwin {
-		// We want /usr/local/osquery.app, not /usr/local/bin/Kolide.app, so we use Dir to strip out `bin`
-		appBundleBinaryPath := filepath.Join(p.packageRoot, filepath.Dir(p.binDir), "osquery.app", "Contents", "MacOS", "launcher")
-		if info, err := os.Stat(appBundleBinaryPath); err == nil && !info.IsDir() {
-			return appBundleBinaryPath
-		}
+// formatVersion formats the version according to the packaging requirements of the target platform
+// currently, only windows and darwin platforms require modification
+func formatVersion(rawVersion string, platform PlatformFlavor) (string, error) {
```
RebeccaMahany, line 70: formatVersion is moved directly from detectLauncherVersion.go -- the only change I made is on line 88, changing `len(matches)` to `len(matches[0])` to fix the error string.

pkg/packaging/helpers_test.go
```diff
@@ -28,35 +28,60 @@ func TestSanitizeHostname(t *testing.T) {
 	}
 }
 
-func Test_osqueryVersionFromVersionOutput(t *testing.T) {
+func TestFormatVersion(t *testing.T) {
```
RebeccaMahany, line 31: This test is moved directly from `detectLauncherVersion_test.go`; it is unmodified.

cesarfda (approved):

zackattack01 (approved):

PR #2687, opened 2026-04-12
fix(actionqueue): guard cancel nil in StopCleanup

directionless:
cancel is only set when StartCleanup is called. If StopCleanup is called first (e.g. during failed startup), cancel is nil and panics.

### DIFF
--- a/ee/control/actionqueue/actionqueue.go
+++ b/ee/control/actionqueue/actionqueue.go
@@ -175,7 +175,9 @@ func (aq *ActionQueue) runCleanup() {
 }
 
 func (aq *ActionQueue) StopCleanup(err error) {
-	aq.cancel()
+	if aq.cancel != nil {
+		aq.cancel()
+	}
 }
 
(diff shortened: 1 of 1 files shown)


ee/control/actionqueue/actionqueue.go
```diff
@@ -175,7 +175,9 @@ func (aq *ActionQueue) runCleanup() {
 }
 
 func (aq *ActionQueue) StopCleanup(err error) {
-	aq.cancel()
+	if aq.cancel != nil {
+		aq.cancel()
+	}
 }
```
RebeccaMahany, line 181: We probably also want a lock around `aq.cancel` to avoid flaky data race failures in tests?

directionless: It passed CI, do you still think we need it?

RebeccaMahany: Yes, might as well if we're updating this area anyway

PR #2691, opened 2026-04-12
Guard against missing name/query keys in scheduled query results

directionless:
osquery_schedule rows are map[string]string; accessing a missing key returns empty string rather than panicking, but passing an empty string as a SQL query to the querier would cause a confusing error. Add explicit presence and emptiness checks before using each value.

### DIFF
--- a/ee/localserver/request-query.go
+++ b/ee/localserver/request-query.go
@@ -101,17 +101,28 @@ func (ls *localServer) requestScheduledQueryHandlerFunc(w http.ResponseWriter, r
 	results := make([]distributed.Result, len(scheduledQueriesQueryResults))
 
 	for i, scheduledQuery := range scheduledQueriesQueryResults {
-		results[i] = distributed.Result{
-			QueryName: scheduledQuery["name"],
+		name, nameOk := scheduledQuery["name"]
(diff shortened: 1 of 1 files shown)


directionless:
This feels a little excessive, but I think, not wrong. Perhaps there's some narrow case it the `[]map[string]string` comes back funny

ee/localserver/request-query.go
```diff
@@ -101,17 +101,28 @@ func (ls *localServer) requestScheduledQueryHandlerFunc(w http.ResponseWriter, r
...
 
 	for i, scheduledQuery := range scheduledQueriesQueryResults {
+		name, nameOk := scheduledQuery["name"]
+		query, queryOk := scheduledQuery["query"]
+		if !nameOk || name == "" || !queryOk || query == "" {
+			ls.slogger.Log(r.Context(), slog.LevelError,
+				"scheduled query result missing name or query field",
+				"query_name", name,
+			)
+			results[i].Status = 1
```
RebeccaMahany, line 111: Is there a const we can use for this? Maybe osquery-go's `ExtensionCode_EXT_FAILED`?

PR #2694, opened 2026-04-13
add zlib compression to osquery log and result publication

zackattack01:
- adds a zstd encoding step prior to our osquery log publication logic
  - per the zstd library docs, we should re-use the encoder for performance benefits when using repeatedly for many small blocks (this allows it to run without significant additional allocations after a warm up period)
- the batching logic is not updated for now, I'll be collecting the compression log stats over a few days to determine an appropriate batch size based on average performance with these logs and results

### DIFF
--- a/ee/osquerypublisher/client.go
+++ b/ee/osquerypublisher/client.go
@@ -14,6 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klauspost/compress/zstd"
 	"github.com/kolide/kit/contexts/uuid"
 	"github.com/kolide/launcher/v2/ee/agent/storage"
 	"github.com/kolide/launcher/v2/ee/agent/types"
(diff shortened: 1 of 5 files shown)


directionless:
Might be worth thinking about the ruby side. Zlib is in the standard library, everything else is 🤷

ee/osquerypublisher/client.go
```diff
@@ -14,6 +14,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klauspost/compress/zstd"
```
directionless: These days, I feel super worried about new dependancies

zackattack01: I had chosen this because we already had to pull it in for the indexeddb transform step, I'll update to use zlib though that will be cleaner with the ruby side!

RebeccaMahany: FWIW, we already use this in https://github.com/kolide/launcher/pull/2644, and even before that PR it was an indirect dependency anyhow

directionless: thank you.

directionless: (I did nose around the ruby side, and I didn't see us pulling in any of the zstd implementations. I also didn't evaluate gems. There are some 🤷 )

RebeccaMahany (approved):

PR #2696, opened 2026-04-15
Trying to reduce secret scan false positives

directionless:
This is a bit if a heavy handed approach to reducing false positives -- it actually parses the file and tosses it. In contrast, the gitleaks developers might suggest being clever with allow regexps. https://github.com/gitleaks/gitleaks/issues/1728

### DIFF
--- a/ee/tables/secretscan/false_positives.go
+++ b/ee/tables/secretscan/false_positives.go
@@ -1,0 +1,65 @@
+package secretscan
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+
+	"github.com/zricethezav/gitleaks/v8/report"
+)
+
+// knownFalsePositive attempts to guess if something is a false
+// positive. Inherently, it's a pile of special cases. It may need
+// to be re-written as we learn more.
+func knownFalsePositive(finding report.Finding) bool {
+	// currently all the known false positives are on the generic rule. Check that first
+	if finding.RuleID == "generic-api-key" {
+		return false
+	}
+
+	// current false positives are under 1000 characters. Conviniently this
+	// reduces risk on the base64 decoder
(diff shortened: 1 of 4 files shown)


ee/tables/secretscan/false_positives.go
```diff
@@ -0,0 +1,65 @@
...
+		// not base64, so not us
+		return false
+	}
+
+	var decoded map[string]string
+	if err := json.Unmarshal(fromBase64, &decoded); err != nil {
+		return false
+	}
+
+	expectedKeys := []string{"cty", "data", "enc", "iv", "kid"}
```
RebeccaMahany, line 52: I think these are all part of the JWE standard -- do we have anything outside of the standard to match against? I am wondering if this would match other items we aren't expecting.

RebeccaMahany: ~~(i.e., _all_ JWEs can't be false positives, right?)~~

If JWEs are all false positives, since they are not plaintext, then we can maybe use the standard to detect them. I took a look and there's a library golang-jwt/jwe similar to golang-jwt/jwt that we already use, but I'm not sure it's stable. But looking at the RFC, I think there are two different ways to serialize JWEs ([JWE Compact](https://datatracker.ietf.org/doc/html/rfc7516#section-7.1) and [JWE JSON](https://datatracker.ietf.org/doc/html/rfc7516#section-7.2)) and maybe we could use those for stricter detection rules here, to make sure we capture both types?

RebeccaMahany: Handled in https://github.com/kolide/launcher/pull/2697

directionless: Just to answer -- Yes, I think all JWEs are false positives

PR #2699, opened 2026-04-16
Ignore empty variable names in secretscan

RebeccaMahany:
Similar to https://github.com/kolide/launcher/pull/2697 -- ignore another common false-positive pattern.

This issue is https://github.com/gitleaks/gitleaks/issues/1828. Since I didn't see a gitleaks workaround at this time, I added another exclusion to our secret processing.

### DIFF
--- a/ee/tables/secretscan/table.go
+++ b/ee/tables/secretscan/table.go
@@ -370,11 +370,7 @@ func isEmptyVariable(finding report.Finding) bool {
 	}
 	before, _, _ := strings.Cut(lineWithSecret, finding.Secret)
 	beforeTrimmed := strings.TrimSpace(before)
-	if beforeTrimmed != "" {
-		return false
-	}
-
-	return true
+	return beforeTrimmed == ""
 }
 
 // findingsToKeyNames attempts to extract the key names (eg: in an .env file) to help understand the context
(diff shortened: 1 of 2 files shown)


ee/tables/secretscan/table.go
```diff
@@ -323,6 +333,46 @@ func isEncryptedJWTFamilyValue(finding report.Finding) bool {
...
+
+	// Next, check for our regex match.
+	if !emptyVariableRegexp.MatchString(finding.Secret) {
+		return false
+	}
+
+	// We expect that this "secret" would be at the start of a line, with either nothing
+	// or whitespace in front of it. However, sometimes our finding.Line will contain
+	// multiple lines -- in this case, it looks like "\nMY_ENV_VAR1=\nMY_ENV_VAR2=".
+	// So first we isolate the actual line we're looking at, then check to see if there's
```
directionless, line 358: You might be able to use the column start field.

directionless (approved):

ee/tables/secretscan/table.go
RebeccaMahany: I tried that and it worked in practice with the examples I'm looking at (restricting to `finding.StartColumn == 2`), but I wanted to cover the case in the `tab before empty variable` test too.

PR #2702, opened 2026-04-20
Move an existing secretscan exception into config.toml

RebeccaMahany:
After updating to use our own override config here: https://github.com/kolide/launcher/pull/2701, we can now move the exceptions I previously added to config.toml.

This PR moves https://github.com/kolide/launcher/pull/2699 into the config.

### DIFF
--- a/ee/tables/secretscan/config.toml
+++ b/ee/tables/secretscan/config.toml
@@ -5,11 +5,15 @@ title = "Kolide secretscan config"
 [extend]
 useDefault = true
 
-# Ignore K8s Sealed Secrets
-# https://github.com/gitleaks/gitleaks/issues/1728
 [[rules]]
 id = "generic-api-key"
 [[rules.allowlists]]
+description = "Ignore K8s Sealed Secrets (https://github.com/gitleaks/gitleaks/issues/1728)"
 condition = "AND"
 regexes = ['''Ag[a-zA-Z0-9+/]{500,}={0,2}''']
 paths = ['''(?i).*\.ya?ml$''']
+[[rules.allowlists]]
+description = "Ignore variable names with empty values (https://github.com/gitleaks/gitleaks/issues/1828)"
+condition = "AND"
+regex_target = "match"
+regexes = ['''^\s*\w[\w-]*=$''']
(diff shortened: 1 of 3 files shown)


ee/tables/secretscan/config.toml
```diff
@@ -5,11 +5,17 @@ title = "Kolide secretscan config"
...
+description = "Ignore K8s Sealed Secrets (https://github.com/gitleaks/gitleaks/issues/1728)"
 condition = "AND"
 regexes = ['''Ag[a-zA-Z0-9+/]{500,}={0,2}''']
 paths = ['''(?i).*\.ya?ml$''']
+[[rules.allowlists]]
+description = "Ignore variable names with empty values (https://github.com/gitleaks/gitleaks/issues/1828)"
+condition = "AND"
+# We limit length to 35 as a proxy for not allowlisting high-entropy values here.
+# Variable names are probably usually under that length.
+regexes = ['''^\s*[A-Z\d][A-Z\d_-]{0,35}=$''']
```
zackattack01, line 21: do these need to be uppercase? I know they usually will be but just double checking since it seems like the existing functionality would have regex matched on either

zackattack01 (approved):

ee/tables/secretscan/config.toml
RebeccaMahany: I made that change to replace our previous requirement for entropy >= 4.0. (The config doesn't let us set entropy per-allowlist -- we would have to override the entropy for the entire generic-api-key rule, which I wasn't sure was a good idea.) Without this and the length restriction, I was getting matches for mixed-case strings that were clearly not empty env vars, like `aBc-a1b2c43..._zXy=`. So, I added in the length restriction and also the consistent case restriction, and that knocks out the high-entropy matches.

I felt like we would be most likely to run into this with all caps `MY_ENV_VAR=`. I thought about updating the regex to be all uppercase or all lowercase, but wasn't sure it'd be worth making the regex more complicated. Though, now that I'm thinking about it, I could also do two separate allowlist entries, one for all caps and one for all lowercase -- that would probably still be readable.

RebeccaMahany: Updated to do two separate allowlists, and added more detail to the comment here

zackattack01: ohhh i see. that approach makes sense to me, thanks!

zackattack01 (approved):

PR #2707, opened 2026-04-23
Remove unused osquery.autoload file

RebeccaMahany:
I believe we used this option before when we shipped extensions (like tables.ext), but we have not shipped these in a very long time. This file is empty and unused. So, no point in creating it.

https://osquery.readthedocs.io/en/stable/deployment/extensions/


zackattack01 (approved):

directionless:
This file was covering for an old osquery bug. IIRC osquery wouldn't properly manage extension connecting to it's socket, if it didn't have that option. And launcher has _always_ connected as an extension. 

I also think this bug is long fixed. But if you see weird behavior where the launcher tables are missing... Suspect this

RebeccaMahany:
@directionless thank you for the additional context -- is https://github.com/osquery/osquery/issues/5680 the issue you were thinking of? Do you remember if the tables were only sometimes missing, or always missing? I tested querying extension tables and didn't run into any issues, but if it was an occasional/transient issue, I can do some additional re-testing.

PR #2711, opened 2026-04-28
Remove unused logs in recursive hot path

cesarfda:
## Summary

Gates the four `slogger.With(...)` call sites in `ee/dataflatten/flatten.go`
behind a `Handler.Enabled(...)` check so we only build a contextual logger when
the level will actually be emitted.

## Why

A flare from a customer host ( showed
the launcher process with:

- `go_mem_bytes` (`Sys − HeapReleased`) sitting around 1.1–1.2 GB
- Lifetime allocations of **311 GB** across 3.18 B mallocs, with NumGC = 3632
- HeapSys peaking at ~38 GB 

There is **no leak** — inuse heap at flare time is 25 MB. The problem is
allocation churn driving GC pressure and intermittent heap growth.

The pprof `alloc_space` profile points unambiguously at one path:

| cum %  | function |
|--------|----------|
| 84.60% | `dataflattentable.(*Table).generate` (251 GB) |
| 77.23% | `dataflatten.flattenJsonl` |
| 71.63% | `dataflatten.(*Flattener).descend` |
| 71.01% | `slog.(*Logger).With` (210 GB) |
| 56.98% | `slog-multi.HandleInlineMiddleware.WithAttrs` |
| 55.49% | `dedup.(*dedupHandler).WithAttrs` (165 GB) |
| 13.55% | `slog.argsToAttrSlice` (40 GB) |
| 10.69% | `slices.Clone` of `[]slog.Attr` (32 GB) |
| 8.89%  | `slog.(*commonHandler).clone` (26 GB) |

### Root cause

`Flattener.descend` is recursive and is the engine behind every `kolide_*`
virtual table query. At the top of every recursion (and inside its inner
loops) it built a contextual logger via `fl.slogger.With(...)` — even though
the subsequent `slogger.Log(...)` calls all ran at
`levelFlattenDebug = slog.LevelDebug - 1` and were filtered out at the
handler.

`Logger.With` is **not** free: it propagates through the entire slog-multi
chain (`utcTimeMiddleware → ctxValuesMiddleWare → dedupHandler →
reportedErrorMiddleware → Fanout → leaf handlers`), and every middleware
clones the attribute slice. For one deep JSON/plist row, `descend` runs
thousands of times, so we ended up doing 6–10 attribute-slice clones per
recursion that produced no output.

### Why remove rather than gate

The first iteration of this fix gated each `With` site behind
`slogger.Enabled(...)` so the allocations only happened when debug logging
was actually requested. After discussing with the team, the call was that
dataflatten has been stable for long enough that the per-step debug logging
is no longer pulling its weight — the times we've actually needed to debug
this code in the last year, the rows themselves and the eventual error
messages have been sufficient. The cleaner fix is to delete the noisy
logging outright (and the `WithDebugLogging` knob, which was only ever wired
to a `tabledebug` flag that nothing sets in production).

The two log calls we kept in `descendMaybePlist` are real failure signals,
not per-iteration tracing, and they're on a cold path (only reached for
string/byte data that looks like a plist).

***Assisted with Claude Code***

### DIFF
--- a/ee/dataflatten/examples/flatten/main.go
+++ b/ee/dataflatten/examples/flatten/main.go
@@ -29,8 +29,6 @@ func main() {
 		flXml   = flagset.String("xml", "", "Path to xml file")
 		flIni   = flagset.String("ini", "", "Path to ini file")
 		flQuery = flagset.String("q", "", "query")
-
-		flDebug = flagset.Bool("debug", false, "use a debug logger")
 	)
 
 	if err := ff.Parse(flagset, os.Args[1:],
@@ -44,9 +42,6 @@ func main() {
 		dataflatten.WithSlogger(multislogger.NewNopLogger()),
 		dataflatten.WithNestedPlist(),
 		dataflatten.WithQuery(strings.Split(*flQuery, `/`)),
-	}
-	if *flDebug {
-		opts = append(opts, dataflatten.WithDebugLogging())
 	}
 
 	rows := []dataflatten.Row{}
--- a/ee/dataflatten/flatten.go
+++ b/ee/dataflatten/flatten.go
@@ -162,13 +162,21 @@ func Flatten(data any, opts ...FlattenOpts) ([]Row, error) {
 func (fl *Flattener) descend(path []string, data any, depth int) error {
 	queryTerm, isQueryMatched := fl.queryAtDepth(depth)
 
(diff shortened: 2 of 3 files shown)


RebeccaMahany (dismissed):

RebeccaMahany (approved):
🔥

PR #2715, opened 2026-05-14
Add indexeddb blob support

RebeccaMahany:
Allows for reading external objects like blobs in Chrome IndexedDBs. Also has a small amount of fixes and cleanup for Chrome IndexedDB parsing to make parsing more complete.

I recommend briefly skimming https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/renderer/modules/indexeddb/idb_value_wrapping.cc and https://github.com/chromium/chromium/blob/main/content/browser/indexed_db/docs/leveldb_coding_scheme.md#externalobject-value (also linked in code) for additional context.


ee/indexeddb/indexeddb.go
```diff
@@ -110,9 +126,49 @@ func QueryIndexeddbObjectStore(ctx context.Context, slogger *slog.Logger, dbLoca
...
+		// Check for external object data associated with this key, in case this value is wrapped --
+		// we want to unwrap it before any additional processing happens.
+		// We can find external object data for this value by taking the portion of the key after the key prefix,
+		// called the "user key", and concatenating it with the blob prefix. If there is external object data
+		// available for this key, it will live at this blob key.
+		userKey := bytes.TrimPrefix(key, keyPrefix)
+		blobKey := slices.Concat(blobPrefix, userKey)
+		var externalObjectFilenames []string
+		// Query for external objects associated with this key.
+		externalObjectsRaw, err := db.Get(blobKey, nil)
```
zackattack01, line 152: could we shift some of the logic from handleWrappedValues here to avoid this call for every record? it seems like the most common path here by far should be ErrNotFound- so just wondering if it would be more efficient to parse the header out and hold it here, and then peek at the first 3 tokens to see if we actually want to do the rest of the blob logic. `handleWrappedValues` could still work the same (minus the header logic) and then we could just update line 169 below to be `append(header, unwrapped...)`

RebeccaMahany: That is reasonable, let me take a look!

zackattack01: i guess it gets kinda messy with the potential double wrapping (although idk if anything could actually be snappy then blob). but even if we just checked against those first 2 bytes I assume it would cut down a ton of these and be safe

RebeccaMahany: Updated!

RebeccaMahany: I went with just checking if it was wrapped (rather than specifically blob) for that reason. It was easy and we still avoid the extra db call for the majority of indexeddb values

zackattack01 (approved):

ee/indexeddb/indexeddb.go
zackattack01: nice! looks great thank you

PR #2716, opened 2026-05-18
Remove go.uber.org/atomic direct dependency

cesarfda:
## Summary
- Add `pkg/atomic` with `Bool`, `Duration`, and `String` wrappers around `sync/atomic` primitives, matching the API we used from `go.uber.org/atomic`.
- Swap the import in `ee/desktop/runner/runner.go`, `ee/localserver/server.go`, `ee/localserver/krypto-ec-middleware.go`, and `ee/control/control.go` — no call-site changes needed.
- `go mod tidy` demotes `go.uber.org/atomic` to `// indirect` (still pulled in by `golang-migrate`, which we don't control).

## Why
Shrinks our direct dependency surface: one less external publisher we implicitly trust to push code into our binaries, and one less candidate for a poisoned-tag-style supply chain attack on a direct dep. The wrapper is ~60 lines and lives under normal code review.

## Test plan
- [x] `go vet ./...`
- [x] `go test ./pkg/atomic/... ./ee/control/... ./ee/desktop/runner/... ./ee/localserver/...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/atomic/atomic.go
+++ b/pkg/atomic/atomic.go
@@ -1,26 +1,14 @@
 // Package atomic provides convenience wrappers around the standard
-// library's sync/atomic primitives for the types launcher uses most
-// frequently: bool, time.Duration, and string.
+// library's sync/atomic primitives for the types it doesn't cover
+// directly: time.Duration and string.
+//
+// For atomic booleans, use sync/atomic.Bool from the standard library.
 package atomic
 
 import (
 	"sync/atomic"
 	"time"
 )
-
-// Bool is an atomic boolean.
-type Bool struct {
-	v atomic.Bool
-}
-
-// Load atomically loads and returns the value.
-func (b *Bool) Load() bool { return b.v.Load() }
-
-// Store atomically stores val.
-func (b *Bool) Store(val bool) { b.v.Store(val) }
-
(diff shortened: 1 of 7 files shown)


pkg/atomic/atomic.go
```diff
@@ -0,0 +1,66 @@
...
+package atomic
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+// Bool is an atomic boolean.
+type Bool struct {
+	v atomic.Bool
```
zackattack01: is there any functional difference between this and just using the upstream `sync/atomic.Bool` directly? just wondering because I thought the uber version wrapped Uint32, so I wonder if upstream atomic.Bool just wasn't there before when this was implemented, and if that's the one we should be using?

cesarfda: It looks like go.uber.org/atomic predates Go 1.19 when sync/atomic.Bool was introduced. I'll update that to use sync/atomic.Bool directly

RebeccaMahany (approved):

pkg/atomic/atomic.go
zackattack01: nice! looks good thank you

zackattack01 (approved):

PR #2717, opened 2026-05-20
dataflatten for claude

cesarfda:
<html><head></head><body><h1>Add <code>data_flatten</code> option to KATC tables</h1>
<h2>Requirement</h2>
<p>Enable osquery queries against Claude's IndexedDB-backed React Query cache, specifically:</p>
<ul>
<li><strong><code>chat_conversation_list</code></strong> — under <code>clientState.state.data.data.*</code>: count of chats and <code>created_at</code> / <code>updated_at</code> for each.</li>
<li><strong><code>chat_conversation_tree</code></strong> — under <code>clientState.state.data.chat_messages.*</code>: count of messages and <code>created_at</code> / <code>updated_at</code> for each.</li>
</ul>
<p>The fixed-columns KATC model couldn't satisfy this because the interesting data is nested inside a JSON-encoded blob (<code>clientState</code>), with shape that varies per React Query entry. Pre-declaring osquery columns for every nested field isn't workable; we need to flatten the deserialized object into key/value rows so users can pivot in SQL.</p>
<h2>Change</h2>
<p>Add <code>data_flatten</code> as a top-level KATC config option (sibling of <code>comparer</code>, JSON tag <code>data_flatten</code>). When enabled on a table:</p>
<ol>
<li>The configured <code>columns</code> are ignored. The table exposes <code>path</code> plus the standard dataflatten columns: <code>fullkey, parent, key, value, query</code>.</li>
<li>After all <code>row_transform_steps</code> complete, each row is run through <code>dataflatten.Flatten</code>. JSON-encoded string values (as produced by <code>indexeddb.DeserializeChrome</code>) are recursively unmarshalled first so nested JSON is expanded — not left as opaque strings.</li>
<li>The standard dataflatten <code>query</code> column constraint is honored, matching the established pattern in <code>secedit</code> / <code>jwt</code> / <code>airport</code>: callers can filter at the dataflatten layer with <code>WHERE query = 'a/b/*'</code> instead of pulling everything and filtering in SQL.</li>
</ol>
<p>Crucially, <code>indexeddb.DeserializeChrome</code>'s output format is <strong>unchanged</strong>, so existing KATC tables that consume it are unaffected (backwards-compatible).</p>
<h2>Example config</h2>
<pre><code class="language-json">{
  "source_type": "indexeddb_leveldb",
  "source_paths": ["%/IndexedDB/https_claude.ai_0.indexeddb.leveldb"],
  "source_query": "keyval-store.keyval",
  "row_transform_steps": ["deserialize_chrome"],
  "data_flatten": true
}
</code></pre>
<p>The <code>columns</code> field is omitted (would be ignored). Resulting osquery columns: <code>path, fullkey, parent, key, value, query</code>.</p>
<h2>Example queries the PR enables</h2>
<p><strong>Count of chats + dates per chat (<code>chat_conversation_list</code>):</strong></p>
<pre><code class="language-sql">SELECT
  parent AS chat_path,
  MAX(CASE WHEN key = 'uuid'       THEN value END) AS uuid,
  MAX(CASE WHEN key = 'name'       THEN value END) AS name,
  MAX(CASE WHEN key = 'created_at' THEN value END) AS created_at,
  MAX(CASE WHEN key = 'updated_at' THEN value END) AS updated_at
FROM katc_ai_claude_desktop_indexed_db
WHERE query = 'clientState/queries/*/state/data/data/*'
  AND key IN ('uuid', 'name', 'created_at', 'updated_at')
GROUP BY parent;

SELECT COUNT(DISTINCT parent) AS chat_count
FROM katc_ai_claude_desktop_indexed_db
WHERE query = 'clientState/queries/*/state/data/data/*/uuid';
</code></pre>
<p><strong>Count of messages + dates per message (<code>chat_conversation_tree</code>):</strong></p>
<pre><code class="language-sql">SELECT
  parent AS message_path,
  MAX(CASE WHEN key = 'uuid'       THEN value END) AS uuid,
  MAX(CASE WHEN key = 'created_at' THEN value END) AS created_at,
  MAX(CASE WHEN key = 'updated_at' THEN value END) AS updated_at
FROM katc_ai_claude_desktop_indexed_db
WHERE query = 'clientState/queries/*/state/data/chat_messages/*'
  AND key IN ('uuid', 'created_at', 'updated_at')
GROUP BY parent;

SELECT COUNT(DISTINCT parent) AS message_count
FROM katc_ai_claude_desktop_indexed_db
WHERE query = 'clientState/queries/*/state/data/chat_messages/*/uuid';
</code></pre>
<p><strong>Probe what's currently cached (helpful for operators — React Query evicts; empty results aren't a bug):</strong></p>
<pre><code class="language-sql">SELECT DISTINCT value
FROM katc_ai_claude_desktop_indexed_db
WHERE query = 'clientState/queries/*/queryKey/0';
</code></pre>
<h2>Files changed</h2>

File | What
-- | --
ee/katc/config.go | New DataFlatten *bool field on katcTableDefinition (JSON tag data_flatten), honored from base config and from overlays.
ee/katc/table.go | Column construction moved to after overlay resolution; when dataFlatten is enabled the table exposes path + dataflattentable.Columns(). generate resolves dataQueries once via tablehelpers.GetConstraints(queryContext, "query", WithDefaults("*")) and loops per row.
ee/katc/dataflatten.go (new) | flattenRow (recursive-unmarshal + dataflatten.Flatten + dataflattentable.ToMap) and recursivelyUnmarshal (depth-capped helper that JSON-unmarshals any string node, passing through if not valid JSON).
ee/katc/config_test.go | Parse case for a config with data_flatten: true.
ee/katc/table_test.go | Functional test against the Chrome IndexedDB fixture: confirms columns are replaced with the dataflatten set, that rows expand (e.g. aliases/0), that the default query is echoed as *, and that WHERE query = 'uuid' narrows to the uuid leaf only.

<h2>Test plan</h2>
<ul>
<li>[x] <code>go test ./ee/katc/...</code> — passes (existing + new cases)</li>
<li>[x] <code>go test ./ee/indexeddb/...</code> — passes (unchanged)</li>
<li>[x] Manual end-to-end against live Claude desktop IndexedDB: confirmed <code>clientState/queries/*/state/data/data/*</code> returns the cached <code>chat_conversation_list</code> payload with the expected fields, that the <code>query</code> column is plumbed and narrows results, and that the existing non-flattened path is unaffected.</li>
</ul>
<h2>Out of scope / follow-ups</h2>
<ul>
<li><code>chat_conversation_tree</code> data is only present while the user has a conversation open in the app (React Query evicts on its own schedule). This is inherent to how Claude caches; not a launcher bug. The probe query above lets operators check what's currently extractable.</li>
<li>No change to <code>indexeddb.DeserializeChrome</code> — its existing output format is preserved (the recursive unmarshal happens only in the flatten path).</li>
</ul></body></html>

### Local test
image
image

### DIFF
--- a/ee/katc/config.go
+++ b/ee/katc/config.go
@@ -183,7 +183,8 @@ type (
 		SourcePaths       *[]string           `json:"source_paths,omitempty"` // Describes how to connect to source (e.g. path to db) -- % and _ wildcards supported
 		SourceQuery       *string             `json:"source_query,omitempty"` // Query to run against each source path
 		RowTransformSteps *[]rowTransformStep `json:"row_transform_steps,omitempty"`
-		Comparer          *comparerOption     `json:"comparer,omitempty"` // LevelDB/indexeddb comparer: "historical_bytewise", "default_bytewise", or "idb_cmp1" (default)
+		Comparer          *comparerOption     `json:"comparer,omitempty"`     // LevelDB/indexeddb comparer: "historical_bytewise", "default_bytewise", or "idb_cmp1" (default)
+		DataFlatten       *bool               `json:"data_flatten,omitempty"` // If true, flatten each post-transform row through the dataflatten package; the configured Columns are then ignored in favor of dataflattentable.Columns()
 	}
 )
 
--- a/ee/indexeddb/indexeddb.go
+++ b/ee/indexeddb/indexeddb.go
@@ -447,7 +447,7 @@ func CopyLeveldb(ctx context.Context, sourceDb string) (string, error) {
 			return fmt.Errorf("error walking blob directory: %w", err)
 		}
 
-		dest := filepath.Join(dbCopyLocation, strings.TrimPrefix(path, blobDir))
+		dest := filepath.Join(dbCopyLocation, strings.TrimPrefix(strings.TrimPrefix(path, blobDir), string(filepath.Separator)))
 		if d.IsDir() {
 			if err := os.MkdirAll(dest, 0755); err != nil {
 				return fmt.Errorf("making directory %s: %w", dest, err)
--- a/ee/katc/config_test.go
+++ b/ee/katc/config_test.go
@@ -243,6 +243,21 @@ func TestConstructKATCTables(t *testing.T) {
 			expectedPluginCount: 1,
 		},
 		{
+			testCaseName: "with data_flatten enabled",
+			katcConfig: map[string]string{
+				"kolide_dataflatten_test": `{
+					"source_type": "indexeddb_leveldb",
+					"columns": ["data"],
+					"source_paths": ["/some/path/to/db.indexeddb.leveldb"],
+					"source_query": "db.store",
+					"row_transform_steps": ["deserialize_chrome"],
+					"data_flatten": true,
+					"overlays": []
+				}`,
+			},
+			expectedPluginCount: 1,
+		},
+		{
 			testCaseName: "invalid comparer option",
 			katcConfig: map[string]string{
 				"kolide_leveldb_test": `{
--- a/ee/katc/dataflatten.go
+++ b/ee/katc/dataflatten.go
@@ -1,0 +1,65 @@
+package katc
+
+import (
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/kolide/launcher/v2/ee/dataflatten"
+	"github.com/kolide/launcher/v2/ee/tables/dataflattentable"
+)
+
+// maxRecursiveUnmarshalDepth caps recursivelyUnmarshal so a pathological input
+// cannot drive unbounded recursion.
+const maxRecursiveUnmarshalDepth = 100
+
+// flattenRow turns one post-transform row into dataflatten-style output rows.
+// Each value is recursively JSON-unmarshalled first so values produced by
+// IndexedDB deserialization (which encode complex types as JSON strings,
+// sometimes nested) are expanded into their structured form before flattening.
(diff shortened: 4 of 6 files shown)


RebeccaMahany (approved):
Nice work!

ee/katc/config.go
```diff
@@ -183,7 +183,8 @@ type (
 		SourcePaths       *[]string           `json:"source_paths,omitempty"` // Describes how to connect to source (e.g. path to db) -- % and _ wildcards supported
 		SourceQuery       *string             `json:"source_query,omitempty"` // Query to run against each source path
 		RowTransformSteps *[]rowTransformStep `json:"row_transform_steps,omitempty"`
-		Comparer          *comparerOption     `json:"comparer,omitempty"` // LevelDB/indexeddb comparer: "historical_bytewise", "default_bytewise", or "idb_cmp1" (default)
+		Comparer          *comparerOption     `json:"comparer,omitempty"`     // LevelDB/indexeddb comparer: "historical_bytewise", "default_bytewise", or "idb_cmp1" (default)
+		DataFlatten       *bool               `json:"data_flatten,omitempty"` // If true, flatten each post-transform row through the dataflatten package; the configured Columns are then ignored in favor of dataflattentable.Columns()
```
directionless, line 187: It looks like this runs row by row? Why add it as a top level, and not in `RowTransformSteps`?

RebeccaMahany: More details in KOL-843, but I asked for it as a top-level option instead of a row transform step because enabling dataflattening has a side effect of also setting the columns to the dataflatten columns. Since that's a change to the table configuration, it felt cleaner to have it elevated to a top-level option.

PR #2723, opened 2026-05-28
Replace serenize/snaker with internal pkg/snake helper

cesarfda:
Adds a small in-tree pkg/snake package providing CamelToSnake and SnakeToCamel helpers (preserving the original initialism handling) and migrates the four call sites to use it, removing the serenize/snaker third-party dependency.

Co-authored with Cursor

### DIFF
--- a/pkg/snake/snake.go
+++ b/pkg/snake/snake.go
@@ -92,14 +92,10 @@ func CamelToSnake(s string) string {
 		words = append(words, s[lastPos:])
 	}
 
-	var b strings.Builder
-	for k, word := range words {
-		if k > 0 {
-			b.WriteByte('_')
-		}
-		b.WriteString(strings.ToLower(word))
-	}
-	return b.String()
+	// Every word is lower-cased and the separator is "_", which ToLower leaves
+	// untouched, so joining then lower-casing once is equivalent to lower-casing
+	// each word individually.
+	return strings.ToLower(strings.Join(words, "_"))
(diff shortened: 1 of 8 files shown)


pkg/snake/snake.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	var b strings.Builder
+	for k, word := range words {
+		if k > 0 {
+			b.WriteByte('_')
+		}
+		b.WriteString(strings.ToLower(word))
+	}
+	return b.String()
```
RebeccaMahany: Could we do `strings.Join(words, "_")` here? Or is strings.Builder preferable for performance reasons?

pkg/snake/snake_test.go
```diff
@@ -0,0 +1,178 @@
...
+		})
+	}
+}
+
+func TestCamelToSnake_RoundTrip(t *testing.T) {
+	t.Parallel()
+
+	inputs := []string{
+		"launcher_kolide_k2_svc",
+		"launcher_kolide_k2_watchdog_task",
```
RebeccaMahany, line 166: I like that we're testing with these specific strings

pkg/snake/snake.go
cesarfda: done, ty

zackattack01 (approved):

RebeccaMahany (approved):

PR #2733, opened 2026-06-04
Restrict access to launcher.db and its backups on Windows

RebeccaMahany:
The ACL on windows for launcher.db should match the permissions we set on posix -- RW for admin/system only, not readable by other users. I set permissions for launcher.db and all its backups in `checkServiceConfiguration`, where we are already setting the same permissions for the enrollment secret file. Since launcher.db and backups can also be created _after_ the `checkServiceConfiguration` call, I updated to set permissions for launcher.db in `runLauncher`, and for the backup dbs inside the database backup saver -- that way, we don't have to wait for a launcher restart and another call to `checkServiceConfiguration` to set the permissions appropriately.

I noticed we have a couple places where we need to set these permissions on Windows (the aforementioned three locations, plus the socket file for the launcher listener), so I moved the function from the launcher listener into a new package that can be reused in all locations.


cmd/launcher/svc_config_windows.go
```diff
@@ -458,112 +475,18 @@ func checkRootDirACLs(logger *slog.Logger, rootDirectory string) {
...
-		return
-	}
-
-	systemSID, err := windows.CreateWellKnownSid(windows.WinLocalSystemSid)
-	if err != nil {
-		logger.Log(context.TODO(), slog.LevelError,
-			"failed getting SYSTEM SID",
-			"err", err,
+	} else {
+		logger.Log(context.TODO(), slog.LevelInfo,
```
directionless, line 493: As a side note in a drive by review, you're logging both in here, and at the caller. Which makes sense, but is duplicative.

RebeccaMahany: Where is the duplicate log?

directionless: Oh, I'm mistaken. I thought this was in `permissions.RestrictFileAccessToRootOnly`, but this is a caller.

cesarfda (approved):

PR #2737, opened 2026-06-09
Write native messaging manifest on launcher startup

RebeccaMahany:
Part 1 of native messaging support --

This PR updates launcher to create a manifest file to register itself as a native messaging host, and writes it to a location known by Chrome.

I moved `allowlistedDt4aOriginsLookup` out of the `localserver` package to avoid an import cycle.

### DIFF
--- a/ee/nativemessaging/manifest.go
+++ b/ee/nativemessaging/manifest.go
@@ -20,7 +20,6 @@ type manifest struct {
 }
 
 const (
-	nativeMessagingHostName        = "com.kolide.agent"
 	nativeMessagingHostDescription = "Device Trust Agent"
 	nativeMessagingInterfaceType   = "stdio" // This is the only possible value for "type"
 )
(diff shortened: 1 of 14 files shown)


ee/nativemessaging/manifest.go
```diff
@@ -0,0 +1,153 @@
...
+type manifest struct {
+	Name           string   `json:"name"`
+	Description    string   `json:"description"`
+	Path           string   `json:"path"`
+	Type           string   `json:"type"`
+	AllowedOrigins []string `json:"allowed_origins"`
+}
+
+const (
+	nativeMessagingHostName        = "com.kolide.agent"
```
zackattack01: do we have any ideas how this will work with dual launcher installations? just wondering if we should leave some room for something like identifier here, but i might be totally misunderstanding how this will be used

RebeccaMahany: Oh hm, good point -- I had figured it would work the same as localserver does (request goes to both launchers, and then launcher rejects or approves based on the munemo in the msgpack), but I don't actually know how Chrome will handle having two native hosts registered with the same name. I will test it out with the extension and either update the code or document it accordingly!

zackattack01: 👍 cool thank you. fine with addressing in a follow up PR too, everything else looks good to me!

zackattack01 (dismissed):

ee/nativemessaging/manifest.go
directionless: (I didn't read the code) Both multi-launcher and multi-tenancy are worth considering, but also fine for followups.

RebeccaMahany: Okay, after testing, it looks like the extension will only talk to the host in the file name -- I have these symlinks:

```
lrwxr-xr-x  1 root  wheel   67 Jun  9 14:51 com.kolide.agent.json -> /var/kolide-nababe-k2/k2device-preprod.kolide.com/nmh-manifest.json
lrwxr-xr-x  1 root  wheel   69 Jun 10 11:17 com.kolidedev.agent.json -> /var/kolide-nababe-k2/k2device-preprod.kolide.com/nmh-manifest-b.json
```

And even though both point to files that _contain_ a host name of "com.kolide.agent", it only follows the top one. So I will update to disambiguate the host name.

RebeccaMahany: Updated!

zackattack01 (approved):

PR #2738, opened 2026-06-11
Native messaging host support

RebeccaMahany:
Part 2 of native messaging support --

This PR updates launcher to detect when it's being called as a native messaging host, and to validate the incoming message appropriately. For now, it only logs the incoming messages.

I suggest briefly skimming https://developer.chrome.com/docs/extensions/develop/concepts/native-messaging (at least https://developer.chrome.com/docs/extensions/develop/concepts/native-messaging#native-messaging-host-protocol) for context about the native messaging protocol.

### DIFF
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -39,7 +39,9 @@ func main() {
 // or running `runLauncher`. We wrap it so that all our deferred calls will execute before we call
 // `os.Exit` in `main()`.
 func runMain() int {
-	// Check to see if this is a native messaging host invocation. We have to check this first,
+	// Check to see if this is a native messaging host invocation, which looks like:
+	// `some-path-to/launcher chrome-extension://hjlinigoblmkhjejkmbegnoaljkphmgo/`
+	// We have to check and handle the native messaging host invocation case first,
 	// before even initializing logging, because we should not write logs to os.Stdout when following
 	// the native messaging protocol.
 	if len(os.Args) == 2 {
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -39,17 +39,14 @@ func main() {
 // or running `runLauncher`. We wrap it so that all our deferred calls will execute before we call
 // `os.Exit` in `main()`.
 func runMain() int {
-	// Check to see if this is a native messaging host invocation, which looks like:
-	// `some-path-to/launcher chrome-extension://hjlinigoblmkhjejkmbegnoaljkphmgo/`
+	// Check to see if this is a native messaging host invocation.
 	// We have to check and handle the native messaging host invocation case first,
 	// before even initializing logging, because we should not write logs to os.Stdout when following
 	// the native messaging protocol.
-	if len(os.Args) == 2 {
-		if _, ok := nativemessaging.AllowlistedDt4aOriginsLookup()[strings.TrimSuffix(os.Args[1], "/")]; ok {
-			// Native messaging host invocation -- continues reading until closed
-			nativemessaging.ReadNativeMessages(context.Background())
-			return 0
-		}
+	if _, err := nativemessaging.ValidateNativeMessagingArgs(os.Args); err == nil {
+		// Native messaging host invocation -- continues reading until closed
+		nativemessaging.ReadNativeMessages(context.Background())
+		return 0
 	}
 
 	systemSlogger, logCloser, err := multislogger.SystemSlogger()
--- a/cmd/launcher/main.go
+++ b/cmd/launcher/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/kolide/launcher/v2/ee/agent"
 	"github.com/kolide/launcher/v2/ee/control/consumers/remoterestartconsumer"
 	"github.com/kolide/launcher/v2/ee/disclaim"
+	"github.com/kolide/launcher/v2/ee/nativemessaging"
 	"github.com/kolide/launcher/v2/ee/tuf"
 	"github.com/kolide/launcher/v2/ee/watchdog"
 	"github.com/kolide/launcher/v2/pkg/contexts/ctxlog"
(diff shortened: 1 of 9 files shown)


cmd/launcher/main.go
```diff
@@ -38,6 +39,17 @@ func main() {
 // or running `runLauncher`. We wrap it so that all our deferred calls will execute before we call
 // `os.Exit` in `main()`.
 func runMain() int {
+	// Check to see if this is a native messaging host invocation. We have to check this first,
+	// before even initializing logging, because we should not write logs to os.Stdout when following
+	// the native messaging protocol.
+	if len(os.Args) == 2 {
```
zackattack01: a comment about the format of invocation for these might be nice here, took me a second to get why this is fine

RebeccaMahany: Updated, lmk if that wasn't what you were looking for!

ee/nativemessaging/host.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// extractIdentifierFromExecutable pulls the identifier (e.g. kolide-k2) out of
+// the path for the current running executable `executablePath`.
+// We're either running out of the original install location (the bin directory)
+// or out of the update directory (inside the root directory). On all OSes, all
+// of these options should contain the identifier for this installation.
+// We check this path to extract the identifier, which will allow us to determine
+// the root directory location.
+func extractIdentifierFromExecutable(executablePath string) string {
```
zackattack01, line 155: nice 🔥 was curious how you were gonna do this

ee/nativemessaging/host.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// Perform per-OS validation
+	if err := validateBrowser(ctx, browserPath, browserProcessName); err != nil {
+		return potentialExtension, fmt.Errorf("validating browser process %s: %w", browserProcessName, err)
+	}
+
+	// Check that the create time is still the same, so that we know the process hasn't died
+	// and had its PID reused by some other process.
+	browserProcessAfterValidation, err := process.NewProcessWithContext(ctx, browserPid)
```
zackattack01, line 226: this seems fine to me but kinda curious, wouldn't just verifying that our PPID is still the same achieve the same?

ee/nativemessaging/host_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Build our identity assertion:
+	// 1. "anchor apple generic" guarantees the signing chain roots are in Apple's CA
+	// 2. "certificate leaf[subject.OU] = <teamIdentifier>" guarantees the team ID matches what we expect
+	identityAssertion := fmt.Sprintf(`anchor apple generic and certificate leaf[subject.OU] = "%s"`, teamIdentifier)
+	verifyCmd, err := allowedcmd.Codesign.Cmd(ctx, "--verify", "--deep", "--verbose", "-R", "="+identityAssertion, browserPath)
+	if err != nil {
+		return fmt.Errorf("creating codesign --verify cmd: %w", err)
+	}
+	verifyOutput, err := verifyCmd.CombinedOutput()
```
zackattack01, line 44: just double checking here, if the verification itself fails it will return an error? no need to check the output?

RebeccaMahany: Correct!

RebeccaMahany: I will add a comment

ee/nativemessaging/host_darwin.go
```diff
@@ -0,0 +1,48 @@
...
+	`Google Chrome Canary`: "EQHXZ8M8AV",
+}
+
+// validateBrowser confirms that the given path is a known browser on our allowlist
+// that has a valid code signature belonging to a known team identifier.
+func validateBrowser(ctx context.Context, browserPath string, browserProcessName string) error {
+	// The path must be in our allowlist
+	teamIdentifier, found := allowlistedBrowsers[browserProcessName]
+	if !found {
+		return fmt.Errorf("name %s for browser process not in allowlisted browser names", browserProcessName)
```
cesarfda, line 31: on macOS Chrome spawns child work from Google Chrome Helper processes which might not pass this check. Based on getBrowserProcess returning the immediate parent

ee/nativemessaging/host.go
RebeccaMahany: It's a little convoluted, but I couldn't think of a better option 😅

ee/nativemessaging/host.go
RebeccaMahany: I think that wouldn't work in the windows cmd.exe case (browser process terminates => child cmd.exe's ppid changes => launcher's ppid does not?) -- we could traverse upwards again in that case, but at that point we're still have to do the work of `NewProcessWithContext`, and this felt a little simpler?

ee/nativemessaging/host_darwin.go
RebeccaMahany: Hm, I am testing on macOS with both Chrome and Chrome Canary, and I can see that I have Google Chrome Helper on disk (for both Chrome and Chrome Canary), but I haven't had launcher spawned via Google Chrome Helper. Do you have a link I could look at?

ee/nativemessaging/host.go
zackattack01: ahh i see. yeah this makes the most sense to me then!

cmd/launcher/main.go
RebeccaMahany: I moved the os.Args check to a reusable function in the nativemessaging package, so I moved this documentation there as well

ee/nativemessaging/host.go
```diff
@@ -176,20 +176,36 @@ func extractIdentifierFromExecutable(executablePath string) string {
...
-	if len(os.Args) != 2 {
-		return "", fmt.Errorf("unexpected number of args: expected 2, got %d", len(os.Args))
+func ValidateNativeMessagingArgs(osArgs []string) (string, error) {
+	// launcher should be called with exactly 1 argument, which is the extension,
+	// or with one additional argument on Windows:
+	// `some-path-to/launcher chrome-extension://hjlinigoblmkhjejkmbegnoaljkphmgo/ --parent-window=0`.
+	if len(osArgs) != 2 && len(osArgs) != 3 {
+		return "", fmt.Errorf("unexpected number of args: expected 2 or 3, got %d", len(osArgs))
+	}
+	if len(osArgs) == 3 {
```
zackattack01: how confident are we that will always be the exact 3rd arg? just wondering if we're discarding that arg anyway if it's worth being this defensive when we might need a release to fix if we find that it's slightly different on some hosts for whatever reason

zackattack01 (dismissed):

ee/nativemessaging/host.go
RebeccaMahany: It looks like it'll be the third arg (Windows only), but that it's not guaranteed to be 0 --

> On Windows, the native messaging host is also passed a command line argument with a handle to the calling Chrome native window: --parent-window=<decimal handle value>. This lets the native messaging host create native UI windows that are correctly parented. Note that this value will be 0 if the calling context is a service worker.

(From https://developer.chrome.com/docs/extensions/develop/concepts/native-messaging.)

I think you're right and it's safer to just discard; the check for the extension below should be sufficient.

zackattack01 (approved):

PR #2739, opened 2026-06-12
Shorten autoupdate interval from 1h to 15m

RebeccaMahany:
In case of rollback, we want launcher to pick up changes to osquery/launcher more quickly.

There shouldn't be a notable impact from this. The automated tests already reduce the interval down to 1m without performance impact.


cesarfda (approved):

directionless:
Because this is all cached in the CDN, this will have almost no impact on our SaaS. It will have the agents checking TUF data more often. But I think that's a pretty light load.

PR #2740, opened 2026-06-12
Accurately check current running osquery version in autoupdater

RebeccaMahany:
The current behavior on rolling back osquery is as follows:

* The autoupdater refreshes TUF metadata
* The autoupdater checks the current running version of osquery, which pulls from the knapsack (LatestOsquerydPath), which pulls from the TUF metadata -- and since this is a _rollback_, the osquery version is already downloaded, and CheckOutLatest will return its path
* The autoupdater believes incorrectly that it's running the rollback version, and does not restart osquery <- this is the bug

This PR fixes the bug by improving how we check the current running version of osquery -- pulling it from our osquery history's latest instance.

### DIFF
--- a/ee/tuf/autoupdate_test.go
+++ b/ee/tuf/autoupdate_test.go
@@ -21,9 +21,12 @@ import (
 
 	"github.com/Masterminds/semver"
 	"github.com/kolide/launcher/v2/ee/agent/flags/keys"
+	"github.com/kolide/launcher/v2/ee/agent/storage/inmemory"
+	"github.com/kolide/launcher/v2/ee/agent/types"
 	typesmocks "github.com/kolide/launcher/v2/ee/agent/types/mocks"
 	tufci "github.com/kolide/launcher/v2/ee/tuf/ci"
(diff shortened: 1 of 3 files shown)


ee/tuf/autoupdate_test.go
```diff
@@ -353,19 +358,38 @@ func TestExecute_downgrade(t *testing.T) {
...
+
 	// Set up autoupdater
 	client := &http.Client{}
 	t.Cleanup(func() {
 		client.CloseIdleConnections()
 	})
-	autoupdater, err := NewTufAutoupdater(t.Context(), mockKnapsack, client, client, WithOsqueryRestart(func(context.Context) error { return nil }))
+	autoupdater, err := NewTufAutoupdater(t.Context(), mockKnapsack, client, client,
+		WithOsqueryRestart(func(context.Context) error {
+			osquerydRestartCount.Add(1)
```
zackattack01, line 388: oh cool, very clean way to do this

zackattack01 (approved):

PR #2741, opened 2026-06-16
support update downloads during autoupdater initial delay

zackattack01:
This updates the autoupdater behavior to not block during the initial delay. Instead, it will support downloading updates, and note any new versions as pending. The initial delay now works as a timer which will check for any pending version updates and restart accordingly.

This is a defensive measure for devices which may see launcher restarting repeatedly or early for any reason. In cases where functionality from an updated version would prevent the restarts, we may not currently outlive the initial delay long enough to see our updates downloaded. This ensures that when launcher restarts (even if it is not intentional) it can still choose the newly selected version on startup. See KOL-945 for additional details.

### DIFF
--- a/ee/tuf/autoupdate_test.go
+++ b/ee/tuf/autoupdate_test.go
@@ -1390,6 +1390,8 @@ func TestFlagsChanged_DuringInitialDelay(t *testing.T) {
 	t.Parallel()
 
 	testRootDir := t.TempDir()
+	tufServerUrl, rootJson := tufci.InitRemoteTufServer(t, "1.2.3")
+
 	mockKnapsack := typesmocks.NewKnapsack(t)
 	mockKnapsack.On("RootDirectory").Return(testRootDir)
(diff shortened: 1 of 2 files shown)


ee/tuf/autoupdate_test.go
```diff
@@ -1193,6 +1390,8 @@ func TestFlagsChanged_DuringInitialDelay(t *testing.T) {
 	t.Parallel()
 
 	testRootDir := t.TempDir()
+	tufServerUrl, rootJson := tufci.InitRemoteTufServer(t, "1.2.3")
```
zackattack01, line 1393: this test just needed a handful of updates because Execute was previously blocking (as expected) on initial delay throughout. with the changes in this PR we were no longer blocking and then started making real requests to `example.com`. This wires it up like the others to mock this behavior out and adds some logic to ensure full cleanup

RebeccaMahany (approved):

PR #2742, opened 2026-06-17
Remove port from ServerName in tls client config

jbielick:
## Purpose

Using `host:port` for KolideServerURL results in a certificate verification failure because go is asserting the `host:port` matches a DNS or IP SAN, but a port is not valid in those names.

In cases where KolideServerURL contains a host+port, this exact string is used for TLS SNI _and_ certificate hostname verification. Since ServerName contains the port, cert verification fails even when the certificate's SANs include the *host* in the ServerName. Go TLS verification asserts the tlsConfig.ServerName matches the certificate SANs *exactly*, but a port is not valid in a SAN (DNS or IP).

## Approach

When deriving a TLS ServerName from the KolideServerURL, we should do what `http.Transport` does and strip the port (if present). Thus, localhost:3443, IP:PORT, or any other address used as KolideServerURL can be verified against a valid certificate's names correctly.

Strip port from host before using for SNI and hostname verification:
https://github.com/golang/go/blob/beed654b97d1fe530df2e7de764079ba4dd5dcc2/src/net/http/transport.go#L1927-L1932

Clone + set ServerName when blank:
https://github.com/golang/go/blob/beed654b97d1fe530df2e7de764079ba4dd5dcc2/src/net/http/transport.go#L1783-L1786

## Alternatives Considered

It would also be possible to drop ServerName entirely and allow http.Transport to set ServerName for us in addTLS, but this behavior is specific to http.Transport and it would be preferable to be explicit so another transport retained the same behavior (like grpc, which originally did this same splitting in 7046f38ba7cf9f5d69acf017c6707908b25048b1).

### DIFF
--- a/pkg/service/tls.go
+++ b/pkg/service/tls.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"log/slog"
+	"net"
 
 	"github.com/kolide/launcher/v2/ee/agent/types"
 )
@@ -15,6 +16,14 @@ import (
 func makeTLSConfig(k types.Knapsack, rootPool *x509.CertPool) *tls.Config {
 
 	hostname := k.KolideServerURL()
+
+	// ServerName must be host-only: certificate SANs (DNS or IP) never encode a
+	// port, and Go matches ServerName against the SANs literally. SplitHostPort
+	// correctly handles bracketed IPv6 literals; it errors when no port is
+	// present, in which case we leave hostname untouched.
+	if host, _, err := net.SplitHostPort(hostname); err == nil {
+		hostname = host
+	}
 
 	conf := &tls.Config{
 		ServerName:         hostname,
--- a/pkg/service/tls_test.go
+++ b/pkg/service/tls_test.go
@@ -1,0 +1,71 @@
+package service
+
+import (
+	"testing"
+
+	typesmocks "github.com/kolide/launcher/v2/ee/agent/types/mocks"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMakeTLSConfig_ServerNameStripsPort(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name               string
+		kolideServerURL    string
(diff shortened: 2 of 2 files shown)


pkg/service/tls.go
```diff
@@ -16,6 +17,14 @@ func makeTLSConfig(k types.Knapsack, rootPool *x509.CertPool) *tls.Config {
 
 	hostname := k.KolideServerURL()
 
+	// ServerName must be host-only: certificate SANs (DNS or IP) never encode a
+	// port, and Go matches ServerName against the SANs literally. SplitHostPort
+	// correctly handles bracketed IPv6 literals; it errors when no port is
+	// present, in which case we leave hostname untouched.
+	if host, _, err := net.SplitHostPort(hostname); err == nil {
+		hostname = host
+	}
```
RebeccaMahany, line 26: I wonder if we'd want this change everywhere? The only other place I can think of that we're making requests using this URL is in the ee/debug/checkup code, so it's maybe not urgent. But if we wanted it everywhere, I think we'd add a `WithSanitizer` around here: https://github.com/kolide/launcher/blob/main/ee/agent/flags/flag_controller.go#L215

RebeccaMahany: Took a look through -- besides the checkups (connectivity.go, dns.go), we also use this url for flare uploads (https://github.com/kolide/launcher/blob/main/ee/debug/shipper/shipper.go). If you also see the same error when doing flare uploads, or if you see errors for the "Connectivity" checkup when grepping logs, probably worth moving?

jbielick: For flares we're using http.Transport so this is done for us. Really anywhere we're using http.Transport I would say we should not do this by hand, but since this code path was originally grpc and there was some history of the transport changing I thought maybe it should be explicit rather than just omitting ServerName. If there were an instance of tls.Dial, grpc transport, or something that did not use http.Transport I could definitely see the need for doing this splitting once (consolidated).

RebeccaMahany (approved):

PR #2743, opened 2026-06-18
Add schema generation and merging for platform-specific tables in CI

cesarfda:
This pull request introduces a cross-platform schema generation and publishing workflow for the `launcher` project. It adds the ability for each platform's CI build to emit its own table schema, then combines these into a single, unioned schema artifact. This combined schema is then attached to GitHub releases for downstream use (e.g., by k2 for Live Query docs and validation). The changes include updates to CI workflows, new schema combination logic in `cmd/launcher/specs.go`, and corresponding tests.

**CI/CD workflow enhancements:**

* Each platform's build job in `.github/workflows/go.yml` now generates and uploads a platform-specific table schema artifact (`launcher-specs-<platform>.json`). [[1]](diffhunk://#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9fR88-R108) [[2]](diffhunk://#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9fR198-R211)
* A new `combine_schema` job is added to the workflow, which downloads all platform schemas, merges them into a single cross-platform schema (`launcher-schema.json`), and uploads it as an artifact. [[1]](diffhunk://#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9fR689-R758) [[2]](diffhunk://#diff-678682767f2477de3e3c546746f8568b9a1942b2c647d32331d7e774b8ff8d9fR840)
* A new workflow (`.github/workflows/release-schema.yml`) is introduced to attach the combined schema artifact to GitHub releases, supporting both automatic and manual (re-)attachment.

**Schema merging logic:**

* The `launcher specs` command gains a `--merge` flag to combine multiple NDJSON schema files into a single JSON array, deduplicating tables by name and unioning their platforms. [[1]](diffhunk://#diff-25f18b1ee67e61e73f0c67c7b15a82bebc7a4e3ebd4db391b12952cb5433abf2R58-R74) [[2]](diffhunk://#diff-25f18b1ee67e61e73f0c67c7b15a82bebc7a4e3ebd4db391b12952cb5433abf2R170-R276)
* New helper functions (`runMergeSpecs`, `mergeSpecFile`, `unionPlatforms`) are implemented in `cmd/launcher/specs.go` to support merging and deduplication logic.

**Testing:**

* Unit tests are added in `cmd/launcher/specs_test.go` to verify correct merging of schemas, handling of missing inputs, and output format.
* Local setup with k2 parsing support for the spec on Mac

Screenshot 2026-06-18 at 8 55 15 AM

image

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -651,12 +651,10 @@ jobs:
           path: build/
           if-no-files-found: error
 
-  # Generate each platform's table schema from its built binary. Tables are
-  # selected at compile time per GOOS, so each runner emits only its own
-  # platform's tables; they're combined in the combine_schema job. This runs the
-  # cached binary (no build needed) on a regular runner matrix, which avoids
-  # both duplicating the step across build/build_linux and colliding with the
-  # reproducible build twins (those produce no schema artifact).
+  # Generate each platform's table schema by running its cached binary (tables
+  # are compile-time per GOOS, so each runner emits only its own). A separate
+  # matrix job, rather than steps in build/build_linux, avoids duplicating the
+  # step and avoids colliding with the reproducible build twins.
   generate_schema:
     permissions:
       contents: read
@@ -695,14 +693,6 @@ jobs:
           path: launcher-specs-${{ runner.os }}.json
           if-no-files-found: error
 
-  # Fan-in: combine the per-platform schemas emitted by generate_schema into a
-  # single cross-platform launcher-schema.json (JSON array, platforms unioned),
-  # which k2 ingests for its table docs. On tag builds this job also attaches the
-  # combined schema to the tag's GitHub release directly -- we're already inside
-  # the tag's ci run with the artifact in hand, so this avoids racing a separate
-  # release-triggered workflow against ci (a freshly published release fires
-  # `release: published` before this ci run finishes) and avoids re-downloading
-  # the artifact.
   combine_schema:
     permissions:
       contents: write # required to attach the schema to the release on tag builds
@@ -754,12 +744,19 @@ jobs:
           merge-multiple: true
 
       - name: Combine schemas
+        shell: bash
         run: |
+          # Glob the downloaded schemas rather than naming each file, so this is
+          # robust to $RUNNER_OS casing or the platform set changing.
+          shopt -s nullglob
+          schema_files=(schemas/launcher-specs-*.json)
+          if [ ${#schema_files[@]} -eq 0 ]; then
+            echo "no per-platform schema files found in schemas/" >&2
+            exit 1
+          fi
           go run ./cmd/launcher specs --merge \
             --output launcher-schema.json \
-            schemas/launcher-specs-macOS.json \
-            schemas/launcher-specs-Linux.json \
-            schemas/launcher-specs-Windows.json
+            "${schema_files[@]}"
 
       - name: Upload combined schema
         uses: actions/upload-artifact@v4
@@ -775,15 +772,21 @@ jobs:
       # release exists.
       - name: Attach schema to release
         if: github.ref_type == 'tag'
+        shell: bash
         env:
           GH_TOKEN: ${{ github.token }}
           TAG: ${{ github.ref_name }}
         run: |
-          if gh release view "$TAG" --repo "$GITHUB_REPOSITORY" >/dev/null 2>&1; then
+          # A missing release is expected and non-fatal (the manual
+          # release-schema.yml backfill handles it); any other error must fail.
+          if err=$(gh release view "$TAG" --repo "$GITHUB_REPOSITORY" 2>&1 >/dev/null); then
             gh release upload "$TAG" launcher-schema.json --repo "$GITHUB_REPOSITORY" --clobber
             echo "Attached launcher-schema.json to release $TAG"
-          else
+          elif printf '%s' "$err" | grep -qi 'release not found'; then
             echo "::notice::No GitHub release found for tag $TAG; skipping schema upload. Use the 'Attach table schema to release' workflow to attach it once the release exists."
+          else
+            echo "failed to look up release $TAG: $err" >&2
+            exit 1
           fi
 
   package_builder_test:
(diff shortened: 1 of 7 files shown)


cmd/launcher/specs.go
```diff
@@ -155,3 +167,110 @@ func runSpecs(systemMultiSlogger *multislogger.MultiSlogger, args []string) erro
...
+func mergeSpecFile(inputPath string, merged map[string]osquerytable.OsqueryTableSpec) error {
+	f, err := os.Open(inputPath)
+	if err != nil {
+		return fmt.Errorf("opening file: %w", err)
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	// Specs can include lengthy descriptions/examples, so allow large lines.
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
```
directionless: Will this fail if the input is pretty printed?

.github/workflows/go.yml
```diff
@@ -85,6 +85,27 @@ jobs:
         run: make github-launcherapp
         if: ${{ contains(matrix.os, 'macos') }}
 
+      # Generate this platform's table schema from the freshly-built binary.
```
RebeccaMahany: Thoughts on making this its own job that depends on [build, build_linux]? The benefits I see are that 1) we don't have to have a workaround for the collision with the reproducible job, and 2) I think we wouldn't have to duplicate the schema generation the way we do here in both build and build_linux -- the new job should be able to run on a regular matrix of [ubuntu-24.04, macos-26, windows-2025].

cmd/launcher/specs.go
```diff
@@ -155,3 +167,110 @@ func runSpecs(systemMultiSlogger *multislogger.MultiSlogger, args []string) erro
...
+		if err := json.Unmarshal([]byte(line), &spec); err != nil {
+			return fmt.Errorf("unmarshalling spec line %q: %w", line, err)
+		}
+
+		existing, ok := merged[spec.Name]
+		if !ok {
+			merged[spec.Name] = spec
+			continue
+		}
+
```
directionless, line 242: Might be worth an error here that, if the table is cross platform, the schemas match. I'm not sure what `merge` should do if they don't, but I think we'll want to know and try to avoid that case.

docs/architecture/table_schema_publishing.md
```diff
@@ -0,0 +1,233 @@
...
+   release tag carries the version.
+
+> Note on immutability: attaching after publish works for mutable releases. If
+> releases are made immutable, this step should attach the asset at
+> release-creation time (e.g. `gh release create --draft ... && gh release edit
+> --draft=false`) or be folded into an atomic tag → build → sign → release step.
+> The artifact that produces the schema does not change — only where it is
+> attached.
+
+## k2 behavior
```
directionless: This does not belong here.

directionless (commented):
nice

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,80 @@
...
+        id: cirun
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          run_id=$(gh run list \
+            --repo "${{ github.repository }}" \
+            --workflow ci \
+            --branch "${{ steps.tag.outputs.tag }}" \
+            --json databaseId,status,conclusion \
+            --jq 'first(.[] | select(.conclusion == "success" and .status == "completed")) | .databaseId')
```
RebeccaMahany, line 45: I am wondering if `on: release: types: [published]` means that this workflow will trigger basically at the same time as the `ci` workflow triggers, so the `ci` won't have completed at this point and we'll hit "no successful ci run found" below?

If that's the case -- could we hook this onto the end of `combine_schema` instead, with an `if: github.ref_type == 'tag'` check? So `combine_schema` combines and uploads the artifact, and then has an extra step to run `gh release upload` if we're inside a tag ci job. (If that works, it might be simpler anyway, since it lets us skip the extra download step here.)

RebeccaMahany (commented):
Very cool!

cmd/launcher/specs.go
cesarfda: mergeSpecFile now auto-detects NDJSON vs. JSON array (compact or pretty-printed) using encoding/json, so pretty-printed input no longer fails and there's no per-line size limit.

cmd/launcher/specs.go
cesarfda: added conflict error during merge

docs/architecture/table_schema_publishing.md
cesarfda: removed k2 related docs

.github/workflows/go.yml
cesarfda: Done

.github/workflows/go.yml
```diff
@@ -686,18 +651,65 @@ jobs:
...
+        with:
+          name: schema-${{ runner.os }}
+          path: launcher-specs-${{ runner.os }}.json
+          if-no-files-found: error
+
+  # Fan-in: combine the per-platform schemas emitted by generate_schema into a
+  # single cross-platform launcher-schema.json (JSON array, platforms unioned),
+  # which k2 ingests for its table docs. On tag builds this job also attaches the
+  # combined schema to the tag's GitHub release directly -- we're already inside
+  # the tag's ci run with the artifact in hand, so this avoids racing a separate
```
RebeccaMahany: Nitpick -- I don't think you need the explanatory comment here ("we're already inside the tag's ci run...") or above ("This runs the cached binary (no build needed) on a regular runner matrix, which avoids...") since they're comments about specific code review feedback.

cmd/launcher/specs.go
RebeccaMahany: Oh interesting, did this just catch a legit error with our tables? `table \"kolide_nix_upgradeable\": column \"uid\" present on [linux] but not [darwin]`

RebeccaMahany (dismissed):
I am good with this as-is, feel free to ping me if you need a re-review/reapproval! Looks great!

RebeccaMahany (approved):

PR #2747, opened 2026-06-25
Native messaging support for Firefox

RebeccaMahany:
Builds on https://github.com/kolide/launcher/pull/2737 and https://github.com/kolide/launcher/pull/2738; adds support for Firefox.

Support is pretty similar: the messaging protocol is identical, and manifest file registration are very similar (only differing in registry key/file locations). One field in the native messaging host manifest differs between Chrome and Firefox, and Firefox cannot handle having an invalid field present, so we do have to maintain two separate native messaging host manifest files. The invocation of launcher is also different between the two; these differences are documented in `ValidateNativeMessagingArgs`.


ee/nativemessaging/host.go
```diff
@@ -177,21 +177,37 @@ func extractIdentifierFromExecutable(executablePath string) string {
...
+	if _, ok := allowlistedDt4aOriginsLookup[potentialExtension]; ok {
+		return potentialExtension, nil
 	}
 
-	return potentialExtension, nil
+	// For Firefox, launcher should be called with 1 or 2 arguments: the first argument
+	// is the path to the app manifest, and the second argument (starting in Firefox 55)
+	// is the extension ID. We require at least Firefox 55 here so that we can validate
+	// the extension ID.
+	if len(osArgs) != 3 {
```
brhoades, line 200: nit: the comment here is slightly murky. We're explicitly excluding support for Firefox <55 at this point, if I understand correctly?

cesarfda (dismissed):

ee/nativemessaging/host_linux.go
```diff
@@ -18,6 +18,7 @@ var allowlistedBrowsers = map[string]struct{}{
 	"chromium":                 {},
 	"chromium-browser":         {},
 	"chromium-browser-privacy": {}, // RPM
+	"xdg-desktop-portal":       {}, // Firefox distributed via Snap or Flatpak
```
brhoades, line 21: Fascinating. If I knew how to trigger this through Firefox I'd check myself, but are we going to support non-flatpack/snap Firefox?

ee/nativemessaging/manifest.go
```diff
@@ -61,11 +73,25 @@ func AllowlistedDt4aOriginsLookup() map[string]struct{} {
...
+}
+
+func launcherChromeManifestFilePath(rootDir string) string {
+	return filepath.Join(rootDir, "chrome-nmh-manifest.json")
 }
 
-func launcherManifestFilePath(rootDir string) string {
-	return filepath.Join(rootDir, "nmh-manifest.json")
+func launcherFirefoxManifestFilePath(rootDir string) string {
+	return filepath.Join(rootDir, "firefox-nmh-manifest.json")
```
brhoades, line 94: Out of curiosity, I see a mix of hardcoded paths and dynamically joined ones. Is there convention for when to grab one over the other?

ee/nativemessaging/manifest.go
```diff
@@ -61,11 +73,25 @@ func AllowlistedDt4aOriginsLookup() map[string]struct{} {
...
+	chromeManifest, firefoxManifest, err := buildManifests(hostName)
+	if err != nil {
+		return fmt.Errorf("building manifests: %w", err)
+	}
+
+	chromeWriteErr := writeManifest(chromeManifest, launcherChromeManifestFilePath(rootDir), chromeManifestFileRegistrationLocations(hostName))
+	firefoxWriteErr := writeManifest(firefoxManifest, launcherFirefoxManifestFilePath(rootDir), firefoxManifestFileRegistrationLocations(hostName))
+	if chromeWriteErr != nil || firefoxWriteErr != nil {
+		return fmt.Errorf("writing manifest files: chrome: %v; firefox: %v", chromeWriteErr, firefoxWriteErr)
+	}
```
brhoades, line 85: nit: to allow unwrapping to an internal error, if you so choose. Not sure if these make it to Sentry; that's usually why I'd bother.


```suggestion
    err := errors.Any(
        writeManifest(chromeManifest, launcherChromeManifestFilePath(rootDir), chromeManifestFileRegistrationLocations(hostName)),
        writeManifest(firefoxManifest, launcherFirefoxManifestFilePath(rootDir), firefoxManifestFileRegistrationLocations(hostName)),
    )
    if err != nil {
        return fmt.Errorf("writing manifest files: %w", err)
    }
    return nil
```

brhoades (commented):
I'm not loaded up on enough context to confidently sign-off here. But I'll selfishly ask questions and gain context :).

As far as I can tell, this looks good to me, though.

ee/nativemessaging/host.go
RebeccaMahany: Correct -- I'll update the comment for clarity

ee/nativemessaging/manifest.go
RebeccaMahany: By hardcoded paths, do you mean the paths in e.g. `chromeManifestFileRegistrationLocations`/`firefoxManifestFileRegistrationLocations`, or something else?

ee/nativemessaging/manifest.go
RebeccaMahany: I am not seeing `errors.Any` in https://pkg.go.dev/errors -- are you looking at a different errors package? (I like the idea of `errors.Join` in the standard errors package, but since it joins error strings with newlines, it muddies our logs.)

To answer your underlying question, these don't go to Sentry. We used to have them go to [Error Reporting](https://docs.cloud.google.com/error-reporting/docs) and would be grouped there, but that's no longer possible for GDPR reasons.

ee/nativemessaging/host_linux.go
RebeccaMahany: Ideally yes, but I didn't go thru the extra setup for this first pass -- I imagine there will be a bunch of extra iteration for Linux support specifically

ee/nativemessaging/manifest.go
brhoades: Oh, yes, sorry. I am contrasting this approach above where the path is dynamically joined versus paths like these:

https://github.com/RebeccaMahany/launcher/blob/e28da48afbe82eba44701e1d5463f200d7903f82/ee/nativemessaging/manifest_linux.go#L23

https://github.com/RebeccaMahany/launcher/blob/e28da48afbe82eba44701e1d5463f200d7903f82/ee/nativemessaging/manifest_darwin.go#L23

I'd guess the distinction is that we've reached for `filepath.Join` here in platform-independent code? But I'm asking about broader best practices :).

ee/nativemessaging/manifest.go
brhoades: Argh, I did mean `errors.Join`, thank you. That's helpful context too on Join looking gross in logs and that we don't report errors anywhere special.

ee/nativemessaging/manifest.go
RebeccaMahany: Oh, I understand! Yeah, we will usually default to `filepath.Join` for anything that could be multi-platform. For the ones you linked above, they're single-platform and it felt tidiest to copy-paste the full path directly from the docs I was pulling from.

ee/nativemessaging/host_linux.go
RebeccaMahany: (Also I am happy to chat thru the setup process if you'd like to try it out, but it is A Process 😅 )

ee/nativemessaging/host.go
RebeccaMahany: Done!

ee/nativemessaging/manifest.go
cesarfda: Maybe we can put this in the Agents.md (static analyzer would be a little to brittle) cause I will probably forget about this and ask again in a few months 😆

cesarfda (approved):

PR #2748, opened 2026-06-25
Add table for phased apt (ubuntu) updates

Micah-Kolide:
Small table addition for tracking Ubuntu phased updates. The output follows the same pattern as `--upgradeable`, so no new test data is required.


RebeccaMahany (approved):

Micah-Kolide:
Had some issues with my commit signing that should be fixed now.

PR #2751, opened 2026-06-29
Remove two launcher-orphaned sloggers,  observe sendbuffer errors

brhoades:
This PR follows a similar goal to #1869. We observe that some logs and attributes do not properly get shipped via logshipper. The prior PR succeeded, now allowing run_group logs to be shipped, but we're still missing others along with some attributes. 

I've identified a handful of places where we mishandle multiple loggers, drop slogger attributes, or orphan loggers. This PR fixes a few and adds a bit of obs.

Bugs addressed:
  * logshipper silently drops logs when configured storage values are exceeded due to an unconfigured `NopLogger`.
    * logshipper requires sendbuffer and a questionable refactor is needed to connect them.
  * `knapsack.GetRunID()` modified the pointer of its internal loggers, which effectively orphaned any previously retrieved sloggers.
    * This was a root cause of #1869, as rungroup setup preceded RunID initialization and logshipper setup. 
  * `runLauncher` juggled an orphaned slogger at the top-level, cloning it with `With`. I believe the cloned one only logged locally, though I'm not sure.
  
Enough logs are still not shipped that I'm certain these fixes aren't a smoking gun. But they're in the way of my investigation.

### DIFF
--- a/pkg/log/logshipper/logshipper.go
+++ b/pkg/log/logshipper/logshipper.go
@@ -52,7 +52,7 @@ func New(k types.Knapsack, baseLogger log.Logger) *LogShipper {
 	sender := newAuthHttpSender()
 
 	sendInterval := defaultSendInterval
-	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval))
+	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval), sendbuffer.WithLogger(baseLogger))
 
 	// setting a ulid as session_ulid allows us to follow a single run of launcher
 	shippingLogger := log.With(log.NewJSONLogger(sendBuffer), "caller", log.Caller(6), "session_ulid", ulid.New())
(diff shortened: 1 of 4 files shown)


pkg/log/logshipper/logshipper.go
```diff
@@ -52,7 +52,7 @@ func New(k types.Knapsack, baseLogger log.Logger) *LogShipper {
 	sender := newAuthHttpSender()
 
 	sendInterval := defaultSendInterval
-	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval))
+	sendBuffer := sendbuffer.New(sender, sendbuffer.WithSendInterval(sendInterval), sendbuffer.WithLogger(baseLogger))
```
RebeccaMahany, line 55: Oh interesting, I wonder why we weren't using this option before

RebeccaMahany (approved):
LGTM!

pkg/log/logshipper/logshipper.go
brhoades: It was pretty easy to overlook. I only saw it after staring at this file for half an hour.

It's possible someone wanted to come back and later plug in the logshipper.

cesarfda (approved):

PR #2753, opened 2026-06-30
logshipper: increase buffer/batch size, fix potential client race, add sendbuffer draining

brhoades:
PR continues log shipping-related fixes from #2751. It chiefly introduces draining to sendbuffer; without it, local testing consistently dropped logs on the floor.

I separated fixes by commit and it should be easier to review that way.

Summary:
  * `authedHttpSender`:
    * Fix potential race in setting token/endpoint
    * Errors now display the _response_ body instead of the request.
  * `SendBuffer`:
    * Tweak and document batch parameters: max buffer 512 KiB -> 1 MiB, max send size 8 KiB -> 512 KiB.
    * Allow context to cancel in flight HTTP requests ( `sender` now plumbs context).
    * Add `Drain` with config, it drains the buffer on `LogShipper.Stop`:
      * Attempts up to 5 sends, waiting 300 ms in between.
      * Does not keep calling if additional logs appear after sending.
      * Is timed out by `LogShipper.Stop` after 5 seconds.

The Drain logic and tests is the bulk of the PR and contained in the final commit. I'm definitely open to feedback on the durations involved: 5 seconds seemed fairly reasonable, but I'm not married to it.

### DIFF
--- a/pkg/sendbuffer/sendbuffer.go
+++ b/pkg/sendbuffer/sendbuffer.go
@@ -13,7 +13,7 @@ import (
 )
 
 type sender interface {
-	Send(r io.Reader) error
+	Send(ctx context.Context, r io.Reader) error
 }
 
 var (
@@ -148,7 +148,7 @@ func (sb *SendBuffer) Run(ctx context.Context) error {
 	}()
 
 	for {
-		if err := sb.sendAndPurge(); err != nil {
+		if err := sb.sendAndPurge(ctx); err != nil {
 			sb.logger.Log("msg", "failed to send and purge", "err", err)
 		}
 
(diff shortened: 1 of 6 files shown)


zackattack01 (dismissed):
nice!

pkg/sendbuffer/sendbuffer.go
```diff
@@ -266,26 +293,64 @@ func (sb *SendBuffer) DeleteAllData() {
...
+// Attempts to drain the send buffer, sending everything present through its configured
+// sender until empty, ctx is done, or it drains at least the number of logs originally
+// present at call.
+func (sb *SendBuffer) Drain(ctx context.Context) error {
+	storedLines := func() int {
+		sb.writeMutex.Lock()
+		defer sb.writeMutex.Unlock()
+		return len(sb.logs)
+	}
+	initialLines, i := storedLines(), 0
```
RebeccaMahany: Nitpick -- maybe `linesRemaining` instead of `initialLines`, since the modification on 309 means it no longer represents the initial stored line count?

RebeccaMahany (dismissed):
One small nitpick -- nice updates! 🔥

pkg/sendbuffer/sendbuffer.go
brhoades: I gave up on naming this initially :sweat_smile:; the hardest part. Will do.

zackattack01 (approved):

RebeccaMahany (approved):

PR #2754, opened 2026-07-01
added buffered transport to osquery

cesarfda:
### What changed
Wrap the osquery extension manager server's Thrift transport in `thrift.NewTBufferedTransportFactory(16*1024)` (via osquery-go's new `WithTransportFactory` option), in both the runtime instance (`osqueryinstance.go`) and interactive mode (`interactive.go`).

### Why / expected impact
osquery-go's extension server serializes results with `TBinaryProtocol` over an **unbuffered** transport, so every Thrift primitive becomes its own `write(2)` to the extension socket. For large table results this means one write syscall per cell — tens of thousands to millions of tiny writes. Buffering coalesces those into ~16 KB socket writes, drastically cutting syscall count (and the associated context-switch/kernel overhead) with no change to the bytes on the wire.

### Measured impact
Built the launcher at this branch and at its parent, then counted `write(2)` syscalls to the **extension socket** on the launcher process while running fixed queries through `launcher interactive` (which loads the kolide tables and exercises the modified path). `count(*)` keeps osqueryd's own output tiny so the count reflects extension→osquery serialization.

| Query (`kolide_system_profiler` datatype) | rows | writes before | writes after | reduction |
|---|---|---|---|---|
| `SPHardwareDataType` ×5 | 13×5 | 2,624 | 17 | ~154× |
| `SPApplicationsDataType` | 3,244 | 113,717 | ~430 | ~265× |
| `SPFontsDataType` | 55,203 | 1,932,354 | 8,039 | ~240× |

Non-socket writes (logs, augeas, certs) were byte-identical between builds, and for results that fit in one write the socket byte totals matched to within 1 byte — confirming the change only affects how writes are batched, not the data sent.

### Trade-off: buffer size vs. socket send buffer
macOS Unix-domain sockets default to an **8 KB** send buffer (`net.local.stream.sendspace = 8192`). A 16 KB flush therefore often can't be accepted in one `write(2)` — the kernel takes ~8 KB and Go's poller writes the remainder, so a full flush can cost two syscalls instead of one. This is an accepted trade-off: even split in two, we're bounded by `total_bytes / send_buffer` writes instead of one-per-primitive, so the order-of-magnitude win stands.

### Why not size the buffer to the socket send buffer (8 KB)?
- **Negligible syscall gain.** The send buffer caps bytes-per-write, so an 8 KB buffer and a 16 KB buffer produce roughly the same number of actual writes; matching 8 KB would only shave the occasional partial-write retry.
- **Not portable.** 8 KB is a macOS default; Linux `AF_UNIX` send buffers are far larger (~200 KB) and Windows uses named pipes. Coupling the constant to a macOS sysctl would be wrong elsewhere.
- A fixed, generous 16 KB is simple and robust across platforms; the more effective lever for further reduction (raising `SO_SNDBUF`) lives in osquery-go and isn't worth it given the current win.

### How to reproduce
1. Build both binaries: build this branch, then `git stash` the change and build the parent (e.g. `launcher-new` / `launcher-old`).
2. Count socket `write(2)`s on the launcher process while it serializes a heavy result. Either:
   - **dtrace (macOS, needs sudo):** run `launcher interactive` under `dtrace`/`dtruss` counting `syscall::write*:entry` for the launcher pid, or
   - **no root:** `DYLD_INSERT_LIBRARIES` a small shim that interposes `write`/`writev` and counts only calls whose fd is a socket (`S_ISSOCK`).
3. Pipe a fixed query on stdin, e.g. `SELECT count(*) FROM kolide_system_profiler WHERE datatype='SPFontsDataType';`, and compare the socket-write totals between the two builds.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/kolide/krypto v0.1.1-0.20251209172506-7738d1f5d7c1
 	github.com/mat/besticon v3.9.0+incompatible
 	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646
-	github.com/osquery/osquery-go v0.0.0-20260226222546-0cc22f415e57
+	github.com/osquery/osquery-go v0.0.0-20260630173615-eb39ad3443df
 	github.com/peterbourgon/ff/v3 v3.1.2
 	github.com/pkg/errors v0.9.1
 	github.com/stretchr/testify v1.11.1
--- a/go.sum
+++ b/go.sum
@@ -328,8 +328,8 @@ github.com/onsi/gomega v1.19.0/go.mod h1:LY+I3pBVzYsTBU1AnDwOSxaYi9WoWiqgwooUqq9
 github.com/opencontainers/go-digest v1.0.0 h1:apOUWs51W5PlhuyGyz9FCeeBIOUDA/6nW8Oi/yOhh5U=
 github.com/opencontainers/go-digest v1.0.0/go.mod h1:0JzlMkj0TRzQZfJkVvzbP0HBR3IKzErnv2BNG4W4MAM=
 github.com/osquery/osquery-go v0.0.0-20210622151333-99b4efa62ec5/go.mod h1:JKR5QhjsYdnIPY7hakgas5sxf8qlA/9wQnLqaMfWdcg=
-github.com/osquery/osquery-go v0.0.0-20260226222546-0cc22f415e57 h1:t6YJWPvNurotG1WBdjycKpVFdHsvL7QqWslqfimhBWA=
-github.com/osquery/osquery-go v0.0.0-20260226222546-0cc22f415e57/go.mod h1:4cBOmXSmmDULG4bTOq0EFvIy5NUMNJMKbLDBMg6lhJE=
+github.com/osquery/osquery-go v0.0.0-20260630173615-eb39ad3443df h1:GEa8SFf/OXRXSHU5EtjUxs4yPz8caf9UlHN/F89/OJA=
+github.com/osquery/osquery-go v0.0.0-20260630173615-eb39ad3443df/go.mod h1:/Nxgn14on72pNnpjn9Fbw9WQc8JKY1sVv5K+SRZknEk=
 github.com/pelletier/go-toml v1.2.0/go.mod h1:5z9KED0ma1S8pY6P1sdut58dfprrGBbd/94hg7ilaic=
 github.com/pelletier/go-toml v1.6.0/go.mod h1:5N711Q9dKgbdkxHL+MEfF31hpT7l0S0s/t2kKREewys=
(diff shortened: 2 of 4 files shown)


zackattack01 (approved):

RebeccaMahany (approved):

cesarfda:
This has been released as part of 2.4.0 and CPU usage has improved considerably:
image

[source](https://console.cloud.google.com/monitoring/metrics-explorer;endTime=2026-07-21T17:31:29Z;startTime=2026-07-14T17:31:29Z?pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesQuery%22:%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22workload.googleapis.com%2Flauncher.cpu.percent%5C%22%20resource.type%3D%5C%22generic_task%5C%22%20metric.label.%5C%22launcher_version%5C%22%3Dmonitoring.regex.full_match(%5C%22%5C%5C%5C%5Cd*%5C%5C%5C%5C.%5C%5C%5C%5Cd*%5C%5C%5C%5C.%5C%5C%5C%5Cd*%5C%22)%22,%22aggregation%22:%7B%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22crossSeriesReducer%22:%22REDUCE_PERCENTILE_95%22,%22groupByFields%22:%5B%22metric.label.%5C%22launcher_version%5C%22%22%5D,%22alignmentPeriod%22:%2260s%22%7D%7D%7D,%22plotType%22:%22LINE%22,%22targetAxis%22:%22Y1%22,%22minAlignmentPeriod%22:%2260s%22%7D,%7B%22timeSeriesQuery%22:%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22workload.googleapis.com%2Flauncher.cpu.percent%5C%22%20resource.type%3D%5C%22generic_node%5C%22%20metric.label.%5C%22launcher_version%5C%22%3Dmonitoring.regex.full_match(%5C%22%5C%5C%5C%5Cd*%5C%5C%5C%5C.%5C%5C%5C%5Cd*%5C%5C%5C%5C.%5C%5C%5C%5Cd*%5C%22)%22,%22aggregation%22:%7B%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22crossSeriesReducer%22:%22REDUCE_PERCENTILE_95%22,%22groupByFields%22:%5B%22metric.label.%5C%22launcher_version%5C%22%22%5D,%22alignmentPeriod%22:%2260s%22%7D%7D%7D,%22plotType%22:%22LINE%22,%22targetAxis%22:%22Y1%22,%22minAlignmentPeriod%22:%2260s%22%7D%5D,%22chartOptions%22:%7B%22mode%22:%22COLOR%22,%22displayHorizontal%22:false%7D,%22thresholds%22:%5B%5D,%22yAxis%22:%7B%22scale%22:%22LINEAR%22%7D%7D%7D&project=kolide-observability-83af)

PR #2760, opened 2026-07-08
ci govulncheck step rework, go 1.26.4 -> 1.26.5, crypto 0.52 -> 0.53 for GO-2026-5856, GO-2026-5932

brhoades:
New vulns in crypto and /x/crypto. [govulncheck](https://github.com/kolide/launcher/actions/runs/28959175820/job/85925986286?pr=2759) is picking it up now in PRs.

See:
 - crypto: https://pkg.go.dev/vuln/GO-2026-5856
 - /x/crypto: https://pkg.go.dev/vuln/GO-2026-5932
 - https://go.dev/doc/devel/release#go1.26.minor
 
 --
 
 The previous golvulncheck CI job fails on vulnerable but unused transitive dependencies. I kept much of the existing logic, but removed the filters since they do not seem relevant any more? It now only runs after confirmation that the standard `govulncheck` fails.
 
 It's possible it may now output irrelevant deps as vulnerable if one already exists, but right now it's clean. For future folks: if `.finding | .[] | .trace[0]` is null, it seems likely that the vuln doesn't affect our codebase.


RebeccaMahany (dismissed):

zackattack01 (dismissed):

.github/workflows/lint.yml
```diff
@@ -64,22 +64,23 @@ jobs:
         shell: bash
 
       - id: govulncheck
-        run: govulncheck -C . -format json ./... > govulncheck.json
-
-      # If we need to exclude future go-tuf findings that do not apply to our version of go-tuf prior to v2,
-      # adding a final select to JQ will do the trick: `select(. != "GO-2024-3166")`
```
zackattack01, line 70: maybe leave this comment for the next time we add exclusions

zackattack01 (dismissed):
nice 🔥

.github/workflows/lint.yml
brhoades: 6d68471b done

zackattack01 (approved):

PR #2769, opened 2026-07-15
Add pkgutil packages macos platform table

danielhoward314:
## Why?

Completes [KOL-667](https://1password.atlassian.net/browse/KOL-667) by adding macOS platform table for pkgutil data.

## Description

While osquery has tables for [package_install_history](https://osquery.io/schema/5.23.1/#package_install_history) and [package_receipts](https://osquery.io/schema/5.23.1/#package_receipts), their results can be inconsistent. See [this bug report](https://github.com/osquery/osquery/issues/8969) as an example. This PR adds a table whose query implementation uses the allowed command `pkgutil` to query the receipts database for installed packages. It returns rows with `package_id` and `volume` columns.


ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+		output, err := pkgutilExecutor.Exec(volume)
+		if err != nil {
+			slogger.Log(ctx, slog.LevelInfo,
+				"pkgutil failed",
+				"err", err,
+			)
+
+			// pkgutil returns exit status 1 for no results
+			if strings.Contains(err.Error(), "exit status 1") {
+				slogger.Log(ctx, slog.LevelInfo,
```
zackattack01: if we're confident the only reason this returns exit status 1 is for no results found we probably don't need to log this, chances are if people are joining users and there are a whole bunch of `/var/empty` volumes it will log this many many times

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+				"pkgutil failed",
+				"err", err,
+			)
+
+			// pkgutil returns exit status 1 for no results
+			if strings.Contains(err.Error(), "exit status 1") {
+				return nil, nil
+			}
+
+			if os.IsNotExist(errors.Cause(err)) {
```
zackattack01: I'd add a comment here that this can only happen like this if the `pkgutil` executable itself is missing from the device, so safe to return in that case

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+			})
+		}
+
+		return results, nil
+	}
+
+	for _, volume := range volumes {
+		output, err := pkgutilExecutor.Exec(volume)
+		if err != nil {
+			slogger.Log(ctx, slog.LevelInfo,
```
zackattack01: similar comment to the spot below, I'd move this log down after checking for known errors so we aren't spamming repeated logs for non-issues

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+func generatePkgutilData(ctx context.Context, queryContext table.QueryContext, pkgutilExecutor executor, slogger *slog.Logger) ([]map[string]string, error) {
+	ctx, span := observability.StartSpan(ctx)
+	defer span.End()
+
+	results := make([]map[string]string, 0)
+
+	volumes := tablehelpers.GetConstraints(queryContext, "volume")
+	// the volume constraint is optional
+	// when absent, query the root volume and return early
+	if len(volumes) < 1 {
```
zackattack01: could we remove the duplicated logic here by just doing something like `if len(volumes) < 1 { volumes = []string{rootVolume} }`?

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+					"pkgutil returned no results for volume",
+					"volume", volume,
+				)
+				continue
+			}
+
+			if os.IsNotExist(errors.Cause(err)) {
+				return nil, nil
+			}
+			return nil, fmt.Errorf("calling pkgutil: %w", err)
```
zackattack01: you had asked about returning errors- I'd like to see what the rest of the team thinks here too but my guess:
looking at the other tables It's pretty rare that we'll return an error without knowing it's a systemic thing, I'm sure there are reasons I'm not thinking of but one is that if this table is joined it will halt the entire query for one issue (e.g. missing directory, like you noted). In this case we know that anyone joining on users will probably hit a missing directory error if their setup looks at all like ours, I'd just add something like
```
// don't fail this if the directory exists, there won't be any packages anyway
if _, err := os.Stat(volume); os.IsNotExist(err) {
		continue
}
```
to the top of the for loop here. I'm not sure that we would want to expose the internal err directly either. Following the pattern of other tables here I would error log any unexpected errors at the end here (so not missing volume, not missing packages), and then just return `results, nil` to allow partial results from the rest of the query to continue.

edited after discussing with Daniel: we should probably separately log the case where the binary itself is missing- even if it will spam when joined we have log dedup and that should be an extremely rare and unexpected error because pkgutil is a built-in, and it could otherwise be confusing why no data is being returned

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) ([]map[string]string, error) {
+	ctx, span := observability.StartSpan(ctx, "table_name", "kolide_pkgutil_packages")
+	defer span.End()
+
+	pkgutilExecutor := &pkgutilExecutor{
+		ctx:     ctx,
+		slogger: t.slogger,
+	}
+
+	return generatePkgutilData(ctx, queryContext, pkgutilExecutor, t.slogger)
```
directionless, line 68: Why have `generatePkgutilData` and `pkgutilExecutor` as an additional layer of indirection over Table and generate?

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+
+		scanner := bufio.NewScanner(bytes.NewReader(output))
+		for scanner.Scan() {
+			line := scanner.Text()
+			results = append(results, map[string]string{
+				"package_id": line,
+				"volume":     rootVolume,
+			})
+		}
```
directionless: This scanner block is duplicated between the rootVolume path and the volumes path. You're already using `tablehelpers.GetConstraints`, it has a `WithDefault` option that makes this clean. 

Maybe there's somewhere we can update godoc so that's more clear?

ee/tables/pkgutil/pkgutil.go
danielhoward314: Updated!

ee/tables/pkgutil/pkgutil.go
danielhoward314: I followed Seph's suggestion below to use the `WithDefaults` and it has the same deduplicating effect.

ee/tables/pkgutil/pkgutil.go
danielhoward314: I've followed all of these suggestions and it makes the output much cleaner when doing a JOIN on the users table.

ee/tables/pkgutil/pkgutil.go
danielhoward314: This does make it all much cleaner.

I think the godoc comment is clear enough. I missed this helper as I was looking at the patterns I found in other tables.

In [go 1.27 release notes](https://go.dev/doc/go1.27#go-doc), it looks like we could start adding examples that would print with the comments.

ee/tables/pkgutil/pkgutil.go
danielhoward314: I used the indirection so I could inject a mock executor in the tests, since `generate` must be of the [GeneratorFunc](https://pkg.go.dev/github.com/osquery/osquery-go/plugin/table#GenerateFunc) type.

ee/tables/pkgutil/pkgutil.go
danielhoward314: I tried to pin down a hard-coded exit status within a disassembly of `pkgutil` where the `_exit` call is made:

```
otool -tV /usr/sbin/pkgutil > /tmp/pkgutil.asm
grep -n "exit" /tmp/pkgutil.asm
515:00000001000012d0	bl	0x100005870 ; symbol stub for: _exit
534:000000010000131c	bl	0x100005870 ; symbol stub for: _exit
4856:00000001000056a4	bl	0x100005870 ; symbol stub for: _exit
4862:00000001000056bc	bl	0x100005870 ; symbol stub for: _exit
```

If I grep around these addresses where the symbol stub is used, I do find 2 spots where a `0x1` is moved into a register right before an `_exit` call, but there's nothing conclusive about the conditions when these addresses are reached. I could use something Hopper/Ida/Ghidra to see if I could turn up something more conclusive.

In practice, exit status 2 seems reserved for when the volume doesn't exist. If it exists, it seems like `pkgutil` appends a few well-known receipt paths to the provided volume and attempts to reads them.

```
# in one terminal tab
sudo fs_usage -w -f pathname pkgutil

# another tab
pkgutil --volume ~ --pkgs # on machine with no user packages
echo $?
1
pkgutil --pkgs=com\.does.\not\.exist
echo $?
1
```

Both of these queries come up with no results and use exit status 1.

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,119 @@
...
+		output, err := pkgutilExecutor.Exec(volume)
+		if err != nil {
+			// pkgutil returns exit status 1 for no results
+			if strings.Contains(err.Error(), "exit status 1") {
+				continue
+			}
+
+			// log that the binary doesn't exist, but don't return an error
+			if os.IsNotExist(errors.Cause(err)) {
+				slogger.Log(ctx, slog.LevelError,
```
zackattack01: this is my fault- I think I told you warn or error level here but upon reflection it should probably be `Warn` level. The reasoning is because we tag all `Error` level logs for GCP Error Reporting in the multislogger middleware [here](https://github.com/kolide/launcher/blob/9f09953108c2c7c3a5e766f1aa46a9fd0003c30b/pkg/log/multislogger/multislogger.go#L227-L244).

So while we definitely want this to be shipped for debugging (anything `Info` level or above will)- and it is unexpected and will cause query failures (I'd escalate to warn for that), there is nothing we can really do about it other than tell the user to install `pkgutil`.

We monitor the error reporting and would eventually just end up either silencing that or moving it back to warn because it doesn't indicate anything systemic or that we can address. The more generic `"pkgutil failed"` however will only contain errors that we aren't expecting right now, that one makes sense to me to keep as Error level because we're not sure why that would fail and might need to change something on our end if we find that is failing a lot

ee/tables/pkgutil/pkgutil_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/kolide/launcher/v2/ee/tables/pkgutil/mocks"
+	"github.com/kolide/launcher/v2/pkg/log/multislogger"
+	"github.com/osquery/osquery-go/plugin/table"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
```
zackattack01, line 19: nice 🔥

ee/tables/pkgutil/pkgutil.go
zackattack01: nice research! that seems good enough to me

ee/tables/pkgutil/pkgutil.go
zackattack01: looks great, thank you!

ee/tables/pkgutil/pkgutil.go
danielhoward314: Thanks for the context! I pushed up the change to `Warn` level.

zackattack01 (approved):
looks great! thank you 🔥

ee/tables/pkgutil/pkgutil.go
directionless: Yeah. that makes sense. In other places there's an `execCC` type which mocks the exec call. I don't have an opinion about which is cleaner.

The other thing, is like the exec-and-parse stuff. If we assume exec works, then we really only need to test the output parsing. So there are some parser types and tests oriented around that.

PR #2771, opened 2026-07-21
Adds macOS pkgutil package info details platform table

danielhoward314:
Expands on https://github.com/kolide/launcher/pull/2769. The previous PR adds a macOS platform table with the results of `pkgutil --volume <volume> --pkgs` with all of the package ids of installed packages. Given all of the package ids, the next useful query is to drill into the details of an individual package with `pkgutil --volume <volume> --pkg-info=<package-id>`.

Example queries:

```
SELECT * FROM kolide_pkgutil_package_info where package_id = 'com.microsoft.VSCode';

SELECT i.* FROM kolide_pkgutil_packages p JOIN kolide_pkgutil_package_info i ON p.package_id = i.package_id;
```


ee/tables/pkgutil/parser.go
```diff
@@ -0,0 +1,93 @@
...
+			continue
+		}
+
+		valueStr, err := pkgInfoValue(value)
+		if err != nil {
+			return nil, fmt.Errorf("parsing %s: %w", column, err)
+		}
+
+		valueStr = strings.TrimSpace(valueStr)
+		if valueStr == "" {
```
zackattack01: not sure how this ends up looking after the query output but it is probably less confusing to still include this with the empty string value than to omit the key/value pair entirely

ee/tables/pkgutil/parser.go
```diff
@@ -0,0 +1,93 @@
...
+			continue
+		}
+
+		if column == "install_time" {
```
zackattack01: let's see what others think too but i'm not sure we need to do this additional processing. especially since we're turning it back into a string anyway it is usually better to just give the data back as is and we can add optimizations for things like this if we find out they're causing some issue downstream later

ee/tables/pkgutil/pkgutil.go
```diff
@@ -0,0 +1,223 @@
...
+	defer span.End()
+
+	results := make([]map[string]string, 0)
+
+	packageIDs := tablehelpers.GetConstraints(queryContext, "package_id")
+	if len(packageIDs) == 0 {
+		slogger.Log(ctx, slog.LevelInfo,
+			"no package_id provided",
+		)
+		return results, nil
```
zackattack01: I think since this is a required query constraint it's appropriate to return an error here

ee/tables/pkgutil/parser.go
danielhoward314: I've removed this check. Results show up with an empty value for the column, which is similar to the `groups` column in the output.

ee/tables/pkgutil/parser.go
danielhoward314: I've removed all the extra processing of the `install-time`.

ee/tables/pkgutil/pkgutil.go
danielhoward314: Makes sense, I changed the code to return `nil, error` isntead and also the log level to `Error`.

ee/tables/pkgutil/pkgutil_test.go
```diff
@@ -106,20 +106,112 @@ func TestGeneratePkgutilData(t *testing.T) {
...
+			queryContext := table.QueryContext{
+				Constraints: map[string]table.ConstraintList{
+					"package_id": {Affinity: table.ColumnTypeText, Constraints: []table.Constraint{{Operator: table.OperatorEquals, Expression: tt.packageID}}},
+					"volume":     {Affinity: table.ColumnTypeText, Constraints: []table.Constraint{}},
+				},
+			}
+
+			got, err := generatePackageInfoData(t.Context(), queryContext, executor, multislogger.NewNopLogger())
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
```
zackattack01: we typically prefer using `require` for all of these

ee/tables/pkgutil/pkgutil.go
zackattack01: thank you! It doesn't matter a ton but just so you know for future reference, I probably wouldn't log this at LevelError because it is a user issue, nothing we can do about it

ee/tables/pkgutil/pkgutil.go
```diff
@@ -117,3 +127,101 @@ func generatePkgutilData(ctx context.Context, queryContext table.QueryContext, p
...
+	return generatePackageInfoData(ctx, queryContext, pkgutilExec, t.slogger)
+}
+
+func generatePackageInfoData(ctx context.Context, queryContext table.QueryContext, pkgutilExec executor, slogger *slog.Logger) ([]map[string]string, error) {
+	ctx, span := observability.StartSpan(ctx)
+	defer span.End()
+
+	results := make([]map[string]string, 0)
+
+	packageIDs := tablehelpers.GetConstraints(queryContext, "package_id")
```
zackattack01: I probably should have noted this for `volume` on the other table as well, but because this is going to be user-supplied we would typically try to add a `tablehelpers.WithAllowedCharacters` here including anything reasonable for a package id

ee/tables/pkgutil/parser.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return result, nil
+}
+
+func pkgInfoValue(value any) (string, error) {
+	switch v := value.(type) {
+	case string:
+		return v, nil
+	case []any:
```
RebeccaMahany, line 58: Is this case for `groups`? (Is the `key_value` package parsing comma-separate strings into arrays?)

ee/tables/pkgutil/pkgutil.go
```diff
@@ -117,3 +127,101 @@ func generatePkgutilData(ctx context.Context, queryContext table.QueryContext, p
...
+				slogger.Log(ctx, slog.LevelError,
+					"failed to parse pkgutil output",
+					"volume", volume,
+					"package_id", packageID,
+					"err", err,
+				)
+				continue
+			}
+
+			if parsed["package_id"] == "" {
```
RebeccaMahany: Worth doing throwing a check for `package_id` existing in here? i.e. `if id, found := parsed["package_id"]; !found || id == "" {`

RebeccaMahany (commented):
Nice!

ee/tables/pkgutil/pkgutil.go
danielhoward314: Is it best to omit the `slogger.Log` call altogether for a user error like this one?

ee/tables/pkgutil/pkgutil.go
danielhoward314: This case is for any key duplicated across lines; the `key_value` package parses duplicate key occurrences into arrays. I haven't seen `pkgutil` list the same key like this across lines.

The only package I could find that has multiple groups is `com.apple.pkg.MobileDeviceDevelopment` on a 2017 Mac (Ventura). It lists multiple groups on the same line, space-separated.

ee/tables/pkgutil/parser.go
danielhoward314: Meant to reply to this comment:

> This case is for any key duplicated across lines; the key_value package parses duplicate key occurrences into arrays. I haven't seen pkgutil list the same key like this across lines.

> The only package I could find that has multiple groups is com.apple.pkg.MobileDeviceDevelopment on a 2017 Mac (Ventura). It lists multiple groups on the same line, space-separated.

ee/tables/pkgutil/pkgutil.go
danielhoward314: Definitely worth doing! I've added this check.

ee/tables/pkgutil/pkgutil.go
danielhoward314: I've added the helper for package IDs. Apple seems to violate [its own documentation](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundleidentifier#Discussion), since I've found packages that include underscores (e.g. `com.apple.pkg.CLTools_SDK_macOS13`).

zackattack01 (approved):
nice!

PR #2772, opened 2026-07-21
localserver: fix kryptoecmiddlware race

brhoades:
## Summary
In kryptoecmiddleware, we close callbackQueue to signal to the callback worker to shut down. However, incoming requests unconditionally send on that channel even once it's closed.

I've seen one CI flake [here](https://github.com/kolide/launcher/actions/runs/29709483013/job/88251553893) where we hit the race condition on test cleanup. And personally, I have seen this when my network connectivity was poor and callbacks piled up: restarting launcher caused this panic, which was caught, but distracting to my investigation for another issue. The test failure can be seen here:
```
panic: send on closed channel

goroutine 234 [running]:
github.com/kolide/launcher/v2/ee/localserver.(*kryptoEcMiddleware).sendCallback(0xc0005b8820, 0xc000162140, 0xc0002a2540)
	D:/a/launcher/launcher/ee/localserver/krypto-ec-middleware.go:319 +0x67e
github.com/kolide/launcher/v2/ee/localserver.Test_sendCallback.func4()
	D:/a/launcher/launcher/ee/localserver/krypto-ec-middleware_test.go:749 +0x155
created by github.com/kolide/launcher/v2/ee/localserver.Test_sendCallback in goroutine 12
	D:/a/launcher/launcher/ee/localserver/krypto-ec-middleware_test.go:746 +0x7c5
FAIL	github.com/kolide/launcher/v2/ee/localserver	5.644s
```

---

This PR:
  - Uses `shutdown` channel to safely signal that the callback worker needs to shut down.
  - Introduces a separate context for the worker's life time, allowing inflight request interruption.
  - Handles a (likely contrived) purge stall out case.

Questions:

1. ~Is the outcoming callback context thoughtfully separate from `sendCallback` and its callers? I think it is? Both presence detection and the middleware itself seem to fire-and-forget callbacks.~ -> It is
   * But I wonder: would we even need a queue if the server only fired once a callback finished? This likely causes more callbacks, even though we are successfully handling some.
   * As a result, I think my `r.Context()` change is moot as written today.
2. ~How much do we care about the potential hang in a "non-blocking" callback hander in sendCallback's `case e.callbackQueue <- req:`. If the queue is truly full, this will just sit there until it isn't.~ -> This should be exceptionally rare and full hang would only happen on a shutdown anyway.

## Testing
Regression test originally failed, now it passes.

I ran through auth on my device and it also seemed happy. Cleaned up a bit:
```
"DEBUG  origin matches allowlist  "
"DEBUG  successful challenge, proxying  "
"DEBUG  sent callback  "
"DEBUG  origin matches allowlist  "
"DEBUG  successful challenge, proxying  "
"DEBUG  sent callback  "
"DEBUG  success  "
```


ee/localserver/krypto-ec-middleware.go
```diff
@@ -156,15 +158,26 @@ type (
...
+			)
+			return
+		}
 
-	// Worker loop
-	for req := range e.callbackQueue {
 		// Anonymous function to avoid piling up defers
 		if err := func() error {
-			ctx, cancel := context.WithTimeout(context.Background(), client.Timeout)
+			ctx, cancel := context.WithTimeout(req.Context(), client.Timeout)
```
RebeccaMahany: I'm not certain we want this (and the reads from `req.Context().Done()` below) -- callbacks should be able to fire even after the request/response lifecycle has completed.

ee/localserver/krypto-ec-middleware.go
```diff
@@ -156,15 +158,26 @@ type (
...
+
+	for {
+		select {
+		case req = <-e.callbackQueue:
+		case <-e.shutdown:
+			e.slogger.Log(context.TODO(), slog.LevelInfo,
+				"callback worker shut down",
+			)
+			return
+		}
```
RebeccaMahany, line 182: Right now, if we have requests in the queue, I think we might still process those and ignore the shutdown request. If that's the case, I think we'd want to prioritize shutdown, so we could have a check for `e.shutdown` first? i.e. --

```golang
// Return if we've already received a shutdown signal
select {
case <-e.shutdown:
	e.slogger.Log(context.TODO(), slog.LevelInfo,
		"callback worker shut down",
	)
	return
default:
}

// Process next request, or shutdown if it comes through first
select {
case req = <-e.callbackQueue:
case <-e.shutdown:
	e.slogger.Log(context.TODO(), slog.LevelInfo,
		"callback worker shut down",
	)
	return
}
```

RebeccaMahany (commented):
Hopefully my comment helps address your first question -- for the second --

> How much do we care about the potential hang in a "non-blocking" callback hander in sendCallback's case e.callbackQueue <- req:. If the queue is truly full, this will just sit there until it isn't.

I think this is okay?

ee/localserver/krypto-ec-middleware.go
brhoades: Argh, that's a good catch. You're right. I'll actually add a test for this case.

brhoades: This is not possible to test cleanly without pulling out the callback worker handle spawned from the constructor. Any approach as written ends up with flakes where we try weakly to detect that the background worker has exited.

I still made the change, but now we leverage a context that spans the worker's lifetime.

ee/localserver/krypto-ec-middleware.go
brhoades: Thanks, that helps. Though this context is not actually plumbed between producer and consumer, it'd be footgun if someone did that. Removed.

ee/localserver/krypto-ec-middleware.go
RebeccaMahany: I think we still need the two select cases to prioritize the `ctx.Done()` signal over `<-e.callbackQueue`? Or am I missing something?

brhoades: Oh. I forgot Golang selects were order-unbiased, so I flipped the case order and assumed it was fine :face_with_peeking_eye:. Since I couldn't test it reliably, I didn't think much of it.

In Rust, you can opt-in to biased selects and I figured that it was less surprising for Go.

Anyway, you're right. I made the loop condition a context without an error. It could potentially get cancelled afterward, but that can happen no matter how we hold this.

RebeccaMahany: select cases happening in "whatever" order has gotten me more than once 😅

RebeccaMahany (approved):
Thank you for fixing, and for adding the regression test as well!

zackattack01 (approved):

PR #2775, opened 2026-07-23
Add `prefilter` column to dataflatten tables

RebeccaMahany:
This PR adds a new `prefilter` column that allows for passing in CEL (https://cel.dev/) expressions to prefilter results from dataflatten tables to reduce memory impact of expensive queries. This PR adds the new column to all dataflatten tables in `dataflattentable.AllTablePlugins`, but not to the EAV tables or the other tables that use dataflatten -- I will handle those in a follow-up PR to keep the diff size manageable.

Performance improvements:

1. `prefilter` allows for specifying multiple conditions/filters in one expression, resulting in only one call to `generate` instead of multiple. Previously, using `query`, a constraint like `query IN ('type=>user/entrypoint', 'type=>user/timestamp')` would spawn two separate calls to the dataflatten table, one for each value. With prefilter, it is possible to include both in one expression (`has(this.type) && this.type == "user" ? { ?"entrypoint": this.?entrypoint, ?"timestamp": this.?timestamp} : {}`), spawning only one call to the dataflatten table.
2. The prefilter discards unmatched fields or objects prior to performing flatten, so we don't have to pass as much data into flatten. (For `kolide_jsonl` specifically, I updated to use a file reader + an iterator, so that as we read each object into memory, we immediately prefilter and discard any we don't care about -- rather than reading all objects into a list first.)

I evaluated using jq (itchyny/gojq) instead of CEL. I found comparable performance (CEL outperformed jq by a small amount); I went with CEL because the package seems more mature and is more widely adopted.

<details>

<summary>Performance benchmarking</summary>

Test cases:

`path` restriction only: 

```sql
WHERE path LIKE 'some/path/to/%/%.jsonl'
```

`path` and `query` restriction (note that this results in four separate table executions):

```sql
WHERE path LIKE 'some/path/to/%/%.jsonl'
AND query IN (
	'type=>user/entrypoint',
    'type=>user/timestamp',
    'type=>user/cwd',
    'type=>user/sourceToolAssistantUUID'
)
```

`path` and `prefilter` restriction:

```sql
WHERE path LIKE 'some/path/to/%/%.jsonl'
AND prefilter = 'has(this.type) && this.type == "user" ? {
  ?"sourceToolAssistantUUID": this.?sourceToolAssistantUUID,
  ?"entrypoint":             this.?entrypoint,
  ?"timestamp":              this.?timestamp,
  ?"cwd":                    this.?cwd
} : {}'
```

| Test data | Test case | Total allocation | Peak HeapSys (sampled every ms) |
|---|---|---|---|
| Single medium file (196 MB) | `path` restriction only | 1.47 GB | 1.1 GB |
| Single medium file (196 MB) | `path` and `query` restriction | 1.56 GB | 1.1 GB |
| Single medium file (196 MB) | `path` and `prefilter` restriction | 439.37 MB | 41 MB |
| Single large file (2 GB) | `path` restriction only | 12.14 GB | 9.6 GB |
| Single large file (2 GB) | `path` and `query` restriction | 13.3 GB | 6 GB |
| Single large file (2 GB) | `path` and `prefilter` restriction | 3.92 GB | 284.3 MB |
| Many files (10k files, 2 GB total size) | `path` restriction only | 12.38 GB | 6.7 GB |
| Many files (10k files, 2 GB total size) | `path` and `query` restriction | 14.9 GB | 133.3 MB |
| Many files (10k files, 2 GB total size) | `path` and `prefilter` restriction | 5.42 GB | 435.3 MB |

</details>


ee/dataflatten/flatten.go
```diff
@@ -117,8 +119,14 @@ func WithQuery(q []string) FlattenOpts {
 	}
 }
 
-// Flatten is the entry point to the Flattener functionality.
-func Flatten(data any, opts ...FlattenOpts) ([]Row, error) {
+// WithPrefilter specifies a CEL prefilter to apply prior to flattening.
+// It can be used as an alterative to `query` in order to further reduce
+// memory impact of a query.
```
directionless:
```suggestion
// WithPrefilter specifies a CEL prefilter to apply prior to flattening.
// It can be used in order to reduce the memory impact of a query.
func WithPrefilter(p *Prefilter) FlattenOpts {
```

ee/tables/dataflattentable/tables.go
```diff
@@ -39,37 +39,38 @@ type DataSourceType struct {
 var allTypes = []DataSourceType{
 	{
 		tableName:        "kolide_json",
-		description:      "Parses JSON files or raw JSON data and returns flattened key-value pairs. Requires a WHERE path = or raw_data = constraint. Supports a query constraint for filtering specific keys. Useful for reading any JSON configuration or data file.",
+		description:      "Parses JSON files or raw JSON data and returns flattened key-value pairs. Requires a WHERE path = or raw_data = constraint. Supports a query or prefilter constraint for filtering specific keys. Useful for reading any JSON configuration or data file.",
```
directionless: For schema stuff, consider also adding to [`EAVNote`](https://github.com/kolide/launcher/blob/main/ee/tables/dataflattentable/exec_and_parse.go#L111) I forget if that's added to these tables, if so we could move the query and prefilter comments. 

(In the schema, Note is an array, so we could swap that to being an array if it felt cleaner)

ee/tables/dataflattentable/tables.go
```diff
@@ -145,6 +146,23 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 		dataflatten.WithNestedPlist(),
 	}
 
+	var prefilterExpr string
+	if dataPrefilter := tablehelpers.GetConstraints(queryContext, "prefilter"); len(dataPrefilter) > 0 {
+		if len(dataPrefilter) > 1 {
+			return results, fmt.Errorf("The %s table allows for a maximum of 1 prefilter constraint", t.tableName)
```
directionless, line 152: Why not more? It'd be weird, but it's just a for loop...

ee/tables/dataflattentable/tables.go
```diff
@@ -155,7 +173,7 @@ func (t *Table) generate(ctx context.Context, queryContext table.QueryContext) (
 
 		for _, filePath := range filePaths {
 			for _, dataQuery := range tablehelpers.GetConstraints(queryContext, "query", tablehelpers.WithDefaults("*")) {
-				subresults, err := t.generatePath(ctx, queryContext, filePath, dataQuery, append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
+				subresults, err := t.generatePath(ctx, queryContext, filePath, dataQuery, prefilterExpr, append(flattenOpts, dataflatten.WithQuery(strings.Split(dataQuery, "/")))...)
```
directionless, line 176: Didn't the prefilter get compiled and attached on line 163? Why also pass it?

directionless: Oh, it's to add it to the returned rows. Wow, that's awkward :|

Might be the simplest thing for now though

directionless (commented):
I didn't read the code super carefully, but I love it

ee/tables/dataflattentable/tables.go
RebeccaMahany: My reasoning --

1. I wasn't sure what correct behavior "should" be when multiple prefilters are passed
2. Since we're providing prefilter as a way to do _all_ filtering in 1 expr instead of in multiple passes with `query`, it felt correct to nudge the user toward the path of 1 expr

I imagined that the correct behavior would probably be process them in order, and if the user provides two conflicting prefilters, oh well, that's their fault -- but maybe there's something nicer we could do, but then given 2) it didn't feel worth spending as much time on.

ee/tables/dataflattentable/tables.go
RebeccaMahany: Will do, thank you!

ee/tables/dataflattentable/tables.go
RebeccaMahany: It felt a little tidier than keeping `prefilterExpr` inside the `Prefilter` struct and then having to expose that field below 🤷‍♀️ agreed, though

ee/tables/dataflattentable/tables.go
RebeccaMahany: Proposed text --

> const EAVNote = "This is an EAV (Entity-Attribute-Value) table. Rather than a column per field, data is returned as rows with fullkey, key, and value columns. fullkey is the slash-separated path to the value (e.g. network/interfaces/0/name), parent is the path of the containing object, and key is the leaf key name.\n\nFor advanced usage, the query constraint can filter by path -- it supports glob patterns (e.g. WHERE query = 'network/interfaces/*/name'). The prefilter constraint accepts CEL expressions that use the variable name `this` to refer to the object being processed; it can be used to reduce the memory impact of a query."

However, I realized that there is a pretty decent amount of threading through that I will have to do to support prefilter in the EAV tables and all other dataflatten tables outside the ones in `dataflattentable.AllTablePlugins`. This is a big enough diff that I'd rather punt that to a subsequent PR -- I will cover all of those there. (Will note in the PR description here too.)

ee/dataflatten/flatten.go
```diff
@@ -224,6 +249,44 @@ func (fl *Flattener) descend(path []string, data any, depth int) error {
 	return nil
 }
 
+func (fl *Flattener) descendArrayElement(path []string, idx int, data any, depth int) error {
```
RebeccaMahany, line 252: This is moved out of the `case []any` case in descend largely untouched -- I updated two variable names (i => idx, e => data for consistency with other function signatures), and I updated the comment to expand on the example.

RebeccaMahany: (I made this change to fix a bug with query filtering -- easiest to see in the specific commit, which also contains the previously-failing now-passing test case for the bug: https://github.com/kolide/launcher/pull/2775/changes/df46d6e289d48e190529e7963fb2db29f4a32c7d.)

ee/tables/dataflattentable/tables.go
directionless: Oh, if this isn't in all the dataflattebn tables, then I retract the suggestion. Less value there. Good call

ee/tables/dataflattentable/tables.go
directionless: Two prefilters would just generate everything twice with different filters. I'm not sure when it would matter, but 🤷

ee/tables/dataflattentable/tables.go
RebeccaMahany: (Technically it _does_ add the new `prefilter` column to all of them, so I figure I might as well add it to the rest of them in a follow-up so that the column doesn't just silently do nothing for those tables.)

ee/dataflatten/jsonc.go
```diff
@@ -26,7 +26,7 @@ func JsoncFile(file string, opts ...FlattenOpts) ([]Row, error) {
 
 	if json.Valid(transformedRawdata) {
 		// We call Jsonl rather than Jsonc because we know it's already valid transformed JSON
-		return Jsonl(transformedRawdata)
+		return Jsonl(transformedRawdata, opts...)
```
RebeccaMahany, line 29: Drive-by fix of a preexisting bug -- `opts` shouldn't have been dropped here

cesarfda (approved):

ee/tables/dataflattentable/tables.go
directionless: Ah. We're both right then... 

I'm not sure what makes the most sense, we use dataflatten in a lot of places, and not in a way it's going to be easy to add the `prefilter` column. (And probably not in places it matters as much)

Maybe it makes sense to _remove_ it from the standard dataflatten columns and only add it to these fileparsing tables?

RebeccaMahany: Forgot to update here -- I went ahead and added it to the other tables here: https://github.com/kolide/launcher/pull/2779. It wasn't too bad!

PR #2776, opened 2026-07-24
Improve performance of kolide_xml table

RebeccaMahany:
While benchmarking tables for https://github.com/kolide/launcher/pull/2775, Claude flagged that `kolide_xml` is slow, and surfaced the fix in this PR. The `mxj` package's reader wrapper does unnecessary work; wrapping our reader in a bufio reader prevents using the `mxj` reader. (See: https://github.com/clbanning/mxj/blob/master/xml.go#L106.)

Benchmarking:

before change: 23166839875 ns/op => 23 seconds

after change: 2299419792 ns/op => 2.3 seconds

### DIFF
--- a/ee/dataflatten/xml.go
+++ b/ee/dataflatten/xml.go
@@ -1,6 +1,7 @@
 package dataflatten
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


directionless (approved):
What a special assumption mxj makes.

PR #2780, opened 2026-07-29
rearrange filewalker filter checks to optimize regexp calls

zackattack01:
We're seeing a device with launcher utilizing excessive CPU, the cpuprofile points to the filewalker logic. The majority of the cycles are coming from syscalls and I think we'll need to address another way (behavior looks expected depending on the number of files for this device), but the regexp matching can also be expensive and this seemed like an easier win for now.

Because all of the filters have to pass for us to retain a filepath entry, and we're sending down ~20 skipDirs per filwalk config, I think it makes sense to do the cheaper checks first. In many cases the files will not match the `fileNameRegex` anyway and we can avoid doing the more costly 20 regexp checks later

### DIFF
--- a/ee/filewalker/filewalker.go
+++ b/ee/filewalker/filewalker.go
@@ -192,19 +192,20 @@ func (f *filewalker) Filewalk(ctx context.Context) {
 					return nil
 				}
 
-				// Check to see if we're in a directory that should be skipped
-				if f.shouldSkip(path) {
-					return fs.SkipDir
-				}
-
-				// If our config restricts file type, check that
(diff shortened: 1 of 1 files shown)


RebeccaMahany (approved):

brhoades (approved):
PR makes sense as an quick optimization.

If we're already using cgo, I wonder if one of the cgo regex engines would buy us much performance? I seem to remember they're quite a bit faster... though it depends on how expensive regexp is compared to the other filewalker syscalls you saw :sweat_smile:.

PR #2782, opened 2026-07-31
[do not merge] repro for CI-only Windows memory bug

brhoades:
A one-off stress test to try to repro Windows memory-related crashes. You can see an example crash here:
https://github.com/kolide/launcher/actions/runs/29606158269/job/88468298622

Cannot repro locally.


brhoades:
https://github.com/actions/runner-images/issues/14483#issuecomment-5185166792

Seems likely to be this issue. I've left a comment.

PR #2784, opened 2026-08-04
Use Cloudsmith for builds

RebeccaMahany:
Uses https://github.com/step-security/cloudsmith-cli-action to authenticate our CI actions to Cloudsmith to pull Golang packages from there.

This PR allows launcher's GH actions to authenticate to Cloudsmith, so that it can pull from the 1Password golang registry during builds. Note that this change only applies to main and tagged builds currently, because we can't mint ID tokens for PRs from forks. There's a workaround described [here](https://github.com/actions/attest-build-provenance/issues/99#issuecomment-2609705859), but it opens us up to a new category of security considerations -- so I am punting on it for now. That means that for now, PRs will _not_ use Cloudsmith when building.

For local development, Kolide engineers should generate a Cloudsmith API key and then configure their GOPROXY to use it, as described here: https://docs.cloudsmith.com/formats/go-registry#private-repositories -- in a follow-up PR, I will define a Makefile target or something to make this easier. For others, the project will continue to build even without Cloudsmith access.


.github/workflows/go.yml
```diff
@@ -133,6 +157,27 @@ jobs:
...
+        # For now, we only use Cloudsmith on builds we're going to distribute: refs/head/main for nightly builds, and 'tag' for stable builds.
+        # cloudsmith-cli-action should mask the API key, but we add additional masks just to be safe.
+        if: github.ref == 'refs/heads/main' || github.ref_type == 'tag'
+        shell: bash
+        run: |
+          echo "::add-mask::${CLOUDSMITH_API_KEY}"
+          [ -z "${CLOUDSMITH_API_KEY:-}" ] && exit 1
+          goproxy="https://token:${CLOUDSMITH_API_KEY}@golang.cloudsmith.io/1password/op-golang/"
+          echo "::add-mask::$goproxy"
+          echo "GOPROXY=$goproxy" >> "$GITHUB_ENV"
```
danielhoward314, line 179: Is there much difference in using `GITHUB_ENV` versus an `env:` block that uses the OIDC token as a step output?

https://github.com/agilebits-inc/b5/blob/cbd599820772e3afd845a1f0594842acd04fdda0/.github/actions/deps_cache_go/action.yaml#L30-L45

RebeccaMahany: Basically the only difference is that I wanted GOPROXY applied to all of the steps in the job, instead of just one. (I can scope it down later, but since I have to merge to main to be able to test this, I wanted to start with it as wide as possible to avoid having to iterate through multiple PRs because I missed setting `env:` on a step.)

danielhoward314 (approved):

zackattack01 (approved):

brhoades (approved):

PR #2785, opened 2026-08-04
Fix queries for kolide_keychain_ tables

RebeccaMahany:
The queries for `kolide_keychain_acls` and `kolide_keychain_items` were swapped, leading to incorrect data returned by both tables -- both the wrong data was returned, and columns were null (type/modified/label/comment/created were all null for `kolide_keychain_items`, and authorizations/keychain_path/description were all null for `kolide_keychain_acls`).

This PR swaps the queries. After making this change, the data returned by each table is correct and complete, with all columns filled in.

@Micah-Kolide confirmed that these tables are not in use by Kolide, nor by any custom checks.


Micah-Kolide (approved):
LGTM. Tested and confirmed that the data is now showing up correctly for each table.

PR #2788, opened 2026-08-10
Remove obsolete launcher tables and fix Mach-O parsing

cesarfda:
## Summary
- remove the legacy Google Drive sync config/history and Chrome login data emails tables
- keep the distinct `kolide_chrome_login_keychain` table registered
- support thin and universal Mach-O binaries, returning one row per architecture
- add fixture-based coverage for both Mach-O formats

## Dependencies
- depends on [kolide/k2#15061](https://github.com/kolide/k2/pull/15061), which migrates Check references from `kolide_chrome_login_data_emails` to `katc_chrome_login_data_emails`; merge that PR first

## Follow-ups
- removal of `kolide_spotlight` and `kolide_onepassword_accounts` moved to [launcher#2791](https://github.com/kolide/launcher/pull/2791), linked to [KOL-1178](https://1password.atlassian.net/browse/KOL-1178)
- retain `kolide_slack_config` for now because it is still in use

## Test plan
- [x] Run `go test ./cmd/launcher ./pkg/osquery/table -count=1`
- [x] Run `make lint`
- [x] Build launcher and query `/bin/ls` through `kolide_macho_info`
- [x] Verify the generated launcher schema no longer contains the removed tables

## Related
- [KOL-1145](https://1password.atlassian.net/browse/KOL-1145)


pkg/osquery/table/platform_tables_darwin.go
```diff
@@ -98,13 +97,10 @@ func platformSpecificTables(k types.Knapsack, slogger *slog.Logger, currentOsque
 		brew_upgradeable.TablePlugin(k, slogger),
 		ChromeLoginKeychainInfo(k, slogger),
 		firmwarepasswd.TablePlugin(k, slogger),
-		GDriveSyncConfig(k, slogger),
-		GDriveSyncHistoryInfo(k, slogger),
 		MDMInfo(k, slogger),
 		macos_software_update.MacOSUpdate(k, slogger),
 		macos_software_update.RecommendedUpdates(k, slogger),
 		MachoInfo(k, slogger),
-		spotlight.TablePlugin(k, slogger),
```
RebeccaMahany, line 107: Did you confirm that these tables aren't in use in any custom checks?

pkg/osquery/table/macho.go
```diff
@@ -34,19 +35,34 @@ func generateMacho(ctx context.Context, queryContext table.QueryContext) ([]map[
 		return nil, errors.New("The kolide_macho_info table requires that you specify a constraint WHERE path =")
 	}
 	path := q.Constraints[0].Expression
-	f, err := macho.Open(path)
+
+	f, thinErr := macho.Open(path)
+	if thinErr == nil {
+		defer f.Close()
+
+		return []map[string]string{machoResult(path, f.Cpu.String())}, nil
```
RebeccaMahany, line 46: I could see us thinking in the future that `if thinErr == nil` is a typo and should be a `if thinErr != nil`, could you throw a quick comment in documenting why we return early here? `// If macho.Open doesn't return an error, then we have a thin file; return its CPU` or something

RebeccaMahany (commented):
Thank you for the cleanup!

pkg/osquery/table/platform_tables_darwin.go
cesarfda: I'll wait on confirmation here: https://kolide.slack.com/archives/CJ3Q1QAJJ/p1786374680528949

pkg/osquery/table/macho.go
cesarfda: Done

RebeccaMahany (dismissed):

pkg/osquery/table/platform_tables_darwin.go
directionless: Heh, I came here to say the same thing. I think we were using it within the last couple years. Which of this or the osquery one works better seems to swap every couple years.

cesarfda:
https://github.com/kolide/k2/pull/15061 merged so we should be unblocked to merge this pr

RebeccaMahany (approved):

PR #2789, opened 2026-08-10
Preserve embedded datetime translations on localization updates

cesarfda:
## Summary

Localized devices were still showing English relative time in the menu bar app (e.g. `Last Menu Update: 5 Minutes Ago` on an `es-ES` device).

The control server only sends the localization sections it owns — currently just the notification "Learn More" label, see [`LocalizationsService`](https://github.com/kolide/k2/blob/main/app/services/agent_control/localizations_service.rb):

```json
{"locale":"es-ES","translations":{"es-ES":{"notifications":{"actions":{"learn_more":"Más información"}}}}}
```

But `LocalizationConsumer` merged that with `maps.Copy`, which replaces a locale's entire `Translations` entry rather than merging fields. So the `es-ES` entry ended up with the correct `learn_more` and every datetime string blanked to `""`, wiping the Spanish strings that ship in `assets/es-ES.json`. `PluralForms.Select` then returned `""`, and `relativeTimeLocalized` fell back to `relativeTimeDefault` (English).

Confirmed on a live device — `localization.json` on disk had `"x_minutes": {"other": ""}` for `es-ES` while the other 122 bundled locales were intact, since `es-ES` was the only one the server sent.

This merges section by section instead, leaving empty incoming sections alone. It also applies to the store-read path in `NewLocalizationConsumer`, so devices that already persisted a blanked entry recover on their next start rather than staying broken.

Only affects the launcher-controlled strings in scope for [KOL-213](https://1password.atlassian.net/browse/KOL-213). The rest of the menu (`About Kolide...`, `View Details...`, etc.) is server-rendered text from `kolide_desktop_menu` and is expected to follow k2's locale handling.

## Test plan

- [x] `TestLocalizationConsumer_NotificationsOnlyUpdate` uses the real payload shape and asserts the embedded datetime strings survive both an `Update` and a consumer restart. Verified it fails against the old behavior (`expected "%{count} minutos", actual ""`) and passes with the fix.
- [x] `go test ./ee/control/consumers/localizationconsumer/ ./ee/desktop/...` passes. (`TestDesktopUserProcessRunner_Execute` times out locally for me on a clean `main` as well — unrelated, it spawns real desktop processes.)
- [ ] Verify on a device with a non-English locale that the menu bar relative time renders localized.

### DIFF
--- a/ee/control/consumers/localizationconsumer/localizationconsumer.go
+++ b/ee/control/consumers/localizationconsumer/localizationconsumer.go
@@ -74,7 +74,7 @@ func NewLocalizationConsumer(slogger *slog.Logger, kvStore types.KVStore) (*Loca
 		return nil, fmt.Errorf("failed to unmarshal localization data from store: %w", err)
 	}
 
-	maps.Copy(t.localizationData.Translations, localizationFromStore.Translations)
+	t.mergeTranslations(localizationFromStore.Translations)
 
 	// set the locale to the one from the store
 	t.localizationData.Locale = localizationFromStore.Locale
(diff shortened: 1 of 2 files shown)


RebeccaMahany (approved):
🔥

PR #2790, opened 2026-08-10
Handle kv.sqlite database contention/incomplete transactions

RebeccaMahany:
A recent update to modernc.org/sqlite makes `_busy_timeout` available. When `_busy_timeout` is set, sqlite will wait for the timeout instead of immediately returning a `SQLITE_BUSY` error. I am hopeful that this will resolve some of the contention-related error logs that we see.

I chose a timeout of 10 seconds based solely on `rungroup.InterruptTimeout`.


brhoades:
Oh interesting. Are the errors mostly Windows? I could see multiple threads attempting concurrent sqlite opens; that may need a separate fix. 

I've seen it on MacOS before too, come to think of it, coming up as EBUSY from the OS.

ee/agent/storage/sqlite/keyvalue_store_sqlite_test.go
```diff
@@ -130,6 +134,56 @@ func TestGetSet(t *testing.T) {
 	require.NoError(t, s.Close())
 }
 
+func TestGetSet_Simultaneous(t *testing.T) {
```
brhoades, line 137: I think testing this with N parallel tests against a single database across platforms might be interesting. Unless we're doing that somewhere.

brhoades (approved):
LGTM. This is a good add. I didn't realize this was a modern feature?

RebeccaMahany:
@brhoades 

> Are the errors mostly Windows?

Windows is overrepresented in some of these specific error logs because it's the only OS where we use the launcher watchdog, so the only place where we're writing to the watchdog logs store in kv.sqlite. For error `writing settings: upserting into startup_settings: database is locked (5) (SQLITE_BUSY)`, slightly less than half are Windows (looking very briefly at logs just over today).

ee/agent/storage/sqlite/keyvalue_store_sqlite_test.go
RebeccaMahany: I don't think we have anything with a shared test database at the moment!

zackattack01 (approved):

brhoades:
Got it. I don't know the proportional fleet size, but if it is a file handle contention problem it'd have to be disproportionate to track with my experience. Just a thought, no blockers from me here.

PR #2791, opened 2026-08-11
Remove obsolete Spotlight and 1Password tables

cesarfda:
## Summary
- remove `kolide_spotlight`, which is redundant with osquery's `mdfind` table
- remove `kolide_onepassword_accounts`, which reads legacy 1Password SQLite formats and is no longer used
- stop registering and publishing both tables in the launcher schema

## Test plan
- [x] Run `go test ./pkg/osquery/table -count=1`
- [x] Run `make lint`

## Dependency
- depends on [KOL-1178](https://1password.atlassian.net/browse/KOL-1178)


brhoades (approved):
LGTM, seems simple enough.

zackattack01 (approved):

PR #2794, opened 2026-08-13
Use dbus => OpenURI to open URLs when possible

RebeccaMahany:
Adding thanks to the suggestion on https://github.com/kolide/launcher/issues/1537 from @smlx.

I see improved behavior on Ubuntu 26.04 -- with previous options, opening URIs would not respect my default browser and open in Chrome instead.

Also added some connection cleanup in the notification handler that was missing before, and added the basic rungroup actor test + goleak detection.

### DIFF
--- a/ee/desktop/user/notify/notify_linux.go
+++ b/ee/desktop/user/notify/notify_linux.go
@@ -7,13 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"strings"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/godbus/dbus/v5"
-	"github.com/kolide/kit/ulid"
 	"github.com/kolide/launcher/v2/ee/allowedcmd"
 )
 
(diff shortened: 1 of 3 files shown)


ee/desktop/user/notify/notify_linux.go
```diff
@@ -98,7 +109,21 @@ func (d *dbusNotifier) Execute() error {
...
+			if err == nil {
+				continue
+			}
+			d.slogger.Log(context.TODO(), slog.LevelWarn,
+				"couldn't open URI via dbus, falling back to exec",
+				"err", err,
+			)
 
 			for _, browserLauncher := range browserLaunchers {
 				cmd, err := browserLauncher.Cmd(context.TODO(), actionUri)
```
brhoades, line 129: If I'm reading this right, we attempt to open the default browser on dbus signal in a loop?

I've been cut by this before with `xdg-open`. I don't think there's a guarantee that the child process forks... and wouldn't that block this whole loop 🤔? You can test this by having it launch a fresh browser instance when one wasn't running, the xdg-open command runs until the child process exits for me.

Let me know if I'm off. I think the dbus mechanism is sound (if it fails quickly). I wonder if handling in a goroutine may be cleaner?

brhoades (commented):
Oh wow, Linux does not make this easy at all.

zackattack01 (approved):

ee/desktop/user/notify/notify_linux.go
brhoades: Oh. This is "Start", not "Exec". So this does not block.

Sorry :).

RebeccaMahany: Correct, it's non-blocking!

PR #2799, opened 2026-08-20
golangci lint 2.13.0 updates

RebeccaMahany:
### SA4023: Impossible comparison of interface value with untyped nil

This check isn't working well cross-platform. For example, here's a Windows failure:

```
 Error: ee\desktop\runner\runner.go:220:3: SA4023(related information): the lhs of the comparison is the 2nd return value of this function call (staticcheck)
		runner.osVersion, err = osversion()
		^
Error: ee\desktop\runner\runner.go:221:6: SA4023: this comparison is always true (staticcheck)
		if err != nil {
		   ^
Error: ee\desktop\runner\runner.go:1358:2: SA4023(related information): the lhs of the comparison is the 2nd return value of this function call (staticcheck)
	currentOsVersion, err := osversion()
```

https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner.go#L220
https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner.go#L1358

`osversion` does indeed always return an error on Windows (https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner_windows.go#L69) -- but it does not on macOS (https://github.com/kolide/launcher/blob/main/ee/desktop/runner/runner_darwin.go#L64). So while it may be a technically correct error when running golangci-lint on Windows, it is not correct for our cross-platform application.

For now, I have `nolint`ed or added build directives where reasonable to ignore these errors. If they get too numerous in the future, we can exclude this check via the config instead.

### SA1019: Using a deprecated function, variable, constant or field

I am continuing to punt on our `(crypto/ecdsa.PublicKey).X` and `(crypto/ecdsa.PublicKey).Y` deprecation for now, since it's going to be more of a pain to untangle. I filed KOL-1205 to handle this later.

```
Error: ee\localserver\jwk_keys.go:66:3: SA1019: (crypto/ecdsa.PublicKey).X has been deprecated since Go 1.26 and an alternative has been available since Go 1.25: modifying the raw coordinates can produce invalid keys, and may invalidate internal optimizations; moreover, [big.Int] methods are not suitable for operating on cryptographic values. To encode and decode PublicKey values, use [PublicKey.Bytes] and [ParseUncompressedPublicKey] or [crypto/x509.MarshalPKIXPublicKey] and [crypto/x509.ParsePKIXPublicKey]. For ECDH, use [crypto/ecdh]. For lower-level elliptic curve operations, use a third-party module like filippo.io/nistec. (staticcheck)
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -120,6 +120,11 @@ linters:
       - linters:
           - staticcheck
         text: QF1001
+      # Impossible comparison of interface value with untyped nil
+      # https://staticcheck.dev/docs/checks/#SA4023
+      - linters:
+          - staticcheck
+        text: SA4023
     paths:
       - test-cmds
       - third_party$
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -120,11 +120,6 @@ linters:
       - linters:
           - staticcheck
         text: QF1001
-      # Impossible comparison of interface value with untyped nil
-      # https://staticcheck.dev/docs/checks/#SA4023
-      - linters:
-          - staticcheck
-        text: SA4023
     paths:
       - test-cmds
       - third_party$
--- a/ee/localserver/jwk_keys.go
+++ b/ee/localserver/jwk_keys.go
@@ -63,8 +63,8 @@ func (j *jwk) ecdsaPubKey() (*ecdsa.PublicKey, error) {
 	// Construct the ECDSA public key.
 	pubKey := &ecdsa.PublicKey{
 		Curve: curve,
-		X:     x,
-		Y:     y,
+		X:     x, //nolint:staticcheck
+		Y:     y, //nolint:staticcheck
 	}
 
 	// this is a little weird, but it's the recommended way to validate a public key,
--- a/ee/localserver/jwk_keys.go
+++ b/ee/localserver/jwk_keys.go
@@ -39,7 +39,7 @@ func parseEllipticCurve(str string) (elliptic.Curve, error) {
 	}
 }
 
-// ecdsaPubKey converts jwk in to ecdsa public key
+// ecdsaPubKey converts jwk into ecdsa public key
 func (j *jwk) ecdsaPubKey() (*ecdsa.PublicKey, error) {
 	curve, err := parseEllipticCurve(j.Curve)
 	if err != nil {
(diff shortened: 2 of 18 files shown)


.golangci.yml
```diff
@@ -120,6 +120,11 @@ linters:
       - linters:
           - staticcheck
         text: QF1001
+      # Impossible comparison of interface value with untyped nil
+      # https://staticcheck.dev/docs/checks/#SA4023
+      - linters:
+          - staticcheck
+        text: SA4023
```
zackattack01: were there a lot of these that weren't working cross-platform? I totally agree with nolinting the one explained in the PR description but in general I've been bitten by the interface != nil thing before (seems like that is what this check is actually supposed to flag) and it was a nightmare to track down what was going on. it seems like this would be useful to flag future offenses but if it's too noisy I'm fine with either approach

zackattack01 (dismissed):

.golangci.yml
RebeccaMahany: It's not terribly noisy. The most annoying ones were pkg/windows/windowsupdate for macOS/Linux (https://github.com/kolide/launcher/actions/runs/32377867072/job/96453539730?pr=2798). For Windows, we have just the desktop runner example above plus ee/powereventwatcher (https://github.com/kolide/launcher/actions/runs/32377867072/job/96453536883?pr=2798). (The powereventwatcher one is an outlier -- `*syscall.LazyProc.Call` does always return an error in the third argument, but the check we have there against `operationSuccessfulMsg` is adequate regardless and I don't want to change the logic unnecessarily.)

The pkg/windows/windowsupdate ones are the only ones that would be annoying to nolint ignore, and I can just change the build directives to windows only for those. 🤞 Trying it out here: https://github.com/kolide/launcher/pull/2799/changes/7884041a00939710616b5e748526da15080f4f38

RebeccaMahany: It works, but I have to also `nolint` the "related information" lines too (https://github.com/kolide/launcher/pull/2799/changes/4be69dd8de30460ad0226af0b241ec348ade02ec). I'm still good with removing the exception from `.golangci.yml` for now; we can revisit if it gets very annoying later.

ee/localserver/jwk_keys.go
```diff
@@ -63,8 +63,8 @@ func (j *jwk) ecdsaPubKey() (*ecdsa.PublicKey, error) {
 	// Construct the ECDSA public key.
 	pubKey := &ecdsa.PublicKey{
 		Curve: curve,
-		X:     x,
-		Y:     y,
+		X:     x, //nolint:staticcheck
+		Y:     y, //nolint:staticcheck
```
brhoades, line 67: Ah, bummer, when the jwk explicitly encodes x/y it's not nearly as easy as https://github.com/kolide/krypto/pull/61

.golangci.yml
zackattack01: looks great, thank you!

zackattack01 (approved):

brhoades (approved):
LGTM.

ee/localserver/jwk_keys.go
RebeccaMahany: I had checked that exact PR in hopes that you'd already found this fix 😅

PR #2801, opened 2026-08-20
Handle `ecdsa.PublicKey.X` and `ecdsa.PublicKey.Y` deprecation

RebeccaMahany:
Handles the following deprecation:

> (crypto/ecdsa.PublicKey).X has been deprecated since Go 1.26 and an alternative has been available since Go 1.25: modifying the raw coordinates can produce invalid keys, and may invalidate internal optimizations; moreover, [big.Int] methods are not suitable for operating on cryptographic values. To encode and decode PublicKey values, use [PublicKey.Bytes] and [ParseUncompressedPublicKey] or [crypto/x509.MarshalPKIXPublicKey] and [crypto/x509.ParsePKIXPublicKey]. For ECDH, use [crypto/ecdh]. For lower-level elliptic curve operations, use a third-party module like filippo.io/nistec. (staticcheck)

> (crypto/ecdsa.PublicKey).Y has been deprecated since Go 1.26 and an alternative has been available since Go 1.25: modifying the raw coordinates can produce invalid keys, and may invalidate internal optimizations; moreover, [big.Int] methods are not suitable for operating on cryptographic values. To encode and decode PublicKey values, use [PublicKey.Bytes] and [ParseUncompressedPublicKey] or [crypto/x509.MarshalPKIXPublicKey] and [crypto/x509.ParsePKIXPublicKey]. For ECDH, use [crypto/ecdh]. For lower-level elliptic curve operations, use a third-party module like filippo.io/nistec. (staticcheck)

### DIFF
--- a/ee/localserver/dt4a_auth_middleware.go
+++ b/ee/localserver/dt4a_auth_middleware.go
@@ -265,16 +265,10 @@ func (c *chain) validate(trustedKeys map[string]*ecdsa.PublicKey) error {
 		}
 	}
 
-	rootKey, ok := trustedKeys[c.Links[0].SignedBy]
+	// Get copy of root key
+	parentEcdsa, ok := trustedKeys[c.Links[0].SignedBy]
 	if !ok {
 		return fmt.Errorf("root key with kid %s not found in trusted keys", c.Links[0].SignedBy)
-	}
-
-	// make a copy of the root key so that we can reassign it
-	parentEcdsa := &ecdsa.PublicKey{
-		Curve: rootKey.Curve,
-		X:     rootKey.X, //nolint:staticcheck
-		Y:     rootKey.Y, //nolint:staticcheck
 	}
 
 	var currentPayload payload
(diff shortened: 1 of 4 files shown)


ee/localserver/dt4a_auth_middleware.go
```diff
@@ -265,18 +265,12 @@ func (c *chain) validate(trustedKeys map[string]*ecdsa.PublicKey) error {
 		}
 	}
 
-	rootKey, ok := trustedKeys[c.Links[0].SignedBy]
+	// Get reference to root key
+	parentEcdsa, ok := trustedKeys[c.Links[0].SignedBy]
```
RebeccaMahany, line 269: We just need a reference to the parent here -- we're not mutating the key itself, so the copy below was unnecessary

brhoades (approved):
LGTM.

What an unfortunate step back as an API consumer, but it appears this is really what Go intends going forward 😞. Thanks for finishing up.

PR #2800, opened 2026-08-20
launcher: log on Windows options parsing, better handle path overrides

brhoades:
Plumbs and uses a logger in launcher options parsing. Previously on Windows, options parsing quietly overrode the configured root directory if a well-known database location existed. This is made more confusing by privileged Windows launchers not logging options parsing _at all_.

Changes on Windows:
  - Log on root directory overrides.
  - Unprivileged processes no longer use restricted override folders.

I'm not very familiar with the history of why this is present, but I believe this is backwards compatible with solving the MSI issue? Please let me know.

## Testing - Unprivileged
Command:
```ps1
Start-Process -FilePath ./launcher.exe -ArgumentList "--config=launcher.flags" -Wait -NoNewWindow -RedirectStandardOutput .\20260824.stdout.log -RedirectStandardError .\20260824.stderr.log
```

### Fresh Run
```jsonl
{"time":"2026-08-24T18:44:29.5801767Z","level":"INFO","msg":"launcher running on windows without elevated permissions, using default stdout instead of eventlog"}
{"time":"2026-08-24T18:44:29.5801767Z","level":"INFO","msg":"launcher starting up","version":"unknown","revision":"unknown"}
{"time":"2026-08-24T18:44:29.5801767Z","level":"INFO","msg":"got new version of launcher to run","old_version":"unknown","new_binary_version":"","new_binary_path":"C:\\Users\\billy\\kolide-2\\launcher.exe"}
{"time":"2026-08-24T18:44:29.5801767Z","level":"INFO","msg":"found newer version of launcher to run","new_binary":"C:\\Users\\billy\\kolide-2\\launcher.exe"}
{"time":"2026-08-24T18:44:29.5801767Z","level":"INFO","msg":"preparing to run command","cmd":"C:\\Users\\billy\\kolide-2\\launcher.exe --config launcher.flags","component":"execwrapper"}
{"time":"2026-08-24T18:44:33.4751211Z","level":"INFO","msg":"launcher running on windows without elevated permissions, using default stdout instead of eventlog"}
{"time":"2026-08-24T18:44:33.4787187Z","level":"INFO","msg":"launcher starting up","version":"unknown","revision":"unknown"}
{"time":"2026-08-24T18:44:33.4808552Z","level":"WARN","msg":"not running elevated, initializing launcher at configured root path and skipping well-known override locations","root_directory":"C:\\Users\\billy\\kolide-2\\"}
[...]

PR #2804, opened 2026-08-26
init ee/currentprocess, centralize current process query APIs, fix unprivileged desktop runner

brhoades:
Why are you reading this:
I promised in #2800 to port and generalize process `IsElevated` logic. It's done. I used it to fix unprivileged Windows desktop runner detection and found a few weird bits to tweak.

Summary:
  * Add `ee/currentprocess` and which has cross-platform `IsElevated() (bool, error)` and `Uid()`
  * Consume new currentprocess API in `flare`, `doctor`, windows syslogger, desktop runner, and launcher options parsing.
  * desktopRunner is now aware of whether it's running privileged.
    * On POSIX platforms, this is minor. It does not run sudo/launchctl.
    * On Windows, it does not query the console users nor use the active user's token; both error.
    * Across all platforms this is passive to privileged execution.  

I've left comments on the review for areas I'm least happy about or context.

Fixes:
  * Unprivileged Windows could not spawn a desktop process at all.
  * Unprivileged Linux from a user service or non-graphical shell launched, but never showed a tray icon.
  * Transient failures in Linux/MacOS on `user.CurrentUser()` are cached, so were permanently fatal to the runner.
  * Dead code: `errors.Is(err, NoExplorerProcessError{})` was always false.

Known remaining bugs:
  * Failure to spawn for any console user can block spawning for others.

## Testing
I ran the general launcher privileged and unprivileged on all three platforms. It detected permissions when it had them or when it didn't. Logs available, but I didn't attach them since I'm worried they may have credentials in them.

On Windows, that led me to discover that unprivileged users can't query console users at all which was fatal:
<details>
<summary>json failure log</summary>
```json
{
  "time": "2026-08-26T20:36:33.257766Z",
  "level": "ERROR",
  "source": {
    "function": "github.com/kolide/launcher/v2/ee/desktop/runner.(*DesktopUsersProcessesRunner).Execute",
    "file": "/home/billy/work/launcher/ee/desktop/runner/runner.go",
    "line": 271
  },
  "msg": "could not run console user desktop process",
  "component": "desktop_runner",
  "err": "getting console users: getting logon session data for LUID: Access is denied."
}
```
</details>

Also worth noting that unprivileged users have a complicated set of requirements [to spawn a process to their own user's graphical session](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessasusera):
<details><summary>json failure log</summary>
```json
{
  "time": "2026-09-11T15:47:16.7741552Z",
  "level": "ERROR",
  "source": {
    "function": "github.com/kolide/launcher/v2/ee/desktop/runner.(*DesktopUsersProcessesRunner).Execute",
    "fil  e": "/home/billy/work/launcher/ee/desktop/runner/runner.go",
    "line": 300
  },
  "msg": "could not run console user desktop process",
  "component": "desktop_runner",
  "err": "spawning  new desktop user process for DESKTOP-1P62ADB\\billy: running desktop command as user: fork/exec C:\\Users\\billy\\kolide\\launcher-edd16a0c-dirty-change-token-bind.exe:  A required privilege is not held by the client."
}
```
</details>

## Questions
1. ~~I assumed we can add flare fields without any changes in k2. I also figured it'd be clearer to add an error key than to push it all into the old field. Is that right?~~
2. ~~Is `ee/currentprocess` the right folder/shape for a common API here? I didn't see clear distinction between pkg and ee.~~


ee/desktop/runner/runner.go
```diff
@@ -857,16 +893,30 @@ func (r *DesktopUsersProcessesRunner) runConsoleUserDesktop() error {
...
-	consoleUsers, err := consoleuser.CurrentUids(ctx)
-	if err != nil {
-		return fmt.Errorf("getting console users: %w", err)
+	var consoleUsers []string
+
+	// Querying console users on Windows requires privilege, other platforms
+	// use session information to establish process configuration even when ran
+	// unprivileged
+	if runtime.GOOS == "windows" && !r.elevated {
+		consoleUsers = []string{r.currentUid}
```
brhoades, line 903: I don't love this, but an alternative distracts from my changeset: extract the loop's body below and call it for `!r.elevated` on all platforms. I can do it if preferred.

ee/desktop/runner/runner_linux.go
brhoades, line 1: The root check was harmful in some cases. When ran as a systemd user unit (which seems plausible) or tmux (which is probably just me), the process won't have DISPLAY.

Also worth noting that `userEnvVars` below worked fine unprivileged on my box.

ee/watchdog/controller_windows.go
```diff
@@ -251,8 +251,11 @@ func (wc *WatchdogController) shouldManageWatchdog() bool {
 		return false
 	}
 
-	// we also don't alter watchdog installation if we're running without elevated permissions
-	if !windows.GetCurrentProcessToken().IsElevated() {
+	if elevated, err := currentprocess.IsElevated(); err != nil {
+		// historical behavior treats failure to check as not elevated
+		wc.slogger.Log(context.TODO(), slog.LevelError, "failed to check if process is elevated", "err", err)
+		return false
```
brhoades, line 257: The historical behavior here is from IsElevated() which treats [a failure to get token info as unprivileged](https://cs.opensource.google/go/x/sys/+/master:windows/security_windows.go;l=761-763;drc=c507910bb52510a2ae06048a4246ad0fe210a872):

```go
// IsElevated returns whether the current token is elevated from a UAC perspective.
func (token Token) IsElevated() bool {
	var isElevated uint32
	var outLen uint32
	err := GetTokenInformation(token, TokenElevation, (*byte)(unsafe.Pointer(&isElevated)), uint32(unsafe.Sizeof(isElevated)), &outLen)
	if err != nil {
		return false
	}
	return outLen == uint32(unsafe.Sizeof(isElevated)) && isElevated != 0
}
```

If written today, I'd rather we fail to configure watchdog when unprivileged than not configure watchdog when privileged and a Windows API misbehaves. 

Any opinions?

RebeccaMahany:
> I assumed we can add flare fields without any changes in k2. I also figured it'd be clearer to add an error key than to push it all into the old field. Is that right?

It should be fine -- you should be able to test by generating a flare and confirming k2 consumes it and displays it appropriately on the flare page.

> Is ee/currentprocess the right folder/shape for a common API here? I didn't see clear distinction between pkg and ee.

Yes, this is a good spot for it. In general, new code lives in `ee` (which has a different license from `pkg`).

ee/debug/checkups/flare-environment-platform-specifics_windows.go
```diff
@@ -2,8 +2,19 @@
...
+	"github.com/kolide/launcher/v2/ee/currentprocess"
+)
 
 func flareEnvironmentPlatformSpecifics(flareEnv map[string]any) {
-	flareEnv["invoked_with_elevated_permissions"] = windows.GetCurrentProcessToken().IsElevated()
+	elevated, err := currentprocess.IsElevated()
+
+	if err != nil {
+		flareEnv["invoked_with_elevated_permissions"] = "unknown"
+		flareEnv["invoked_with_elevated_permissions_err"] = fmt.Sprintf("failed to check if elevated: %v", err)
```
RebeccaMahany, line 16: Oh yeah, this is completely safe to add. Just gets thrown into a data file in the flare.

ee/desktop/runner/runner.go
RebeccaMahany: I am fine with this -- since the launcher parent process being unprivileged is a new case (i.e. we always currently expect the parent process to be privileged), we can define whatever behavior we want here.

ee/desktop/runner/runner_darwin.go
```diff
@@ -16,15 +16,16 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// For notifications to work, we must run in the user context with launchctl asuser.
```
RebeccaMahany, line 19: Can we retain this info in the comment below? `If the uid differs from the user running the current process, runAsUser uses launchctl runas to start cmd in user context (required for notifications to work)`

ee/watchdog/controller_windows.go
RebeccaMahany: Seems fine to me -- @zackattack01 ?

RebeccaMahany (dismissed):
LGTM, one small documentation request

ee/debug/checkups/flare-environment-platform-specifics_windows.go
RebeccaMahany: This lives in the top-level `metadata.json` in the flare zip if you want to confirm

ee/desktop/runner/runner_darwin.go
brhoades: Absolutely. I missed that context getting dropped in my prereview.

a04cd28b1ff5f2fbdaa41392c5bee9554ed8702a

ee/debug/checkups/flare-environment-platform-specifics_windows.go
brhoades: Oh, great, thank you. I made a few flares and they were fine, I didn't know if k2 did more.

```
$ cat metadata.json
{"goarch":"amd64","goos":"windows","invoked_with_elevated_permissions":true,"mode":"standalone","version":{"version":"unknown","branch":"unknown","revision":"unknown","go_version":"go1.26.7","build_date":"unknown","build_user":"unknown"}}
```
So we're good.

RebeccaMahany (approved):

ee/watchdog/controller_windows.go
zackattack01: sorry for the delay here, yes that seems okay to me!

zackattack01 (approved):
this LGTM! i'm fine with this as is or if you want to change the watchdog controller behavior go for it, the task itself is installed to run as SYSTEM regardless of the permissions of the invoking process so it should fail. and subsequent startups with elevated permissions should notice the missing task and try again

PR #2816, opened 2026-09-09
Remove debian 11 and add debian 13 to exec tests

RebeccaMahany:
Updating per failure here: https://github.com/kolide/launcher/actions/runs/34297323501/job/102457655799?pr=2815

> E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired (invalid since 1d 14h 53min 6s). Updates for this repository will not be applied.


zackattack01 (approved):

brhoades (approved):
Ahh, I noticed this yesterday but didn't realize it was done.

Bummer that we can't pin to stable/unstable :(.

PR #1606, opened 2024-02-15
Update some exec calls to use `tablehelpers.Exec`

directionless:
Relates To: #1321

I started chasing these down and converting them.

PR #2817, opened 2026-09-10
mdm: use tablehelpers.RunSimple for profiles exec

### LINKED ISSUE #1321: Launcher should report errors from all exec tables

techreign:
Fixes #1321

## What was wrong

`kolide_mdm_info`'s `getMDMProfile` and `getMDMProfileStatus` (in `pkg/osquery/table/mdm.go`) called `allowedcmd.Profiles.Cmd(ctx, ...)` directly and read the result with `cmd.Output()`, instead of going through `tablehelpers` like the rest of the tables migrated in #1606 and #1696.

## What changed

Both functions now call `tablehelpers.RunSimple(ctx, slogger, 10, allowedcmd.Profiles, args)`, preserving the existing 10 second timeout and args exactly. `MDMInfo` didn't previously have a `*slog.Logger` threaded down to its generate function, so a small closure now passes `slogger` through `generateMDMInfo` -> `getMDMProfile`/`getMDMProfileStatus`, mirroring how the `gsettings` migration in #1696 threaded `slogger` into its helper. `mdm_test.go` was updated to pass `multislogger.NewNopLogger()`, per this repo's test conventions.

## Inventory of remaining `allowedcmd`/exec call sites in `ee/tables/**` and `pkg/osquery/table/**`

I grepped both trees for every place a table package creates and directly runs a command (`allowedcmd.*.Cmd(...)` followed by `.Run()`/`.Output()`/`.CombinedOutput()`/`.Start()`, and any raw `exec.Command`/`exec.CommandContext`, which `forbidigo` already forbids repo-wide). Everything else already goes through `tablehelpers.Run`/`tablehelpers.RunSimple` (directly, or via `dataflattentable.NewExecAndParseTable`, which itself calls `tablehelpers.Run`).

| Location | Pattern | Action |
|---|---|---|
| `pkg/osquery/table/mdm.go` (`getMDMProfile`, `getMDMProfileStatus`) | `allowedcmd.Profiles.Cmd` + `cmd.Output()` | **Migrated** to `tablehelpers.RunSimple` (this PR) |
| `ee/tables/windowsupdatetable/windowsupdate.go` (`callQueryWindowsUpdatesSubcommand`) | `allowedcmd.Launcher.Cmd` + `cmd.CombinedOutput()` | **Left alone** — see below |
| `ee/tables/homebrew/upgradeable.go` (`generate`) | `allowedcmd.Brew.Cmd(ctx)` | **Left alone** — not actually executed |
| `ee/tables/tablehelpers/disclaim_darwin.go`, `ee/tables/tablehelpers/exec.go` | `allowedcmd.Launcher.Cmd` / `execCmd.Cmd` | **Left alone** — this *is* the tablehelpers implementation |

### Why `windowsupdate.go` was left alone

`callQueryWindowsUpdatesSubcommand` re-execs `launcher.exe query-windowsupdates` as a workaround for a memory leak in the underlying `Search` COM call. It's deliberately different from a plain "run and capture" exec:

- It has no timeout of its own by design — the doc comment says "If a timeout has been set on `ctx`, it will be respected by the underlying exec," and one caller (`cache_windows.go`) relies on this to set an intentional 20-minute deadline for a background cache refresh, far longer than any table-appropriate default `tablehelpers.Run` timeout. Picking an arbitrary `timeoutSeconds` for `tablehelpers.Run` would either truncate that 20-minute cache path or silently stop enforcing any bound at all elsewhere, either way changing behavior instead of just moving code.
- After creating the command it validates `cmd.Path` actually ends in `launcher.exe` before running it (a safety check against running an unexpected binary, e.g. in tests), which doesn't fit the `RunSimple`/`Run` signature without adding a bespoke `ExecOps` just for this one call site.

This is likely the "windowsupdate might be all that's left" case @RebeccaMahany flagged on the issue; given the above I think it's a legitimate exception rather than something to force through the helper, but happy to revisit if maintainers prefer a dedicated `ExecOps` for the path check.

### Why `homebrew/upgradeable.go` was left alone

`allowedcmd.Brew.Cmd(ctx)` there is only used to `os.Stat` the binary and find its owning UID (so `brew` can later be run as that user via `tablehelpers.WithUid`); the command is never actually executed at that call site. The real exec a few lines down already goes through `tablehelpers.Run`.

## forbidigo lint rule

Per the issue discussion, I looked at whether a one-line `forbidigo` pattern could enforce "no raw `allowedcmd.*.Cmd()` outside `tablehelpers`" going forward. `forbidigo` patterns in this repo's `.golangci.yml` apply repo-wide (as @James-Pickett noted, per-path linter config isn't supported), and the three intentional exceptions above (plus `tablehelpers`'s own implementation) mean a blanket pattern would immediately flag legitimate code and need per-call-site `//nolint` comments to reach zero violations. That's beyond a one-line config change, so I left it out rather than add a rule with exceptions baked in — flagging it here per the issue instead.

## Verification

```
gofmt -l pkg/osquery/table/mdm.go pkg/osquery/table/mdm_test.go   # no output

go build ./...                                                    # native Windows, exit 0

go vet ./pkg/osquery/table/...                                    # native Windows, no output
go test ./pkg/osquery/table/...                                   # ok  	github.com/kolide/launcher/v2/pkg/osquery/table	0.098s

GOOS=darwin go vet ./ee/tables/...   # only pre-existing failures, unrelated to this PR (see below)
GOOS=linux  go vet ./ee/tables/...   # only pre-existing failures, unrelated to this PR (see below)

golangci-lint run ./pkg/osquery/table/...                         # 0 issues
golangci-lint run ./...                                           # 0 issues
```

`mdm.go` carries a `//go:build darwin` tag, so it can't be exercised with `go build`/`go vet` from this (non-macOS) machine — the whole `pkg/osquery/table` package transitively imports macOS-cgo-only code (`github.com/knightsc/system_policy/sp`, `secureenclave`, `kolide/systray`'s darwin internals) that needs a real `clang` toolchain to cross-compile. I confirmed via `git stash` that this same failure reproduces identically on an unmodified checkout, so it's pre-existing and unrelated to this change. With that one darwin-cgo file temporarily moved out of the directory (not committed), `GOOS=darwin go vet ./pkg/osquery/table/...` type-checks the rest of the package, including the new `mdm.go`, cleanly.

Similarly, `GOOS=darwin`/`GOOS=linux go vet ./ee/tables/...` (I didn't touch anything under `ee/tables` in this PR, since the only remaining migration candidate turned out to be in `pkg/osquery/table`) only fails on `ee/tables/cryptsetup` and `ee/tables/fscrypt_info` (pull in `google/fscrypt`, cgo-only) and `ee/tables/knightsc/kextpolicy` (same darwin-cgo issue as above) — all pre-existing per the same `git stash` check, and excluding just those three packages leaves the rest of `ee/tables` vetting clean for both GOOS values.


RebeccaMahany:
@techreign thank you for the contribution -- happy to review once the CLA is signed

